PR #400, opened 2021-11-08
📨 Show send and receive activity for all tokens

henryboldi:
### Send and receive activity
This PR adds send and receive activity for all tokens functionality to the activity panel on the home tab. This helps us move towards the goal of a complete home tab. 

Of note, this adds a `determineAsset` function that attempts to extract the asset data from a transactions' input data. This uses a variety of heuristics, which isn't ideal but seems to work pretty well. A part of this function deals with swaps, and that's probably the fuzziest part. I can totally chop out all the swap-related code here

Screen Shot 248


henryboldi:
I might sneak swap activities in here — it's pretty close 😬
Swap

ui/components/Wallet/WalletActivityListItem.tsx
```diff
@@ -35,20 +37,40 @@ export default function WalletActivityListItem(props: Props): ReactElement {
...
           </div>
           <div className="right">
-            <div className="outcome">
-              From:
-              {` ${activity.from.slice(0, 6)}...${activity.from.slice(37, 41)}`}
-            </div>
+            {activity.isSent ? (
+              <div className="outcome">
+                To:
+                {` ${activity?.to?.slice(0, 6)}...${activity?.to?.slice(
```
henryboldi, line 60: Perhaps this formatting should be moved to the selector

ui/utils/transactionInputInterpretionHelper.ts
```diff
@@ -0,0 +1,20 @@
+import { ActivityItem } from "@tallyho/tally-background/redux-slices/activities"
+
+export function determineActivityValue(activityItem: ActivityItem) {
+  const { asset } = activityItem
+  let { value } = activityItem
+  // Derive value from transaction transfer input if not ETH
+  if (asset && !asset.symbol.includes("ETH") && activityItem?.input) {
+    const isSwap =
+      activityItem.input.includes("0x38") || activityItem.input.includes("0x18")
```
henryboldi, line 9: Tip 44: Don't Program by Coincidence 😬

background/services/chain/index.ts
```diff
@@ -524,6 +525,39 @@ export default class ChainService extends BaseService<Events> {
...
+    const assetAddressInputPrefix =
+      "0000000000000000000000000000000000000000000002000000000000000000000000"
+    const { data } = result
+
+    // Heuristic for determining a swap
+    const isSwap = data?.includes("0x38") || data?.includes("0x18")
+
+    let asset = ETH
+    try {
+      let meta: any = false
```
henryboldi, line 538: 🚩

ui/components/Wallet/WalletActivityListItem.tsx
```diff
@@ -35,20 +37,40 @@ export default function WalletActivityListItem(props: Props): ReactElement {
...
+                symbol={activity.asset.symbol}
+                size="small"
+              />
             </div>
             <div className="amount">
               <span className="bold_amount_count">
-                {`${convertToEth(activity.value)}`.substring(0, 6)}
+                {`${
+                  Number(determineActivityValue(activity)) /
+                  10 ** activity.asset.decimals
```
henryboldi, line 50: Likely should move this decimal conversion to inside of `determineActivityValue`, another until function, or even the selector

henryboldi (commented):
- Too many violations of `Tip 44: Don't Program by Coincidence`
- Could move more logic into the selector or reducer

⚠️ Updates to address my self-review coming soon. Feel free to review as well now

background/services/chain/index.ts
```diff
@@ -540,7 +574,9 @@ export default class ChainService extends BaseService<Events> {
         // TODO make this multi network
         const result = await this.pollingProviders.ethereum.getTransaction(hash)
 
-        const tx = txFromEthersTx(result, ETH, getEthereumNetwork())
+        const asset = await this.determineAsset(result)
+
+        const tx = txFromEthersTx(result, asset, getEthereumNetwork())
```
Shadowfiend, line 579: Transaction asset is meant to be the base asset of the underlying network (this became clearer [after the type refactor](https://github.com/tallycash/tally-extension/blob/main/background/networks.ts#L95)), not any asset involved in the transaction! We're going to need to surface information about assets involved *in* the transaction differently.

That said, I don't know if that kind of enrichment belongs in the database vs in redux. Feels like we could do it in `adaptForUI`, where the `data` field gets adapted to contain the transaction payload + additional derived information.

ui/utils/transactionInputInterpretionHelper.ts
Shadowfiend: Hahaha yeah. So here's how we will actually know if this is a swap:

- The `to` address will be the address of the swap contract. This will be a constant when we release.
- The function data will be decodable using the swap ABI. Once that's included for the swap work, we can use the Ethers [parsing functions](https://docs.ethers.io/v5/api/utils/abi/interface/#Interface--parsing) to parse transaction data into function call data. The call will be to `swapByQuote`, and may require some additional translation from 0x terms to ours.

I think if we want to get into this in the current work, we should be thinking about how we scalably add different interpreters, rather than specifically interpreting some of the bits that aren't fully fleshed out yet. But it also feels out of scope for now, and I do lean towards dropping this entirely until we've got more of the swap stuff finished, then revisiting.

ui/utils/transactionInputInterpretionHelper.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    if (!isSwap) {
+      value = BigInt(parseInt(activityItem.input.slice(-64), 16))
+    } else {
+      value = BigInt(parseInt(activityItem.input.slice(10).slice(0, 64), 16))
```
Shadowfiend, line 14: In general you should never use `parse*` with `BigInt`, because it risks losing precision if the string represents a number greater than `Number.MAX_SAFE_INTEGER`. Converting to `BigInt` after the fact just wraps the lost precision into a format that implies we didn't lose the precision.

To avoid the `parseInt` hop, you can pass a string directly to `BigInt` to parse it, and prefix with `0x` to ensure `BigInt` parses the string as a hexadecimal number.

background/redux-slices/utils.ts
```diff
@@ -173,6 +174,11 @@ export type ActivityItem = AnyEVMTransaction & {
       valueDetail: string
     }
   }
+  asset: {
```
Shadowfiend, line 177: Above, should `isSent?` become `isSent` now that we're populating it consistently?

Shadowfiend (changes requested):
Few blockers here, left some notes!

background/redux-slices/utils.ts
henryboldi: Ah, totally! Thank you

background/redux-slices/utils/activity-utils.ts
```diff
@@ -0,0 +1,159 @@
...
+  blockHeight: number
+  infoRows: {
+    [name: string]: {
+      label: string
+      value: unknown
+      valueDetail: string
+    }
+  }
+  token: FungibleAsset
+  tokenDecimalValue: number
```
Shadowfiend, line 45: My concern with this structure is that it walks away from the existing way of capturing an asset + amount (`AssetAmount`), as well as the existing way of including decimal values alongside that information (`CompleteAssetAmount` from the accounts slice), which means any component that wants to render `ActivityItem`'s version of this information will have to be written to *only* work with `ActivityItem`'s version.

A good option here IMO is to use a name like `activityAssetAmount: CompleteAssetAmount`. This makes it clear that it's specific to the activity, and carries the `CompleteAssetAmount` information. Once #469 lands there'll be one more PR that adds the `enrich*` helpers that will make it trivial to turn an `AssetAmount` into a `CompleteAssetAmount`.

background/redux-slices/utils/activity-utils.ts
```diff
@@ -0,0 +1,159 @@
...
+}
+
+export function determineActivityDecimalValue(activityItem: ActivityItem) {
+  const { token } = activityItem
+  let { value } = activityItem
+
+  // Derive value from transaction transfer input if not ETH
+  if (token && !token.symbol.includes("ETH") && activityItem.input) {
+    value = BigInt(`0x${activityItem.input.slice(10).slice(0, 64)}`)
+  }
```
Shadowfiend, line 155: A transaction can transfer both ETH and a token—here we explicitly choose the token to display, but is that correct, or should we be treating such a transaction as a generic contract interaction instead of presenting partial information?

background/redux-slices/utils/activity-utils.ts
```diff
@@ -0,0 +1,159 @@
...
+  },
+}
+
+export async function determineToken(result: AnyEVMTransaction) {
+  const assetAddressInputPrefix =
+    "0000000000000000000000000000000000000000000002000000000000000000000000"
+  const { input } = result
+
+  // Heuristic for determining a swap
+  const isSwap = input?.includes("0x38") || input?.includes("0x18")
```
Shadowfiend, line 118: Just realized I didn't cover how this plays out in my original note, so I want to be clear that this implementation risks misidentifying non-swap transactions as swap transactions, since it's just pattern-matching against the first byte of a function selector, which has no uniqueness guarantees and, as the result of a hash function, doesn't really have any relevance guarantees either. In that scenario, the code below will pull the first swap asset address as, at worst, `0x` followed by the 70th-110th characters in the input (since indexOf returns -1 if the search string isn't found, and substring returns the whole string for negative start indices), which could be anything. Even though that will *most likely* not match a real token address, and Alchemy will *most likely* return no metadata in that scenario, you're absolutely right that a positive outcome like that is purely programming by coincidence. We should not use it.

This does bring up a design question though: is the expectation that this information will purely work for swaps using Tally swap, or that it will also appear for other swaps? If the latter, we'll need to do a lot of lifting to correctly handle the various swap possibilities–again pointing to focusing on setting up the framework for extra interpretation of transaction data rather than doing that interpretation right now.

background/redux-slices/utils/activity-utils.ts
```diff
@@ -0,0 +1,159 @@
...
+      }
+    } catch (err) {
+      logger.error(`Error getting token metadata`, err)
+    }
+  }
+
+  return asset
+}
+
+export function determineActivityDecimalValue(activityItem: ActivityItem) {
```
Shadowfiend, line 148: All these linter warnings are here for a reason :) With the exception of `adaptForUI`, we should try to fill in return types for the exported functions here.

There are some other ones above as well.

Shadowfiend (changes requested):
Love the introduction of `activity-utils` 🙌 

Still have some concerns about structure + implementation.

background/redux-slices/utils/activity-utils.ts
henryboldi: Is it catastrophic to risk misidentifying non-swap transactions as swap transactions? What is a rigorous way to identify a swap transaction? Couldn't this line be replaced with something more robust later on?

background/redux-slices/utils/activity-utils.ts
henryboldi: Sounds good!

background/redux-slices/utils/activity-utils.ts
henryboldi: How would you go about determining when a transfer is of both ETH and a token?

background/redux-slices/utils/activity-utils.ts
mhluongo: Big discussion I'm helicoptering into here, but tx introspection has been on my mind!

> Is it catastrophic to risk misidentifying non-swap transactions as swap transactions?

Maybe not catastrophic, but not great. This is people's money, and seeing the wrong thing (versus having a more generic "contract interfaction" category) can be pretty unsettling. Imagine you thought you just staked $1M in a token and your wallet tells you you swapped that for $0, for example.

> What is a rigorous way to identify a swap transaction?

In the general case, it's impossible to rigorously categorize whether an EVM transaction "is a swap" :) In our particular product though, I think what we want to do is break this idea into a few sub-problems

* Which transactions are swaps using the in-wallet swap? **For these, we should use local wallet data**
* Which transactions are swaps using a couple common protocols? **For these, you a function selector database + a contract address allowist**. This approach will catch swaps on Uniswap, Sushi, Curve, Saddle, etc.
* Which transactions are swaps using services we don't know? **This is where things become really interesting** from a product perspective, and impossible using pure engineering.

> Couldn't this line be replaced with something more robust later on?

Yes! But we want to keep `main` releasable, so this would need to be feature-flagged or made to work correctly for one of the above categories, I think? I'm sure there's some false positive % we're okay with, but I'm not sure what it is yet.

background/redux-slices/utils/activity-utils.ts
Shadowfiend: Any transaction with a nonzero `value` transmits ETH, irrespective of what else it does, basically.

mhluongo: I'd lean toward generic if it does both. When we include asset transfers we can make this much fancier

henryboldi: What would a generic asset look like in this case? @mhluongo

background/redux-slices/utils/activity-utils.ts
henryboldi: @Shadowfiend, which function(s) did you have in mind here from #469? I don't believe it includes something for `localizedDecimalAmount`. Anyway, for now, I pulled in a type for the assets to suggest the intent of consistency
```
tokenDecimalValue: AssetDecimalAmount["decimalAmount"]
```
<br/>

Alternatively, I could use `CompleteAssetAmount` with some missing values? But this screen doesn't display main currency values, so potentially `CompleteAssetAmount` isn't need

henryboldi:
I might have broke something here on the latest version 😬. Fix incoming

background/redux-slices/utils/activity-utils.ts
mhluongo: Oh, I meant a generic contract interaction, not a generic asset. You can imagine a tx type that's called "Contract interaction" with an icon that's a pencil or something, which has a link to Etherscan. Down the road, we break this into individual asset transfers ala Etherscan

Recent example from my own adventures. As you can see, this involved many many assets.. and while this one didn't have a base ETH value, if it were a pool with WETH in it it might have.

Screenshot from 2021-11-29 15-58-49

mhluongo: Unless we originate the transaction, we don't quite know what's useful to display. Eg in Etherscan's attempt, I'd argue it has a bunch of stuff most users don't care about in a wallet. That LP tokens were minted from `0x0` is irrelevant to me as the depositor... what's relevant is that I put in a bunch of different tokens and amounts, and got out a single LP token

henryboldi: Given the current design for a generic contract interaction,
contract

Would we drop the asset line, or include the ETH and ETH value?

mhluongo: Good question. We should def change "from" to "to" there... and I guess we can include the asset line with ETH initially, but we should switch to having a name for the function call rather than an asset sooner rather than later.

background/redux-slices/utils/activity-utils.ts
Shadowfiend: See #490, but not going to block this PR on this. We can follow up with it.

background/redux-slices/utils/activity-utils.ts
Shadowfiend: Matt basically said everything I was going to lol, sorry for not getting back sooner 🙆

background/main.ts
```diff
@@ -258,8 +259,15 @@ export default class Main extends BaseService<never> {
...
       this.store.dispatch(updateAccountBalance(accountWithBalance))
     })
-    this.chainService.emitter.on("transaction", (payload) => {
+    this.chainService.emitter.on("transaction", async (payload) => {
       const { transaction } = payload
+      const enrichedPayload = {
+        ...payload,
+        transaction: {
+          ...transaction,
+          token: await determineToken(transaction),
```
Shadowfiend, line 268: Not blocking, but I think we should do this in an async thunk in the activities slice. Main should be almost exclusively wiring between services and slices, and stuff like data enrichment should really stay in the slices.

Think “where would I do this if there was no main”, basically.

background/redux-slices/utils/activity-utils.ts
```diff
@@ -0,0 +1,148 @@
...
+export async function determineToken(
+  result: AnyEVMTransaction
+): Promise<FungibleAsset | SmartContractFungibleAsset | null> {
+  const { input } = result
+  let asset = ETH
+  if (input) {
+    try {
+      let meta: SmartContractFungibleAsset | null = null
+      if (result?.to) {
+        meta = await getTokenMetadata(pollingProviders.ethereum, result.to)
```
Shadowfiend, line 122: Not blocking, but it should feel like an antipattern to directly interact with any external service from the slices if we are also interacting with that service in the services.

In this particular case, we should lean on our assets slice to help us find the token information in our own data format—remember that the assets slice sees every asset the indexing service sees, and every asset in the assets slice that is a `SmartContractFungibleAsset` knows its contract address, so there's no need to go to a server to pull that information every time we see a transaction.

Shadowfiend (approved):
Couple of remaining notes on how to structure this overall, but I think we're at “good enough to ship the current thing and iterate”.

Issue #411, opened 2021-11-11
Resolve ENS names to addresses on Ethereum


mhluongo:
- [x] Users should be able to add read-only wallets by specifiying an ENS name
- [x] Users should be able to send ETH and ERC-20s to an ENS name
  - Names should be resolved before the user can send, helping prevent mistakes
  - Resolution should include name avatars in the UI to help avoid mistakes

Issue #225, opened 2021-09-12
Name / content resolution service


mhluongo:
We need to be able to resolve...

- [ ] IPFS content
- [x] ENS names to
  - [x] #411
  - [x] Addresses on other networks
  - [ ] #412 
  - [x] #413
- [x] #414
- [x] Unstoppable Domain names to
  - [x] Addresses on Ethereum
  - [x] Addresses on other networks
  - [ ] Websites / dApps
- [ ] Handshake/HNS names to websites

... in rough priority order

PR #415, opened 2021-11-11
📛 The Names of Things: Look up ENS names and avatars for new wallets

### LINKED ISSUE #12: NFTs make good account avatars
Would be interesting to use HashMasks, CryptoPunks, and ENS names to start

### LINKED ISSUE #414: Resolve tracked Ethereum addresses to ENS names & avatars
Refs #12

mhluongo:
An implementation of reverse ENS resolution, enabling ENS nicknames and avatars in the menu.

Includes

- [x] A basic name service (#225)
- [x] Reverse name and avatar resolution when 

Doesn't include

- A way to add a new read-only wallet by ENS name or to send funds to an ENS name (#411)
- Any sort of "address book" functionality

Screenshot from 2021-11-11 18-36-57

Screenshot from 2021-11-11 18-17-45

Closes #12
Closes #414

### DIFF
--- a/background/constants/index.ts
+++ b/background/constants/index.ts
@@ -18,5 +18,10 @@ export const ALARMS = {
   },
 }
 
+export const SECOND = 1000
+export const MINUTE = 60 * SECOND
+export const HOUR = 60 * MINUTE
+export const DAY = 24 * HOUR
+
 export * from "./currencies"
 export * from "./networks"
--- a/background/lib/erc721.ts
+++ b/background/lib/erc721.ts
@@ -1,0 +1,77 @@
+import { Contract } from "ethers"
+import { BaseProvider } from "@ethersproject/providers"
+import { fetchJson } from "@ethersproject/web"
+
+import logger from "./logger"
+import { HexString, URI } from "../types"
+import { jtdValidatorFor } from "./validation"
+
+const abi = [
+  "function tokenURI(uint256 _tokenId) external view returns (string)",
+]
+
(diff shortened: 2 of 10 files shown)


henryboldi (changes requested):
Looks good! Two questions,

- Why is the new service called just `name`?
- Should this service be added to the architecture section of the readme?

mhluongo:
> Why is the new service called just `name`?

The thought here is that DNS is Domain Name Service, ENS is Ethereum Name Service, UNS is Unstoppable Name Service, etc etc, so I thought `NameService` was appropriate. We could also do something like `NameResolutionService` or `DomainAndAddressResolutionService`... which are perhaps easier to understand? Unfortunately none of these names captures metadata like avatars and other text records :shrug: 

ENS launched what it calls a ["metadata service"](https://metadata.ens.domains/), though that also seems limiting and has overlap with `IndexingService`.

> Should this service be added to the architecture section of the readme?

I'd actually like to replace the architecture diagram with something a bit denser than ASCII art... but felt like it was outside the scope of this work lol. @Shadowfiend if you have a link to regenerate that diagram I can make a point to keep it updated

henryboldi (approved):

Shadowfiend:
I may well not have a great way to regenerate, will poke around and see. More likely we want to think about a different way to maintain it, or we want to keep it higher level rather than trying to describe all service interrelationships in diagram form in a set of primarily-text docs.

PR #425, opened 2021-11-16
Add contract utility functions and distributor abi

wojciech-turek:
This PR introduces basic contract utility functions like `getAlchemyProvider` and `getContract`, as well as current distributor abi for future interactions.


background/redux-slices/contract-abis/merkle-distributor.ts
```diff
@@ -0,0 +1,118 @@
...
+      {
+        internalType: "address",
+        name: "",
+        type: "address",
+      },
+    ],
+    stateMutability: "view",
+    type: "function",
+  },
+]
```
itsrachelfish, line 116: How was this generated? :thinking:

wojciech-turek: The abi is generated after compiling a smart contract with a tool like `hardhat`.

background/redux-slices/utils/contract-utils.ts
```diff
@@ -0,0 +1,19 @@
+import { AlchemyProvider, getNetwork } from "@ethersproject/providers"
+import { Contract, ethers } from "ethers"
+import { getEthereumNetwork } from "../../lib/utils"
+
+export function getAlchemyProvider(): AlchemyProvider {
```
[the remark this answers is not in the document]
Shadowfiend: Nope, this is temporary.

I would rename this `getProvider` and have it use `Provider` or `JsonRpcProvider` as the return type (even if it internally instantiates an `AlchemyProvider`).

background/redux-slices/contract-abis/merkle-distributor.ts
Shadowfiend: @wojciech-turek I think maybe the question here is more precisely “where did you get this from and how did you generate it”, rather than “how would one go about generating this kind of thing in general” :)

background/redux-slices/utils/contract-utils.ts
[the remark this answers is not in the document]
wojciech-turek: "This looks like a job for me" - Eminem - Micheal Scott

background/redux-slices/contract-abis/merkle-distributor.ts
wojciech-turek: Whopps sorry misunderstood, in order to create the abis you need to go to our contracts repo [here](https://github.com/tallycash/tally-contracts) and setup it locally. After running `npm install` you can run `npx hardhat compile` that will generate `artifacts` directory in which you can find compiled `contracts`. The result of a compilation is a .json file which contains amoung others two crucial fields `abi` and `bytecode`.

Shadowfiend: One small note: you should be using yarn instead of npm or npx in our repos. Not a huge deal for this, but for consistency with repo setup, lockfiles, etc, that's the way to go.

background/redux-slices/utils/contract-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+  const provider = new AlchemyProvider(
+    getNetwork(Number(getEthereumNetwork().chainID)),
+    process.env.ALCHEMY_KEY
+  )
+  return provider
+}
+
+export const getContract = async (
+  address: string,
+  abi: any[]
```
Shadowfiend, line 15: Can this be `ContractInterface` instead of `any[]`?

background/redux-slices/utils/contract-utils.ts
```diff
@@ -0,0 +1,19 @@
+import { Provider, AlchemyProvider, getNetwork } from "@ethersproject/providers"
+import { Contract, ethers } from "ethers"
+import { getEthereumNetwork } from "../../lib/utils"
+
+export function getProvider(): Provider {
+  const provider = new AlchemyProvider(
+    getNetwork(Number(getEthereumNetwork().chainID)),
+    process.env.ALCHEMY_KEY
+  )
+  return provider
```
Shadowfiend, line 10: Feels like we should only instantiate one provider and then return it, no? I.e. have a global `let` that starts `undefined` but gets populated the first time `getProvider` gets called?

Shadowfiend (changes requested):
Couple of additional notes here sorry, didn't get a chance to go deep yesterday.

background/redux-slices/utils/contract-utils.ts
wojciech-turek: That is an interesting question 🤔 , what about if the user changes network. We would need to add a detection mechanism. to re-do the provider. I have tried looking for any information whether a standalone provider caches anything but to no success. By global `let` do you mean we should store provider in redux?

background/redux-slices/utils/contract-utils.ts
wojciech-turek: Absolutely, very good comment.

background/redux-slices/utils/contract-utils.ts
Shadowfiend: Nah, directly in this file.

I think it's fine for now, since we don't know the final shape that the provider will take anyway.

background/redux-slices/utils/contract-utils.ts
```diff
@@ -12,7 +30,7 @@ export function getProvider(): Provider {
 
 export const getContract = async (
   address: string,
-  abi: any[]
+  abi: ContractInterface
```
Shadowfiend, line 33: Yikes, sorry—I meant the ethers type (`import { ContractInterface } from "ethers"`).

wojciech-turek: Changed :)

Shadowfiend (approved):
Lgtm, letting @itsrachelfish have the last look + merge privileges :grin:

background/redux-slices/contract-abis/merkle-distributor.ts
itsrachelfish: Thanks @wojciech-turek, I was able to replicate this output by running `yarn install` and `yarn hardhat compiled` :grinning: :+1:

itsrachelfish (approved):

Issue #420, opened 2021-11-12
Community PR tracking issue


Shadowfiend:
Tracking issue for any larger PR that is temporarily closed while the first release(s) go out the door.

PRs listed here will be revisited in Q1 2022 to consider how they can be slotted in, what updates might be needed, etc.

0xDaedalus:
Closed b/c stale and unused

PR #435, opened 2021-11-18
Add ENS validation and support for adding a read-only mode account

Umiiii:
image

Currently, we are only able to enter the Ethereum address(0x...) form in this preview mode, this patch added an ENS validation so that the ENS address can be also used.

### DIFF
--- a/ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
+++ b/ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
@@ -19,8 +19,12 @@ export default function OnboardingViewOnlyWallet(): ReactElement {
     return /^(0x){1}[0-9a-fA-F]{40}$/i.test(checkAddress)
   }
 
+  function checkIfPlausibleENSAddress(checkAddress: string) {
+    return /^[a-z0-9-]{1,63}(.eth)$/i.test(checkAddress)
+  }
+
   const handleSubmitViewOnlyAddress = useCallback(async () => {
-    if (checkIfPlausibleETHAddress(address)) {
+    if (checkIfPlausibleETHAddress(address) || checkIfPlausibleENSAddress(address)) {
       await dispatch(
         addAddressNetwork({
           address,
@@ -29,7 +33,7 @@ export default function OnboardingViewOnlyWallet(): ReactElement {
       )
       setRedirect(true)
     } else {
-      alert("Please enter a valid address")
+      alert("Please enter a valid address or ENS address")
     }
   }, [dispatch, address])
 


Shadowfiend:
Awesome touch @Umiiii, thanks for the contribution! I've been slapping around some ENS code so I'll give this a test drive today and see if we can land the code as is. 

In the meantime, would you mind updating the commit message you've used to clarify that the ENS support is specifically for the read only address form? Actually it could be the same as the PR description and would be 💯

Umiiii:
> Awesome touch @Umiiii, thanks for the contribution! I've been slapping around some ENS code so I'll give this a test drive today and see if we can land the code as is.
> 
> In the meantime, would you mind updating the commit message you've used to clarify that the ENS support is specifically for the read only address form? Actually it could be the same as the PR description and would be 💯

Hello, thanks for pointing out. I've already made a clearer PR description. Also, I tried your ENS implementation and it worked perfectly with this patch.

Umiiii:
> One more here, didn't see it when I was looking on mobile sorry—it looks like the first commit in the series isn't signed, could you also fix that up? I'm going to run the build here in the meantime 💪

Signed!

Shadowfiend:
Ok, poked at this locally a little—I think it may need a bit more support in the backend. Right now dropping an ENS address in will look up ETH balance correctly, but resolving other assets fails because we don't convert to an address before passing it on to certain APIs that expect an address. The logs for the extension list some of these errors.

Separately, there's also a UX question here of what happens when you put in a read-only ENS address—do we follow ENS if the address the name points to changes? What if you then promote the address to non-read-only?

I think if we resolved the ENS name to an address before starting to track it, we would get most of the way there… But that requires some cross-cutting changes to the redux slice (using something other than `AddressNetwork`), UI (renamed to match the slice), and the chain service (to `addAccountToTrack` to take things that are not necessarily addresses and attempt to resolve them to addresses before passing them on to the rest of the system).

If you're willing to take these changes on I can probably describe them in a little more detail; otherwise I would love to close this PR and track it in #420 so we can come back to it after the first release or two of the wallet.

Umiiii:
Totally agreed with you. So what can I do? Could you describe a little bit more about the details so that I can follow?

Shadowfiend:
Awesome! Will do in the morning Eastern :)

Shadowfiend:
Ok, so I think we can solve this in a short-term way by assuming the ENS name is resolved only once when you add the account, and if the ENS name updates the wallet sticks with the original address. I think it would be nice to follow the ENS name updates, but that would require a lot more thinking through of how we indicate whether an ENS name is being followed or whether it's the result of doing a reverse-lookup on the hex address, etc, as well as potentially thinking through a flow where the wallet sees the ENS name has been updated and prompts you to decide whether you want to follow the update.

With that in mind, here are the changes I think we need to make:
- Add a background async thunk to `background/redux-slices/accounts.ts`, `account/addAccountByName`, that is called if the field has an ENS-like object instead of an address; it should take the domain name + the network in the same way that `addAccount` takes the address + the network.
- In that thunk, emit an `addAccountByName` event (like `accounts/addAccount` emits `addAccount`).
- In `MainService.connectChainService` (`background/main.ts`), after the handling of `addAccount`, add a handler for `addAccountByName`. This handler should:
  - Call `this.nameService.lookUpEthereumAddress` with the passed account.
  - If the return value is not `undefined`, call `this.chainService.addAccountToTrack(addressNetwork)` with the result.
  - If the return value is undefined, or if there is an exception (try/catch), log an error (`logger.error(\`Failed to resolve new account by name: ${accountName}, ${network}.\`)` or so).

This last piece will lead to some weird behavior if the ENS name can't be resolved, but we don't have a consistent way to do back-and-forth error handling yet in the UI. That should be filling in soon, at which point we should be able to plug this into it.

ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
```diff
@@ -19,8 +19,12 @@ export default function OnboardingViewOnlyWallet(): ReactElement {
     return /^(0x){1}[0-9a-fA-F]{40}$/i.test(checkAddress)
   }
 
+  function checkIfPlausibleENSAddress(checkAddress: string) {
+    return /^[a-z0-9-]{1,63}(.eth)$/i.test(checkAddress)
+  }
+
   const handleSubmitViewOnlyAddress = useCallback(async () => {
-    if (checkIfPlausibleETHAddress(address)) {
+    if (checkIfPlausibleETHAddress(address) || checkIfPlausibleENSAddress(address)) {
```
henryboldi, line 27:
```suggestion
    if (
      checkIfPlausibleETHAddress(address) ||
      checkIfPlausibleENSAddress(address)
    ) {
```

This should resolve the linting error

Shadowfiend:
@Umiiii going to close this PR for now, but please feel free to reopen if you decide to move further on this! We'd be excited to add it to the wallet :muscle:

PR #437, opened 2021-11-19
Move the content script into manifest.development.json

greg-nagy:
The content script is not yet ready to be released so we want to keep it confined into development mode. 

For this separation we use modes to handle different setups in the manifest.json
https://github.com/tallycash/tally-extension/blob/main/webpack.config.ts#L153

This PR moves the content script from the main manifest.json file into manifest.development.json so it get's picked up in development mode, but won't be included when we release a new version to the chrome store.

### DIFF
--- a/manifest/manifest.development.json
+++ b/manifest/manifest.development.json
@@ -1,5 +1,13 @@
 {
   "background": {
     "scripts": ["dev-utils/extension-reload.js"]
-  }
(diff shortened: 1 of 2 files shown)


wojciech-turek (commented):
Looks good to me!

wojciech-turek (approved):

PR #440, opened 2021-11-19
🦁  ROR: Initial Window Provider / EIP-1193 implementation

greg-nagy:
https://en.wikipedia.org/wiki/Ether
> Ethers are a class of organic compounds that contain an ether group—an oxygen atom connected to two alkyl or aryl groups. They have the general formula R–O–R′, where R and R′ represent the alkyl or aryl groups.

CleanShot 2021-11-19 at 17 05 28

---

This implementation is based on the usages in onboard.js and web3-react demos.

Probably more details will surface and more robust solutions will come when we have a better integration with the background parts and we will have a real way to test this.

This implementation is NOT intended to work on it's own. It will need the `ProviderBridgeService` and the `InternalEthereumProviderService`.


src/window-provider.ts
```diff
@@ -7,42 +9,110 @@ const WINDOW_PROVIDER_TARGET = "tally-window-provider"
...
+      function listener(
+        this: TallyWindowProvider,
+        event: {
+          origin: string
+          source: unknown
+          data: { id: string; target: string; payload: { result: unknown } }
+        }
+      ) {
+        if (
+          event.origin !== unsafeOrigin || // we want to recieve messages only from the provider-bridge script
```
Shadowfiend, line 62: Can we change this comment to “filter to messages claiming to be from the provider-bridge script”? I think that makes it clearer that it's about where the message *claims* to come from rather than where it's actually from.

src/window-provider.ts
```diff
@@ -7,42 +9,110 @@ const WINDOW_PROVIDER_TARGET = "tally-window-provider"
...
+          event.source !== window || // we want to recieve messages only from the provider-bridge script
+          event.data.target !== WINDOW_PROVIDER_TARGET
+        ) {
+          return
+        }
```
Shadowfiend, line 67: I like early returns in general, but I wonder if this would be easier to understand if we inverted the condition to a series of conjunctive affirmatives instead of a series of disjunctive negatives. I always find the latter harder to reason about. What that might look like:

```suggestion
        // Only handle messages (claiming to be) from the provider bridge.
        if (
          // only messages (claiming to be) from the same origin
          event.origin === unsafeOrigin &&
          // only messages (claiming to be) from the same window
          event.source === window &&
          // only messages (claiming to be) for the window provider
          event.data.target === WINDOW_PROVIDER_TARGET
        ) {
          // ... the rest of the code ...
        }
```

src/window-provider.ts
```diff
@@ -7,42 +9,110 @@ const WINDOW_PROVIDER_TARGET = "tally-window-provider"
...
-    )
-      return
-
-    // to demonstrate how it works it was necessary. Will remove later
-    // eslint-disable-next-line no-console
-    console.log("inpage: ", event.data)
-  })
+    }
+
+    // ‼️ Always include target origin to avoid unwanted attention
```
Shadowfiend, line 49: This isn't unwanted attention, but rather accidentally triggering message handlers other than our own, right? That is, we can still get unwanted attention even if we include the unsafe origin.

src/window-provider.ts
```diff
@@ -7,42 +9,110 @@ const WINDOW_PROVIDER_TARGET = "tally-window-provider"
...
+          payloadMethod === "eth_accounts" &&
+          Array.isArray(result) &&
+          result.length !== 0
+        ) {
+          const [address] = result
+          if (this.selectedAddress !== address) {
+            this.selectedAddress = address
+            this.emit("accountsChanged", [this.selectedAddress])
+          }
+        }
```
Shadowfiend, line 104: Wondering if it's worth splitting out custom handlers by payload method. Something like, e.g.:

```typescript
  chainHandlers = {
    "eth_chainId": handleChainID.bind(this),
    "eth_accounts": handleAccounts.bind(this)
  }
  ...
      payloadHandlers[payloadMethod]?.(result)
  ...
  handleChainID(chainIDPayload: number): void {
    if (!this.isConnected) ...
  }

  handleAccounts(accountsPayload: unknown): void {
    if (Array.isArray(result) && result.length > 0) {
      ...
    }
  }
```

Not sure if there will be more of them/enough to justify, but if we anticipate having more, or frankly even in the current form because of how orthogonal they are to the job of the event handler, it feels like it might be a good refactor.

src/window-provider.ts
```diff
@@ -7,42 +9,110 @@ const WINDOW_PROVIDER_TARGET = "tally-window-provider"
...
+        const { result } = event.data.payload
+
+        if (payloadMethod === "eth_chainId") {
+          if (!this.isConnected) {
+            this.isConnected = true
+            this.emit("connect", { chainId: result })
+          }
+
+          if (this.chainId !== result) {
+            this.chainId = Number(result)
```
Shadowfiend, line 88: These can't both be true, in the sense that if we have to coerce `result` to a `Number`, then `!==` will not pass since it checks type *and* value. Unless what we want here is `if (typeof result === "number" && this.chainId != result)`, which would ensure we only deal with numbers here.

src/window-provider.ts
```diff
@@ -1,3 +1,5 @@
+import EventEmitter from "events"
```
Shadowfiend, line 1: Can we use `Emittery` here? Not sure if it's API-compatible in a way that will work with EIP-1193, but if it is then we can use typing information to make the event dispatch a little safer on our end.

src/window-provider.ts
```diff
@@ -7,42 +9,110 @@ const WINDOW_PROVIDER_TARGET = "tally-window-provider"
...
+            this.selectedAddress = address
+            this.emit("accountsChanged", [this.selectedAddress])
+          }
+        }
+
+        resolve(result)
+      }
+
+      this.bridgeListeners.set(sendData.id, listener.bind(this))
+      unsafeAddEventListener("message", this.bridgeListeners.get(sendData.id))
```
Shadowfiend, line 110: I believe you asked this when you were prototyping—now that we're out of prototyping, what do you think about using a single unified listener that uses the `event.data.id` field to index into a `requestResolvers: { [requestID: string]: (resolveValue: unknown)=>void }` map of Promise resolvers? In this model we would `unsafeAddEventListener` once in the constructor, then the returned `Promise` here would basically be a `return new Promise((resolve) => this.requestResolvers[sendData.id] = resolve)`.

I'm less concerned about resource usage here, but I think with this and the `chainHandlers` bit I suggested above, we end up with a few functions with smaller responsibilities that are a bit easier to follow.

src/window-provider.ts
```diff
@@ -7,42 +9,110 @@ const WINDOW_PROVIDER_TARGET = "tally-window-provider"
...
+    })
+  }
+}
+
+if (enabled === "true") {
+  // @ts-expect-error I don't really have any way to know the exact type of window.ethereum so it's better to expect error than lie
```
Shadowfiend, line 116: This is what we would do if we set Tally as the default wallet, right?

src/window-provider.ts
```diff
@@ -7,42 +9,110 @@ const WINDOW_PROVIDER_TARGET = "tally-window-provider"
...
+  // onboard.js relies on this, or uses a deprecated api. It seemed to be a reasonable workaround for now.
+  chainId: number | undefined = 1
+
+  selectedAddress: string | undefined
+
+  isConnected = false
+
+  isTally = true
+
+  bridgeListeners = new Map()
```
Shadowfiend, line 30: Why use a `Map` instead of a plain object here? IDs are strings, right?

Shadowfiend (approved):
Left a bunch of notes, but none are blocking per se given we're moving through this feature in an ongoing fashion.

That said, I'm going to wait and see if you want to restructure anything based on my comments before merging here @greg-nagy .

src/window-provider.ts
```diff
@@ -7,42 +9,110 @@ const WINDOW_PROVIDER_TARGET = "tally-window-provider"
...
-    // eslint-disable-next-line no-console
-    console.log("inpage: ", event.data)
-  })
+    }
+
+    // ‼️ Always include target origin to avoid unwanted attention
+    unsafePostMessage(sendData, unsafeOrigin)
+
+    return new Promise((resolve) => {
+      function listener(
```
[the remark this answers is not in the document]
greg-nagy: I rely on the closure containing the `resolve` function here. 

To have the same effect with an arrow function it would need to be wrapped when registering and passing in the `resolve` (and later the `reject`) explicitly.  `this.bridgeListeners.set(sendData.id, (event) => listener(evenr, resolve))`

I liked the bind solution more partially for personal preference, partially because it makes it possible to move the listener out of the Promise later. Might refactor later when I find the correct place for the listener.

src/window-provider.ts
greg-nagy: It's correct in that sense that it's just a narrowing down the options of a listener, and it's also a good practise to it like this.

src/window-provider.ts
greg-nagy: Thanks for bringing this up! It's a pain point for me :)

This will be refactored for sure, when I have better understanding of how this will look like. 
So I left it like this (and stopped myself mid refactor 3 times ... :D) for now.

I need to progress with the service part to have some real life examples before doing this refactor.

src/window-provider.ts
greg-nagy: Thanks, I was aiming for this, jsut did not come to mind and went with the first working solution.

greg-nagy: oh and :( 

CleanShot 2021-11-20 at 10 33 47

So I will leave this like here for now. What is the proper way to deal with situations like this? Eg linter forcing me to do things that does not make sense in given context?

src/window-provider.ts
greg-nagy: Would not use this in-page context. 
a) It's 44.1 kb 
b) Our codebase won't rely on the emmited events, and the dApps won't be able to use the typing information as I understand.

That being said, I am not sure if I fully understand your idea and what benefits you see in using `emmitery`.
Based on it's documentation it's main feature is async event emmit. I don't see how that is useful for us here.

src/window-provider.ts
greg-nagy: It's api was a better fit for the task.

src/window-provider.ts
greg-nagy: In essence yes. But this is just a really initial implemenation of that.

src/window-provider.ts
greg-nagy: Absolutely agree! Added TODOs for this and the other refactors too. 
Tried to do this but got stuck at some point and I felt that I need to understand the problem a bit more before adding this extra complexity.

src/window-provider.ts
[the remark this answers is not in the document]
Shadowfiend: In this structure I was suggesting keeping the placement of the listener function the same, but changing the syntax. That shouldn't change the access to the `resolve` value. Agree that waiting to see what the final form of the listener setup will be makes sense though.

src/window-provider.ts
Shadowfiend: `!==` is still right in the version I suggested, sorry. We can always discuss linterisms that seem wrong, but in this case we lose nothing by doing the stricter version 🤷

src/window-provider.ts
Shadowfiend: We do use types in that when we emit the event in our script it's type-checked... But I don't really want to pull another 44k in (how?!?), even if it is handled locally, esp for the very small typing win.

src/window-provider.ts
Shadowfiend: Yeah just wanted to make sure I had my understanding straight. Thanks!

src/window-provider.ts
Shadowfiend: How so?

If we do stick to a Map, let's give it the proper typing:

Shadowfiend (approved):
Looks good, let's do it 🚀

src/window-provider.ts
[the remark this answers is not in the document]
greg-nagy: My understanding about the arrow functinos were partial and did not understand what you meant by lexical `this` binding. I do now! Thank you!

PR #441, opened 2021-11-19
Delete legacy release workflow

Shadowfiend:
This workflow is no longer in use, and releases are cut directly from the build
workflow's artifacts uploaded on the main branch.


greg-nagy (approved):
CleanShot 2021-11-20 at 09 15 33

Issue #390, opened 2021-10-29
`InternalEthereumProviderService` tracking issue


Shadowfiend:
This issue will be the tracking issue for the `InternalEthereumService`.

The responsibility of this block is to provide a standardised facade for our internal background code and for the ethereum communication layer. 

This will be used by our internal dApps and the `ProviderBridgeService`.

greg-nagy:
This became stale and probably is for the best not to keep this tracking issue open.

Issue #389, opened 2021-10-29
`ProviderBridgeService` tracking issue


Shadowfiend:
This issue is to track the `ProviderBridgeService` related PRs. 

The responsibility of this block is to 
- set up communication with the content script
- handle permissions (request, show page, pending, allow, not allowed etc)
- validate the passing through communication
- integrate with the redux store
- provide a persistant Db solution to support the above functionality

greg-nagy:
This became stale and probably is for the best not to keep this tracking issue open.

PR #451, opened 2021-11-22
🔉Oddball: Initial ProviderBridgeService + RPC method list

### LINKED ISSUE #462: PBS - init setup (demos are working)

### LINKED ISSUE #463: IES - init setup

greg-nagy:
CleanShot 2021-11-22 at 13 43 58

https://en.wikipedia.org/wiki/Kelly%27s_Heroes
```
...
The Shermans fight their way through the German lines, destroying a railway depot in the process 
but the bridge they need to cross is blown up by Allied fighter-bombers. 

Oddball contacts an engineer unit to build him a bridge, 
and the engineers in turn bring in even more men for support.
...
```

This PR creates the `ProviderBridgeService` and connects it into the `background/main.ts` flow. 
There is a list of RPC methods in the form of `case` statements. This is intended as discussion material about dApp connectins. 

These methods will be moved into the `InternalEthereumProviderService` and this service will be only responsible for the communication validation (ajv + jtd schemas) and port creation.

The full list for API discovery [can be found here](https://github.com/tallycash/tally-extension/blob/a54920ee6a34b47e0cfac2204f6ad40add0e737b/provider-bridge/API_DISCOVERY.txt )

See #389, see #390 
Closes #462, closes #463


background/services/provider-bridge/index.ts
```diff
@@ -0,0 +1,179 @@
...
+  ServiceLifecycleEvents,
+} from ".."
+import logger from "../../lib/logger"
+import BaseService from "../base"
+
+type Events = ServiceLifecycleEvents & {
+  // connect
+  // disconnet
+  // account change
+  // networkchange
```
Shadowfiend, line 14: These events will be requests, right? e.g., when the content script requests connection, this service will emit a connection request. Similarly, if the content script requests an account or network change, this service will emit that request as an event?

background/services/provider-bridge/index.ts
```diff
@@ -0,0 +1,179 @@
...
+          ),
+        }
+        logger.log("background response:", response)
+
+        port.postMessage(response)
+      })
+    })
+  }
+
+  async router(method: string, params?: Array<unknown>): Promise<unknown> {
```
Shadowfiend, line 61: I would name this to make it clear what we're routing, e.g. `routeContentScriptRPCRequest` or something.

background/services/provider-bridge/index.ts
```diff
@@ -0,0 +1,179 @@
...
+      case "+wallet__highlighted":
+      case "+wallet__list":
+      case "+wallet_addethereumchain":
+      case "+wallet_default_hd_path":
+      case "+wallet_getpermissions":
+      case "+wallet_registeronboarding":
+      case "+wallet_route":
+      case "+wallet_type":
+      default:
+        break
```
Shadowfiend, line 176: Our default here, like in the EIP bridge, should be to forward to our provider.

The simplest (and least secure) way to make this work immediately is to forward to `this.chainService.pollingProviders.ethereum`. Since the bridge only exposed on dev builds, I think we can probably do that now.

Eventually these will directly flow through the DAppConnectionService, and the bridge service should only be responsible for validating JSON and blocking or permitting calls.

Shadowfiend (commented):
Couple more notes.

background/services/provider-bridge/index.ts
greg-nagy: I have moved these to the `InternalEthereumProviderService` because the eip-1193 clearly defines connected as `The Provider is said to be “connected” when it can service RPC requests to at least one chain.` and I was not aware of this when we talked about it. 

I am not married to this structure, just seemed to be logical.

background/services/provider-bridge/index.ts
greg-nagy: Updated the PR a bit with the `InternalEteherumProviderService` and put this code there.

greg-nagy:
@Shadowfiend This PR is ready now for another round. The scope of this PR changed a bit to set up the whole communication flow and have a workiing integration. 

I have also updated the api list with what alchemy provides and ran a new sanity check on the MM repo for the custom methods with a better workflow. 

Sorry for this PR getting a bit bigger than originally intended but could not complete the workin setup in parts
When I tried to split it I realised the process was error prone and realised that this might make sense to be in a single PR.

background/services/internal-ethereum-provider/index.ts
```diff
@@ -0,0 +1,115 @@
...
+type Events = ServiceLifecycleEvents & {
+  // connect
+  // disconnet
+  // account change
+  // networkchange
+}
+
+export default class InternalEthereumProviderService extends BaseService<Events> {
+  static create: ServiceCreatorFunction<
+    ServiceLifecycleEvents,
```
Shadowfiend, line 17:
```suggestion
    Events,
```

Just to make sure we don't lose the provider-specific events in the types and get confused later :)

background/services/internal-ethereum-provider/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    [Promise<ChainService>]
+  > = async (chainService) => {
+    return new this(await chainService)
+  }
+
+  private constructor(private chainService: ChainService) {
+    super()
+  }
+
+  async routeSafetRPCRequest(
```
Shadowfiend, line 28:
```suggestion
  async routeSafeRPCRequest(
```

background/services/internal-ethereum-provider/index.ts
```diff
@@ -0,0 +1,115 @@
...
+      case "eth_subscribe":
+      case "eth_syncing":
+      case "eth_uninstallFilter":
+      case "eth_unsubscribe":
+      case "net_listening":
+      case "net_version":
+      case "web3_clientVersion":
+      case "web3_sha3":
+        return this.chainService.pollingProviders.ethereum.send(method, params)
+      case "--- important wallet methods ---":
```
Shadowfiend, line 76: Feels like these should be comments.

background/services/internal-ethereum-provider/index.ts
```diff
@@ -0,0 +1,115 @@
...
+      case "metamask_logWeb3ShimUsage":
+      case "metamask_unlockStateChanged":
+      case "metamask_watchAsset":
+      case "net_peerCount":
+      case "wallet_accountsChanged":
+      case "wallet_addEthereumChain":
+      case "wallet_registerOnboarding":
+      case "wallet_switchEthereumChain":
+      default:
+        throw new Error(`unsupported method: ${method}`)
```
Shadowfiend, line 112: If we're routing a safe method, should we ever be the ones throwing the unsupported method error? In my mental model, we would explicitly enumerate the methods we want custom behavior for, and literally everything else would go through to the underlying provider, which may throw an error about not supporting the method.

In that model, if we wanted to gate certain methods as inaccessible/unsupported by Tally for external dApps, we would do it explicitly in the provider bridge and never hit the internal provider.

Thoughts?

background/services/provider-bridge/db.ts
```diff
@@ -0,0 +1,79 @@
+import Dexie from "dexie"
+
+/**
+ * An object describing a set of permissions granted to a given provider.
+ * Generally tracked by dApp as part of DAppPermissions.
```
Shadowfiend, line 5:
```suggestion
 * Generally tracked by dAppID as part of DAppPermissions.
```

Pretty sure I messed this one up in my PR haha.

background/services/provider-bridge/index.ts
```diff
@@ -0,0 +1,123 @@
...
+type Events = ServiceLifecycleEvents & {
+  accountAccessRequestedByDAppID: string
+}
+
+/**
+ * The ProviderBridgeService is responsible for the communication with the
+ * provider-bridge (content-script).
+ *
+ * The main purpose for this service/layer is to provide a transition
+ * between the untrusted communiction from the window-provider - which runs
```
Shadowfiend, line 21:
```suggestion
 * between the untrusted communication from the window-provider - which runs
```

background/services/provider-bridge/index.ts
```diff
@@ -0,0 +1,123 @@
...
+ * in user space and can be modifed by many attacks - and our internal service
+ * layer.
```
Shadowfiend, line 23:
```suggestion
 * in shared dapp space and can be modified by other extensions - and our
 * internal service layer.
```

Changing “user” to “dApp” because the mental model is a little different than the usual userspace/kernel space distinction. Does this seem like a fair rephrasing?

background/services/provider-bridge/index.ts
```diff
@@ -0,0 +1,123 @@
...
+ * - Provide connection interface - handle port communication, connect, disconnect etc
+ * - Validate the incoming communication and make sure that what we receive is what we expect
+ */
+export default class ProviderBridgeService extends BaseService<Events> {
+  // /**
+  //  * Tracks pending account requests by id and the function that can be used to
+  //  * respond to them when a response is provided via
+  //  * `respondToPendingAccountRequest`.
+  //  */
+  // #pendingAccountRequests: {
```
Shadowfiend, line 35: Just one thing I want to call out here because it wasn't obvious in the original PR and the commit and PR were WIP without proper messages/descriptions: I set this up as a true ECMAScript private variable because it feels more security-sensitive than most of the rest of the background script; in particular, it seems safest to guarantee that only the provider bridge can modify the handling of pending account requests. It's a small layer of protection given the complexity of the action, but I think it's a reasonable one nonetheless.

background/services/internal-ethereum-provider/index.ts
```diff
@@ -0,0 +1,115 @@
...
+      case "eth_sendRawTransaction":
+      case "eth_subscribe":
+      case "eth_syncing":
+      case "eth_uninstallFilter":
+      case "eth_unsubscribe":
+      case "net_listening":
+      case "net_version":
+      case "web3_clientVersion":
+      case "web3_sha3":
+        return this.chainService.pollingProviders.ethereum.send(method, params)
```
Shadowfiend, line 75: Let's not wreck the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) here once we move to a more final implementation. In particular, let's add a method to `chainService` that does the call to the Ethereum provider rather directly than reaching in through the providers collection.

Not blocking though.

background/services/provider-bridge/index.ts
```diff
@@ -0,0 +1,123 @@
...
+  //   [dAppID: string]: (grantedAccounts: string[]) => void
+  // } = {}
+
+  static create: ServiceCreatorFunction<
+    Events,
+    ProviderBridgeService,
+    [Promise<ChainService>, Promise<InternalEthereumProviderService>]
```
Shadowfiend, line 42: I can see this going two ways, but want to ask the question: should we call `ChainService` for the accounts information below, or should `InternalEthereumProviderService` do that on our behalf?

provider-bridge/index.ts
```diff
@@ -21,30 +24,28 @@ export function connectProviderBridge() {
...
 
-      port.postMessage({
-        message: `ping ${event.data.message}`,
-      })
+      port.postMessage(event.data)
     }
   })
 
-  port.onMessage.addListener((payload) => {
+  port.onMessage.addListener((data) => {
```
Shadowfiend, line 31: What prompted this change, out of curiosity?

Shadowfiend (approved):
I'm tentatively approving here but not merging as I've left some typo and other small fixes/notes.

@greg-nagy let me know if you want me to merge this because you've applied changes in a forthcoming branch, and I'll do it in the morning. Meanwhile feel free to open a follow-up PR with this branch as base, and GitHub will retarget it when this PR merges.

background/services/internal-ethereum-provider/index.ts
greg-nagy: Wanted to keep the separator as a line to strucutre visually better and if I do that linter slaps me with a `no-fallthrough`.

Will move it to line comments

background/services/internal-ethereum-provider/index.ts
greg-nagy: Thanks 😌

background/services/internal-ethereum-provider/index.ts
greg-nagy: In the end this will be the outcome and I will remove this from here. But for now I would like to keep them here. All the currently unsupported methods in the same place so I have a clear list that I need to reduce to 0 and then remove the error.

Right now there are 3 kind of methods that are listed here. 
- Some of them will be implemented here eg `eth_sign`
- Some of them will be moved to the `ProvidersBridgeSerivce` and throw unsupported error (with the proper error format and message) 
- Some of them will be implemented in provider bridge (probably `metmask_sendDomainMetaData`)

At the moment I did not want to come up with the exact plan to be able to categorise them, but wanted to progress with the implementation and move them as I come across them. The reason for this is for example `eth_personalSign` probably should be implemented here, it's used by the `web3-sample-official` demo app, but I think we won't implement it until later, because it does not feel to be an important feature for now. But I am not sure.

background/services/provider-bridge/index.ts
greg-nagy: Yes, I noticed, I was like 
`👀  Does it make any sense, because I am sure we compile it to something else` -> 
-> I looked into our compile target and checked the availability of this feature. ->
-> `Damn, this is good 😎`

background/services/internal-ethereum-provider/index.ts
greg-nagy: I was, until somehow the `sendEthereumProvider()` method dissapeared from the chain service. :)
Not sure what happened though. 

Totally agree.

background/services/provider-bridge/index.ts
greg-nagy: My plan was to keep the permission related things totally out of `InternalEteherumProviderService` and handle those concerns here. This was the reason for this implemenatation first. 

Now that I understand how the internal dApp will communicate I have moved the account logic into the `InternalEthereumProviderService`. 

But I am not sure we can totally get by here without the `ChainService` though.  But that's for the future to decide. Now, we don't need it here, so removed it.

provider-bridge/index.ts
greg-nagy: Tried to unify nameing throughout the communication flow, and it came from the `event.data` api used in the `window-provider`. I am still not happy with the api/nameing though, but don't see yet what would be an optimal solution.

greg-nagy:
@Shadowfiend Fixed the notes in this PR, please review it again and merge it if it's good.

background/main.ts
```diff
@@ -200,7 +210,17 @@ export default class Main extends BaseService<never> {
...
-    private nameService: NameService
+    private nameService: NameService,
+    /**
+     * A promise to the internal ethereum provider service, which acts as
+     * web3 / ethereum provider for the internal and external dApps to use.
+     */
+    private internalEthereumProviderService: InternalEthereumProviderService,
+    /**
+     * A promise to the provider bridge service, handling and validating
+     * the communication coming from dApps according to EIP-1193 and some tribal knowledge
```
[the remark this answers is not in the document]
Shadowfiend: We still missed the period here haha. I know, I'm a monster 😅

background/services/internal-ethereum-provider/index.ts
Shadowfiend: Ah that's rough. Did block comments also not work? Might be okay with disabling the linter rule once the implementation is filled out, since that seems like a bug.

background/services/chain/index.ts
```diff
@@ -382,6 +382,10 @@ export default class ChainService extends BaseService<Events> {
     }
   }
 
+  async send(method: string, params: unknown[]) {
```
Shadowfiend, line 385: This is going to need a more descriptive name as `ChainService` is currently envisioned to handle multiple chains---but no need to block on it for now.

Shadowfiend (approved):
Let's go 🚀

background/main.ts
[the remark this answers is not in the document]
greg-nagy: :lolsob: :lovehate: :mostlygrateful: 

https://github.com/tallycash/tally-extension/pull/477/commits/04d877ccaf6f65f841055fbcb737e7741730f7d3

background/services/internal-ethereum-provider/index.ts
greg-nagy: Did not check that, but probably not that important because my goal is to reduce this list to 0.

background/services/chain/index.ts
greg-nagy: I was wondering how to name this and went with a nameing based on the other methods. My guess is that we will refactor these methods when we want to be multichain and then we will have a better understandind of the constraints/needs.

background/main.ts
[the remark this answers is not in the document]
greg-nagy: Just realised that the PR title in which this was fixed fits this "issue" like a glove.

PR #471, opened 2021-11-24
Signing transactions

wojciech-turek:
The Send page now allows users to select a network fee from the available options coming from blocknative and select a token amount and destination address to create a partial transaction request.

Transaction-construction was updated to accept sendTransactionRequest coming from the Send component and triggers updateOptions function in main.ts that loads other data like gasLimit and nonce to create a full transaction request object and saves it in redux store. 

That transaction request can be then used on the Sign page where user can click Sign button to trigger keyRingService signTransaction function and have the signedTx broadcasted via chainservice.


background/services/chain/index.ts
```diff
@@ -375,6 +375,12 @@ export default class ChainService extends BaseService<Events> {
       const blockPrices = await this.blocknative?.getBlockPrices()
       this.emitter.emit("blockPrices", blockPrices)
 
+      // TODO set new alarm with new periodInMinutes?
+      // const pollBlockPrices = () => {
+      //   this.pollBlockPrices()
+      // }
+      // browser.alarms.onAlarm.addListener(pollBlockPrices)
```
Shadowfiend, line 382: You can do this with the `super` call in the constructor[^1], just add an entry there that calls this function. This function can then focus on just doing the check instead of scheduling, and can be renamed to reflect that it's doing the action rather than setting up the polling. 

Also looked a little deeper and saw that `Main` is the one that currently sets this off. `internalStartService` here is the one that should call this polling function once to get the party started instead, at which point Main's call can be removed. 

[^1]: https://github.com/tallycash/tally-extension/blob/19ad0b15cf5b99d2af0671805b5382e5b557b59b/background/services/chain/index.ts#L131-L149

ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -231,10 +222,11 @@ export default function SharedAssetInput(
             )}
             <input
               className="input_amount"
-              type="text"
+              type="number"
+              step="any"
               placeholder="0.0"
               value={amount}
-              onChange={assetAmountChanged}
+              onChange={(e) => onAmountChanged(e.target.value)}
```
Shadowfiend, line 229: Why not `useCallback`?

ui/components/Shared/SharedSlideUpMenu.tsx
```diff
@@ -4,17 +4,20 @@ interface Props {
...
-  const { isOpen, close, size, children } = props
+  const { isOpen, close, size, children, customSize } = props
 
   let menuHeight = "536px"
   if (size === "large") {
     menuHeight = "600px"
   } else if (size === "small") {
     menuHeight = "268px"
+  } else if (size === "custom") {
+    menuHeight = customSize || "600px"
```
Shadowfiend, line 20: Not blocking, but are these real sizes or are they all just "be tall enough to contain your contents"? If the latter, can we let the browser do the math for us somehow?

ui/pages/Send.tsx
```diff
@@ -84,6 +294,30 @@ export default function Send(): ReactElement {
...
           }
+          .networkfee {
+            display: flex;
+            justify-content: space-between;
+            font-size: 14px;
+            line-height: 16px;
+            color: var(--green-40);
+            margin-bottom: 12px;
+          }
+          .networkfee__settings {
```
Shadowfiend, line 305: I don't think we've been using this kind of class name structure fwiw. We *have* been using underscores for eg network_feee.

Shadowfiend (commented):
Quick thoughts!

ui/components/Shared/SharedSlideUpMenu.tsx
henryboldi: cc @VladUXUI

wojciech-turek: We could add "auto" for the height but we would need to change the animation, currently it breaks.

ui/pages/Send.tsx
wojciech-turek: Understood, I will correct it and follow same in the future.

background/services/chain/index.ts
wojciech-turek: I think I got it, will be in my new commit :)

ui/components/Shared/SharedAssetInput.tsx
wojciech-turek: I think there is no benefit of using `useCallback` here, the input of the address onChange function does not use it as well, so the second reason was consistency.

ui/pages/Send.tsx
Shadowfiend: Worth noting here that styled-jsx ensures the selectors are scoped to their containing component, which also reduces the arguments for some of the heavier CSS naming strategies as well.

background/redux-slices/transaction-construction.ts
```diff
@@ -1,58 +1,105 @@
...
+  updateOptions: Partial<EIP1559TransactionRequest>
+  gasEstimates: BlockPrices
+  signRequest: EIP1559TransactionRequest
+}
+
+export const emitter = new Emittery<Events>()
+
+// Async thunk to pass transaction options from the store to the background via an event
+export const updateTransactionOptions = createBackgroundAsyncThunk(
+  "transaction/options",
```
Shadowfiend, line 34: Let's make this `update-options`. I would like to stay closely aligned between the action name and thunk variable.

Similarly, we should update the thunk prefix to be `transaction-construction` rather than just `transaction`.

background/redux-slices/transaction-construction.ts
```diff
@@ -1,58 +1,105 @@
...
+export const { transactionOptions, signed, gasEstimates } =
+  transactionSlice.actions
 
 export default transactionSlice.reducer
 
-export type Events = {
-  updateOptions: Partial<EIP1559TransactionRequest>
-  gasEstimates: BlockPrices
-}
+export const selectGasEstimates = createSelector(
```
Shadowfiend, line 83: One thing I'll flag is that we should probably update all our stuff about “gas estimates” to specifically say “gas price estimates” or “per gas fee estimates”, because transactions care both about *gas* (the compute units expended, for which we can set the `gasLimit`) and the gas *price* (expressed as `*FeePerGas` in EIP1559 and `gasPrice` in legacy transactions).

I would suggest we explore a naming that follows EIP1559, where we do something like `estimateFeePerGas`, `estimatedFeePerGas`, `selectEstimatedFeePerGas`, etc.

background/services/chain/index.ts
```diff
@@ -145,6 +145,14 @@ export default class ChainService extends BaseService<Events> {
           this.handleHistoricAssetTransferAlarm()
         },
       },
+      blockPrices: {
+        schedule: {
+          periodInMinutes: 2,
```
Shadowfiend, line 150: We may still want to use `process.env.BLOCKNATIVE_POLLING_FREQUENCY` to allow for customization here, though it needs to be in minutes.

background/services/chain/index.ts
```diff
@@ -145,6 +145,14 @@ export default class ChainService extends BaseService<Events> {
...
+        },
+        handler: () => {
+          this.handleBlockPricesAlarm()
```
Shadowfiend, line 153: This can directly call `this.pollBlockPrices()` rather than indirecting through a handler function, IMO. I think the handler functions only make sense if there's some additional work to do specific to the alarm, but here we're just calling the method.

ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -231,10 +222,11 @@ export default function SharedAssetInput(
             )}
             <input
               className="input_amount"
-              type="text"
+              type="number"
```
Shadowfiend, line 225: Bravissimo!

ui/components/Shared/SharedSlideUpMenu.tsx
```diff
@@ -4,24 +4,27 @@ interface Props {
...
   }
 
   return (
     <div
-      className={`slide_up_menu
-        ${size === "large" ? " large" : ""}
-        ${!isOpen ? " closed" : ""}`}
+      className={`slide_up_menu ${size === "large" ? " large" : ""}${
+        !isOpen ? " closed" : ""
+      }`}
```
Shadowfiend, line 27: Let's use `classNames` here:
```suggestion
      className={classNames("slide_up_menu", { large: size === "large", closed: !isOpen })`}
```

(Plus run `prettier` :grin:)

ui/components/Wallet/WalletAccountBalanceControl.tsx
```diff
@@ -48,7 +48,8 @@ export default function WalletAccountBalanceControl(
     (state) => state.keyrings.importing
   )
   const [openReceiveMenu, setOpenReceiveMenu] = useState(false)
-  const hasSavedSeed = window.localStorage.getItem("hasSavedSeed")
+  // ! REMOVE ME - for testing
+  const hasSavedSeed = true
```
Shadowfiend, line 52: :wave:

In general I would suggest pulling this from an env variable OR using the build mode (`process.env.NODE_ENV === "development"`) if you want to approach something in an interim way, otherwise we risk putting something in a public build that we don't want.

Another approach, which is still error-prone, is to make the change but never commit it (by using `git add`/`git add -i`/`git add -p` to ensure only other changes land).

ui/pages/SignTransaction.tsx
```diff
@@ -1,5 +1,12 @@
 import React, { ReactElement, useState } from "react"
 import { useHistory, useLocation } from "react-router-dom"
+import { useDispatch, useSelector } from "react-redux"
```
Shadowfiend, line 3: You should pretty much always be using `useBackgroundDispatch` and `useBackgroundSelector` in the UI, as they enrich the pipeline with typing from our slices.

ui/pages/SignTransaction.tsx
```diff
@@ -54,6 +65,20 @@ export default function SignTransaction(): ReactElement {
     },
   }
 
+  const handleConfirm = async () => {
+    if (
+      signContent[signType].confirmButtonText === "Sign" &&
```
Shadowfiend, line 70: Hmmm… Why are we keying off of the button text here instead of just the `SignType`?

ui/pages/Send.tsx
```diff
@@ -1,25 +1,305 @@
...
-import SharedNetworkFeeGroup from "../components/Shared/SharedNetworkFeeGroup"
+import SharedInput from "../components/Shared/SharedInput"
+import SharedSlideUpMenu from "../components/Shared/SharedSlideUpMenu"
+import { useBackgroundSelector } from "../hooks"
 
 interface SendLocationState {
-  token: string
+  token: {
+    name: string
+  }
```
Shadowfiend, line 20: We should really be using the term `asset`, and `name` here is actually `symbol`. This should align better with naming across the stack.

`ETH` is not a token, for example :)

If we're only tracking the symbol, it feels reasonable to just have `assetSymbol` rather than an object structure with one property though?

ui/pages/Send.tsx
```diff
@@ -1,25 +1,305 @@
...
+}
+interface GasOption {
```
Shadowfiend, line 22: This is something we've been sluggish to do but are finally zeroing in on: prefer using `type` over `interface`:

```suggestion
type GasOption = {
```

ui/pages/Send.tsx
```diff
@@ -1,25 +1,305 @@
...
+            background: #002522;
+            display: flex;
+            box-sizing: border-box;
+            justify-content: flex-end;
+            padding: 20px 16px;
+          }
+        `}
+      </style>
+    </div>
+  )
```
Shadowfiend, line 290: What do you think about pulling this whole thing out to its own component that emits a single event, e.g. `onPerGasFeeSelected`, with the appropriate fee information? The `Send` component would directly track `maxFeePerGas` and `maxPriorityFeePerGas` in that scenario, and that's what would get updated with `onPerGasFeeSelected`.

ui/pages/Send.tsx
```diff
@@ -45,29 +324,59 @@ export default function Send(): ReactElement {
...
+              </button>
+            </div>
             <div className="divider" />
             <div className="total_footer standard_width_padded">
               <div className="total_amount">
                 <div className="total_label">Total</div>
-                <div className="total_amount_number">{amount}</div>
+                <div className="total_amount_number">{`${
+                  amount || 0
+                } ${token}`}</div>
```
Shadowfiend, line 357: Is this denominated in the asset being transferred, or is it the total main currency (= USD) spend from sending the token + fees?

ui/pages/Send.tsx
```diff
@@ -84,6 +393,30 @@ export default function Send(): ReactElement {
...
           }
+          .network_fee {
+            display: flex;
+            justify-content: space-between;
+            font-size: 14px;
+            line-height: 16px;
+            color: var(--green-40);
+            margin-bottom: 12px;
+          }
+          .network_fee__settings {
```
Shadowfiend, line 404: Again, we've been relying on the styled-jsx scoping to keep our selectors simple here. Let's use `.network_fee_settings`, `.network_fee > .settings` or just `.settings`, as makes the most sense.

Shadowfiend (changes requested):
Left a lot of notes, but I think the only one that can't be handled in a new PR is the `hasSavedSeed` override, which we need to fix before merging to `main`.

background/main.ts
```diff
@@ -338,20 +339,29 @@ export default class Main extends BaseService<never> {
           type: 2 as const,
           chainID: "1",
           nonce: resolvedNonce,
-          gasPrice:
-            await this.chainService.pollingProviders.ethereum.getGasPrice(),
         }
 
         // We need to convert the transaction to a EIP1559TransactionRequest before we can estimate the gas limit
```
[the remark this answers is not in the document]
wojciech-turek: Its gone

background/main.ts
```diff
@@ -338,20 +339,29 @@ export default class Main extends BaseService<never> {
...
           transaction
         )
-
-        await this.keyringService.signTransaction(options.from, transaction)
+        this.store.dispatch(transactionOptions(transaction))
       }
     })
 
+    transactionSliceEmitter.on(
+      "signRequest",
```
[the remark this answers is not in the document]
wojciech-turek: Changed as suggested.

background/redux-slices/transaction-construction.ts
```diff
@@ -1,58 +1,105 @@
...
+  BlockPrices,
+  EIP1559TransactionRequest,
+  SignedEVMTransaction,
+} from "../networks"
 import { createBackgroundAsyncThunk } from "./utils"
 
 type TransactionConstruction = {
-  transactionRequest: Partial<EIP1559TransactionRequest>
+  status: string
+  signedTx: Partial<SignedEVMTransaction>
```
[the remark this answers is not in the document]
wojciech-turek: Correct and I am not sure if we need to keep the signed transaction in the state at all. I think not, I will remove it.

background/redux-slices/transaction-construction.ts
```diff
@@ -1,58 +1,105 @@
...
-    maxPriorityFeePerGas: BigInt(21000),
-  },
+  status: "idle",
+  signedTx: {},
   gasEstimates: null,
 }
 
+export type Events = {
+  updateOptions: Partial<EIP1559TransactionRequest>
+  gasEstimates: BlockPrices
```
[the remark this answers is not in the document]
wojciech-turek: We're not, got rid of it.

background/redux-slices/transaction-construction.ts
wojciech-turek: 100%

background/redux-slices/transaction-construction.ts
wojciech-turek: I have changed that but since we get multiple options I have uses plural form of fee, thought its even more descriptive.

background/redux-slices/transaction-construction.ts
```diff
@@ -1,58 +1,105 @@
...
+)
+
 const transactionSlice = createSlice({
   name: "transaction-construction",
   initialState,
   reducers: {
     transactionOptions: (
       immerState,
-      { payload: options }: { payload: Partial<EIP1559TransactionRequest> }
+      { payload: options }: { payload: EIP1559TransactionRequest }
```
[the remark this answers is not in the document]
wojciech-turek: Absolutely

background/redux-slices/transaction-construction.ts
```diff
@@ -1,58 +1,105 @@
...
+export const selectIsTransactionLoaded = createSelector(
+  (state: { transactionConstruction: TransactionConstruction }) =>
+    state.transactionConstruction.status,
+  (status) => status === "loaded"
+)
+
+export const selectIsTransactionSigned = createSelector(
+  (state: { transactionConstruction: TransactionConstruction }) =>
+    state.transactionConstruction.status,
+  (status) => status === "signed"
```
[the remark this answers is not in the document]
wojciech-turek: I find it very clear that I fetch a true/false value from redux rather than getting a status and then comparing inside the component if it matches desired state. If we use `selectIsTransactionLoaded ` in more than one place in the future we reduce the code size. My vote is on keeping it this way.

background/services/chain/index.ts
wojciech-turek: Oh, very good catch. Sometimes I blindly follow what was used before and miss obvious choices I'd normally make. 🙇🏻

ui/components/Shared/SharedSlideUpMenu.tsx
wojciech-turek: Oh, neat, I was not using `classNames` before!

ui/pages/SignTransaction.tsx
wojciech-turek: Copy that

ui/pages/SignTransaction.tsx
wojciech-turek: Hmmm good question, changed to `SignType.Sign === signType`

ui/pages/SignTransaction.tsx
```diff
@@ -83,7 +108,12 @@ export default function SignTransaction(): ReactElement {
         >
           Reject
         </SharedButton>
-        <SharedButton type="primary" iconSize="large" size="large">
+        <SharedButton
+          type="primary"
+          iconSize="large"
+          size="large"
+          onClick={handleConfirm}
```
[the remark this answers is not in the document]
wojciech-turek: Agree, once we have the loader designed from @VladUXUI we can base the loading status based on status from transaction-construction slice.

ui/pages/Send.tsx
wojciech-turek: I have moved that out in another PR but without setting an event, rather than that I just pass required items and use local state for handling selected gas. Works well but we can talk about a restructure if needed.

ui/pages/Send.tsx
```diff
@@ -45,29 +324,59 @@ export default function Send(): ReactElement {
...
+                type="button"
+                onClick={openSelectFeeModal}
+                style={{
+                  background: `linear-gradient(90deg, var(--green-80) ${(
+                    ((selectedGas?.gwei || minGas) / maxGas) *
+                    100
+                  ).toFixed()}%, rgba(0, 0, 0, 0) ${(
+                    ((selectedGas?.gwei || minGas) / maxGas) *
+                    100
+                  ).toFixed()}%)`,
```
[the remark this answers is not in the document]
wojciech-turek: I have made an assumption, that its a kind of a progress bar to show the current gas and max in a visual way, @VladUXUI please let me know how do you see it work.

ui/pages/Send.tsx
```diff
@@ -45,29 +324,59 @@ export default function Send(): ReactElement {
...
+                />
+              </button>
+            </div>
             <div className="divider" />
             <div className="total_footer standard_width_padded">
               <div className="total_amount">
                 <div className="total_label">Total</div>
-                <div className="total_amount_number">{amount}</div>
+                <div className="total_amount_number">{`${
+                  amount || 0
```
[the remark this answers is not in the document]
wojciech-turek: I would incline towards leaving it as `||` as if amount ever happens to be an empty string I would prefer it to show 0 rather than nothing. What do you think?

ui/pages/Send.tsx
wojciech-turek: Should be transfer amount + fees. That will land in a new PR.

ui/pages/Send.tsx
wojciech-turek: Got it, moving out of habits is not so easy haha

background/services/chain/index.ts
```diff
@@ -147,10 +147,11 @@ export default class ChainService extends BaseService<Events> {
       },
       blockPrices: {
         schedule: {
-          periodInMinutes: 2,
+          periodInMinutes:
+            Number(process.env.BLOCKNATIVE_POLLING_FREQUENCY) / 60,
```
Shadowfiend, line 151: We should probably ` ?? 120` or something of the sort in case it's unset, otherwise we'll get a period of 0 (which I'm pretty sure still works and gets us a fixed 1-minute period, but…).

background/redux-slices/transaction-construction.ts
```diff
@@ -1,88 +1,87 @@
...
+    await emitter.emit("requestSignature", transaction)
   }
 )
 
 const transactionSlice = createSlice({
   name: "transaction-construction",
   initialState,
   reducers: {
-    transactionOptions: (
+    setTransactionRequest: (
```
Shadowfiend, line 50: So, what we've done elsewhere is not have the `set` prefix for the actions. I'm ambivalent about that approach and haven't had a chance to see if there's a preferred path suggested by redux-toolkit or a go-to approach in the community, but we're doing it in all slices right now… So I'm tempted to stick with that here.

background/redux-slices/transaction-construction.ts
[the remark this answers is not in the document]
Shadowfiend: :thumbsup: I'm comfortable with keeping this for now. There's something about it that bugs me, but… I can't quite put my finger on it, so won't be pushing until I get a better feel for what I'm worried about.

ui/components/Wallet/WalletAccountBalanceControl.tsx
Shadowfiend: Haha I don't think you should, fwiw—feature flags are a great strategy, and this is basically a feature flag, we can just avoid hard-coding them and use the existing env mechanism :muscle:

ui/pages/Send.tsx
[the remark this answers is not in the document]
Shadowfiend: I actually didn't realize this was a string, my bad.

I'm not that worried due to the fact that we're reworking this anyway to be the proper total, at which point this will look very different.

ui/pages/Send.tsx
Shadowfiend: This also applies to settings__image -> again, either `.settings_image` or `.settings > img`. Basically avoid double underscores :grin:

Shadowfiend (approved):
Left some tiny notes, but the main thing is fixed and almost everything else got hit, too! 🎊 

Let's get this sucker in :sunglasses:

background/services/chain/index.ts
wojciech-turek: Changed added, `?? 2`

ui/pages/Send.tsx
```diff
@@ -81,16 +75,22 @@ export default function Send(): ReactElement {
   // I mean how do i know when its going to refresh when I enter this screen
 
   const updateGasOptions = useCallback(() => {
-    if (gas) {
-      const instant = gas.estimatedPrices.find((el) => el.confidence === 99)
-      const express = gas.estimatedPrices.find((el) => el.confidence === 90)
-      const regular = gas.estimatedPrices.find((el) => el.confidence === 70)
+    if (estimatedFeesPerGas) {
+      const instant = estimatedFeesPerGas.estimatedPrices.find(
```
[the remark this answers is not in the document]
wojciech-turek: Noted for future PR

ui/pages/Send.tsx
wojciech-turek: Changed that particular one, but theres more places like `fees__option__subtext` that needs to be changed, I will address that in a new PR.

background/redux-slices/transaction-construction.ts
wojciech-turek: `set` no more :)

background/services/chain/index.ts
```diff
@@ -148,7 +148,7 @@ export default class ChainService extends BaseService<Events> {
       blockPrices: {
         schedule: {
           periodInMinutes:
-            Number(process.env.BLOCKNATIVE_POLLING_FREQUENCY) / 60,
+            Number(process.env.BLOCKNATIVE_POLLING_FREQUENCY) / 60 ?? 2,
```
Shadowfiend, line 151: I don't think this will have the desired effect, as the `??` binds to the result of `Number(process.env.BLOCKNATIVE_POLLING_FREQUENCY) / 60`, which will never be `undefined`. Instead, I think we want this:

```suggestion
            Number(process.env.BLOCKNATIVE_POLLING_FREQUENCY ?? "2") / 60,
```

Shadowfiend (approved):
One minor nit, but since env defaults include the blocknative polling frequency I'm not worried about it. Let's land this now and handle after :rocket:

Shadowfiend (approved):
This is the part where I approve my own merge so when things are broken we all know it's my fault.

background/services/chain/index.ts
wojciech-turek: True that, I have corrected that in network fees PR. Thanks!

PR #474, opened 2021-11-24
🗺 Quick overview tab style update

henryboldi:
Previously, the top section felt to users as if something was cut off the top of the screen. So the design has been updated. Based on the changes to the design in Figma,
Mostly updated overview tab


mhluongo (approved):
:clap: :clap: :clap:

Issue #388, opened 2021-10-29
`provider-bridge` (content script) tracking issue


Shadowfiend:
This ticket is for tracking `provider-bridge` related work. 

This is the name of the package for the content-script that injects the `window-provider` into the page and sets up a communication proxy between the background script `ProviderBridgeService` and the dApp.

greg-nagy:
This became stale and probably is for the best not to keep this tracking issue open.

Issue #387, opened 2021-10-29
`TallyWindowProvider` tracking issue


Shadowfiend:
The purpose of this issue is to track and collect every PR that relates to the window provider. - Greg

Window provider is responsible for providing the dApp with the window.ethereum / window.tally object with a standard EIP-1193 implemantation, so the dApps can communicate with our wallet through a standard way, hopefully without any changes necessary to their codebase.

greg-nagy:
This became stale and probably is for the best not to keep this tracking issue open.

PR #477, opened 2021-11-29
🤗 🔈  Embrace your internal oddball: Provider for the internal dApps

### LINKED ISSUE #478: Internal dApp provider for swaps (and other internal dApps)
- [x] port communication + ethers wrapper + getProvider
- [x] `provider-bridge-shared` package 
- [x] ?maybe? `window-provider` package (if I can salvage the code for the internal provider)

greg-nagy:
CleanShot 2021-12-04 at 16 38 06

### Internal provider implementation based on the requirements below

- The internal dapps can use this directly without going through the window.ethereum field, and will skip the connect handshake
- The provider will be resolved for internal dapps through getProvider here: https://github.com/tallycash/tally-extension/blob/main/background/redux-slices/utils/contract-utils.ts

### Main changes

- Refactored the `window-provider` api so 
  - it supports the requirements of both `ethers.js` and `web3-react` in terms of the deprecated `.send()` method
  - The main entry point is the standard `.request()` method for the communication
  - The communication layer is decoupled from the EIP-1193 implementations so the same code can be used in the in-page and popup.html context as well (using window.postMessage in the first and using port communication in the later)
- Implemented the provider in the `contract-utils.ts` and it connects directly to the `InternalEthereumProviderService` through a dedicated internal port
- Created 2 new packages
  - `provider-bridge-shares` to share connection constants and typings with minimal coupling among all the contexts they are needed 
  - `window-provider` so it can be used in the UI thread 
  
  
### Note
I was aiming for getting the main building blocks into the proper places with roughly the shape that they will have. They will need some love and care down the road. I track these here: https://github.com/tallycash/extension/issues/366

Closes #478 
See #387 See #388  See #390


background/package.json
```diff
@@ -32,6 +32,8 @@
     "@ethersproject/transactions": "^5.4.0",
     "@ethersproject/web": "^5.4.0",
     "@tallyho/hd-keyring": "^0.3.2",
+    "@tallyho/provider-bridge-shared": "0.0.1",
```
Shadowfiend, line 35: Hmmm… This package should purely coordinate the child packages, whereas the shared package seems like it should only be a dependency of those (window-provider and provider-bridge), as it were… Anyway, not a huge issue right now.

background/redux-slices/utils/contract-utils.ts
```diff
@@ -1,8 +1,29 @@
...
+  const provider = new TallyWindowProvider({
+    postMessage: port.postMessage,
+    addEventListener: port.onMessage.addListener,
+    removeEventListener: port.onMessage.removeListener,
+    origin: window.location.origin,
+  })
+
+  // @ts-expect-error web3-react needs send() with a specific syntax. Not sure why but the lib falls trhough and uses that unmaintained implementation which collides with the ethers.js ExternalProvider type
+  return new Web3Provider(provider)
+}
```
Shadowfiend, line 24: Really the expect-error reason here is that `Web3Provider` takes more than just an `ExternalProvider`—it can also handle other providers under the covers, including EIP1193 providers.

I would still prefer to hook this up via internal events rather than through an extra port, tbh… But I don't think that's a blocker by any means, and this is easier to start with.

background/redux-slices/utils/contract-utils.ts
```diff
@@ -1,8 +1,29 @@
...
+
+export function getAlchemyProvider(): Provider {
```
Shadowfiend, line 26: We do not need an alchemy provider, let's nuke it from orbit.

provider-bridge-shared/index.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export type PortListenerFn = (callback: unknown, ...params: unknown[]) => void
+export type PortListener = (listener: PortListenerFn) => void
+export type PortTransport = {
+  postMessage: (data: unknown) => void
+  addEventListener: PortListener
+  removeEventListener: PortListener
+  origin: string
+}
```
Shadowfiend, line 37: Real mixed feelings about this package lol. It's tiny, and I hate it, but also I like that we've got constants in one place… Rough.

src/window-provider.ts
```diff
@@ -1,121 +1,19 @@
...
-  // @ts-expect-error I don't really have any way to know the exact type of window.ethereum so it's better to expect error than lie
-  window.ethereum = new TallyWindowProvider()
+  // The window object is considered unsafe, because other extensions could have modified them before this script is run.
+  // For 100% certainty we could create an iframe here, store the references and then destoroy the iframe.
+  //   something like this: https://speakerdeck.com/fransrosen/owasp-appseceu-2018-attacking-modern-web-technologies?slide=95
+  window.tally = new TallyWindowProvider({
+    postMessage: (data: unknown) => {
+      return window.postMessage(data, window.location.origin)
+    },
+    addEventListener: window.addEventListener.bind(this),
```
Shadowfiend, line 14: Is `this` meant to be `window` here? If so, maybe we use it explicitly, since we're relying on it being consistent in this particular file?

Shadowfiend (commented):
Path looks good. Left a couple of notes.

background/redux-slices/utils/contract-utils.ts
greg-nagy: What do you mean by internal events? 

Afaik there is no connection between ui and background other than port communication. 

In my understanding we have only 2 options here. Use port communication directly or use port communication through webext-redux.

background/redux-slices/utils/contract-utils.ts
greg-nagy: CleanShot 2021-12-04 at 09 14 39

https://github.com/tallycash/tally-extension/commit/7dd3d0a36dbe4610910e86eb46130911e82e1b02

provider-bridge-shared/index.ts
greg-nagy: I know, I know ... heh

The reason I went with this is the complex dependency graph, error prone nature and the expected volume of the content.

src/window-provider.ts
greg-nagy: https://github.com/tallycash/tally-extension/commit/0de2d0bfdc716617a49a600e07ff80fd914b58d6

background/package.json
greg-nagy: We have the same love hate relationship to this package :) 

But these constants are used in the background too, so it's a necessary dependency here.

Shadowfiend: … I def thought this was the top-level package.json, not background's. This makes complete sense, my bad haha.

background/redux-slices/utils/contract-utils.ts
Shadowfiend: Oh of course you're right 😅 We'd need a dedicated slice to shuttle this kind of stuff in that scenario, hook that slice up to the service, etc etc. Yeah, I still think we may run into a situation in the future where that approach is desirable, but this is absolutely the right way to go now 👍

window-provider/index.ts
```diff
@@ -25,14 +33,64 @@ export default class TallyWindowProvider extends EventEmitter {
...
+        (error) => paramsOrCallback(error, null)
+      )
+    }
+
+    return Promise.reject(new Error("Unsupported function parameters"))
   }
 
-  send(method: string, params: Array<unknown> = []): Promise<unknown> {
+  request(arg: RequestArgument): Promise<unknown> {
+    if (typeof arg.method !== "string")
```
Shadowfiend, line 86: Isn't this already checked below?

window-provider/index.ts
```diff
@@ -0,0 +1,177 @@
...
+      request: {
+        method,
+        params,
+      },
+    }
+
+    this.transport.postMessage(sendData)
+
+    const isWindowTransport = (t: ProviderTransport): t is WindowTransport => {
+      return !t.origin.startsWith("chrome-extension://")
```
Shadowfiend, line 109: This will potentially break in non-Chrome browsers. It also seems trivially circumventable from a `WindowTransport`, if someone wanted to try to circumvent, which is a problem if bypassing any of the checks below could yield any security issues—I haven't fully modeled that out to figure out if it's a real compromise, but would rather avoid entirely.

Can the `PortTransport` ship some sort of tag property that identifies it? Or can this code itself check for whether it's running inside an extension context, and if it is assume the transport is a `PortTransport`? Another option is, port listeners get a `runtime.MessageSender` as a second parameter, which we may be able to check as `instanceof MessageSender` or similar. Regular event handlers I believe don't get a second parameter, though a malicious implementation of `addEventListener` may be able to evade that (but maybe not in a way that escapes the `instanceof` check, since `runtime` won't exist in the first place on a non-extension page).

window-provider/index.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    return new Promise((resolve) => {
+      // TODO: refactor the listener function out of the Promise
+      const listener = (event: WindowResponseEvent & PortResponseEvent) => {
```
Shadowfiend, line 114: This should be an `|`, not an `&`, which bleeds down into the handling below… We should probably assert (with `as`, which makes it clear it's not type-system-enforced) that `event` is expected to be of type `WindowResponseEvent` in one branch and `PortResponseEvent` in the other.

The parameter type also really isn't true, this parameter is `unknown` and we make assumptions about what it will actually be below based on the transport. We could still receive an unexpected message irrespective of transport.

Shadowfiend (commented):
So, I'm somewhat concerned about the joint handling of window providers and port providers in the same place… It leads to less duplication, but we've declared these two communication endpoints as having different security models, and by using one spot to handle both types of messages, I'm worried we will fail to correctly handle those two different models.

A lot of our difference in treatment isn't on the window provider side, of course, it's on the service side, so that's reassuring... On the other hand, if that's the case I'm not sure why we're trying to track `origin` for the port providers at all (though that's maybe explained by the `isWindowTransport` check).

Anyway, I still think this is directionally correct… Just a few details I need to finish wrapping my head around.

window-provider/index.ts
```diff
@@ -25,14 +33,64 @@ export default class TallyWindowProvider extends EventEmitter {
...
+  ): Promise<unknown> | void {
+    if (
+      typeof methodOrRequest === "string" &&
+      typeof paramsOrCallback !== "function"
+    ) {
+      return this.request({ method: methodOrRequest, params: paramsOrCallback })
+    }
+
+    if (
+      Object.prototype.toString.call(methodOrRequest) === "[object Object]" &&
```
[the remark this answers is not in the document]
greg-nagy: I wanted to do that but that is passes for arrays and dates too and here I wanted to be exact because it's the entry point for the whole communication handling. 

Would like to keep the strictness if you are ok with it, but happy to change to a more elegant and typesafe solution but all the other options seemed like an overshoot for this single case.

window-provider/index.ts
greg-nagy: removed

window-provider/index.ts
greg-nagy: This was a totally broken approach, thanks for flagging this! This comment and the next one bruised my ego quite a bit :) Luckily I am quick to recover and come up with fixes, which happened now. Details below.

window-provider/index.ts
greg-nagy: > The parameter type also really isn't true, this parameter is unknown and we make assumptions about what it will actually be below based on the transport. We could still receive an unexpected message irrespective of transport.

This was a really approach from my part. It should have checked the event object for the correct type and make decisions based on that. I have written some simple type guards and composed a runtime type checking function for both event types. 

Did not want to include lodash or ajv for size reasons and the code to run these checks are simple.
See `window-provider/utils.ts`

I really appreciate this catch!

greg-nagy:
In https://github.com/tallycash/extension/pull/519 I ended up moving the type checking functions to `provider-bridge-shared` because it felt more appropriate to keep the compile time and runtime type checks together.

Shadowfiend (approved):
Still nervous about our type checking, but not like “things are going to catch fire” nervous, just “there are edge cases we haven't played out all the way yet” nervous. I think we're good enough to ship as-is and keep refining.

window-provider/index.ts
```diff
@@ -33,27 +38,11 @@ export default class TallyWindowProvider extends EventEmitter {
...
-   *           request?: (request: { method: string, params?: Array<any> }) => Promise<any>
-   *       }
-   *       ```
-   * Had to come up with a way to support both send methods
-   */
+  // deprecated EIP1193 send for web3-react injected provider Send type:
+  // https://github.com/NoahZinsmeister/web3-react/blob/d0b038c748a42ec85641a307e6c588546d86afc2/packages/injected-connector/src/types.ts#L4
   send(method: string, params: Array<unknown>): Promise<unknown>
+  // deprecated EIP1193 send for ethers.js Web3Provider > ExternalProvider:
+  // https://github.com/ethers-io/ethers.js/blob/73a46efea32c3f9a4833ed77896a216e3d3752a0/packages/providers/src.ts/web3-provider.ts#L19
```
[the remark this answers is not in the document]
greg-nagy: We need this to satisfy it's typing definition, not becasue it will be used.

- ethers uses [buildEip1193Fetcher](https://github.com/ethers-io/ethers.js/blob/73a46efea32c3f9a4833ed77896a216e3d3752a0/packages/providers/src.ts/web3-provider.ts#L83)
- which requires [ExternalProvider](https://github.com/ethers-io/ethers.js/blob/73a46efea32c3f9a4833ed77896a216e3d3752a0/packages/providers/src.ts/web3-provider.ts#L13)
- which defines [send](https://github.com/ethers-io/ethers.js/blob/73a46efea32c3f9a4833ed77896a216e3d3752a0/packages/providers/src.ts/web3-provider.ts#L19) as optional parameter

Because we are forced to have `send` to be compatible with `web3-react` injectd provider — which is really old and not standard implementation.

I could have forced only the typing somehow but that's like working a lot to lie to your future self. It seemed to be a more robust solution to have an actual working implementation so if in any case it's called by ethers — which probably won't happen — then it's has the correct signature.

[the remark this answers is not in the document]
Shadowfiend: Ah, got it! That's rough… Feels like we should be able to fix ethers's typing, but that's for another time.

provider-bridge-shared/index.ts
```diff
@@ -51,3 +51,81 @@ export type PortTransport = {
...
+export function isArray(arg: unknown): arg is Array<unknown> {
+  return getType(arg) === "Array"
+}
+
+export function isUndefined(arg: unknown): arg is undefined {
+  return typeof arg === "undefined"
+}
+
+export function isString(arg: unknown): arg is string {
+  return getType(arg) === "String"
```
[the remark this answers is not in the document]
greg-nagy: Wanted to do that — that was my first implementation — then I read a bit about the issue :( 

```
typeof new String("asdfasd")
'object'
```

provider-bridge-shared/index.ts
```diff
@@ -51,3 +51,81 @@ export type PortTransport = {
...
+    public message: string,
+    public code: number = PROVIDER_ERROR_CODES.unknown,
+    public data?: unknown
+  ) {
+    super(message)
+  }
+}
+
+export function getType(arg: unknown) {
+  return Object.prototype.toString.call(arg).slice("[object ".length, -1)
```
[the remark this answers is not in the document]
greg-nagy: I see now, that in some cases I had tunnel vision with this solution. 
But in other cases this is sadly seem to be a necessary move, so the thruth lies somewhere between us. 

My goal here was not to be "untrickable" but to be able to provide a reliable layer for typing information — and create a  place in the codebase where we can accumulate our learning about type checks. `provider-bridge-shared` is probably not it's final place, but it works for now.

Both instanceof and typeof is really cumbersome and unreliable in javascript — and typescript is only a layer on top of it. Suffering from being frame dependant, not working logically etc. Sadly the best argument against `Any reason we can't use instanceof for anything that isn't an Object check?` is it's mdn page itself. 

Agree though that this `Object.prototype.toString` solution should only be used in special cases. I kep this for objects and strings and used native solution for the rest of the checks.

window-provider/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+} from "@tallyho/provider-bridge-shared"
+
+export function isObject(
+  arg: unknown
+): arg is Record<string | number | symbol, unknown> {
+  return Object.prototype.toString.call(arg) === "[object Object]"
+}
+
+export function isArray(arg: unknown): arg is Array<unknown> {
+  return Object.prototype.toString.call(arg) === "[object Array]"
```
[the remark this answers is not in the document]
greg-nagy: https://github.com/tallycash/extension/pull/519/commits/42fc7d3e88e68afd5def6c03a79178b05a8f5f83

window-provider/index.ts
```diff
@@ -105,16 +84,12 @@ export default class TallyWindowProvider extends EventEmitter {
...
     this.transport.postMessage(sendData)
 
-    const isWindowTransport = (t: ProviderTransport): t is WindowTransport => {
-      return !t.origin.startsWith("chrome-extension://")
-    }
-
     return new Promise((resolve) => {
       // TODO: refactor the listener function out of the Promise
-      const listener = (event: WindowResponseEvent & PortResponseEvent) => {
+      const listener = (event: WindowResponseEvent | PortResponseEvent) => {
```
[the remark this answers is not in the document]
greg-nagy: https://github.com/tallycash/extension/pull/519/commits/42fc7d3e88e68afd5def6c03a79178b05a8f5f83

provider-bridge-shared/index.ts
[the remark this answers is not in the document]
greg-nagy: Checking the current version of the type checks I think we have put together a pretty decent and robust solution https://github.com/tallycash/extension/blob/42fc7d3e88e68afd5def6c03a79178b05a8f5f83/provider-bridge-shared/index.ts#L83

I expect this to mature really well too.

PR #490, opened 2021-11-29
An Embarrassment of Enriches: Add a few helpers to factor non-coordination work out of selectAccountAndTimestampedActivites

Shadowfiend:
`selectAssetPricePoint` handles price point lookup in the asset slice,
`enrichAssetAmountWithMainCurrencyValues` adds main currency values to
an `AnyAssetAmount`, and `enrichAssetAmountWithDecimalValues` adds
localized decimal values to an `AnyAssetAmount`.

`selectAccountAndTimestampedActivities` is now purely a coordinator.


henryboldi (approved):
These asset utils are valuable in a number of places outside of the asset slice. The current folder structure might semi-incorrectly suggest that the utils directly correspond with the slice,

activities slice → activity-utils
assets slice → assets-utils
etc

Shadowfiend:
Yeah… I think the missing piece here is a `redux-store`-type package that has a `slices/...` and a *sibling* `helpers/...`, and exports the helpers at the top level.

Def something I think will be coming in the future, but not sure now's the moment.

PR #493, opened 2021-12-01
Fix Fairy: Fix a few nits and bugs in current main

Shadowfiend:
During testing, we realized the asset price point conversions landed
in #490 weren't working quite as expected. This PR fixes that issue, as
well as a handful of others:
- All remaining React warnings in common paths are gone.
- Certain lookups that were waiting 1 minute to run due to being hooked up
  as alarms now run immediately.
- Initial balance checks for a new account are immediately followed by price
  checks for those assets.

### DIFF
--- a/background/services/indexing/index.ts
+++ b/background/services/indexing/index.ts
@@ -75,6 +75,7 @@ export default class IndexingService extends BaseService<Events> {
           periodInMinutes: 30,
         },
         handler: () => this.handleTokenAlarm(),
+        runAtStart: true,
       },
       prices: {
         schedule: {
@@ -82,6 +83,7 @@ export default class IndexingService extends BaseService<Events> {
           periodInMinutes: 10,
         },
         handler: () => this.handlePriceAlarm(),
+        runAtStart: true,
       },
     })
   }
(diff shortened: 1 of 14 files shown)


background/services/indexing/index.ts
```diff
@@ -210,6 +212,9 @@ export default class IndexingService extends BaseService<Events> {
         // default list and add any non-zero tokens to the tracking list
         const balances = await this.retrieveTokenBalances(addressNetwork)
 
+        // FIXME Refactor this to only update prices for tokens with balances.
+        await this.handlePriceAlarm()
```
Shadowfiend, line 216: @mhluongo want your eyes on this specific change. It's overall fairly small, but want to make sure there aren't any concerns about rate limiting that I've missed.

mhluongo: I'm not a huge fan without that `FIXME` getting `FIXED`... but I don't think this specifically will be a rate limiting issue. Accounts aren't added frequently enough.

More general rate limiting concerns are price checks are still a thing, of course.

henryboldi (changes requested):
Great changes! The home tab loads everything in much faster!

However, after testing this locally, it looks like there's now an issue displaying the price in the overview tab,

Loading
Timeout

Shadowfiend:
Heh I was trying to figure this out and then realized it's because the enrichment functions didn't include unit price 😅 Oh well, good opportunity to move another field name away from the term “token” :grimacing:

Shadowfiend:
This should be fixed now.

henryboldi (approved):
Beautiful!

PR #494, opened 2021-12-01
:pie:  Swap slice refactor

itsrachelfish:
This PR creates a dedicated redux slice for the 0x swap code.

While testing I noticed that some of the assets we are storing in redux have incorrect contract addresses. For example when loading the swap page you can see this warning for the 0x token:

contract-address-discrepancy

I think this issue may have been introduced by my testnet changes, so I'll look into it.


ui/pages/Swap.tsx
```diff
@@ -53,94 +39,96 @@ export default function Swap(): ReactElement {
...
-              .toString(),
-          }
-        })
+        dispatch(
+          setSwapAmount({
+            sellAmount: inputValue,
+            // TODO: Use a safe math library
+            buyAmount: (
+              floatValue / parseFloat(swap.buyToken.price)
+            ).toString(),
```
Shadowfiend, line 94: Should this update trigger a price check for the specific swap via the `/price` endpoint as well? I think just mathing out the swapped amount can only get us a very rough approximation, since changes in the sell or buy amounts can result in routing through different exchanges based on available liquidity, and thus different prices.

Separately, realizing that I need to add a division function to the fixed-point lib.

ui/pages/Swap.tsx
```diff
@@ -53,94 +39,96 @@ export default function Swap(): ReactElement {
...
-
-      if (Number.isNaN(inputValue) || swap.price.isZero()) {
-        setSwapAmount(() => {
-          return {
-            from: event.target.value, // We have to preserve the original input value, otherwise users won't be able to type decimals
-            to: "0",
-          }
-        })
+      const inputValue = event.target.value.replace(/[^0-9.]/g, "") // Allow numbers and decimals only
+      const floatValue = parseFloat(inputValue)
```
Shadowfiend, line 75: `parseToFixedPointNumber` from fixed-point is the jam here.

ui/pages/Swap.tsx
```diff
@@ -53,94 +39,96 @@ export default function Swap(): ReactElement {
...
+            sellAmount: "0",
+            buyAmount: inputValue,
+          })
+        )
+      } else {
+        dispatch(
+          setSwapAmount({
+            // TODO: Use a safe math library
+            sellAmount: (
+              floatValue * parseFloat(swap.buyToken.price)
```
Shadowfiend, line 123: You'll be able to use `multiplyFixedPointNumbers` for this.

background/redux-slices/0x-swap.ts
```diff
@@ -0,0 +1,171 @@
...
+interface ZrxPrice {
+  symbol: string
+  price: string
+}
+
+export const fetchTokens = createBackgroundAsyncThunk(
+  "0x-swap/fetchTokens",
+  async (_, { getState }) => {
+    const state = getState() as { assets: AssetsState }
+    const assets = state.assets as Asset[]
```
Shadowfiend, line 47: This `as` isn't necessary, since `AssetsState` is an alias for `SingleAssetState[]` which is an `Asset[]`. If you *want* to downcast the type, you can do it by declaring the type of the variable instead: `const assets: Asset[] = state.assets`. `as` is generally dangerous and should only be used to add type information where it's missing (e.g. in the `getState` call above).

That said, I have some thoughts on how this could get restructured a bit as I think we should avoid pulling in other slices during action handling, partly due to the missing type information.

background/redux-slices/0x-swap.ts
```diff
@@ -0,0 +1,171 @@
...
+          return false
+        })
+
+        // TODO: What if multiple assets match?
+        if (matchingTokens.length) {
+          return matchingTokens[0]
+        }
+
+        return false
+      })
```
Shadowfiend, line 96: I would consider moving all of this to a selector, since it depends on another slice. In general I'm trying to drive us towards using selectors for cross-slice coordination. For this case, this makes it so we don't have to worry about notifying the swap slice if something happens to update the assets slice (for example, the user changes their asset lists)—the selector will automatically rerun and the 0x data will be presented through the lens of the user's preferred assets.

The approach then would be:
```
dispatch(fetchTokens())
--
fetchTokens: API call to 0x
fetchTokens: validate JSON + store an `zrxTokens: ZrxToken[]` field in the slice state
--
In the component:

const availableAssets = useBackgroundSelector(selectSwappableAssets)
```

Then the selector might be something like:

```typescript
export const selectSwappableAssets = createSelector(
  (state: { assets: AssetsState; swap: SwapState }) => ({ assets, zrxTokens: swap.zrxTokens })
  ({ assets, zrxTokens }) => {
    const filteredAssets = ...
    ...
    return filteredAssets
  }
)
```

A similar selector could do the filtering currently happening in `fetchSwapPrices.fulfilled` below. The idea with that one would be that we stash the latest prices we've gotten for the swap in this slice, then the selector takes those and the swappable asset list and filters down matching assets + fills in their prices in a single pass:

```typescript
export const selectSwapPrices(
  (state: { assets: AssetsState; swap: SwapState }) => ({
    swappableAssets: selectSwappableAssets(state),
    zrxPrices: assets.zrxPrices
  }),
  ({ swappableAssets, zrxPrices }) => {
    return zrxPrices.map...
  }
)
```

background/redux-slices/0x-swap.ts
```diff
@@ -0,0 +1,171 @@
...
+              }
+
+              return false
+            })
+
+            if (matchingAsset.length) {
+              return { ...asset, price: matchingAsset[0].price }
+            }
+
+            return { ...asset, price: 0 }
```
Shadowfiend, line 155: Should we let 0 prices in? Or do we want to filter them out, since what it really means is we don't have an asset price, which I think implies we can't swap?

Shadowfiend (approved):
Excited to see this refactor come together!

I left a bunch of thoughts, some architectural and others stylistic, but I don't think any of them are blocking. A lot of the architectural thoughts are coming out of other PRs over the last week or two as a common way to approach some repeated architectural questions.

I'm going to take this for a test drive! 🏎

background/redux-slices/0x-swap.ts
Shadowfiend: Realized we *do* need the type here to be `Asset[]` rather than `SingleAssetState[]` for the `isSmartContractFungibleAsset` type narrowing to take effect below—to which end I think declaring the type of `assets` is the right way to go.

Shadowfiend: … I see that also fights with eslint's prefer-destructuring rule, and there's an issue about it in typescript-eslint: typescript-eslint/typescript-eslint#723 haha. If we go down that path, I think we can eslint-ignore that with a link to that issue.

itsrachelfish: Yeah I ended up writing and re-writing this slice logic like 4 times to deal with a bunch of linter errors and cyclic redundancies :sweat_smile: 

At first I was using a background selector to get the asset list in the swap component and passing it to the thunk being dispatched. But that meant I would have to pass the asset list variable as a dependency to the `useEffect` function, which makes the `useEffect` function dispatch the 0x API call over and over, each time the asset list changes. 

The recommended way of dispatching an action "on page load" in react is by using `useEffect` without any dependencies.
https://reactjs.org/docs/hooks-effect.html

> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. 

By pulling in the asset list within the slice I can dispatch this thunk when the swap page first loads, by calling `useEffect` without any dependencies.

background/redux-slices/0x-swap.ts
```diff
@@ -0,0 +1,171 @@
+import { createSlice } from "@reduxjs/toolkit"
+import { fetchJson } from "@ethersproject/web"
+
+import { createBackgroundAsyncThunk } from "./utils"
+import { isSmartContractFungibleAsset, Asset } from "../assets"
+import { AssetsState } from "./assets"
+import logger from "../lib/logger"
+
+interface SwapAsset extends Asset {
+  price?: string
```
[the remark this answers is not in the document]
itsrachelfish: The main reason I didn't immediately use the `PricePoint` type was because I didn't have good math functions to work with. Should be easy now though :smile:

background/redux-slices/0x-swap.ts
itsrachelfish: You're right, assets without a price should be filtered out!

background/redux-slices/0x-swap.ts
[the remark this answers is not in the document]
Shadowfiend: Yesssss, that's delightful news :grin:

background/redux-slices/0x-swap.ts
Shadowfiend: < insert Aussie reverse conga party parrot emoji >

PR #495, opened 2021-12-01
Create animation to show fees timout

wojciech-turek:
Move out chosing fees code to separate component to make the
send component more readable. This PR adds a new value in redux slice
to track last time gasEstimates were updated.

ezgif com-gif-maker

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -2,7 +2,7 @@ import browser from "webextension-polyfill"
 import { alias, wrapStore } from "webext-redux"
 import { configureStore, isPlain, Middleware } from "@reduxjs/toolkit"
 import devToolsEnhancer from "remote-redux-devtools"
-import ethers from "ethers"
+import { ethers } from "ethers"
 
 import { decodeJSON, encodeJSON, getEthereumNetwork } from "./lib/utils"
 import logger from "./lib/logger"
--- a/background/redux-slices/transaction-construction.ts
+++ b/background/redux-slices/transaction-construction.ts
@@ -14,11 +14,13 @@ export type TransactionConstruction = {
   status: TransactionConstructionStatus
   transactionRequest?: EIP1559TransactionRequest
   estimatedFeesPerGas: BlockPrices | null
+  lastGasEstimatesRefreshed: number
 }
 
 export const initialState: TransactionConstruction = {
   status: TransactionConstructionStatus.Idle,
   estimatedFeesPerGas: null,
+  lastGasEstimatesRefreshed: new Date().getTime(),
 }
 
 export type Events = {
@@ -64,7 +66,11 @@ const transactionSlice = createSlice({
       immerState,
       { payload: estimatedFeesPerGas }: { payload: BlockPrices }
     ) => {
-      return { ...immerState, estimatedFeesPerGas }
+      return {
+        ...immerState,
+        estimatedFeesPerGas,
+        lastGasEstimatesRefreshed: new Date().getTime(),
+      }
     },
   },
   extraReducers: (builder) => {
@@ -83,6 +89,12 @@ export const selectEstimatedFeesPerGas = createSelector(
   (state: { transactionConstruction: TransactionConstruction }) =>
     state.transactionConstruction.estimatedFeesPerGas,
   (gasData) => gasData
+)
+
+export const selectLastGasEstimatesRefreshTime = createSelector(
+  (state: { transactionConstruction: TransactionConstruction }) =>
+    state.transactionConstruction.lastGasEstimatesRefreshed,
+  (updateTime) => updateTime
 )
 
 export const selectTransactionData = createSelector(
--- a/background/redux-slices/transaction-construction.ts
+++ b/background/redux-slices/transaction-construction.ts
@@ -1,7 +1,11 @@
 import { createSlice, createSelector } from "@reduxjs/toolkit"
 import Emittery from "emittery"
 
-import { BlockPrices, EIP1559TransactionRequest } from "../networks"
+import {
+  BlockEstimate,
+  BlockPrices,
+  EIP1559TransactionRequest,
+} from "../networks"
 import { createBackgroundAsyncThunk } from "./utils"
 
 const enum TransactionConstructionStatus {
@@ -13,7 +17,12 @@ const enum TransactionConstructionStatus {
 export type TransactionConstruction = {
   status: TransactionConstructionStatus
   transactionRequest?: EIP1559TransactionRequest
-  estimatedFeesPerGas: BlockPrices | null
+  estimatedFeesPerGas: {
+    baseFeePerGas: bigint
+    instant: BlockEstimate | undefined
+    express: BlockEstimate | undefined
+    regular: BlockEstimate | undefined
+  } | null
   lastGasEstimatesRefreshed: number
 }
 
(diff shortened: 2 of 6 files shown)


Shadowfiend:
Insta-glorious. Going to hold off on reviewing here until #471 has its one issue fixed + I can merge it.

henryboldi:
Amazing! It might be nice to upload a GIF of it here @wojciech-turek  :)

ui/pages/Send.tsx
```diff
@@ -298,7 +155,16 @@ export default function Send(): ReactElement {
...
+          {feeModalOpen && (
+            <NetworkFeesChooser
+              gasOptions={gasOptions}
+              activeFeeIndex={activeFeeIndex}
+              handleSelectGasOption={handleSelectGasOption}
+              gasLimit={gasLimit}
+              setGasLimit={setGasLimit}
+              saveUserGasChoice={saveUserGasChoice}
+            />
+          )}
```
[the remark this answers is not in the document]
wojciech-turek: Could you please explain what is the benefit of this change? Visually it seems less readable since with `&&` I know it renders fees chooser or nothing and with `?` I also need to care about false scenario.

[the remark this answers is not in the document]
Shadowfiend: You care about the false scenario, you're just relying on the fact that `&&` will return it as `null` to implicitly handle it :) The idea here is to be explicit about the alternate path, basically.

background/redux-slices/transaction-construction.ts
```diff
@@ -13,12 +17,19 @@ const enum TransactionConstructionStatus {
 export type TransactionConstruction = {
   status: TransactionConstructionStatus
   transactionRequest?: EIP1559TransactionRequest
-  estimatedFeesPerGas: BlockPrices | null
+  estimatedFeesPerGas: {
+    baseFeePerGas: bigint
+    instant: BlockEstimate | undefined
+    express: BlockEstimate | undefined
+    regular: BlockEstimate | undefined
+  } | null
```
Shadowfiend, line 25: For consistency, let's move to `| undefined` here.

ui/components/NetworkFees/Chooser.tsx
```diff
@@ -0,0 +1,207 @@
+import { selectLastGasEstimatesRefreshTime } from "@tallyho/tally-background/redux-slices/transaction-construction"
+import React, { ReactElement, useCallback, useEffect, useState } from "react"
+import { useSelector } from "react-redux"
+import SharedButton from "../Shared/SharedButton"
+import SharedInput from "../Shared/SharedInput"
+
+interface GasOption {
```
Shadowfiend, line 7: `type`*

ui/components/NetworkFees/Chooser.tsx
```diff
@@ -0,0 +1,207 @@
...
+interface NetworkFeesChooserProps {
+  gasOptions: GasOption[]
+  activeFeeIndex: number
+  handleSelectGasOption: (number: number) => void
+  gasLimit: string | number
+  setGasLimit: React.Dispatch<React.SetStateAction<string>>
+  saveUserGasChoice: () => void
+}
+
+export default function NetworkFeesChooser({
```
Shadowfiend, line 25: Naming-wise, I believe this filename should match the function name, so `NetworkFeesChooser.tsx`.

ui/pages/Send.tsx
```diff
@@ -71,42 +69,46 @@ export default function Send(): ReactElement {
...
           {
             name: "Regular",
-            time: "~10 Min",
-            gwei: Number(estimatedFeesPerGas?.baseFeePerGas / 1000000000n),
+            confidence: `${regular.confidence}%`,
+            gwei: Number(
+              (BigInt(regular.maxFeePerGas) +
+                BigInt(regular.maxPriorityFeePerGas)) /
+                1000000000n
+            ),
```
Shadowfiend, line 87: These should all use the ethers `formatUnits` function, something like:

```typescript
formatUnits(
  BigInt(regular.maxFeePerGas) + BigInt(regular.maxPriorityFeePerGas),
  "gwei"
)
```

ui/pages/Send.tsx
```diff
@@ -71,42 +69,46 @@ export default function Send(): ReactElement {
...
+      const regular = estimatedFeesPerGas?.regular
       if (!!instant && !!express && !!regular) {
         const updatedGasOptions = [
           {
             name: "Regular",
-            time: "~10 Min",
-            gwei: Number(estimatedFeesPerGas?.baseFeePerGas / 1000000000n),
+            confidence: `${regular.confidence}%`,
+            gwei: Number(
+              (BigInt(regular.maxFeePerGas) +
```
Shadowfiend, line 84: We should really make `BlockPrices` carry `BigInt` all the way through. It's not what we get from Blocknative, but we should convert from the Blocknative type before we propagate it through the system.

Obv not for this PR hehe.

ui/pages/Send.tsx
```diff
@@ -120,11 +122,10 @@ export default function Send(): ReactElement {
...
   const findMinMaxGas = useCallback(() => {
     if (estimatedFeesPerGas) {
-      const values = estimatedFeesPerGas.estimatedPrices.map((el) =>
-        Number(BigInt(el.maxFeePerGas) / 1000000000n)
-      )
       setMinGas(Number(estimatedFeesPerGas?.baseFeePerGas / 1000000000n))
-      setMaxGas(Math.max(...values) + 40)
+      setMaxGas(
+        Number(estimatedFeesPerGas.instant?.maxFeePerGas) / Number(1000000000n)
+      )
```
Shadowfiend, line 128: Why not just read both of these directly or set them in `updateGasOptions` instead of having a separate callback? They're pretty tightly coupled to the estimated fees, so having it be separate feels odd.

ui/pages/Send.tsx
```diff
@@ -71,42 +69,46 @@ export default function Send(): ReactElement {
...
-        (el) => el.confidence === 90
-      )
-      const regular = estimatedFeesPerGas.estimatedPrices.find(
-        (el) => el.confidence === 70
-      )
+      const instant = estimatedFeesPerGas?.instant
+      const express = estimatedFeesPerGas?.express
+      const regular = estimatedFeesPerGas?.regular
       if (!!instant && !!express && !!regular) {
         const updatedGasOptions = [
```
Shadowfiend, line 79: If you push this formatting responsibility into the fee chooser, you can make it self-contained so that the send page is only concerned with what fee was selected, rather than trying to map between a set of fee options and an index.

In this model, you would pass `[estimatedFeePerGas.instant, estimatedFeePerGas.express, estimatedFeesPerGas.regular]` directly to the chooser, and the chooser would internally do all of the formatting in this block. Then it would have a single external interaction point, `onFeeChange` or `onChange`, which would pass the selected `BlockPrices` object.

To put it a different way: there will be other components that will use the fee chooser as part of their transaction construction (e.g., swap), and we don't want them to have to redo the gas option management here.

Shadowfiend (approved):
This is good enough to ship, but I left a few additional architectural notes to iterate on.

background/main.ts
```diff
@@ -2,7 +2,7 @@ import browser from "webextension-polyfill"
 import { alias, wrapStore } from "webext-redux"
 import { configureStore, isPlain, Middleware } from "@reduxjs/toolkit"
 import devToolsEnhancer from "remote-redux-devtools"
-import ethers from "ethers"
+import { ethers } from "ethers"
```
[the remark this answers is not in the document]
wojciech-turek: I was getting ethers.utils is undefined without doing that :(

ui/components/NetworkFees/Chooser.tsx
wojciech-turek: Haha you got me trying to sneak in another interface

ui/pages/Send.tsx
wojciech-turek: Thanks for that, I was sure there is a better way to do this, appreciate it.

ui/pages/Send.tsx
wojciech-turek: I have updated the selector in a new PR to convert values to bigInt

ui/pages/Send.tsx
wojciech-turek: Hmm, moved it to updateGasOptions, works fine with less code so 💣

ui/pages/Send.tsx
wojciech-turek: I have done something alike in https://github.com/tallycash/extension/pull/518.

PR #499, opened 2021-12-01
⛵️ Onboarding flow updates

henryboldi:
### Community edition onboarding

For the community edition, our onboarding aims to explain the community-owned nature of Tally and guide users in getting their wallet set up. This PR includes,

- [x] Introduction info flow
- [x] A shared progress indicator
- [x] Add wallet options screen
- [x] Enter read-only address page
- [x] Updated MetaMask import page
- [x] Working back buttons

welcome

<hr/>

### Introduction info flow
Lets users know Tally is a DAO and that this is an early version to test with a small amount of tokens. Featuring lovely illustrations from @VladUXUI 
Introduction info flow
<br/><br/>
### Add wallet options screen
For now, we're leaving out new seed phrase creation. So users can either preview and address in view-only more, or import an existing seed phrase from MetaMask.
options
<br/><br/>

### Enter read-only address page
read-only
<br/><br/>

### Updated MetaMask import page
import

🤘🎸


ui/pages/Onboarding/OnboardingInfoIntro.tsx
```diff
@@ -0,0 +1,171 @@
...
+      extraStyles: `margin-top: 25px;`,
+    },
+    title: "Community Edition",
+    body: `You're about to enjoy an early, pre-release version of Tally. Only test with small amounts and join our Discord!`,
+    buttonCopy: "Get started",
+  },
+]
+
+export default function OnboardingInfoIntro(): ReactElement {
+  const [activeStep, setActiveStep] = useState(1)
```
Shadowfiend, line 44: Above, `steps` is an array of objects, but here `step` refers to a number.

Why don't we make `activeStep`/`setActiveStep` work on an index instead of a number (`activeStepIndex`/`setActiveStepIndex`), start at 0, and then split out a `const activeStep = steps[activeStepIndex]`? It looks like that would clean things up below a bit.

Similarly, doing this in `SharedProgressIndicator` would allow us to eliminate a lot of the `- 1`s that aren't really doing much to clarify the code.

Thoughts?

ui/pages/Onboarding/OnboardingInfoIntro.tsx
```diff
@@ -0,0 +1,171 @@
...
+              padding-bottom: 32px;
+              flex-direction: column;
+              align-items: center;
+              justify-content: space-between;
+              height: 230px;
+              text-align: center;
+            }
+          `}
+        </style>
+        <style jsx>
```
Shadowfiend, line 141: Why make this a separate element?

ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
```diff
@@ -39,143 +38,57 @@ export default function OnboardingViewOnlyWallet(): ReactElement {
...
-          `}
-        </style>
-      </section>
-    </CorePage>
+    <section>
+      <div className="wordmark" />
+      <BackButton />
+      <h1 className="serif_header">Explore Tally</h1>
+      <div className="subtitle">
+        Add an Ethereum address or ENS name to view an existing wallet in Tally.
```
Shadowfiend, line 46: We should drop “or ENS name” until we actually support those… Right now this form does not accept them, pending some version of #435 landing.

ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
```diff
@@ -39,143 +38,57 @@ export default function OnboardingViewOnlyWallet(): ReactElement {
...
+        You can upgrade to a full account anytime from Accounts
```
Shadowfiend, line 47: We should drop this text as well until it's true.

ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
```diff
@@ -39,143 +38,57 @@ export default function OnboardingViewOnlyWallet(): ReactElement {
...
+      </div>
+      <div className="input_wrap">
+        <SharedInput placeholder="ETH address" onChange={setAddress} />
```
Shadowfiend, line 50: Not blocking but… Placeholders are not labels! Let's talk to @VladUXUI about how we want this label to work at some point.

ui/pages/Onboarding/OnboardingAddWallet.tsx
```diff
@@ -0,0 +1,119 @@
...
+  return (
+    <section>
+      <div className="wordmark" />
+      <BackButton />
+      <div className="primary_wrap">
+        <div className="choice_wrap">
+          <h1 className="serif_header">Explore Tally</h1>
+          <div className="subtitle">
+            Use one of your existing addresses to explore Tally. You will not be
+            able to sign transaction untill you upgrade it.
```
Shadowfiend, line 15: until* but also we should probably drop mention of upgrading here as well.

ui/pages/Onboarding/OnboardingAddWallet.tsx
```diff
@@ -0,0 +1,119 @@
...
+            linkTo="/onboarding/viewOnlyWallet"
+          >
+            Add address
+          </SharedButton>
+        </div>
+        <div className="or_divider">or</div>
+        <div className="choice_wrap">
+          <h1 className="serif_header">Import seed</h1>
+          <div className="subtitle">
+            Use an existing Metamask seed or any web3 wallet seed.
```
Shadowfiend, line 29:
```suggestion
            Use an existing Metamask seed; other web3 wallets coming soon!
```

Something like this :grimacing:

Shadowfiend (approved):
Left a few notes, none blocking. Will merge later today.

ui/pages/Onboarding/OnboardingInfoIntro.tsx
henryboldi: I believe putting it in a separate element improves rendering. Once separated, the styles which change based on the state won't cause the other css to rerun. Possibly this is voodoo, but I've found it to be true.

Shadowfiend: I was mostly curious here so don't care too much, but in general if you have a performance improvement that isn't also an improvement in code structure, it's best left until there's an observed and known performance issue to fix. Also expressed as “make it work, make it right, make it fast” (Kent Beck) and “premature optimization is the root of all evil” (Donald Knuth) :)

ui/pages/Onboarding/OnboardingInfoIntro.tsx
```diff
@@ -0,0 +1,171 @@
...
+            onClick={() => {
+              if (activeStep < steps.length) {
+                setActiveStep(activeStep + 1)
+              } else {
+                setRedirectToAddWallet(true)
+              }
+            }}
+          >
+            {steps[activeStep - 1].buttonCopy}
+          </SharedButton>
```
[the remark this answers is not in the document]
Shadowfiend: Separately separately, why are we doing this state/redirect pattern instead of using react-router's `useNavigate` again?

Shadowfiend (approved):
Rest is gravy, gave this a test drive, let's roll!

PR #500, opened 2021-12-01
Don't bubble out empty token balances from Alchemy

Shadowfiend:
In certain cases, Alchemy returns token balance data with a token
balance of `0x`. In these cases, Alchemy has no data for the token
balance in question—however, that data may be available elsewhere.

Originally, this result was converted into a 0 balance, which would
result in updating the recorded balance elsewhere in the extension
to 0, manifesting in the token “disappearing”. getTokenBalances now
discards these entries entirely, letting any existing balance stand
on its own.

### DIFF
--- a/background/lib/alchemy.ts
+++ b/background/lib/alchemy.ts
@@ -200,27 +200,33 @@ export async function getTokenBalances(
   }
 
   // TODO log balances with errors, consider returning an error type
-  return json.tokenBalances
-    .filter(
-      (
-        b
-      ): b is typeof json["tokenBalances"][0] & {
-        tokenBalance: Exclude<
-          typeof json["tokenBalances"][0]["tokenBalance"],
-          null
-        >
-      } => b.error === null && b.tokenBalance !== null
-    )
-    .map((tokenBalance) => {
-      let balance = tokenBalance.tokenBalance
-      if (balance.length > 66) {
-        balance = balance.substring(0, 66)
-      }
-      return {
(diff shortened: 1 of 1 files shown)


background/lib/alchemy.ts
```diff
@@ -211,14 +211,15 @@ export async function getTokenBalances(
         >
       } => b.error === null && b.tokenBalance !== null
     )
+    .filter(({ tokenBalance }) => tokenBalance !== "0x")
```
mhluongo, line 214: I'm confused... how can we tell if the balance is actually `0`?

mhluongo (commented):
Questionssss

background/lib/alchemy.ts
mhluongo: Is it `0x0` rather than `0x`? :facepalm:

mhluongo: And if so... can we get that provider-specific behavior in a comment?

mhluongo:
I very nearly added `💭 ` to your PR title for you...

background/lib/alchemy.ts
Shadowfiend: It's not really provider specific. `0x` isn't a valid hex number--eg this is also how an ethereum node expresses an empty data slot for an address.

Shadowfiend: I added a note about this when I rebased against the fixes in #493.

mhluongo (approved):
:rocket:

PR #501, opened 2021-12-02
🗺 Replace onboarding coordination page w/ react router

henryboldi:
Previously, I had set up a funky page that managed the steps of the onboarding flow.

The impetus of this refactor was to make the shared back button, which operates on top of React Router, work within the onboarding steps. So previously, adding the regular shared back button to an onboarding page didn't work, and extra finagling needed to be done. No longer must we suffer! :)


ui/pages/Wallet.tsx
```diff
@@ -23,7 +23,7 @@ export default function Wallet(): ReactElement {
   // If an account doesn't exist, display view only
   // onboarding for the initial test release.
```
Shadowfiend, line 24: Sounds like this is probably incorrect now, right?

Shadowfiend (approved):
The magic strings make me a little twitchy with the router, which isn't new and is something to ponder for the future.

This looks good, let's do it.

Issue #363, opened 2021-10-28
Redux store refactor


Shadowfiend:

henryboldi:
To be more specific, I think mainly the `accounts` slice could use some love. The other ones seem to be in decent shape

Shadowfiend:
This ended up being a little general, but I think we said we felt good about the accounts slice now @henryboldi ? I'm thinking we close here and open more specific issues once we have more specific things we have opinions about.

henryboldi:
Yes, agreed!

PR #502, opened 2021-12-02
Wen Account Selector: Fill in the rest of the account selector data, adjust the wallet tab to only display current account

Shadowfiend:
This PR fils in the remaining data in the account selector, as
well as adjusting the wallet and single asset tabs to only display
balances and activity related to the currently-selected account. In
the process, almost all of the combined activity-related code in the
accounts slice is removed, since it was meant to populate this view,
which was never meant to display data across accounts.

The other large refactor here was the introduction of the
`accountsSelectors` file, following @henryboldi's example with the
`activitiesSelectors` file. Both of these are exposed at the top level
via a re-exporting `background/redux-slices/selectors` package that
exposes all internal selectors for the UI to use, which has allowed
some cleanup in typing dependencies to take place (in particular,
allowing the selectors to reference the full state without having to
resort to redeclaring it to avoid circular dependencies).

A few additional tweaks include trying to resolve avatars where
possible, changing the fixed Trezor header to display read-only mode
(for now), and not displaying a name when an account has no associated
ENS name. Also, the single asset page no longer displays placeholder
“this asset is on Arbitrum” language, which was confusing users and is
rooted in unimplemented functionality. These elements are hidden for
now.

Updated account selector

### DIFF
--- a/background/redux-slices/index.ts
+++ b/background/redux-slices/index.ts
@@ -8,7 +8,7 @@ import swapReducer from "./0x-swap"
 import transactionConstructionReducer from "./transaction-construction"
 import uiReducer from "./ui"
 
-export default combineReducers({
+const mainReducer = combineReducers({
   account: accountsReducer,
   assets: assetsReducer,
   activities: activitiesReducer,
@@ -17,3 +17,7 @@ export default combineReducers({
   transactionConstruction: transactionConstructionReducer,
   ui: uiReducer,
 })
+
+export default mainReducer
+
+export type RootState = ReturnType<typeof mainReducer>
--- a/background/redux-slices/selectors/accountsSelectors.ts
+++ b/background/redux-slices/selectors/accountsSelectors.ts
@@ -32,21 +32,6 @@ export const selectAccountAndTimestampedActivities = createSelector(
   getAssetsState,
   selectHideDust,
   (account, assets, hideDust) => {
-    // Derive activities with timestamps included
-    const activity = account.combinedData.activity.map((activityItem) => {
-      const isSent =
-        activityItem.from.toLowerCase() ===
-        Object.keys(account.accountsData)[0].toLowerCase()
-
-      return {
-        ...activityItem,
-        ...(activityItem.blockHeight && {
-          timestamp: account?.blocks[activityItem.blockHeight]?.timestamp,
-        }),
-        isSent,
-      }
-    })
-
     // Keep a tally of the total user value; undefined if no main currency data
     // is available.
     let totalMainCurrencyAmount: number | undefined
@@ -102,10 +87,8 @@ export const selectAccountAndTimestampedActivities = createSelector(
               desiredDecimals
             )
           : undefined,
-        activity: account.combinedData.activity,
       },
       accountData: account.accountsData,
-      activity,
     }
   }
 )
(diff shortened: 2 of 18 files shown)


Shadowfiend:
There's an issue remaining with the back button on the single asset page, but it's present on main due to the onboarding flow changes that made it `position: fixed`. I should have flagged that, `position: fixed` in a shared component is always a recipe for disaster down the line haha.

I tried to fix it up but ran into some walls… @henryboldi could use some help with it, maybe as a separate PR?

henryboldi:
> I should have flagged that, position: fixed in a shared component is always a recipe for disaster down the line haha. I tried to fix it up but ran into some walls… @henryboldi could use some help with it, maybe as a separate PR?

Totally my bad! Yes, I'll open up a separate PR with a fix @Shadowfiend

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -0,0 +1,225 @@
...
+            mainCurrencySymbol,
+            totalMainCurrencyAmount,
+            desiredDecimals
+          )
+        : undefined,
+    }
+  }
+)
+
+export type AccountTotal = {
```
henryboldi, line 163: The `AccountTotal` naming is a tad confusing to me. To someone unfamiliar with the codebase, it might give the impression that it's some sort of total amount

henryboldi (changes requested):
This is *chef's kiss*! A few questions and comments though,
1) After adding a new account, the wallet should probably switch to it by default

2) The avatar border should be rounded
Avatar

3) How about hiding this "add address" button? It could be confusing 
add address

4) How do we feel about hiding the notifications tab in this PR?

Shadowfiend:
1. Yes 2. Yes but not doing it for this release in the interest of time 3. Yes 4. Yes.

background/redux-slices/selectors/accountsSelectors.ts
Shadowfiend: It's the account info + it's total balance, which is in line with how we've done eg AccountNetwork.

Would love a better name though--I don't love that pattern, but I couldn't come up with anything great after giving it some thought. 

Either way a jsdoc couldn't hurt.

Shadowfiend:
Ended up doing all of them + fixing a React console error. Notably, wanting to select the newly-added account led to also removing some hacky code that was in use for initial wallet startup :tada:

mhluongo (approved):
Non-controversial, LGTM!

background/redux-slices/index.ts
```diff
@@ -17,3 +17,7 @@ export default combineReducers({
   transactionConstruction: transactionConstructionReducer,
   ui: uiReducer,
 })
+
+export default mainReducer
+
+export type RootState = ReturnType<typeof mainReducer>
```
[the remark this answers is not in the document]
Shadowfiend: It's overall better, otherwise we'd have to update every time we added a slice 🤷 DRY and all.

henryboldi:
Should this close #363?

Shadowfiend:
I'm interested in your take on that. I didn't set out to refactor the accounts slice as a goal, so there may still be some work to do if it's considered from that perspective.

PR #503, opened 2021-12-03
⚙️ Update settings tab to community edition design

henryboldi:
Since we don't have many settings yet — one to be exact, haha — we're filling up space with a CTA to our Discord server. Chloe mentioned that the Discord CTA button's label and shape make it uncleared that it's a button. And I know Vlad was leaning towards a different version anyway. That said, I believe a replacement for this design hasn't been finalized yet. So while we ulti<i>mutt</i>ly are deciding against this version, I think it's worth including for now!

settings2


ui/pages/Menu.tsx
```diff
@@ -111,51 +63,69 @@ export default function Menu(): ReactElement {
...
+            <h2 className="serif_header">Community release!</h2>
+            <p>Join our discord to give us feedback!</p>
+            <button
+              type="button"
+              aria-label="discord"
+              className="mega_discord_chat_bubble_button"
+              onClick={() => {
+                window.open(`https://chat.tally.cash/`, "_blank")?.focus()
+              }}
+            />
```
Shadowfiend, line 82: Could this not be an `<a href="https://chat.tally.cash/" target="_blank">`? Worth flagging because I imagine we'll have the link regardless.

Shadowfiend (approved):
One small note, but nothing that prevents a solid :shipit:

PR #505, opened 2021-12-03
Make blockprices run at app start

wojciech-turek:
The block prices should be pulled immediately after we load the app so there isn't a window of 2 minutes when fees aren't ready.


Shadowfiend (approved):
Looks good, does what it says on the tin 🥫

PR #506, opened 2021-12-03
ARCHIVED: 🐝  the change: let's give our beloved OGs some dApp goodies

greg-nagy:
This PR will integrate the window-provider and provider-bridge into the published extension. 
The `window-provider` is still behind a feature flag so it does not introduce issues until the integration is ready. 
To enable the integration the `isTallyWindowProviderEnabled` has to be set to `true` in the local storage of the given page.

* moved the content script into the main manifest.json — solving the CORS issues as well
* added typing for the window provider because that's always a cosy feeling
* added window.tally
* switched to the websocket provider so subscriptions can also work

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -387,7 +387,8 @@ export default class ChainService extends BaseService<Events> {
   }
 
   async send(method: string, params: unknown[]) {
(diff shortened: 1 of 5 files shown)


greg-nagy:
This PR had to have an emoji in the branch name too. There was no other way :D

PR #510, opened 2021-12-03
👻 Fix back button gore

henryboldi:
I had committed the sin of using a fixed position in a shared component. This ended up breaking its display on the single asset page. No longer must we suffer after PR!

For now, it leaves the "back" button text in for all uses. We could add a prop to turn this on and off.


Shadowfiend (approved):
Yussssss

PR #511, opened 2021-12-03
🚧 Temporarily feature flag navigating to import seed

henryboldi:
Entirely hiding the add seed choice makes the onboarding feel strange. It'd feel redundant going from "Explore Tally" to another page that says "Explore Tally" just with an input field. Adding accounts from the account switcher would encounter this same issue.

As a solution, I've given the add seed button the disabled appearance and added an alert. The alert is not ideal, but this feature flag is very temporary. 🤞

Disable add seed

Alert


Shadowfiend (approved):
Good enough for government work. 

I need to test drive this before merging, so will have to wait until later.

ui/pages/Onboarding/OnboardingAddWallet.tsx
```diff
@@ -32,8 +32,18 @@ export default function OnboardingAddWallet(): ReactElement {
           </div>
           <SharedButton
             type="primary"
-            linkTo="/onboarding/importMetamask"
+            isDisabled
```
Shadowfiend, line 35: `isDisabled` should really also be tied to `procss.env.HIDE_ADD_SEED === "true"`.

```suggestion
            isDisabled={procss.env.HIDE_ADD_SEED === "true"}
```

But not going to block this for a dev tweak obv.

wojciech-turek:
@henryboldi @Shadowfiend  The button is disabled visually but still works fine for those to dare to click anyway haha
ezgif com-gif-maker (1)

henryboldi:
@wojciech-turek, I just tested it locally and it seems to be working 🤔

You're certain `HIDE_ADD_SEED=true`?

wojciech-turek:
100%, I just reinstalled it and works same as the vid above lol

henryboldi:
Dang it 😆

I'm not sure how to reproduce it, but I'll investigate. Thank you

henryboldi:
I discussed this with @wojciech-turek, and I believe restarting `yarn start` did the trick.

Issue #423, opened 2021-11-15
Wallet Locking and Unlocking


Shadowfiend:
Implementation notes: 
- We only ask for a pw when you try to import a wallet. No reason to have pw with read-only wallets.
- Wallet lock/unlock is for transaction signing/keyring interactions, not for the whole wallet.
- The wallet always displays balances, etc. We *may* spin up a separate feature to allow toggling visibility of these, but it is out of scope for this issue.
- A wallet unlock lasts for one “session”. Definition of “session” not yet decided.

PR #517, opened 2021-12-06
Custodian: Add keyring service autolock and track status in keyrings slice

Shadowfiend:
From the first commit message:

The service supports autolocking depending on internal and external
idle times. Currently both are set to 30 minutes, but although the
markOutsideActivity method exists on the service, it is not yet being
called by any external caller, so the keyring is locked for max 30
minutes no matter what.

This PR also bumps our PBKDF2 iterations per the Least Authority audit.

See #423 .

### DIFF
--- a/background/redux-slices/keyrings.ts
+++ b/background/redux-slices/keyrings.ts
@@ -16,11 +16,13 @@ type Keyring = {
 type KeyringsState = {
   keyrings: Keyring[]
   importing: false | "pending" | "done"
+  status: "locked" | "unlocked" | "uninitialized"
 }
 
 export const initialState: KeyringsState = {
   keyrings: [],
   importing: false,
+  status: "uninitialized",
 }
 
 export type Events = {
@@ -47,6 +49,8 @@ const keyringsSlice = createSlice({
   name: "keyrings",
   initialState,
   reducers: {
+    keyringLocked: (state) => ({ ...state, status: "locked" }),
+    keyringUnlocked: (state) => ({ ...state, status: "unlocked" }),
     updateKeyrings: (state, { payload: keyrings }: { payload: Keyring[] }) => ({
       ...state,
       keyrings,
@@ -69,7 +73,8 @@ const keyringsSlice = createSlice({
   },
 })
 
-export const { updateKeyrings } = keyringsSlice.actions
+export const { updateKeyrings, keyringLocked, keyringUnlocked } =
+  keyringsSlice.actions
 
 export default keyringsSlice.reducer
 
--- a/background/services/keyring/encryption.ts
+++ b/background/services/keyring/encryption.ts
@@ -86,7 +86,7 @@ export async function deriveSymmetricKeyFromPassword(
     {
       name: "PBKDF2",
       salt: encoder.encode(salt),
-      iterations: 10000,
+      iterations: 1000000,
       hash: "SHA-256",
     },
     derivationKey,
--- a/background/tests/keyring-integration.test.ts
+++ b/background/tests/keyring-integration.test.ts
@@ -1,7 +1,13 @@
 import "mockzilla-webextension"
 
 import { webcrypto } from "crypto"
-import KeyringService, { Keyring } from "../services/keyring"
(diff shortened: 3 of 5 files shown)


background/redux-slices/keyrings.ts
```diff
@@ -16,11 +16,13 @@ type Keyring = {
 type KeyringsState = {
   keyrings: Keyring[]
   importing: false | "pending" | "done"
+  status: "locked" | "unlocked" | "uninitialized"
```
mhluongo, line 19: Any value to making this an enum?

background/services/keyring/encryption.ts
```diff
@@ -86,7 +86,7 @@ export async function deriveSymmetricKeyFromPassword(
     {
       name: "PBKDF2",
       salt: encoder.encode(salt),
-      iterations: 10000,
+      iterations: 1000000,
```
mhluongo, line 89: Phwew :sweat_smile:

background/tests/keyring-integration.test.ts
```diff
@@ -64,12 +73,18 @@ function expectBase64String(
   )
 }
 
+const mockAlarms = (mock: MockzillaDeep<Browser>) => {
+  mock.alarms.create.mock(() => ({}))
+  mock.alarms.onAlarm.addListener.mock((_, __) => ({}))
```
mhluongo, line 78: eslint disable?

background/tests/keyring-integration.test.ts
```diff
@@ -274,7 +291,7 @@ describe("KeyringService when saving keyrings", () => {
     })
   })
 
-  it("loads encrypted data at instantiation time", async () => {
+  it.only("loads encrypted data at instantiation time", async () => {
```
mhluongo, line 294: :thinking:

background/tests/keyring-integration.test.ts
```diff
@@ -315,3 +332,129 @@ describe("KeyringService when saving keyrings", () => {
...
+    expect(service.locked()).toEqual(false)
+
+    callAutolockHandler(MAX_OUTSIDE_IDLE_TIME - 10)
+    expect(service.locked()).toEqual(false)
+
+    callAutolockHandler(MAX_OUTSIDE_IDLE_TIME)
+    expect(service.locked()).toEqual(true)
+  })
+
+  it.each([
```
mhluongo, line 389: I need to learn more about our testing framework it appears

background/tests/keyring-integration.test.ts
```diff
@@ -315,3 +332,129 @@ describe("KeyringService when saving keyrings", () => {
...
+        await service.importLegacyKeyring(validMnemonics.metamask[0])
+      },
+    },
+    {
+      action: "generating a keyring",
+      call: async () => {
+        await service.generateNewKeyring(KeyringTypes.mnemonicBIP39S256)
+      },
+    },
+  ])("will bump keyring activity idle time when $action", async ({ call }) => {
```
mhluongo, line 413: ah this is cool

mhluongo (approved):
Excellent work. I've only done a static review, so trusting the build here on this merge and saving manual QA for the next PR.

Onward!

background/redux-slices/keyrings.ts
Shadowfiend: Still fuzzy on when which is better tbh. I like the simplicity of type-checked constants but also the declarativeness of enums.

background/services/keyring/encryption.ts
Shadowfiend: It's real noticeable in practice haha.

background/tests/keyring-integration.test.ts
Shadowfiend: ... Yeah this is not good. I thought eslint yelled at us about it, need to make sure it starts doing so.

background/tests/keyring-integration.test.ts
Shadowfiend: No I think I just forgot my JS here. Can just drop the params.

PR #518, opened 2021-12-06
Move fee selection logic to NetworkFeesChooser

wojciech-turek:
Since we want network fees chooser component to be reusable
in different parts of the app, this PR makes the chooser
handle most of the logic internally so it doesn't have to be passed
in every time.

Chooser now tracks its own active index and remembers after being dismissed.


background/redux-slices/transaction-construction.ts
```diff
@@ -17,18 +17,20 @@ const enum TransactionConstructionStatus {
...
   status: TransactionConstructionStatus
   transactionRequest?: EIP1559TransactionRequest
-  estimatedFeesPerGas: {
-    baseFeePerGas: bigint
-    instant: BlockEstimate | undefined
-    express: BlockEstimate | undefined
-    regular: BlockEstimate | undefined
-  } | null
+  estimatedFeesPerGas:
+    | {
```
wojciech-turek, line 21: Forced by prettier, not sure why.

background/services/chain/utils.ts
```diff
@@ -73,6 +73,7 @@ export function ethersTxFromSignedTx(
     to: tx.to,
     from: tx.from,
     data: tx.input || "",
+    type: tx.type,
```
wojciech-turek, line 76: Not related in the context of this PR description, its for the transaction type being passed correctly.

background/redux-slices/transaction-construction.ts
Shadowfiend: We need not understand why, that's the beauty of letting the code formatter worry about formatting :grin:

background/redux-slices/transaction-construction.ts
```diff
@@ -108,7 +110,34 @@ export default transactionSlice.reducer
 export const selectEstimatedFeesPerGas = createSelector(
   (state: { transactionConstruction: TransactionConstruction }) =>
     state.transactionConstruction.estimatedFeesPerGas,
-  (gasData) => gasData
+  (gasData) => {
+    if (
+      gasData?.instant &&
+      gasData.express &&
+      gasData.regular &&
+      gasData.baseFeePerGas
```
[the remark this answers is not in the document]
Shadowfiend: Actually… These are maybe already always `bigint`??

I think we can drop the `number |` in the `BlockPrices` type and remove all of this type fiddling.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -46,6 +78,72 @@ export default function NetworkFeesChooser({
...
+        ]
+        setGasOptions(updatedGasOptions)
+        setSelectedGas(updatedGasOptions[0])
+      }
+    }
+  }, [estimatedFeesPerGas, setSelectedGas])
+
+  useEffect(() => {
+    updateGasOptions()
+  }, [updateGasOptions])
```
Shadowfiend, line 139: Why not just update the gas options in line, without any callback/effect intermediaries? It feels like a very cheap operation.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -46,6 +78,72 @@ export default function NetworkFeesChooser({
...
+            gwei: parseInt(
+              formatUnits(
+                regular.maxFeePerGas + regular.maxPriorityFeePerGas,
+                "gwei"
+              ),
+              10
+            ),
+            dollarValue: "$??",
+            maxFeePerGas: regular.maxFeePerGas,
+            maxPriorityFeePerGas: regular.maxPriorityFeePerGas,
```
Shadowfiend, line 100: Why don't we extract this to a function, since we're doing the same thing for each option?

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -46,6 +78,72 @@ export default function NetworkFeesChooser({
...
+          {
+            name: "Regular",
+            confidence: `${regular.confidence}%`,
+            gwei: parseInt(
+              formatUnits(
+                regular.maxFeePerGas + regular.maxPriorityFeePerGas,
+                "gwei"
+              ),
+              10
+            ),
```
Shadowfiend, line 97: We should either use a `BigInt` or stick to a string here, tbh.

ui/pages/Send.tsx
```diff
@@ -70,106 +62,50 @@ export default function Send(): ReactElement {
...
-              gasLimit={gasLimit}
-              setGasLimit={setGasLimit}
-              saveUserGasChoice={saveUserGasChoice}
-            />
-          ) : (
-            <></>
-          )}
+          <NetworkFeesChooser
+            setFeeModalOpen={setFeeModalOpen}
+            setSelectedGas={setSelectedGas}
```
Shadowfiend, line 103: Feels like these should be events (`onSave`, `onEstimatedFeePerGasSelected`), maybe?

ui/pages/Send.tsx
```diff
@@ -70,106 +62,50 @@ export default function Send(): ReactElement {
...
+            selectedGas={selectedGas}
```
Shadowfiend, line 104: Let's continue to use the right terminology here, e.g. `selectedEstimatedFeePerGas`.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -14,23 +15,54 @@ interface GasOption {
...
-  gasOptions: GasOption[]
-  activeFeeIndex: number
-  handleSelectGasOption: (number: number) => void
+  setFeeModalOpen: React.Dispatch<React.SetStateAction<boolean>>
+  setSelectedGas: React.Dispatch<React.SetStateAction<GasOption | undefined>>
+  selectedGas?: GasOption
   gasLimit: string | number
   setGasLimit: React.Dispatch<React.SetStateAction<string>>
-  saveUserGasChoice: () => void
+  estimatedFeesPerGas: {
```
Shadowfiend, line 23: Maybe we can name the estimated fees type (`type EstimatedFeesPerGas = { baseFeePerGas: bigint, ... }`) in `transaction-construction` and reference it here?

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -14,23 +15,54 @@ interface GasOption {
 }
 
 interface NetworkFeesChooserProps {
-  gasOptions: GasOption[]
-  activeFeeIndex: number
-  handleSelectGasOption: (number: number) => void
+  setFeeModalOpen: React.Dispatch<React.SetStateAction<boolean>>
+  setSelectedGas: React.Dispatch<React.SetStateAction<GasOption | undefined>>
+  selectedGas?: GasOption
```
Shadowfiend, line 20: If we name the `EstimatedFeesPerGas` type, what about using `Omit<"baseFeePerGas", keyof EstimatedFeesPerGas>` here? That is, for the selected option to be `instant`/`express`/`regular` instead of a full object? The caller doesn't know anything about `GasOption`, nor really should they. They should deal in `EstimatedFeesPerGas` and its sub types, IMO.

Shadowfiend (changes requested):
Big realization that we're doing a lot of unnecessary type fiddling. I think if we remove all of that it'll clean this PR up a bit.

Also a few other notes on the interface for the fees chooser.

background/redux-slices/transaction-construction.ts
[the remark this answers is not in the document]
wojciech-turek: I think they are always `bigint`, true! I will make necessary changes.

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: I was thinking it would save us some calculation on re-render since this component is re-rendering a lot due to the moving of fees timeout. If you think the performance won't be affected I am happy to remove it.

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: Done, in my next commit you will see the result.

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: Modified gwei to be a string, `formatUnits` return value with 2 decimals so I had to slice them off.

ui/pages/Send.tsx
wojciech-turek: I went with `onSaveGasChoice` as I find it most descriptive, since we're in a chooser :)

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: Great idea, reduced code length :)

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: This made sense, I went ahead and it turned out to be a bigger refactor but I think I've succeeded (at least partially 😅)

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -8,44 +12,25 @@ import SharedInput from "../Shared/SharedInput"
...
+  price: bigint
+  maxFeePerGas: bigint
+  maxPriorityFeePerGas: bigint
 }
 
 interface NetworkFeesChooserProps {
   setFeeModalOpen: React.Dispatch<React.SetStateAction<boolean>>
-  setSelectedGas: React.Dispatch<React.SetStateAction<GasOption | undefined>>
-  selectedGas?: GasOption
+  onSaveGasChoice: React.Dispatch<React.SetStateAction<BlockEstimate>>
```
Shadowfiend, line 24: Let's just make this a `(BlockEstimate)=>void`. The chooser shouldn't care if the caller is setting state or not.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -78,69 +68,50 @@ export default function NetworkFeesChooser({
...
+      70: "Regular",
+      95: "Express",
+      99: "Instant",
+    }
+    return {
+      name: names[confidence],
+      confidence: `${confidence}`,
+      gwei: Number(
+        formatUnits(option.maxFeePerGas + option.maxPriorityFeePerGas, "gwei")
+      ).toFixed(),
```
Shadowfiend, line 83: This is a dangerous thing to do in general, as the parse to `Number` can lose precision. Are you just trying to lose the fractional component here? If so, it may be worth leaning on the fact that ethers always uses a `.` for decimals and just do a `.split(".")[0]` here.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -78,69 +68,50 @@ export default function NetworkFeesChooser({
...
+      maxPriorityFeePerGas: option.maxPriorityFeePerGas,
+    }
+  }
+
   const updateGasOptions = useCallback(() => {
     if (estimatedFeesPerGas) {
       const instant = estimatedFeesPerGas?.instant
       const express = estimatedFeesPerGas?.express
       const regular = estimatedFeesPerGas?.regular
       if (!!instant && !!express && !!regular) {
```
Shadowfiend, line 96: Again, we should be using `typeof === "undefined"` here. Cryptic fiddling like `!!` is the real underlying reason why we avoid type coercion-based checks, tbh.

ui/components/NetworkFees/NetworkFeesChooser.tsx
Shadowfiend: Got it. I mean, in general it feels like premature optimization, I really don't care *that* much hehe.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -78,69 +68,50 @@ export default function NetworkFeesChooser({
     }
   })
 
+  const getGasOptionFormatted = (option: BlockEstimate) => {
```
Shadowfiend, line 71: Perhaps even `formatBlockEstimate`? :)

ui/pages/Send.tsx
Shadowfiend: I would challenge the use of `save` here, as we're not really saving anything. `select` (i.e., `onSelectGasChoice`) feels like the correct choice, and in line with other selection events in other components.

I would also ask why the `NetworkFeesChooser` is selecting a `GasChoice`—we should either be choosing a fee, or it should be a gas chooser :grin:

ui/components/NetworkFees/NetworkFeesChooser.tsx
Shadowfiend: There's still some stuff I don't love about passing around the whole `BlockEstimate`, but I see why we're doing it. I'll noodle on it for later, maybe there'll be an interesting refactor down the line.

Shadowfiend (approved):
I left a few additional notes, but nothing is blocking. Let's 🚢

ui/pages/Send.tsx
wojciech-turek: That makes sense 🙈  ill do some ranaming

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: Thanks, but isn't that even less precise? When using `split` `2.99` will show as `2`? ``Math.round()` is a viable option?

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: Hahah, yeah, I like it.

Issue #391, opened 2021-10-29
dApp Permission request UI tracking issue


Shadowfiend:
This task is to track the permission related UI tasks and the integration of this window with the `ProviderBridgeService`. 

This has an overlapping scope with the `ProviderBridgeService`. Anything in the service that is related to permission handling should be tracked here. 

The reason for this distinction is the pragmatic consideration, that the feature can "work" without this part implemented at all. Should not - especially after release - but can. This will be important in the upcoming weeks, so I tried to make this as clear cut as possible.

PR #519, opened 2021-12-06
❌ 🧙🏻‍♂️ You shall not pass: dApp permission request UI initial setup

### LINKED ISSUE #482: dApp permission request UI initial setup
- [x] window open
- [x] redux communication
- [x] initial url storage
- [x] allow button

greg-nagy:
CleanShot 2021-12-06 at 12 19 33

# Purpose
The purpose of this PR is to set up the initial skeleton, for the dApp permission requests. 

# Main changes
- Implemented a pop up window on `eth_requestAccounts` and it's positioned to the top right corner of the current window
- App reads `?page=<page-name>` on startup, and if defined loads the specified page instead of the `/` route
- dummy permission window
- `ProviderBridgeService` <> `background/main.ts` <> `providerBridgeSlice.ts` <> `Permission.tsx` communication fow
- in memory permission store in `ProviderBridgeService`

# Demo
CleanShot 2021-12-08 at 13 36 01

# Test
This can be tested with the [web3-react-official demo](https://github.com/tallycash/tally-dapp-integration-samples)

## Setup
- start the demo app 
```
git clone https://github.com/tallycash/tally-dapp-integration-samples.git
cd tally-dapp-integration-samples/web3-react-official
yarn install
yarn start
```
- open http://localhost:3000
- set `isTallyWindowProviderEnabled` to true in localstorage 
- reload the page 
- press injected button

See #391 
Closes #482


ui/pages/Popup.tsx
```diff
@@ -22,54 +22,81 @@ import EarnDeposit from "./EarnDeposit"
 import Menu from "./Menu"
 import Send from "./Send"
 import Swap from "./Swap"
+import Permission from "./Permission"
+
+function transformLocation(inputLocation: Location): Location {
+  // Poor man's query string parse in a somewhat controlled environment
+  // The inputLocation is not populated with the actual query string — even though it should be
+  // so I need to grab it from the window
+  const maybePage = window.location.href.split("page=")[1]
```
Shadowfiend, line 31: You can use [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get) to do this with proper parsing.

greg-nagy: https://github.com/tallycash/tally-extension/commit/8184ee8030ac3fa6f5bb48ee5493f02d1416b602

background/services/chain/index.ts
```diff
@@ -387,7 +388,7 @@ export default class ChainService extends BaseService<Events> {
     }
   }
 
-  async send(method: string, params: unknown[]): Promise<unknown> {
+  async send(method: string, params: RPCRequest["params"]): Promise<unknown> {
```
Shadowfiend, line 391: Not sure we should set this parameter based on the provider bridge's requirements. `ChainService` shouldn't know about the provider bridge, it's just exposing access to the chain functionality irrespective of who uses it.

background/services/provider-bridge/index.ts
```diff
@@ -1,13 +1,19 @@
...
+  PortRequestEvent,
+  PortResponseEvent,
+  RPCRequest,
+} from "@tallyho/provider-bridge-shared"
 import { ServiceCreatorFunction, ServiceLifecycleEvents } from ".."
 import logger from "../../lib/logger"
 import BaseService from "../base"
 import InternalEthereumProviderService from "../internal-ethereum-provider"
 import { getOrCreateDB, ProviderBridgeServiceDatabase } from "./db"
+import { PermissionRequest } from "../../redux-slices/provider-bridge"
```
Shadowfiend, line 13: Redux should depend on services, not vice versa.

Really redux and services both should depend on types at the top of `background`, but there are a couple of cases where that's getting dicey without some more domain design we don't have time for yet.

background/services/provider-bridge/index.ts
```diff
@@ -51,36 +65,93 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
+        await ProviderBridgeService.showDappConnectWindow()
+
+        // @ts-expect-error unblockResolve is assigned value in the Promise but don't know how to convince ts about this (assigning default value did not feel right)
+        this.pendingPermissions[permissionRequest.url] = blockResolve
+        await blockUntilUserAction
+      }
+
+      // TBD @Antonio:
+      // I copied the way MM works here — I return `result: []` when the url does not have permission
+      // According to EIP-1193 it should return a `4100` ProviderRPCError but felt that dApps probably does not expect this.
```
Shadowfiend, line 114: My default is to break the dApps and walk it back if we hear about it and can't get them to update their behavior, given the exceptional case this scenario represents.

background/services/provider-bridge/index.ts
```diff
@@ -125,23 +114,37 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
       delete this.allowedPages[permission.url]
-      this.pendingPermissions[permission.url]("Time to move on")
-      delete this.pendingPermissions[permission.url]
+      this.#pendingPermissionsRequests[permission.url]("Time to move on")
+      delete this.#pendingPermissionsRequests[permission.url]
     }
   }
 
-  async checkPermission(url: string): Promise<boolean> {
+  async permissionCheck(url: string): Promise<boolean> {
```
Shadowfiend, line 147: Not sure why we're moving from verb-noun order here but… We should not unless I'm missing something? :grimacing:

`checkPermission`, `denyOrRevokePermission`, `grantPermission`, `requestPermission` should be the right name for the methods, `permissionRequest` and `PermissionRequest` should be the right names for events and data, etc.

background/services/chain/index.ts
greg-nagy: Wanted to have a single type to flow through the communcation, but it should not bleed into here, so I removed it. Thanks

background/services/provider-bridge/index.ts
greg-nagy: Moved this type to `provider-bridge-shared` because this is truly a cross cutting code.

background/services/provider-bridge/index.ts
greg-nagy: Was pondering about it, went with the other because it was there, but this feels better.

background/services/provider-bridge/index.ts
```diff
@@ -51,36 +58,108 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
+    const blockUntilUserAction = new Promise((resolve) => {
+      blockResolve = resolve
+    })
+
+    this.emitter.emit("permissionRequest", permissionRequest)
+    await ProviderBridgeService.showDappConnectWindow()
+
+    // ts compiler does not know that we assign value to blockResolve so we need to tell him
+    this.#pendingPermissionsRequests[permissionRequest.url] = blockResolve!
+    return blockUntilUserAction
```
Shadowfiend, line 133: This is working accidentally I believe. The promise constructor isn't guaranteed to run synchronously, and it's only because of the intervening `await` that this gets set correctly. We should set this inside the promise function and skip the intermediate variable. 

```suggestion
    this.emitter.emit("permissionRequest", permissionRequest)
    ProviderBridgeService.showDappConnectWindow()

    // This promise will be resolved by a call to grantPermission
    // or denyOrRevokePermission.
    return new Promise((resolve) => {
      this.#pendingPermissionsRequests[permissionRequest.url] = resolve
    })
```

I think this is overall clearer.

background/services/provider-bridge/index.ts
```diff
@@ -51,36 +58,108 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
+          favIconUrl,
+          state: "request",
+        }
+
+        const blockUntilUserAction = await this.requestPermission(
+          permissionRequest
+        )
+        await blockUntilUserAction
+
+        if (await this.checkPermission(url)) {
```
Shadowfiend, line 103: If you make the permissions request resolve with the actual permission outcome, you can tighten this up:

```suggestion
        if (await this.requestPermission(
          permissionRequest
        )) {
```

background/services/provider-bridge/index.ts
```diff
@@ -94,25 +173,19 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
-  //   this.emitter.emit("accountAccessRequestedByDAppID", dAppID)
-  //   return requestPromise
-  // }
+  static async showDappConnectWindow(): Promise<browser.Windows.Window> {
+    const { left = 0, top, width = 1920 } = await browser.windows.getCurrent()
+    const popupWidth = 400
+    const popupHeight = 600
+    const internalPageName = "permission"
+    return browser.windows.create({
+      url: `${browser.runtime.getURL("popup.html")}?page=${internalPageName}`,
```
Shadowfiend, line 182: Ultimately I think we're going to want to do this through redux (in line with some of what @henryboldi wants to do in terms of preserving UI state), but we can park that for later.

ui/pages/Popup.tsx
```diff
@@ -25,64 +25,84 @@ import Swap from "./Swap"
...
+
+function transformLocation(inputLocation: Location): Location {
+  // The inputLocation is not populated with the actual query string — even though it should be
+  // so I need to grab it from the window
+  const params = new URLSearchParams(window.location.search)
+  const maybePage = params.get("page")
+
+  return {
+    ...inputLocation,
+    pathname: maybePage ? `/${maybePage}` : inputLocation.pathname,
```
Shadowfiend, line 38: We should probably whitelist the pages that can be injected this way while we're using this approach. Something about allowing this makes me itchy.

background/services/provider-bridge/index.ts
```diff
@@ -51,36 +58,108 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
+
+    this.emitter.emit("permissionRequest", permissionRequest)
+    await ProviderBridgeService.showDappConnectWindow()
+
+    // ts compiler does not know that we assign value to blockResolve so we need to tell him
+    this.#pendingPermissionsRequests[permissionRequest.url] = blockResolve!
+    return blockUntilUserAction
+  }
+
+  async grandPermission(permission: PermissionRequest): Promise<void> {
```
Shadowfiend, line 136:
```suggestion
  async grantPermission(permission: PermissionRequest): Promise<void> {
```

background/redux-slices/index.ts
```diff
@@ -7,6 +7,7 @@ import keyringsReducer from "./keyrings"
 import swapReducer from "./0x-swap"
 import transactionConstructionReducer from "./transaction-construction"
 import uiReducer from "./ui"
+import providerBridgeReducer from "./provider-bridge"
```
Shadowfiend, line 10: This reducer feels to me like it should be named something much more about its user function, eg dAppReducer or dAppPermissionsReducer. Thoughts?

background/redux-slices/provider-bridge.ts
```diff
@@ -0,0 +1,96 @@
...
+    await emitter.emit("permissionDenied", permission)
+    return permission
+  }
+)
+
+const providerBridgeSlice = createSlice({
+  name: "provider-bridge",
+  initialState,
+  reducers: {
+    newPermissionRequest: (
```
Shadowfiend, line 45: I believe we've just been naming these the bare type (eg `permissionRequest`).

background/services/provider-bridge/index.ts
greg-nagy: Loved the explicitness and the `checkPermission` is desinged to work from indexeddb and `await this.checkPermission(url)` had the additional meta meaning,  that the info has been persisted into the indexeddb and has been retrieved properly so there was no place for race conditions. 

But I also like your solution, let's give it a go and return back to this if needed.

background/services/provider-bridge/index.ts
```diff
@@ -51,36 +58,108 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
+
+  async denyOrRevokePermission(permission: PermissionRequest): Promise<void> {
+    if (this.#pendingPermissionsRequests[permission.url]) {
+      delete this.allowedPages[permission.url]
+      this.#pendingPermissionsRequests[permission.url]("Time to move on")
+      delete this.#pendingPermissionsRequests[permission.url]
+    }
+  }
+
+  async checkPermission(url: string): Promise<boolean> {
```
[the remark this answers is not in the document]
greg-nagy: This will work from indexeddb soon, so I wanted to have the final foorprint so only the internals of the funcition need to change when I get there.

background/services/provider-bridge/index.ts
greg-nagy: Was thinking about something similar —wanted to move the creation of the window into the `main.ts` — because it's somewhat a shared concern between the background and the UI. 

Don't know what Henry has in mind, but def let's get back to this, when we are there.

ui/pages/Popup.tsx
greg-nagy: Kind of proud of this :)

```
export const ALLOWED_QUERY_PARAM_PAGE = {
  permissions: "/permission",
  signTransaction: "/signTransaction",
  dappConnect: "/dapp-connect",
} as const

export type AllowedQueryParamPage =
  typeof ALLOWED_QUERY_PARAM_PAGE[keyof typeof ALLOWED_QUERY_PARAM_PAGE]

export function isAllowedQueryParamPage(
  url: unknown
): url is AllowedQueryParamPage {
  // The typing for Array.includes in `lib.es.2016.array.include.ts` does not make any sense -> Object.values<string>
  // interface Array<T> { ... includes(searchElement: T, fromIndex?: number): boolean; ...
  return Object.values<unknown>(ALLOWED_QUERY_PARAM_PAGE).includes(url)
}
```

background/services/provider-bridge/index.ts
greg-nagy: lol ... sniped it before got to this comment :) thanks

background/redux-slices/index.ts
greg-nagy: Tbh I find it really helpful to have a single nameing schema for the feature set that I am working on throughout the codebase. 

Your suggestion has it's merits but would prefer to keep the nameing to reduce my mental load if it's ok :)

background/redux-slices/provider-bridge.ts
greg-nagy: Get a nameing collision when wanted to do that. Will think about it later.

background/redux-slices/index.ts
Shadowfiend: I think it's fine for now. To clarify my thinking, for me redux slices need to be operating in the user domain, because they're exclusively UI code. The user will never think about a provider bridge, so it follows that the UI code generally wouldn't talk about a provider bridge—instead it's just thinking about dApp permissions, just like the UI components are DAppInfo, DAppConnectRequest, ... rather than ProviderBridgeInfo, etc.

background/redux-slices/provider-bridge.ts
Shadowfiend: Tbh I'm skeptical of our current approach, but do want to strive for consistency so we can potentially switch to something else… Consistently :grin:

Anyway, def fine for now.

background/redux-slices/index.ts
greg-nagy: Giving it a bit more thoughts and observing my feelings and concerns: I am not in favour of this nameing at all, I just a) don't see this as important for now — **definitely important later** b) I have 0 information about how we structure the domain for the ui, backend <> ui. 

If we could spend some time sync/async on the domain model or rough guidelines at least that I would feel more comfortable in questions like this. 

Things I have considered when designing this solution: 
- Shall I send in the url through a query param, or have a different html which prepopulas the state with the root url for the popup? Who holds the truth? window.location or state? 
- Is redux store a full reflection of the background OR is it to serve the ui and should be structured by the UI needs (re: store segregation / performance issue because we have a ton of data in store that UI does not need)
- Do we have a consistent rule of thumb to designing this?  Or  it's just my lack of tribal knowledge? 
- If I were to store the state of the popup window where would it go? In this slice? But popup will be opened from multiple places and it's a different domain concern from permission requests.
- Do I store the pages that have permission in this slice? If I do I need to maintain the data syncrhonised. If I don't I can have a single source of truth but pay in complexity of back and forth messaging. 
- Do we design for our current needs or do we have some kind of vision? 
- Do I put signing in this slice or that should live somewhere else? It belongs to dApps so it would make sense to track them here but I saw you implemented it in the internal provider. So it might make sense to use a different slice for that. 

All in all: in the current situation with so much concerns about what is the store for us I went with a structure that felt the safest — and easiest tbh. 

I am __**really passionate**__ aboout having at least rudementally defined domain boundaries and having an efficient store structure. In all my previous projects one of these caused 80% of headaches along the road and they became obious only when it was next to impossible to fix them.

Sorry for the lengthy response. I am just really passionate about getting this right and makeing it consistent as fast as possible but not sure on the how.

background/redux-slices/provider-bridge.ts
greg-nagy:
```
    this.providerBridgeService.emitter.on(
      "permissionRequest",
      (permissionRequest: PermissionRequest) => {
        this.store.dispatch(permissionRequest(permissionRequest))
      }
    )
```

When using the same nameing in the ui and in the background it collides at integration. How/where do we resolve this issue?

background/services/provider-bridge/index.ts
Shadowfiend: Not sure I follow, but will hold my tongue until the IndexedDB bits land.

background/redux-slices/index.ts
greg-nagy: https://github.com/tallycash/extension/pull/547/commits/289bf997dad81ec28e7749e3a62374f315930102

background/redux-slices/provider-bridge.ts
greg-nagy: https://github.com/tallycash/extension/pull/547/commits/289bf997dad81ec28e7749e3a62374f315930102

greg-nagy: > Tbh I'm skeptical of our current approach

And you were correct in that. Just needed some time to realize it.

PR #520, opened 2021-12-06
🤡 Fix back button position on view-only step

henryboldi:
It appears I missed fixing the back button position on the view-only input page. Here's the fix!

Before,
before

After,
after


Shadowfiend (approved):
Looks good :shipit:

PR #528, opened 2021-12-07
Affinity Transforms: Correctly group account types in account switcher

### LINKED ISSUE #524: Close the account selector when an account is switched
The account selector not closing after a new selection is confusing, and makes the user think the account didn't change. After a selection, close the selector and return to the wallet page.

Shadowfiend:
The design for the switcher is still not aligned with the
latest from @VladUXUI in Figma, but I need to park that for
now and move to the next thing.

Also included in this are a couple of snipes:
- `selectCurrentAccountActivitiesWithTimestamps` returns `undefined`
  if no current account exists, with callers handling this themselves.
  This is more accurate to what can happen in practice.
- The UI slice refers to `currentAccount` instead of `selectedAccount`.
- The account selector dismisses when an address is clicked.

145106493-604527a7-9cea-4c3b-9b84-959e4d4725b6

------

Closes #524.


mhluongo:
Undrafted this after what I think was sync permission from @Shadowfiend? :sweat_smile:

Shadowfiend:
Yep!

background/redux-slices/keyrings.ts
```diff
@@ -61,10 +61,20 @@ const keyringsSlice = createSlice({
...
     keyringUnlocked: (state) => ({ ...state, status: "unlocked" }),
-    updateKeyrings: (state, { payload: keyrings }: { payload: Keyring[] }) => ({
-      ...state,
-      keyrings,
-    }),
+    updateKeyrings: (state, { payload: keyrings }: { payload: Keyring[] }) => {
+      // When the keyrings are locked, we receive updateKeyrings with an empty
+      // list as the keyring service clears the in-memory keyrings. For UI
+      // purposes, however, we want to continue tracking the keyring metadata,
+      // so we ignore an empty list if the keyrings are locked.
```
henryboldi, line 68: What would happen if a user decides to remove their added accounts?

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -163,21 +164,29 @@ export const selectCurrentAccountBalances = createSelector(
...
+      .map(([address, accountData]) => {
         const shortenedAddress = `${address.slice(0, 6)}...${address.slice(-4)}`
 
         if (accountData === "loading") {
-          return { address, shortenedAddress }
+          return {
+            address,
+            shortenedAddress,
+            accountType: AccountType.ReadOnly,
+          }
```
henryboldi, line 189: Why does `accountData === "loading"` mean it's read-only in the account switcher category?

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -208,9 +217,16 @@ export const selectAccountTotals = createSelector(
           })
           .reduce((total, assetBalance) => total + assetBalance, 0)
 
+        const resolvedAccountType =
```
henryboldi, line 220: Why is this determined in the selector rather than the reducer?

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
```diff
@@ -72,39 +102,78 @@ function WalletName() {
...
+    dispatch(setCurrentAccount(address))
+  }
+
+  useEffect(() => {
+    if (
+      pendingSelectedAddress !== "" &&
+      pendingSelectedAddress === selectedAccount
+    ) {
+      onCurrentAddressChange(pendingSelectedAddress)
+      setPendingSelectedAddress("")
```
henryboldi, line 133: Why is `pendingSelectedAddress` needed here?

henryboldi (changes requested):
I left a couple of questions. My main concern is that we're doing too much heavy lifting in the selectors rather than the reducers. Firstly, relying on the selectors makes composability more difficult, since grabbing the enriched info in a new way needs to dubiously latch on to the existing selector. I suppose pulling the logic out of the selectors into shared utils helps. But why not save the enriched data to the store rather than run it on every update?

background/redux-slices/keyrings.ts
Shadowfiend: They'd need to unlock the keyrings first.

There's actually no way to remove accounts at the moment, to be clear, only a way to start with a clean slate (i.e., “erase” all accounts)—but the result is the same, the keyrings are unlocked when the event with the empty list goes out.

background/redux-slices/selectors/accountsSelectors.ts
Shadowfiend: Whoopsies, it absolutely doesn't! This should be running the same logic as `resolvedAccountType` below.

background/redux-slices/selectors/accountsSelectors.ts
Shadowfiend: Because it has to cross slices to conclude the account type–an account's data is stored in the accounts slice, while its type is (effectively) read from the keyrings slice.

We could hook up keyring-related events to the accounts slice, but it's not immediately clear to me that that is better in any material way.

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
Shadowfiend: We need to clean up state for the cycle so that the user can interact with it again. If the pending address remains set, then if the user tries to bring up the account selector again this effect will run again and `onCurrentAddressChange` will be called again (because `pendingSelectedAddress === selectedAccount` will be `true`). In practice that means the account selector will immediately be dismissed again 🙊

Shadowfiend:
> relying on the selectors makes composability more difficult, since grabbing the enriched info in a new way needs to dubiously latch on to the existing selector

Hmm… What's dubious about composing selectors instead of reducers?

The tradeoff for me is basically stored size vs runtime computation, but I don't see a significant composability tradeoff. I've been tilting towards runtime computation in my recent PRs because most of the enrichments and other determinations being made are either cacheable (by `createSelector`) or fairly cheap and cross slices, and a smaller store means cheaper serialization costs that we're having to pay in a few places. Charging selectors with inter-slice data coordination is also allowing for better separation of concerns, in particular leading to a bit less duplication of data and data handling (which tend to be opportunities for parts of the store to get out of sync due to bugs elsewhere).

henryboldi:
Good points, I stand corrected and informed 🙇‍♂️

I'll keep in mind the value of deriving data in the selectors. Reductively, given their efficiency, maybe I should think of selectors as creating a ✨free✨ new version of part of the store? Still, something about chunky selectors strikes me as tricky to deal with. There's an added complexity, in terms of dealing with similar yet different objects, when a foundational(?) selector isn't used consistently.

Anyway, the spacing here feels tight,
spacing
Particularly the space between "accounts" and "imported". This doesn't need to be blocking, but it feels funky. We could create a version that doesn't involve perfectly emulating the design, but also doesn't look like a bug. If you'd like, so you can move on to the next thing, I could apply a bit of makeup here in a follow-up PR.

henryboldi (approved):

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -163,21 +164,37 @@ export const selectCurrentAccountBalances = createSelector(
...
+  accountType: AccountType
   name?: string
   avatarURL?: string
   localizedTotalMainCurrencyAmount?: string
 }
 
-export const selectAccountTotals = createSelector(
+export type CategorizedAccountTotals = { [key in AccountType]?: AccountTotal[] }
+
+export const selectAccountTotalsByCategory = createSelector(
```
henryboldi, line 175: In retrospect, for orthogonality, would it make sense for `selectAccountTotalsByCategory` to be separate from and build on top of `selectAccountTotals`?

PR #529, opened 2021-12-07
Pablo Escrollar: Fix unexpected scrollbars in Firefox

### LINKED ISSUE #339: Fix scroll bars in Firefox
ROUNDED CORNERS
^ From @itsrachelfish

Shadowfiend:
Quite a few places were using 100vw where 100% was the correct thing to
use. Additionally, some overflow: scrolls were specified where letting
the browser auto-scroll when needed was the right thing to do.

Finally, explicitly add overflow-y: auto to the wallet asset/activity
panel so that the whole wallet screen doesn't scroll when many assets or
activities appear. That behavior may best be rolled back, depending on
UX preferences.

----

To test this on macOS, go to System Preferences -> General and switch the “Show scroll bars” option to “Always” to observe the issue pre this branch and see it gone post.

Closes #339.


mhluongo (approved):
Built and confirmed the fix on Firefox Dev Edition as well as Chromium.

PR #530, opened 2021-12-08
⚡️ dApp connection approval UI

henryboldi:
Here are the UI components to slot into what Greg has been working on with the dApp connection approval. Mostly this is ordinary, but I'd like to point out potential oddities,
- `AccountsNotificationPanelAccountItem` is turned into `SharedPanelAccountItem` and more generalized to display as the account info here.
- The popup window routing and additional HTML file in `/public` I imagine will get replaced by @greg-nagy's existing work.

connect-ui


ui/pages/DAppConnectRequest.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  const accountTotals = accountTotalsByCategory["read-only"]
+  if (typeof accountTotalsByCategory?.imported !== "undefined") {
+    accountTotals?.concat(accountTotalsByCategory?.imported)
+  }
+
+  const currentAccountTotal = accountTotals?.filter(
+    (accountTotal) => accountTotal?.address === currentAccount
+  )[0]
+  if (!currentAccountTotal) return <></>
```
mhluongo, line 62: Small nit but I try to avoid braceless if statements... small syntax issue that makes maintenance more difficult (later adding more code to the success block)

ui/pages/Popup.tsx
```diff
@@ -75,7 +76,11 @@ export default function Popup({ store }: { store: Store }): ReactElement {
             <Swap />
           </Route>
           <Route path="/">
-            <Wallet />
+            {document.location.href.includes("dapp-connect") ? (
```
mhluongo, line 79: This feels awkward, aren't we using different HTML entrypoints for this? Is there a way the approval HTML can push this into the JS rather than the JS auto-discovering a location string that might change?

Mostly feels weird because of how the extension works, btw. I'd be surprised if I changed the HTML file name and this JS broke

mhluongo (commented):
Took a quick pass. Small comments

ui/pages/Popup.tsx
henryboldi: I agree. I imagine what Greg has worked on for opening this window would replace this. I can remove it.

greg-nagy: I have a solution for this in https://github.com/tallycash/extension/pull/519

> App reads ?page=<page-name> on startup, and if defined loads the specified page instead of the / route
dummy permission window

https://github.com/tallycash/extension/pull/519/files#diff-67347e171ecce754b7e623a056ca13d1b49d7a75fbcffefdcadf2ea80399a0b2R50

ui/components/Core/CorePage.tsx
```diff
@@ -21,7 +21,12 @@ export default function CorePage(props: Props): ReactElement {
 
   function handleOpenHiddenDevMenu(e: React.MouseEvent) {
     if (process.env.NODE_ENV === "development" && e.detail === 3) {
-      setIsDevToolsOpen(true)
+      // setIsDevToolsOpen(true)
+      window.open(
```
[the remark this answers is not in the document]
henryboldi: I can remove it here, but don't want to cause you merge conflicts either. What do you think?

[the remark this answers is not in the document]
greg-nagy: Just get this merged, don't worry abbout anything, I will take it from main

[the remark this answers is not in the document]
henryboldi: Sweet! Can you grant me an approval? :)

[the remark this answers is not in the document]
greg-nagy: lol :) yes

greg-nagy (approved):
let's get this bad boy merged! Looks good, will handle the rest in the wiring up pr

ui/pages/DAppConnectRequest.tsx
greg-nagy: https://github.com/tallycash/extension/pull/547/commits/bfef7a1b9a3ce671fc25c108e3632bc7da287640

PR #531, opened 2021-12-08
Attempt at simplifying the feature assertions

wojciech-turek:
Having feature assertion done in a single place and exporting it as a boolean already saves some time and code.

We no longer have to write `process.env.FEATURE === 'true'`, we can use just `FEATURE` as a boolean and expect it to be true when it is set to be 'true' in the env.

I saw we are also doing reverse checks for same variable like:

`HIDE_SOMETHING === "true"` and somewhere `HIDE_SOMETHING === "false"` which is error prone and subject to bad results due to spelling mistakes and type verification.

I have only used variable that are currently used in more than once.


mhluongo (approved):
Yes yes yes.

PR #533, opened 2021-12-08
v0.8.1: Patch fix to drop notifications permission

Shadowfiend:
This permission isn't currently used by the extension, which has
led to rejection in the Chrome store. Drop it since we don't need
it.

Also included: 

* 🎨 New onboarding and settings paint
* 🚧 Improved infrastructure for autolocking and account listing
* 🐛 Firefox scrollbars, buggy button icon, and weird back button positions begone!
* 🛑 Drop notifications permission to enable Chrome store approval

### DIFF
--- a/manifest/manifest.json
+++ b/manifest/manifest.json
@@ -1,6 +1,6 @@
 {
   "name": "Tally",
-  "version": "0.8.0",
+  "version": "0.8.1",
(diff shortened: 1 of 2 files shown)


mhluongo (approved):
:green_circle: :rocket:

PR #534, opened 2021-12-08
Send page 🐛 to 🦋

wojciech-turek:
Send page was missing come elements from the design like selecting maximum available balance or displaying localized decimal value, this PR will add those details not only to the send page, but also to the sign page so it is more closely aligned with the design.

Some variables and functions have been re-named to better reflect their purpose or be more descriptive on what they do.


wojciech-turek:
Self-notes (might be action points later):

- Currently even if you set your own gasLimit its going to get overriden by the `estimateGasLimit` function in main.ts.

ui/pages/SignTransaction.tsx
```diff
@@ -40,6 +40,8 @@ export default function SignTransaction(): ReactElement {
   const isTransactionDataReady = useBackgroundSelector(
     selectIsTransactionLoaded
   )
+
+  // TODO the below should return a promise that resolves once tx is signed
```
Shadowfiend, line 44: This won't happen, because whatever is there will be shuttled over the webext-redux bridge.

Instead, you should use the `status` in the redux slice + `useEffect` to redirect when the status transitions to signed. See the import flow for an example:

https://github.com/tallycash/extension/blob/5708042663f226fbaedf072750cbcf655e291cf8/ui/pages/Onboarding/OnboardingImportMetamask.tsx#L59-L69

wojciech-turek: Thank you, I will implement it in the new PR.

ui/pages/Send.tsx
```diff
@@ -120,12 +164,29 @@ export default function Send(): ReactElement {
           </h1>
           <div className="form">
             <div className="form_input">
+              <div className="balance">
+                Balance: {`${currentBalance.substr(0, 8)}\u2026 `}
+                <span
+                  className="max"
+                  onClick={setMaxBalance}
```
henryboldi, line 171: * Included in the proposed move to `SharedAssetInput`

ui/pages/Send.tsx
```diff
@@ -137,6 +198,7 @@ export default function Send(): ReactElement {
                 defaultToken={{ symbol: assetSymbol, name: assetSymbol }}
                 amount={amount}
               />
+              <div className="value">${getTotalLocalizedValue()}</div>
```
henryboldi, line 201: In my opinion, this also belongs in `SharedAssetInput`

henryboldi (approved):
🛳🛳🛳

ui/pages/Send.tsx
```diff
@@ -120,12 +164,29 @@ export default function Send(): ReactElement {
           </h1>
           <div className="form">
             <div className="form_input">
+              <div className="balance">
+                Balance: {`${currentBalance.substr(0, 8)}\u2026 `}
```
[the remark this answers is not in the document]
wojciech-turek: Yes, I am 100% up for it.

ui/pages/Send.tsx
```diff
@@ -120,12 +164,29 @@ export default function Send(): ReactElement {
...
+                <span
+                  className="max"
+                  onClick={setMaxBalance}
+                  role="button"
+                  tabIndex={0}
+                  onKeyDown={() => {}}
```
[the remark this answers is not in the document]
wojciech-turek: The linter was giving me an error if there was not key event specified. It didn't for you?

background/services/keyring/index.ts
```diff
@@ -16,8 +16,8 @@ import { EIP1559TransactionRequest, SignedEVMTransaction } from "../../networks"
 import BaseService from "../base"
 import { ETH, MINUTE } from "../../constants"
 
-export const MAX_KEYRING_IDLE_TIME = 60 * MINUTE
-export const MAX_OUTSIDE_IDLE_TIME = 60 * MINUTE
+export const MAX_KEYRING_IDLE_TIME = 0.2 * MINUTE
```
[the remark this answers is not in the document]
wojciech-turek: Good catch, missed it 🙇🏻

PR #535, opened 2021-12-08
🔁 Fix having to go through adding a read-only wallet twice

### LINKED ISSUE #527: [Bug]: had to go through explore onboarding twice
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/917871963077218324

Chrome

trying to onboard into the wallet and transfer my assets

### What did not work?

the "explore tally" button on the last onboarding slide

### Version

OG Edition v0.0.7

### Relevant log output

_No response_

henryboldi:
Only one line change lol

Closes #527


ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
```diff
@@ -21,7 +21,7 @@ export default function OnboardingViewOnlyWallet(): ReactElement {
 
   const handleSubmitViewOnlyAddress = useCallback(async () => {
     if (checkIfPlausibleETHAddress(address)) {
-      dispatch(
+      await dispatch(
```
Shadowfiend, line 24: Ooofff remote store 1, @Shadowfiend 0.

Shadowfiend (approved):
Approving, giving this a test drive, then merging.

ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
Shadowfiend: The fun bit here is that you can put this `await` on *either* of the dispatches, because all we need is to make sure that the account has been added in the background script and made it into the account data slice in the UI, and to do that we need to wait for one full round-trip in webext-redux. Which round-trip doesn't matter!

In the current formulation with the `await` around `addAddressNetwork` there's a short moment _before_ the current account is set in the UI but _after_ the Wallet component renders... Fortunately the Wallet component can handle this situation gracefully and updates immediately once the current account lands, so it's imperceptible.

Bigg29 (commented):
[Github](github.com)

PR #536, opened 2021-12-08
fix provider-bridge package name in `provider-bridge.ts`

greg-nagy:


greg-nagy:
Local environment dependency  issue, not a real problem

Issue #368, opened 2021-10-28
Transaction List


Shadowfiend:

henryboldi:
Temporarily for the community edition, how about we just include sent, received, and contract interactions (displaying one asset)? Contract interactions would cover swaps to avoid the impression that tokens were lost

mhluongo:
Personally, I think that's great for a first release because we can do it accurately.

I think @Shadowfiend has something he's working on that goes a tiny bit further, recognizing ERC-20 transfers and approves. Whether we want to display those specially in the tx list or stick with the general contract interaction is unclear to me.

henryboldi:
Handled by #538

PR #537, opened 2021-12-08
📜 Display contract interaction transactions

henryboldi:
For now, it covers swaps to avoid the impression that tokens were only sent.

Activity item

I removed the title from the slide of menu. "Contract interaction" doesn't fit in the header, so without design work, this looks decent for now,

Details

Frankly, we could link directly to Etherscan from the activity item, and reintroduce this details screen later.

Context in #368 comments,
GitHub comments


Shadowfiend:
I pulled the styling work into #538, which is the PR I've been promising that reworks how and most especially where we handle contract data recognition for transactions.

I'll close this PR in the morning if there are no objections.

henryboldi:
Sounds good, @Shadowfiend 🙌

PR #538, opened 2021-12-08
Deconstructive Activities: Rework contract data enrichment

Shadowfiend:
Few things here:
- Contract data enrichment is now handled via a stub for a future enrichment
  service. The stub is just a function.
- The activities slice now tracks transactions with null block heights, meaning
  pending transactions.
- Contract info is now added to a separate field in the `ActivityItem` type.
  The activity list items have a harness for treating different resolved
  contract info differently, heavily based on the work in #537 but meant to
  illustrate an idea of a future direction. It's still a bit dirty and could
  use a little more abstraction in the future.
- The accounts slice no longer tracks any transaction activity. The existing
  data was unused anyway.
- The activities reducer is now typed correctly and the `infoRows` it adds are
  now handled correctly in the UI. They were already rendering roughly
  correctly, but it was an accident rather than intentional proper use of the
  data.


background/redux-slices/activities.ts
```diff
@@ -1,23 +1,28 @@
...
 } from "./utils/activity-utils"
 
 export { ActivityItem }
 
 const activitiesAdapter = createEntityAdapter<ActivityItem>({
   selectId: (activityItem) => activityItem.hash,
   sortComparer: (a, b) => {
-    if (a.blockHeight === b.blockHeight) {
+    // null means pending, pending is always sorted above everything.
+    if (a.blockHeight === null && b.blockHeight === 0) {
```
mhluongo, line 16: ... what if they're both `null`? Is that `0` a mistake?

background/redux-slices/utils/activity-utils.ts
```diff
@@ -110,39 +162,76 @@ export const keysMap: UIAdaptationMap<ActivityItem> = {
...
+    typeof matchingFungibleAsset !== "undefined" &&
+    contractInput.length >= 74 &&
+    contractInput.startsWith("0xa9059cbb") // transfer selector
+  ) {
+    return {
+      type: "asset-transfer",
+      contractLogoURL,
+      assetAmount: enrichAssetAmountWithDecimalValues(
+        {
+          asset: matchingFungibleAsset,
```
mhluongo, line 199: Note that if this tx is a `transfer()` that internally moves ERC-20 tokens around, this `matchingFungibleAsset` might not be the same as the contract that has has `transfer()` called on it. Might want to confirm `matchingFungibleAsset`'s contract address is the same as the `to` of the tx, otherwise return a general contract interaction

background/redux-slices/utils/activity-utils.ts
```diff
@@ -110,39 +162,76 @@ export const keysMap: UIAdaptationMap<ActivityItem> = {
...
+  assets: AnyAsset[],
+  transaction: AnyEVMTransaction,
+  desiredDecimals: number
+): AnyEVMTransaction & { contractInfo?: ContractInfo | undefined } {
+  if (transaction.input === null || transaction.input === "0x") {
+    // This is _almost certainly_ not a contract interaction, move on. Note that
+    // a simple ETH send to a contract address can still effectively be a
+    // contract interaction (because it calls the fallback function on the
+    // contract), but for now we deliberately ignore that scenario when
+    // categorizing activities.
```
mhluongo, line 224: This is a beautiful comment. I'd love to see it for any similarly intentional stuff about, eg like including the `contractLogoURL` with a general `contract-interaction`

background/redux-slices/utils/activity-utils.ts
```diff
@@ -110,39 +162,76 @@ export const keysMap: UIAdaptationMap<ActivityItem> = {
...
+    return transaction
+  }
+
+  return {
+    ...transaction,
+    contractInfo: resolveContractInfo(
```
mhluongo, line 230: I can't believe this isn't `async`, we have so much more to do here heh

mhluongo (commented):
Left a couple thoughts. Tested locally, and clearly this thing hasn't been styled yet.... it's a draft, after all! But I do see some of my erroneous tx labelling has been replaced by "contract interaction" which is nice.

background/redux-slices/activities.ts
Shadowfiend: ... Yep 😅

background/redux-slices/utils/activity-utils.ts
Shadowfiend: You're saying if the resolved target is the same as the sender? The matching asset is resolved against the transaction to address.

background/redux-slices/utils/activity-utils.ts
Shadowfiend: Imo this shouldn't need to be async even when we're looking up more stuff--it should just emit an updated thing once we resolve whatever else we want to resolve.

background/redux-slices/utils/activity-utils.ts
mhluongo: Oh I'd assumed we got the asset from Alchemy metadata. If it's already from the transaction `to` this isn't an issue ofc

background/redux-slices/utils/activity-utils.ts
mhluongo: That's two steps ahead 😁

mhluongo (approved):
I'm happy with this but looking for another :+1: before merge. @henryboldi?

henryboldi (changes requested):
This is amazing! 🧙‍♂️ I've got a couple questions and things I'd like to flag.

1) I believe a contract interaction shouldn't be listed as a sent asset in the activity detail panel

contract interaction not displayed

2) To be clear, we're not marking swaps in this PR?

3) Does this change retain the token data for displaying non-ETH transactions? It wasn't exhaustive but we used to be able to display (in #537),
Zero
Is that not worth keeping?

background/main.ts
```diff
@@ -289,24 +287,19 @@ export default class Main extends BaseService<never> {
...
-      ) {
-        this.store.dispatch(transactionConfirmed(transaction))
-      } else {
-        this.store.dispatch(transactionSeen(transaction))
-      }
-      this.store.dispatch(activityEncountered(enrichedPayload))
+      const enrichedTransaction = enrichTransactionWithContractInfo(
+        this.store.getState().assets,
+        transaction,
+        2 /* TODO desiredDecimals should be configurable */
```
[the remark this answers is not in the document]
Shadowfiend: It's all string amounts for formatting.

background/services/enrichment/index.ts
```diff
@@ -0,0 +1,129 @@
...
+    contractInput.length >= 74 &&
+    contractInput.startsWith("0xa9059cbb") // transfer selector
+  ) {
+    return {
+      type: "asset-transfer",
+      contractLogoURL,
+      assetAmount: enrichAssetAmountWithDecimalValues(
+        {
+          asset: matchingFungibleAsset,
+          amount: BigInt(`0x${contractInput.slice(10, 10 + 64)}`),
```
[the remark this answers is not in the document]
Shadowfiend: Yep, see the TODO above.

Shadowfiend:
> To be clear, we're not marking swaps in this PR?

Correct. We have the infra for it, but not the identification or full rendering for now.

> I believe a contract interaction shouldn't be listed as a sent asset in the activity detail panel

Whoops, failed to pull in your header adjustments. Fixed.

> Does this change retain the token data for displaying non-ETH transactions?

Only when we are reasonably certain that it's an ERC20 token transfer (by checking the function selector and input length for something we've identified as a smart contract fungible asset). Otherwise we don't actually know the data we're reading is correct, because we have no idea what the function called in the transaction is or what its input ABI looks like—we're just offsetting into input data and hoping the function being called takes a single uint256 that is a token amount, and that this means it's a token transfer. That is the salient piece of this identifier: https://github.com/tallycash/extension/pull/538/commits/6b69209b6d51ec032537ed0b9670669766e94809#diff-e21d963357c1054efb4597ad83510c25be9ca92fd5280c0b53d0c2ab945466fbR190-R192 .

As an example, as far as I can tell, the ZERO transaction in that list isn't correct. That transaction resulted in 100 ZERO being minted, whereas 0.1 is the ETH sent with the transaction. If there had been a 0 ETH value in the transaction, we would have tried to pull the value from the first parameter to the function call—but the function called in that transaction has as its first parameter an address, not an amount, so we would have pulled the amount 1.410102024832666e+48 or so (after dividing out the 18 decimals). The amount that particular function takes is the third parameter—and it's actually (for reasons that are unclear to me, since this info is available without a function parameter) the ETH amount sent with the transaction.

The potential to display some transactions incorrectly is a greater problem than the utility of displaying some other ones more comprehensively. This PR's goal was to set us up for a good way to comprehensively categorize transactions based on deeper understanding of their input structure.

**With all of that said**… The ERC20 recognizer was wrong as-was, and I absolutely would not have noticed if you hadn't asked 😅 It was checking for too short an input length (`transfer` transaction parameters are 128 characters long), and it was trying to pull the value from the first rather than second parameter (the first 64 characters are the recipient address). I've pushed a fix 🙌 

While fixing this, I realized that the first transaction inserted into an account always gets lost due to something weird in how `upsertOne` works with the result of `getInitialState`. I also realized that we were hiding 0-ETH activities, which will lose all asset transfers in the list. I fixed both of those.

Last but not least, I've added recipient information to the asset transfer so that the To in the activity list is correctly the target of the transfer.

You can test this one by adding account 0x2a8405140dc94965478c51c522d84615ff96dc92, whose latest transaction is a UNI transfer.

background/services/enrichment/index.ts
```diff
@@ -0,0 +1,131 @@
...
+      asset.contractAddress.toLowerCase() === contractAddress.toLowerCase()
+  )
+
+  const contractLogoURL = matchingFungibleAsset?.metadata?.logoURL
+
+  // Derive value from transaction transfer if not sending ETH
+  // FIXME Move to ERC20 parsing using ethers.
+  if (
+    typeof matchingFungibleAsset !== "undefined" &&
+    contractInput.length === 138 &&
```
mhluongo, line 81: :ok_hand:

mhluongo (approved):
Transactions look good on my local wallet. Can't wait to write more tx recognizers. Let's get this merged!

PR #539, opened 2021-12-09
:recycle:  Swap confirmation page

itsrachelfish:
This PR wires up the swap confirmation template with data from redux, adds JTD validation, and updates the swap redux slice to use selectors as requested in @Shadowfiend's feedback on #494 

swap-confirm-page

I still need to finish integrating the shared gas selector component and broadcast the transaction, but that will have to wait for the next PR :sweat_smile:


.env.defaults
```diff
@@ -1,3 +1,4 @@
+DEBUG=false
```
mhluongo, line 1: Did you commit this intentionally?

ui/pages/Swap.tsx
```diff
@@ -89,26 +107,21 @@ export default function Swap(): ReactElement {
...
 
-    [dispatch, swap]
+    [dispatch, buyPrice]
   )
 
   const toAmountChanged = useCallback(
-    (event) => {
-      const inputValue = event.target.value.replace(/[^0-9.]/g, "") // Allow numbers and decimals only
+    (amount) => {
+      const inputValue = amount.replace(/[^0-9.]/g, "") // Allow numbers and decimals only
```
mhluongo, line 121: We should change the input type to `number` with a very small `step` as well. No reason to let people make this mistake in the first place

ui/pages/Swap.tsx
```diff
@@ -89,26 +107,21 @@ export default function Swap(): ReactElement {
           setSwapAmount({
             sellAmount: inputValue,
             // TODO: Use a safe math library
-            buyAmount: (
-              floatValue / parseFloat(swap.buyToken.price)
-            ).toString(),
+            buyAmount: (floatValue / parseFloat(buyPrice)).toString(),
```
mhluongo, line 110: The float math here has me itchy

ui/pages/Swap.tsx
mhluongo: Note that this regex will also allow "......", "127.0.0.1", etc

ui/components/Swap/SwapQuote.tsx
```diff
@@ -1,11 +1,44 @@
...
+
+            return false
+          }),
+        }
+      }
+
+      // We should always have a quote by the time we get to this page, but just in case!
+      return {
+        ...state.swap,
+        sources: [],
```
mhluongo, line 38: Should `sources` come first here, or are you meaning to overwrite a potential `state.swap.sources` var?

mhluongo (commented):
I'm catching up on the swap UI here, but wanted to drop a few comments in the meantime.

ui/components/Swap/SwapQuote.tsx
itsrachelfish: There is no `state.swap.sources` variable in redux, it's being generated in this selector based on `state.swap.quote.sources`, which could potentially be undefined. By including it here I can avoid adding a bunch of undefined type checks in the template.

ui/pages/Swap.tsx
itsrachelfish: Yeah eventually I want to convert all of this logic to use @Shadowfiend's fixed-point utility functions, but there isn't a helper function for doing division yet

.env.defaults
itsrachelfish: Yes, this is being used to suppress all of the swap token discrepancy warnings by default. Since I converted the swap code to use a redux selector, any time our assets list changes we get hundreds of these warning messages.

https://github.com/tallycash/extension/pull/539/files#diff-b5f84327fc0a3e82c8daa46b4ed8eabc57e1f7cf6e6df5e20444285432ff36b4R320-R346

Shadowfiend: This is what I was getting at in flowdock when I suggested we (1) go back to using `console[thing]` in our logger and (2) start using `logger.debug`. You can then filter in-browser based on log level (at least in FF). It seems to work with grouping though the group isn't suppressed (but it's much less noisy).

itsrachelfish: A downside to using `console[thing]` is that `console.error` always produces a stack trace in chrome, but nothing else does. That's why we were occasionally seeing two stack traces previously.

background/redux-slices/0x-swap.ts
```diff
@@ -40,132 +34,360 @@ interface ZrxPrice {
...
+      "Swap quote API call didn't validate, did the 0x API change?",
+      apiData,
+      isValidSwapPriceResponse.errors
+    )
+
+    return undefined
+  }
+)
+
 const swapSlice = createSlice({
```
greg-nagy, line 246: If there is another version of redux stored in cache, this does get's initalized because the reducer keeps the `swap` name. 

If we change the name of the slice in the `redux-slices/index.ts` to something else and update the main selector, then it works.

background/redux-slices/0x-swap.ts
```diff
@@ -40,132 +34,360 @@ interface ZrxPrice {
...
 
-        return false
-      })
+const isValidSwapAssetsResponse = jtdValidatorFor(swapAssetsJTD)
 
-    return filteredAssets
+export const fetchSwapAssets = createBackgroundAsyncThunk(
+  "0x-swap/fetchAssets",
+  async () => {
+    const apiData = await fetchJson(`https://api.0x.org/swap/v1/tokens`)
```
greg-nagy, line 114: This fetchJson() is awesome! :)

background/redux-slices/0x-swap.ts
```diff
@@ -40,132 +34,360 @@ interface ZrxPrice {
...
+  buyAmount: "",
+  zrxAssets: [],
+  zrxPrices: [],
+}
 
-        // TODO: What if multiple assets match?
-        if (matchingTokens.length) {
-          return matchingTokens[0]
-        }
+const swapAssetsJTD = {
```
greg-nagy, line 94: That's a huge amount of JTD! Congrats 💪

ui/pages/Swap.tsx
```diff
@@ -146,43 +160,37 @@ export default function Swap(): ReactElement {
...
                 onAssetSelect={fromAssetSelected}
                 onAmountChange={fromAmountChanged}
-                amount={swap.sellAmount}
+                amount={sellAmount}
                 label="Swap from:"
               />
             </div>
             <div className="icon_change" />
             <div className="form_input">
               <SharedAssetInput
```
greg-nagy, line 173: As a next step we could disable this if there is not swap options based on the from selection.

greg-nagy (commented):
When github collapses your code because it's just too much ... that's when you know that you have just written way too much JTD and interface defintions! 

There are a couple of follow up issues, but I think this is good to go. 
Will give this tomorrow another look.

.env.defaults
Shadowfiend: My 2c is that this is fine, esp because Chrome's stack trace is probably more actionable/clickable than our otherwise-logged one (which might be considered more of a fallback), and especially because we use console.error relatively rarely and only when things are more severe. But I see why this might be annoying :/

ui/pages/Swap.tsx
Shadowfiend: Def a both, but yeah, using `type="number"` is :100: @wojciech-turek had a great example of this where he also hid the additional chrome browsers will add since it's less necessary for our use case.

Shadowfiend (dismissed):
Given the feature-flagged nature of this and the fact that two separate folks have had a look here, I'm going to approve + merge and leave additional notes tomorrow as I wasn't able to load up full context today. Let's keep the train rolling!

background/redux-slices/0x-swap.ts
```diff
@@ -40,132 +34,360 @@ interface ZrxPrice {
...
+        properties: {
+          address: { type: "string" },
+          decimals: { type: "int8" },
+          name: { type: "string" },
+          symbol: { type: "string" },
+        },
+      },
+    },
+  },
+}
```
Shadowfiend, line 107: What do you think about pulling these JTDs out to their own file? Perhaps `utils/0x-utils.ts` with `resolveValidSwapAssets(json: unknown): ZrxAsset[] | undefined`, `resolveValidZrxQuote(json: unknown): ZrxQuote | undefined`, etc?

In this scenario, `0x-utils` could also export e.g. `ZrxAsset` as a type derived directly from the JTD definition, provided you use `as const`, which is the next part…

Separately from the code split, you can add `as const` to the JTD declarations to get more precise types from the validator type guards. If you add that here, in `fetchSwapAssets` below you'll be able to drop the `as ZrxAsset[]` coercion because the types will already be correct. You can make sure this doesn't break in the future by giving the async action handler a `: Promise<ZrxAsset[]>` return type.

Once you have that, you can actually define `ZrxAsset` *in terms of the JTD* and compose its definition into other parts of the API, which is particularly helpful to avoid duplication for something big like the quote return type:

```typescript
const zrxAssetJTD = {
  properties: {
    address: { type: "string" },
    decimals: { type: "int8" },
    name: { type: "string" },
    symbol: { type: "string" },
  }
} as const

const swapAssetsJTD = {
  properties: {
    records: {
      elements: zrxAssetsJTD
    }
  }
} as const

const isValidSwapAssetsResponse = jtdValidtorFor(swapAssetsJTD)

export type ZrxAsset = JTDDataType<typeof zrxAssetJTD>

export const fetchSwapAssets = createBackgroundAsyncThunk(
   "0x-swap/fetchAssets",
   async (): Promise<ZrxAsset[]> => {
     const apiData = await fetchJson(`https://api.0x.org/swap/v1/tokens`)

     if (isValidSwapAssetsResponse(apiData)) {
       return apiData.records
     }

     logger.warn(
       "Swap asset API call didn't validate, did the 0x API change?",
       apiData,
       isValidSwapAssetsResponse.errors
     )

     return []
   }
 )
```

Shadowfiend (commented):
Let a quick comment on JTD organization/composition/typing. More tomorrow.

background/redux-slices/0x-swap.ts
itsrachelfish: I was wondering why the other JTD definitions used `as const`, this makes sense :+1:

Shadowfiend: Yeah `as const` here raises the strings (like `number` in `type: "number"`) to actual types, so Ajv's generic types can then use them to make decisions about which JS types they correspond to. With `as const`, the type of `{ type: "number" }` is `{ type: "number" }`, whereas without it it's just `{ type: string }`, without the type system knowing the specific string.

ui/pages/Swap.tsx
wojciech-turek: What I did there was to add type `number` and add the step as `any`, also add `min="0"`. 

To hide the number handlebars I've used:
```
 input[type="number"] {
      -moz-appearance: textfield;
 }
```

Shadowfiend (approved):

PR #541, opened 2021-12-09
⚡️ Connected dApp status UI

henryboldi:
This PR gives @greg-nagy a UI to plug into for displaying an active dApp connection. 😊

Status


ui/components/TopMenu/TopMenuConnectedDAppInfo.tsx
```diff
@@ -0,0 +1,111 @@
...
+          type="button"
+          className="disconnect_icon"
+        />
+      </div>
+      <button
+        aria-label="Background close"
+        type="button"
+        className="void_space"
+        onClick={close}
+      />
```
Shadowfiend, line 33: Just attach a click handler to the `bg` div, no need for the extra button IMO. Accessibility-wise we just need to make sure there *is* a close button that can be read—having two makes it strictly worse. Letting the background close things is a visual convenience without a real auditory equivalent.

Shadowfiend (approved):
Couple of small notes; we also probably need an onDisconnect.

:shipit:

ui/components/Core/CorePage.tsx
```diff
@@ -32,6 +35,15 @@ export default function CorePage(props: Props): ReactElement {
 
   return (
     <main>
+      {isActiveDAppConnectionInfoOpen ? (
+        <TopMenuConnectedDAppInfo
+          title="SushiSwap | Sushi"
+          url="https://app.sushi.com"
```
greg-nagy, line 41: I will also need the faviconUrl but will add them soon

ui/components/TopMenu/TopMenu.tsx
```diff
@@ -6,10 +6,15 @@ import { useBackgroundSelector } from "../../hooks"
 interface Props {
   toggleOpenProtocolList: () => void
   toggleOpenNotifications: () => void
+  toggleOpenDAppConnectionInfo: () => void
 }
 
 export default function TopMenu(props: Props): ReactElement {
-  const { toggleOpenProtocolList, toggleOpenNotifications } = props
```
greg-nagy, line 12: Is there any default value for these props?

greg-nagy (commented):
Thanks for this build 🧡

ui/components/Core/CorePage.tsx
henryboldi: Basing the `faviconUrl` off of the `url` here, but that might not cover all cases

greg-nagy: Hm, my thinking was focused by the fact that I get `faviconUrl` from chrome and changed the implementation when integrateing it. 

This approach would work most of the times, but I will go with the info coming from the chrome tab because that should be set from the html tag in the page which feels to be more robust. Not a big problem in any way.

PR #542, opened 2021-12-09
Reverse Safety: Fix autolock clearing keyring metadata for good

Shadowfiend:
Had a reversed condition here. While poking around, also found
some quick wins on optimizing selectors, and while looking at that
also found an issue with how we were setting the `from` on send
transactions.


mhluongo (approved):
Optimizations look good, appears to work on a local build. Doing the thing.

mhluongo:
I should've been more specific: I tested the selectors, etc, I didn't wait around to see if it locks.

PR #543, opened 2021-12-09
🧹 Visual housekeeping

henryboldi:
This PR only fixes the account switcher spacing and wallet tab width so far. Small stuff.

⚠️ I've got a few other issues in mind to include fixes for, but still, it's mergeable now.

width
.


Shadowfiend:
@henryboldi dealers choice here--would you rather accumulate a few more or land this now? Don't really mind either way on my end.

henryboldi:
I'd rather accumulate a few more! That said, if we want to bundle up the testing edition very soon, it might make sense to include now

Shadowfiend:
Cool cool, I'll plan on merging here in an hour or so.

ui/pages/Wallet.tsx
```diff
@@ -39,7 +39,7 @@ export default function Wallet(): ReactElement {
   }
 
   return (
-    <div className="wrap">
```
Shadowfiend, line 42: Probably want to nuke the `wrap` class below as well.

Shadowfiend: Actually peeking more closely, every other page has a `main` element up here (with no further class). We should stick to that pattern here.

Shadowfiend (approved):
Fixes de tings.

ui/pages/Wallet.tsx
henryboldi: > Actually peeking more closely, every other page has a main element up here (with no further class). We should stick to that pattern here.

They do?

PR #544, opened 2021-12-09
Remove dist from prettier checks

greg-nagy:
In some situations the precommit hook did not lint properly the codebase because it failed, causing error in the CI

Dist folder caused the precommit hook to fail in local if the files were not properly formatted. Then prettier did fail and linter did not ran. Causing failures in the CI.


Shadowfiend (approved):
> Causing failures in the CI.

Still not convinced CI had any issues related to this, fwiw. The only CI failures I'm seeing are related to ts-nocheck and import ordering issues in real files.

Either way though, this aligns our prettier config to our eslint config, so let's do itttt.

greg-nagy:
It's not that the CI is wrong. On the contrary! CI works as it should, BUT the action fails because the code that's pushed is wrong — because it was not linted by the precommit hook. [See this example](https://github.com/tallycash/extension/actions/runs/1560292739)

The steps I have taken to fix this PR:
- Ran `yarn lint` and saw it fail on the prettier step because of `chrome/dist/manifest.json` — and eslint was not even run.
- Removed `dist/chrome`
- Ran `yarn lint` and saw the same errors as in the CI output 
- Fixed the errors pushed and [CI passed](https://github.com/tallycash/extension/actions/runs/1560454525)

Shadowfiend:
Ah! We want to add the `-c` flag to the `run-p` invocation so it doesn't abort all tasks when one errors out. You should def see all lint errors when running `yarn lint` otherwise we're wasting everybody's time 😁

PR #545, opened 2021-12-09
🤬 Add various UI error messages

henryboldi:
This PR includes error messages for,
- [x] Password matches + minimum length (8) in the create password screen.
- [x] Incorrect password in the unlock screen
- [x] Insufficient balance for asset selectors

Plus a spinner for,
- [x] Create password screen

error

error 2

error 3

error 4


ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -21,11 +22,31 @@ export default function KeyringSetPassword(): ReactElement {
     }
   }, [history, areKeyringsUnlocked])
 
+  const validatePassword = (): boolean => {
+    if (password.length < 8) {
```
greg-nagy, line 26: 🤔 If we want to have stronger password requirements?

ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -170,6 +173,19 @@ export default function SharedAssetInput(
     [onAssetSelect]
   )
 
+  const selectedTokenAssetAmount = assetAmounts?.filter((token) => {
+    return token.asset.symbol === selectedToken.symbol
+  })[0]
+
+  const getErrorMessage = (givenAmount: string): string | undefined => {
+    return (!isTypeDestination &&
+      selectedTokenAssetAmount &&
```
greg-nagy, line 182: It nice have trust, but it's even better when we make sure everything is as expected 🧡

ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -179,10 +195,12 @@ export default function SharedAssetInput(
           setOpenAssetMenu(false)
         }}
       >
-        <SelectTokenMenuContent
-          assets={assets}
-          setSelectedTokenAndClose={setSelectedTokenAndClose}
-        />
+        {assets && (
+          <SelectTokenMenuContent
+            assets={assets}
```
greg-nagy, line 200: Could it cause any issue here if the assets were the default `[{ symbol: "ETH", name: "Example Asset" }]`?

greg-nagy (dismissed):
This is a nice PR, thanks for this! Let's merge this bad boy :)

greg-nagy (approved):
let's merge this after conflicts have been resolved

ui/components/Shared/SharedAssetInput.tsx
Shadowfiend: Yaaasss. Only notes I'll add:
- avoid ternary operators outside of places they're critical.
- typeof === "undefined" is the explicit way we're trying to do undefined checks.

I'm having to flag the second one too much so if I can't find a linter rule for it we might drop it.

PR #546, opened 2021-12-10
Dapp signing

### LINKED ISSUE #131: End-to-end send ETH flow
- [x] Key import interface (#237)
- [x] Transaction signature API (#248)
- [x] Transaction signature backend (#158)
- [ ] Confirm the UI shows transaction confirmations (#233)
- [ ] Wire up the signature page (#231)
- [ ] EIP-1559 gas selection design (#229)
- [x] #294
- [ ] Request tx signature from backend (#231)
- [ ] Broadcast transaction to network (#258)
- [x] Kick off account tracking when a new key is added (#241)
- [x] Recovery phrase UI hookup (#242)

### LINKED ISSUE #231: Confirm button requesting signature in backend

### LINKED ISSUE #629: [BUG] Unable to send ETH b/c of gas selection issues
Build #1129

I'm unable to send ETH b/c of gas selection issues:

1 - Gas initially defaults to zero

Screen Shot 2021-12-14 at 6 19 51 PM

2 - I click gear icon to set gas ... gas limit is set to 21000 ... I click to increase it but 'save' button does not work. I need to click "x" to close pop-up.
 
Screen Shot 2021-12-14 at 6 20 23 PM

3 - I am able to submit txn with no gas and txn is stuck in pending

Screen Shot 2021-12-14 at 6 21 27 PM

Screen Shot 2021-12-14 at 6 22 32 PM

### LINKED ISSUE #644: Not able to send ETH (beachballing on final confirmation view)
I get stuck belong; pending txns don't appear in Activity

Screen Shot 2021-12-14 at 9 53 38 PM

Shadowfiend:
Implements dApp signing and a base sign transaction page. The existing
ETH transfer signature page is moved to a separate transaction block that
is specific to transfers. The internal Ethereum provider service now sets
transaction state and presents the sign page on eth_signTransaction and
eth_sendTransaction requests. Signed transactions are correctly broadcast
and listed in the activity list.

Closes #644, closes #629, closes #131, closes #231.

Introduces #646


background/services/internal-ethereum-provider/index.ts
```diff
@@ -98,9 +109,14 @@ export default class InternalEthereumProviderService extends BaseService<Events>
         return this.chainService
           .getAccountsToTrack()
           .then(([account]) => [account.address])
-      case "eth_sign": // --- important wallet methods ---
       case "eth_sendTransaction":
+        return this.signTransaction(
```
greg-nagy, line 113: This might cause double signing, depending on what dApp expects.

src/window-provider.ts
```diff
@@ -5,9 +5,11 @@ import {
 } from "@tallyho/provider-bridge-shared"
 import TallyWindowProvider from "@tallyho/window-provider"
 
-const enabled = window.localStorage.getItem(WINDOW_PROVIDER_FLAG)
+const windowProviderEnabled = process.env.ENABLE_WINDOW_PROVIDER === "true"
+const windowProviderDefaultEnabled =
```
greg-nagy, line 9: Webpakc replaces `process.env...` with constants build time I think, so this probably will have the effect of turning the runtime feature flag into compile time feature flag.

src/window-provider.ts
```diff
@@ -5,9 +5,11 @@ import {
 } from "@tallyho/provider-bridge-shared"
 import TallyWindowProvider from "@tallyho/window-provider"
 
-const enabled = window.localStorage.getItem(WINDOW_PROVIDER_FLAG)
+const windowProviderEnabled = process.env.ENABLE_WINDOW_PROVIDER === "true"
```
greg-nagy, line 8: Webpakc replaces `process.env...` with constants at build time I think, so this probably will have the effect of turning the runtime feature flag into compile time feature flag.

background/services/internal-ethereum-provider/index.ts
Shadowfiend: Can you elaborate a bit?

src/window-provider.ts
Shadowfiend: Yeah I'm not meaning to leave this in without further discussion, just needed to do a few things that were eased by it.

background/services/internal-ethereum-provider/index.ts
greg-nagy: Given the code below if a dApp calls `eth_signTransacion` first then with the payload calls the `eth_sendTransaction` then the signing should happen twice. 

Not sure if it's a real possibility at all though. 

```
      case "eth_sendTransaction":
        return this.signTransaction(
          params[0] as EIP1559TransactionRequest
        ).then((signed) => this.chainService.broadcastSignedTransaction(signed))
      case "eth_signTransaction":
        // FIXME Convert to standard Ethereum signed tx string.
        return this.signTransaction(params[0] as EIP1559TransactionRequest)
      case "eth_sign": // --- important wallet methods ---
```

Shadowfiend: `eth_signTransaction` should be returning a hex string (that's the FIXME), which is then used with `eth_sendRawTransaction`. I don't think `eth_signTransaction` output passed to `eth_sendTransaction` is supported.

src/window-provider.ts
Shadowfiend: Dropped this with the removal of the feature flag.

background/services/internal-ethereum-provider/index.ts
```diff
@@ -98,9 +110,14 @@ export default class InternalEthereumProviderService extends BaseService<Events>
         return this.chainService
           .getAccountsToTrack()
           .then(([account]) => [account.address])
-      case "eth_sign": // --- important wallet methods ---
       case "eth_sendTransaction":
+        return this.signTransaction(params[0] as EthersTransactionRequest).then(
+          (signed) => this.chainService.broadcastSignedTransaction(signed)
+        )
       case "eth_signTransaction":
+        // FIXME Convert to standard Ethereum signed tx string.
```
mhluongo, line 118: Why not just join the sig components here?

background/main.ts
```diff
@@ -481,6 +463,22 @@ export default class Main extends BaseService<never> {
...
   }
 
+  async connectInternalEthereumProviderService(): Promise<void> {
+    this.internalEthereumProviderService.emitter.on(
+      "transactionSignatureRequest",
+      async ({ payload, resolver }) => {
+        this.store.dispatch(updateTransactionOptions(payload))
+        // TODO force route?
+
+        const signedTransaction = await this.keyringService.emitter.once(
```
mhluongo, line 473: Trying to keep up here. What if there were two of these events? Do we not want them all broadcast?

If we want to maintain this 1:1 relationship, do sig requests need IDs or anything else tying them together? Currently this looks like a recipe for an event handling mistake to lead to a backlog of signed transactions that haven't been broadcast, or a mismatch between signing requests in the UI and responses.

ui/components/SignTransaction/SignTransactionSignBlock.tsx
```diff
@@ -1,19 +1,85 @@
...
 
 interface Props {
-  token: string
-  amount: number
+  transactionDetails: EIP1559TransactionRequest
 }
 
-export default function SignTransactionSignBlock(props: Props): ReactElement {
-  const { token, amount } = props
+export default function SignTransactionSignBlock({
```
mhluongo, line 21: This name is unfortunate

ui/pages/SignTransaction.tsx
```diff
@@ -35,22 +37,36 @@ export default function SignTransaction(): ReactElement {
...
+    if (isTransactionSigning && isTransactionSigned) {
+      history.push("/")
+    }
+  }, [history, isTransactionSigned, isTransactionSigning])
 
   if (!areKeyringsUnlocked) {
     return <></>
   }
 
+  if (typeof transactionDetails === "undefined") {
```
mhluongo, line 64: When might this be the case?

ui/pages/SignTransaction.tsx
```diff
@@ -68,20 +84,33 @@ export default function SignTransaction(): ReactElement {
       component: () => <SignTransactionApproveSpendAssetBlock />,
       confirmButtonText: "Approve",
     },
+    [SignType.SignTransfer]: {
+      title: "Sign Transfer",
+      component: () => (
+        <SignTransactionTransferBlock
+          token={assetSymbol ?? ""}
+          amount={amount ?? 0}
```
mhluongo, line 92: Not sure we should allow an undefined amount through. Hard to imagine a 0 amount transfer being a good idea outside nonce management maybe?

mhluongo (commented):
Getting familiar here, glad I started the review in draft

background/services/internal-ethereum-provider/index.ts
Shadowfiend: Not sure what you mean… But the issue here is that we round-trip through an area of the code where we don't provide the signed hex. Maybe you mean just restringifying here, which is one of the options I was considering.

background/main.ts
Shadowfiend: Yep, this is extremely brittle… But actually the most brittle thing is the signing screen, so I'm not worried about this part quite yet. As designed, the signing page won't clear until this happens.

ui/components/SignTransaction/SignTransactionSignBlock.tsx
Shadowfiend: Iz true.

ui/pages/SignTransaction.tsx
Shadowfiend: Any situation in which the transaction details don't get correctly set in redux. “This shouldn't happen”™.

ui/pages/SignTransaction.tsx
Shadowfiend: I'm not touching transfer signatures here.

background/services/internal-ethereum-provider/index.ts
Shadowfiend: Have a look at https://github.com/tallycash/extension/pull/546/commits/ea6001fee81307182f5b4e99748c7622a1ef4a9a and see if that's what you had in mind.

Shadowfiend: Oh! I understand now. No, the response for `eth_signTransaction` is the raw transaction hex of the signed transaction, i.e., the input to `eth_sendRawTransaction`, not just the hex of the signature itself.

background/services/internal-ethereum-provider/index.ts
```diff
@@ -1,14 +1,30 @@
...
+} from "../chain/utils"
+
+type DAppRequestEvent<T, E> = {
+  payload: T
+  resolver: (result: E | PromiseLike<E>) => void
+}
 
 type Events = ServiceLifecycleEvents & {
+  transactionSignatureRequest: DAppRequestEvent<
+    Partial<EIP1559TransactionRequest> & { from: string },
```
mhluongo, line 25: `HexString`?

ui/components/SignTransaction/SignTransactionSignBlock.tsx
```diff
@@ -1,43 +1,90 @@
...
 
   return (
     <div className="sign_block">
       <div className="container">
-        <div className="label">Send to</div>
-        <div className="send_to">{truncateAddress(destination)}</div>
+        {typeof transactionDetails.to === "undefined" ? (
+          <>
+            <div className="label">Send to:</div>
+            <div className="send_to">Contract creation</div>
```
mhluongo, line 61: Deserves a TODO... let's come back for this later.

ui/components/SignTransaction/SignTransactionSignBlock.tsx
```diff
@@ -1,43 +1,90 @@
...
+          </>
+        ) : (
+          <>
+            <div className="label">Send to:</div>
+            <div className="send_to">
+              {transactionDetails.to.slice(0, 6)}...
```
mhluongo, line 67: We do this a lot, isn't there a util?

ui/components/SignTransaction/SignTransactionTransferBlock.tsx
```diff
@@ -0,0 +1,84 @@
...
+  destination: string
+  localizedValue: string | number
+}
+
+export default function SignTransactionTransferBlock(
+  props: Props
+): ReactElement {
+  const { token, amount, destination, localizedValue } = props
+
+  function truncateAddress(address: string): string {
```
mhluongo, line 15: And we see it again. Opened an issue to follow up

https://github.com/tallycash/extension/issues/647

ui/components/SignTransaction/SignTransactionTransferBlock.tsx
```diff
@@ -0,0 +1,84 @@
...
+    return `${address.slice(0, 6)}...${address.slice(37, 42)}`
+  }
+
+  function truncateAmount(value: number): string {
```
mhluongo, line 19: https://github.com/tallycash/extension/issues/647

mhluongo (approved):
Non-blocking comments. Did my best to catch all the issues this closes, as well as a couple it opens.

Well done ser.

mhluongo (approved):
Repeated from a dismissed review:

Non-blocking comments. Did my best to catch all the issues this closes, as well as a couple it opens.

Well done ser.

ui/components/SignTransaction/SignTransactionSignBlock.tsx
Shadowfiend: What should the TODO be?

ui/components/SignTransaction/SignTransactionSignBlock.tsx
Shadowfiend: 😬 

It's terrible. We probably want a component that does this + a copy button + a `title` with the full address.

ui/components/SignTransaction/SignTransactionTransferBlock.tsx
Shadowfiend: Yes, but this one is subtly different. 

😅🤡🤪

ui/components/SignTransaction/SignTransactionSignBlock.tsx
mhluongo: Opened an issue on [displaying contract deployment intentionally](https://github.com/tallycash/extension/issues/648).

ui/components/SignTransaction/SignTransactionSignBlock.tsx
mhluongo: Your wish is my quick write-up

https://github.com/tallycash/extension/issues/649

ui/pages/SignTransaction.tsx
```diff
@@ -65,6 +68,13 @@ export default function SignTransaction(): ReactElement {
     })
 
   const [panelNumber, setPanelNumber] = useState(0)
+  const [isTransactionSigning, setIsTransactionSigning] = useState(false)
+
+  useEffect(() => {
+    if (isTransactionSigning && isTransactionSigned) {
```
wojciech-turek, line 74: Hee this may be confusing, we should consider renaming it.

ui/pages/SignTransaction.tsx
```diff
@@ -65,6 +68,13 @@ export default function SignTransaction(): ReactElement {
...
+      history.push("/")
```
wojciech-turek, line 75: Why do we push to '/' and not to singleAsset page where user can see the activity page?

Shadowfiend: I definitely lost that during the rebase/conflict resolution process. Bring it back!

PR #548, opened 2021-12-10
feat(CI): add more github actions

xwartz:
- Add GitHub Action to assign labels to PRs based on title
- Add GitHub Action to automerge dependency updates from Dependabot

### DIFF
--- a/.github/workflows/generate-labels.yml
+++ b/.github/workflows/generate-labels.yml
@@ -1,0 +1,14 @@
+name: 'Generate Labels'
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize]
+
+jobs:
+  label:
+    name: 'Label PR based on title'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: srvaroa/labeler@v0.8
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
--- a/.github/workflows/merge-dependencies.yml
+++ b/.github/workflows/merge-dependencies.yml
@@ -1,0 +1,23 @@
+name: 'Merge Dependencies'
+
+# https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/
+# https://github.com/ahmadnassri/action-dependabot-auto-merge/issues/60#issuecomment-806027389
+on: [pull_request_target]
+
+jobs:
+  auto-merge:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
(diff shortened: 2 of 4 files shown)


.github/workflows/generate-labels.yml
```diff
@@ -0,0 +1,14 @@
...
+    types: [opened, edited, synchronize]
+
+jobs:
+  label:
+    name: 'Label PR based on title'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: srvaroa/labeler@v0.8
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
```
Shadowfiend, line 14: Can you help me understand what this brings to the table? Adding labels based on info already searchable from the title doesn't seem immediately useful on its face, but very possible I'm missing something.

.github/workflows/merge-dependencies.yml
```diff
@@ -0,0 +1,23 @@
...
+    if: github.actor == 'dependabot[bot]'
+    steps:
+      - name: 'Checkout repository'
+        uses: actions/checkout@v2.3.4
+      - name: 'Automerge dependency updates from Dependabot'
+        uses: ahmadnassri/action-dependabot-auto-merge@v2.4.0
+        # Guarantee that commit comes from Dependabot (don't blindly trust external GitHub Actions)
+        if: github.actor == 'dependabot[bot]'
+        with:
+          github-token: ${{ github.token }}
```
Shadowfiend, line 23: I'm not sure I'm comfortable auto merging dependency updates, at least not yet—much as it pains me. The potential for supply chain attacks (not to mention introducing bugs) with something automated like dependabot isn't something I feel we've fully explored yet.

Shadowfiend (commented):
Thank you for the contribution! Had a few questions and maybe a thing or two we need to discuss on the team before sending these through due to the nature of the changes; left my questions below 😊

.github/workflows/generate-labels.yml
xwartz: We can use labels to manage and filter PRs, not only to help better search, but also to help reviewers distinguish PR categories at a glance, just like issues.

.github/workflows/merge-dependencies.yml
xwartz: ok, I understand what you mean, let's remove the automatic merge.

mhluongo:
@xwartz I'm closing this for now — we don't want to auto-merge for security reasons, and we prefer users and devs input issue labels themselves rather than doing it via a title and an action that tries to guess their intent.

I imagine we'll want to reconsider that last decision down the road though.

Thanks for the suggestion and keep them coming!

PR #567, opened 2021-12-10
🤽 Enrichment Activities: Flesh out an early EnrichmentService

mhluongo:
Pull transaction enrichment into a dedicated service.


background/services/enrichment/index.ts
```diff
@@ -52,80 +48,162 @@ export type ContractInfo =
...
+        contractInfo = {
+          type: "contract-interaction",
+          // Include the logo URL if we resolve it even if the interaction is
+          // non-specific; the UI can choose to use it or not, but if we know the
+          // address has an associated logo it's worth passing on.
+          contractLogoURL,
+        }
+      }
+
+      this.emitter.emit("resolvedContractInfo", {
```
Shadowfiend, line 170: My 2c: keep resolve pure and let enrich emit both events.

background/services/index.ts
```diff
@@ -1,19 +1,15 @@
...
-}
+export { default as BaseService } from "./base"
+export { default as ChainService } from "./chain"
+export { default as EnrichmentService } from "./enrichment"
+export { default as IndexingService } from "./indexing"
+export { default as KeyringService } from "./keyring"
+export { default as NameService } from "./name"
+export { default as PreferenceService } from "./preferences"
+export { default as ProviderBridgeService } from "./provider-bridge"
+export { default as InternalEthereumProviderService } from "./internal-ethereum-provider"
```
Shadowfiend, line 15: Ooh de fancy is shmancy.

Shadowfiend (commented):
ADD A FEW THINGS ALREADY

But really maybe adding things in a separate PR?

background/services/enrichment/index.ts
mhluongo: I have.. other plans, but noted

mhluongo: Since I didn't hit said plans in this PR... the teaser is that I want to decompose "transaction decoding", "subsequent transaction lookup", "address lookup", and a few other ideas. This method isn't going to survive in it's current form.

ui/components/Wallet/WalletActivityListItem.tsx
```diff
@@ -24,25 +28,58 @@ export default function WalletActivityListItem(props: Props): ReactElement {
...
+        ...renderDetails,
+        label: renderDetails.isSend ? "Sent" : "Received",
         recipient: truncateAddress(activity.contractInfo.recipientAddress),
         assetLogoURL: activity.contractInfo.contractLogoURL,
         assetSymbol: activity.contractInfo.assetAmount.asset.symbol,
         assetValue: activity.contractInfo.assetAmount.localizedDecimalAmount,
       }
       break
+    case "asset-approval":
+      renderDetails = {
```
henryboldi, line 72: Shouldn't this logic be handled in a selector?

ui/components/Wallet/WalletActivityListItem.tsx
```diff
@@ -24,25 +28,58 @@ export default function WalletActivityListItem(props: Props): ReactElement {
...
+    label: "Contract interaction",
     recipient: activity.toTruncated,
     assetLogoURL: undefined,
     assetSymbol: activity.asset.symbol,
     assetValue: activity.localizedDecimalValue,
+    isSend: false,
+  }
+
+  if (activity.value && !activity.input) {
+    if (activity.to && sameEVMAddress(asAccount, activity.to)) {
```
henryboldi, line 43: Again, seems like this is more appropriate in a selector?

ui/components/Wallet/WalletActivityListItem.tsx
mhluongo: I actually think this entire file should be burned to the ground and salted. Trying to kick it to the next PR though... if you see what was there before it has the same issues.

ui/components/Wallet/WalletActivityListItem.tsx
mhluongo: Similar comment above :) this needs to be rethought, but I'm planning to blow away the whole data structure / odd render logic after this release so I figured I'd wait.

ui/components/Wallet/WalletActivityListItem.tsx
henryboldi: Why should the entire file be burned to the ground?

Shadowfiend: I'm assuming that remark means the same as the TODO at the top of the renderDetails block :grin:

mhluongo: > Why should the entire file be burned to the ground?

All 3 of us have taken a swing at this pinata without a longer-term plan. We've got an imperative style that doesn't fit the problem, and a mishmash of render logic and categorization / business rules that need to be separated.

Good example is the old `isSent` activity property... which was being used in a few ways, but doesn't actually represent whether an activity is a send. I've band-aided that bit out for now, but this needs to be scrapped and moved to a model where the tx is run through a pipeline of categorization rules, then the render just... renders.

Anyway, normal thing that happens.

mhluongo:
Okay, tactical retreat to a smaller PR. I've stripped out everything after the `EnrichmentService` and ABIs as I think they're non-controversial and already got a light review. I'll follow up with another PR that may or may not continue the game of `WalletActivityListItem` hot potato... but hoping this is mergeable in the meantime.

Shadowfiend:
> But really maybe adding things in a separate PR?

So that was a no? 😄

mhluongo:
It doesn't anymore, I force-pushed all the parts you guys didn't like away

henryboldi:
So are #642 and #588 handled now? You force-pushed them into main?

mhluongo:
> You force-pushed them into main?

I'm not a barbarian, no!

> So are #642 and #588 handled now? 

They are now hiding in another branch for a follow-on PR I'm working on now

henryboldi:
> I'm not a barbarian, no!

Sometimes I wonder ahahaha


> They are now hiding in another branch for a follow-on PR I'm working on now

Awesome :)

mhluongo:
I like to walk a careful line on the edge of civilized society :grin: 

If you or @Shadowfiend or anyone really give this a go-ahead, the one big comment from this state of the PR is about functional purity... and something that'll be irrelevant in a PR or 2. Should now be uncontroversial

henryboldi (approved):
LGTM

ui/components/Wallet/WalletActivityListItem.tsx
mhluongo: Kicked this to the next PR.

ui/components/Wallet/WalletActivityListItem.tsx
mhluongo: Kicked this to the next PR. Note that it's still not handled in a selector for what I think are good reasons.

PR #569, opened 2021-12-11
Small improvements from dApp signing confirmation work

Shadowfiend:
Start tracking network alongside account in the UI slice's
currentAccount info, use origin instead of full URL for dApp
permissions, and abstract the current account total to a
selector as used by `DAppConnectRequest`.

Tagging @greg-nagy and @mhluongo, either one should feel free
to merge if changes look good.


background/redux-slices/selectors/dappPermissionSelectors.ts
```diff
@@ -20,6 +20,6 @@ export const selectPendingPermissionRequests = createSelector(
 export const selectCurrentPendingPermission = createSelector(
   selectPendingPermissionRequests,
   (permissionRequests) => {
-    return permissionRequests[0]
+    return permissionRequests.length > 0 ? permissionRequests[0] : undefined
```
greg-nagy, line 23: I think this is not necessary. If there is no elementt in `permissionRequest` then ` permissionRequests[0] ` will return `undefined` anyway. 

Not sure if the intention was to check for array type, but if so I would use `isArray()` or `Array.isArray()`.

ui/pages/DAppConnectRequest.tsx
```diff
@@ -57,19 +57,7 @@ function RequestingDAppBlock(props: {
...
-
-  const accountTotals = accountTotalsByCategory["read-only"]
-  if (typeof accountTotalsByCategory?.imported !== "undefined") {
-    accountTotals?.concat(accountTotalsByCategory?.imported)
-  }
-
-  const currentAccountTotal = accountTotals?.filter(
-    (accountTotal) => accountTotal?.address === currentAccount
-  )[0]
+  const currentAccountTotal = useBackgroundSelector(selectCurrentAccountTotal)
```
greg-nagy, line 60: Noice Borrklyn99

greg-nagy (approved):
Nice changes, let's do this 🐶

background/redux-slices/selectors/dappPermissionSelectors.ts
Shadowfiend: It will, but the type my won't reflect it. This is a historical decision typescript made to preserve compatibility with tons of existing code---but we want to know about the difference so we can act if needed.

PR #570, opened 2021-12-11
🧙🏻‍♂️ You shall not pass: Persisted permissions

### LINKED ISSUE #481: dApp permission request UI persisted permission handling implementation
- states: store, retrieve, validate, pending
- favicon/assets store
- error handling

### LINKED ISSUE #564: [Bug]: Dapp connections UI displaying dummy 'sushiswap' connection
### Discord Discussion Link

_No response_

Chrome

-

### What did not work?

-

Screen Shot 2021-12-10 at 3 54 08 PM

### Version

_No response_

### Relevant log output

_No response_

greg-nagy:
CleanShot 2021-12-11 at 15 45 53

# Purpose

- [x] handle the persistent storage of the permissions in IndexedDB 
- [x] initialiy populate redux store from the saved list
- [x] wire the [⚡️ Connected dApp status UI](https://github.com/tallycash/extension/pull/541) parts

# Note
‼️ I have added the `activeTab` permission, so I can query the url and other info for the current page. Should not cause any issue, but wanted to mention. 

See #391 
Closes #481, closes #564


background/services/provider-bridge/db.ts
```diff
@@ -1,25 +1,25 @@
...
 
+function keyBy(
+  permissionsArray: Array<PermissionRequest>,
+  key: keyof PermissionRequest
+): Record<string, PermissionRequest> {
+  return permissionsArray.reduce((previous, current) => {
+    // Reduce is meant to be used like this.
+    /* eslint-disable no-param-reassign */
+    previous[current[key]] = current
+    return previous
```
Shadowfiend, line 17: Reduce can be used like this, but is not meant to :) The airbnb style guide explicitly requires either a special name (e.g. `acc`) or non-mutating updates, e.g.:

```suggestion
    // Reduce is meant to be used like this.
    /* eslint-disable no-param-reassign */
    previous[current[key]] = current
    return {
      ...previous,
      [current[key]]: current
    }
```

background/services/provider-bridge/index.ts
```diff
@@ -58,6 +55,7 @@ export default class ProviderBridgeService extends BaseService<Events> {
     private internalEthereumProviderService: InternalEthereumProviderService
   ) {
     super()
+    this.init()
```
Shadowfiend, line 58: This should be in `startService`. The store is not hooked up before the constructor runs, so if the emitted event is handled correctly here it's accidental because `init` is declared async and therefore wrapped in a promise that is scheduled later. The constructor/`startService` distinction exists for this reason; see the `Service` docs:

https://github.com/tallycash/extension/blob/7d2cceb8f7e6d2a387f4eeee49178a02a6f97587/background/services/types.ts#L30-L36

background/services/provider-bridge/index.ts
```diff
@@ -71,6 +69,13 @@ export default class ProviderBridgeService extends BaseService<Events> {
     // TODO: on internal provider handlers connect, disconnect, account change, network change
   }
 
+  async init() {
+    this.emitter.emit(
+      "initializeAllowedPages",
+      await this.db.getAllPermission()
+    )
```
Shadowfiend, line 76: Worth noting: we don't actually do this with any other service right now. The systemic assumption is that the redux store has been persisted and we don't need to rehydrate it from indexeddb at startup.

We do want to change that eventually, but haven't established a set pattern for it.

ui/components/Core/CorePage.tsx
```diff
@@ -16,11 +23,59 @@ interface Props {
...
+  }, [initPermissionAndOrigin])
+
+  const deny = useCallback(async () => {
+    if (typeof currentPermission !== "undefined") {
+      await dispatch(
+        denyOrRevokePermission({ ...currentPermission, state: "deny" })
+      )
+    }
+    window.close()
+  }, [dispatch, currentPermission])
```
Shadowfiend, line 78: Yikes, this is a huge broadening of the responsibility of `CorePage`. I suggest we move it into `TopMenuConnectedDAppInfo`—even though generally I think params are better than selectors, I think if the selectors are inherent to the functionality of the component it should do internal lookup, otherwise you end up with serious responsibility creep on the parent components.

`TopMenuConnectedDAppInfo` can also be in charge of returning `<></>` if there is no dApp connection. That places all responsibility on what to do with the actual dApp connection data squarely in there, instead of up here.

ui/public/popup.html
```diff
@@ -1,12 +0,0 @@
...
-		<meta charset="utf-8" />
-		<title>Tally</title>
-		<link rel="stylesheet" href="./fonts.css" />
-		<link rel="stylesheet" href="./index.css" />
-	</head>
-	<body>
-		<div id="tally-root" />
-		<script src="../ui.js" type="text/javascript"></script>
-	</body>
-</html>
```
Shadowfiend, line 12: This... Was probably a mistake?

greg-nagy: Shit ... I meant to delete [dapp-connect-popup.html](https://github.com/tallycash/extension/blob/3a9ede12963646c927dc3251a39c218a56a2d925/ui/public/dapp-connect-popup.html)

background/services/provider-bridge/index.ts
greg-nagy: Thanks, I meant to do this. 

I believe you meant  [internalStartService()](https://github.com/tallycash/extension/blob/fda29b3e4dadbbea77cfb7cd15e07df83f2b479c/background/services/base.ts)

background/services/provider-bridge/db.ts
greg-nagy: Can you please ellaborate on the reasons behind the non-mutating updates _here_? 

As per my current understanding it's meant to protect the initial object from being modified or if the internal steps shared their object instance.

I changed the variable name to `acc` so our linter is happy now.
I chose previous as I thought it to be more descriptive/easier to read, but `acc` is what I usually use.

Just for reference: my initial implementation :) 
`then((p) => p.reduce((a, c) => ((a[c.origin] = c), a), {}))`

background/services/provider-bridge/db.ts
```diff
@@ -28,32 +28,29 @@ export class ProviderBridgeServiceDatabase extends Dexie {
 
     this.version(1).stores({
       migrations: "++id,appliedAt",
-      dAppPermissions: "&dAppID,permissions.allowedAccounts",
+      dAppPermissions: "&origin,faviconUrl,title,state",
```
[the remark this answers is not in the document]
greg-nagy: I don't know what an account means here so I will need a bit more info to be able to use that here. 

I was wondering a lot what could you mean by dAppID as we don't have anything unique to a dapp. It seemed to be a better idea not to introduce a totally new term just for the db.

background/services/provider-bridge/index.ts
mhluongo: Indexing does a poor job of it with cached assets. Nice to see us try a few different approaches before formalizing

background/services/provider-bridge/db.ts
```diff
@@ -28,32 +28,29 @@ export class ProviderBridgeServiceDatabase extends Dexie {
...
     })
   }
 
-  /**
-   * Look up existing permissions for a given dAppID.
-   */
-  async getDAppPermissions(
-    dAppID: string
-  ): Promise<ProviderPermissions | undefined> {
+  async getAllPermission() {
```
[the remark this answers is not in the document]
greg-nagy: It's kind of the format that it's stored in. I expected the `.toCollection()` to return this but that's not the case sadly. 
I was pondering about too, but went with this, because everywhere else this format is used consistently and it's also kind of close to the internal data format too.

henryboldi (commented):
Amazing!! This is fantastic 🧙👏

I gave this a test drive, and it appears to work correctly! Specially, I tested connecting to a dApp and seeing the connected to dApp info. I haven't noticed any side effects outside of the added features here. 

That said, I wouldn't consider my testing totally exhaustive. And I haven't reviewed the code here. I want to give this "emergency merge authorization" but would first like someone else's approval.

background/services/provider-bridge/db.ts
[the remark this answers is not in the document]
greg-nagy: I have refactored the code to have `accountAddress` in the schema, and it's set to the current account address on every permission grant.

ui/components/TopMenu/TopMenu.tsx
```diff
@@ -8,13 +8,15 @@ interface Props {
   toggleOpenProtocolList: () => void
   toggleOpenNotifications: () => void
   toggleOpenDAppConnectionInfo: () => void
+  isConnectedToDApp: boolean
```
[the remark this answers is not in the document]
greg-nagy: Added this to https://github.com/tallycash/extension/issues/366 as a follow up task.

ui/components/Core/CorePage.tsx
greg-nagy: Added this to https://github.com/tallycash/extension/issues/366

background/services/provider-bridge/db.ts
```diff
@@ -28,32 +28,29 @@ export class ProviderBridgeServiceDatabase extends Dexie {
...
-  ): Promise<string> {
-    return this.dAppPermissions.put({
-      dAppID,
-      permissions,
-    })
+  async setPermission(
+    permission: PermissionRequest
+  ): Promise<string | undefined> {
+    if (permission.state !== "allow") return Promise.resolve(undefined)
+    return this.dAppPermissions.put(permission)
```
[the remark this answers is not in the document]
greg-nagy: Removed this completly because I felt that the db layer should not be responsible for business concerns. Moved this check into the `ProviderBridgeService`

provider-bridge-shared/index.ts
```diff
@@ -3,7 +3,7 @@ export * from "./eip-1193"
 
 export type PermissionRequest = {
   origin: string
-  favIconUrl: string
+  faviconUrl: string
```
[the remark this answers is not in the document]
greg-nagy: Added this to https://github.com/tallycash/extension/issues/366 as a follow up task.

henryboldi (approved):
🤪

background/services/provider-bridge/db.ts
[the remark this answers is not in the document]
Shadowfiend: … One other thing I forgot when reviewing here is *we don't need all of this in the schema*. We only look things up by origin, that's all we should have in the schema, otherwise we force migrations for things that are completely unnecessary in the future.

PR #597, opened 2021-12-13
🖨️ Temporarily use Etherscan instead of semi-working detail menu

henryboldi:
⚠️ I'll add a description here with context. I saw some complaints in the OG channel.
Issue


henryboldi:
Closing this in favor of fixing the activity detail panel opening strangely on return to the activity list

PR #599, opened 2021-12-13
Network fees ranges and gas limit handling

wojciech-turek:
This PR updates the fee selection logic with updated multipliers, we now show the estimated amount along with a maximum that can be charged.

The animation of fees timeout has been reversed to be as per the design.

Users can now specify their custom gasLimit to be set in the transaction provided it's more than `21000`. If the user mistakenly sets lower amount, we will estimate the needed gas and update.

ezgif com-gif-maker (5)

To be merged after #534

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -338,8 +338,11 @@ export default class Main extends BaseService<never> {
       }
 
       try {
-        // We use estimateGasLimit only if user did not specify the gas explicitly
-        if (options.gasLimit === 0n) {
+        // We use estimateGasLimit only if user did not specify the gas explicitly or it was set below minimum
+        if (
+          typeof options.gasLimit === "undefined" ||
+          options.gasLimit < 21000n
+        ) {
           transaction.gasLimit = await this.chainService.estimateGasLimit(
             getEthereumNetwork(),
             transaction
--- a/background/main.ts
+++ b/background/main.ts
@@ -338,10 +338,14 @@ export default class Main extends BaseService<never> {
       }
 
       try {
-        transaction.gasLimit = await this.chainService.estimateGasLimit(
-          getEthereumNetwork(),
-          transaction
-        )
+        // We use estimateGasLimit only if user did not specify the gas explicitly
+        if (options.gasLimit === 0n) {
+          transaction.gasLimit = await this.chainService.estimateGasLimit(
+            getEthereumNetwork(),
+            transaction
+          )
+        }
+        // TODO If the user does specify gas explicitly, test for success.
 
         this.store.dispatch(
           transactionRequest({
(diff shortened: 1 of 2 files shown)


background/main.ts
```diff
@@ -342,11 +342,13 @@ export default class Main extends BaseService<never> {
           chainID: "1",
           nonce: resolvedNonce,
         }
-
-        transaction.gasLimit = await this.chainService.estimateGasLimit(
-          getEthereumNetwork(),
-          transaction
-        )
+        // We use estimateGasLimit only if user did not specify the gas explicitly or it was set below minimum
+        if (options.gasLimit === 0n || options.gasLimit < 21000n) {
```
Shadowfiend, line 346: These are redundant to each other.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -83,32 +85,37 @@ export default function NetworkFeesChooser({
...
-          Number(
-            (baseFee * feeOptionData.multiplier[confidence]) / 10n
-          ).toFixed()
-        ) + option.maxPriorityFeePerGas,
-        "gwei"
-      ).split(".")[0]
+      const formatToGwei = (multiplier: bigint) => {
+        return formatUnits(
+          BigInt(Number((baseFee * multiplier) / 10n).toFixed()) +
+            option.maxPriorityFeePerGas,
```
Shadowfiend, line 110: This is now doing fixed-point math. Time to use the tools for that -> `background/lib/fixed-point.ts`, `multiplyFixedPointNumbers`.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -200,23 +212,31 @@ export default function NetworkFeesChooser({
...
+        <div className="info">
+          <div className="limit">
+            <label className="limit_label" htmlFor="gasLimit">
+              Gas limit
+            </label>
+            <SharedInput
+              id="gasLimit"
+              value={gasLimit}
+              onChange={(val) => setGasLimit(val)}
+              placeholder="21000"
```
Shadowfiend, line 230: This should be populated from the estimated gas limit. How to do this is a separate question, but basically if the gas limit is not set by the user, we should display the estimate.

Shadowfiend (approved):
Couple of notes here, which I'll be attending to as I'm going to take over this PR (-> because it's blocking dApp signing).

Shadowfiend (approved):
Undrafted, rebased, approved. Let's roll it in and keep moving ⏩

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: Understood, will do that, thanks.

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: That's an interesting problem, happy to tackle this later on.

PR #600, opened 2021-12-13
Refactor the network fees to be reusable

wojciech-turek:
My goal here is to be able to update fees from the Sign page, that requires a significant refactor on how do we handle updating and selecting the gas fees. 

In this PR the network fee selection is stored in redux so it can be retrieved on other pages and updated on other pages.

### DIFF
--- a/background/redux-slices/transaction-construction.ts
+++ b/background/redux-slices/transaction-construction.ts
@@ -19,6 +19,8 @@ export type TransactionConstruction = {
   transactionRequest?: EIP1559TransactionRequest
   estimatedFeesPerGas: EstimatedFeesPerGas | undefined
   lastGasEstimatesRefreshed: number
+  gasLimitSelected: bigint
+  selectedNetworkFee?: BlockEstimate
 }
 
 export type EstimatedFeesPerGas = {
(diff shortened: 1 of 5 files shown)


Shadowfiend:
It's not clear to me why this is necessary. Can't we figure out which fee is selected from the info we already store?

wojciech-turek:
> It's not clear to me why this is necessary. Can't we figure out which fee is selected from the info we already store?

@Shadowfiend, my reasoning was that if we want to be able to change the selected fee on the Sign page or anywhere else for that matter we need to be holding the selected option in redux so that I can show the currently selected option in `NetworkFeesChooser` on another page.

Then on change of the option I would need to dispatch an update to the transaction request in `transaction-construction` slice.

Also thought that we need to clean the state after each transaction.

Am I thinking too broad about it? Do you see a simpler solution?

Shadowfiend:
See what I did in #624 here. It's not the whole solution but I think it's moving in the direction I had in mind with a bit less state coordination.

I think the thing we'll need to store in redux is the confidence level the user has selected, per discussions with @VladUXUI . Then we can use that to keep the fees up-to-date before submitting.

PR #615, opened 2021-12-14
👻 Ghostbuster: Only store the latest dApp request in store

### LINKED ISSUE #606: [Bug]: Dapp connection pop-up (sometimes) displays app info from a previous dapp
### Discord Discussion Link

_No response_

Chrome

I log into a new dapp (sushi, curve, yearn) and click to connect Tally. I expect to see a pop-up with sushi, curve, or yearn attempting to connect ... 

### What did not work?

... but instead the icon and name of a previous dapp I used appears:

(note: iirc yearn and curve are both onboard.js but sushi appears to be web3react)

Compound displays instead of Curve
Screen Shot 2021-12-13 at 3 23 06 PM

Compound displays instead of Sushi
Screen Shot 2021-12-13 at 3 24 03 PM

Compound instead of yearn
Screen Shot 2021-12-13 at 3 23 29 PM

Uniswap displays instead of Curve
https://user-images.githubusercontent.com/7005061/145883681-ca12399f-c5c8-4cc1-a1bf-7a9edd6d0687.mov

### Version

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #613: [Bug]: Not able to connect to 1inch ... or am I?
### Discord Discussion Link

_No response_

Chrome

-

### What did not work?

not able to establish a connection! 

connection pop-up displays yearn (instead of 1inch), then when I click confirm it flashes 'balancer' (where I previously connected but am no longer conencted - see 8/9 seconds in video. no connection is made

https://user-images.githubusercontent.com/7005061/145904009-d4bd01b1-5708-446d-be4c-24cc580a7425.mov

### Version

_No response_

### Relevant log output

_No response_

greg-nagy:
I have set up the store in a way to be able to handle multiple dApp requests, but we are just showing the latest in the UI. 
This change overwrites the whole `permissionRequests` object with the incoming request, so we keep track only the latest request.  

The source of the problem seems to be that the store was really busy and user interaction come quicker then it could remove the previous request from the queue. The remove reducer is ran on the fulfillment of an async thunk. It's more than possible that something keeps that from finishing.

Closes #606, closes #613

### DIFF
--- a/background/services/provider-bridge/index.ts
+++ b/background/services/provider-bridge/index.ts
@@ -92,7 +92,7 @@ export default class ProviderBridgeService extends BaseService<Events> {
 
     // a port: browser.Runtime.Port is passed into this function as a 2nd argument by the port.onMessage.addEventListener.
     // This contradicts the MDN documentation so better not to rely on it.
-    logger.log(`background: request payload: ${JSON.stringify(event.request)}`)
+    // logger.log(`background: request payload: ${JSON.stringify(event.request)}`)
 
     const response: PortResponseEvent = { id: event.id, result: [] }
 
@@ -123,7 +123,7 @@ export default class ProviderBridgeService extends BaseService<Events> {
       response.result = new EIP1193Error(EIP1193_ERROR.unauthorized)
     }
 
-    logger.log("background response:", response)
+    // logger.log("background response:", response)
 
     port.postMessage(response)
   }
--- a/background/services/provider-bridge/index.ts
+++ b/background/services/provider-bridge/index.ts
@@ -90,10 +90,6 @@ export default class ProviderBridgeService extends BaseService<Events> {
     const faviconUrl = tab?.favIconUrl ?? ""
     const title = tab?.title ?? ""
 
-    // a port: browser.Runtime.Port is passed into this function as a 2nd argument by the port.onMessage.addEventListener.
-    // This contradicts the MDN documentation so better not to rely on it.
-    // logger.log(`background: request payload: ${JSON.stringify(event.request)}`)
-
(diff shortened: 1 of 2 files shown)


background/services/provider-bridge/index.ts
```diff
@@ -92,7 +92,7 @@ export default class ProviderBridgeService extends BaseService<Events> {
 
     // a port: browser.Runtime.Port is passed into this function as a 2nd argument by the port.onMessage.addEventListener.
     // This contradicts the MDN documentation so better not to rely on it.
-    logger.log(`background: request payload: ${JSON.stringify(event.request)}`)
+    // logger.log(`background: request payload: ${JSON.stringify(event.request)}`)
```
wojciech-turek, line 95: If we no longer need to log it, why keep it commented out in the code?

background/services/provider-bridge/index.ts
```diff
@@ -123,7 +123,7 @@ export default class ProviderBridgeService extends BaseService<Events> {
       response.result = new EIP1193Error(EIP1193_ERROR.unauthorized)
     }
 
-    logger.log("background response:", response)
+    // logger.log("background response:", response)
```
wojciech-turek, line 126: Same as before, if we no longer need to log it, why keep it commented out in the code?

background/services/provider-bridge/index.ts
greg-nagy: removed

background/services/provider-bridge/index.ts
greg-nagy: removed

wojciech-turek (approved):

PR #616, opened 2021-12-14
Put the overwriting of `window.ethereum` behind a feature flag

greg-nagy:
The overwriting of the `window.ethereum` without the possibility to turn off or any warning is quite hostile.

Note this will break most of the integrations where Tally has not been added explicitly.


src/window-provider.ts
```diff
@@ -16,4 +16,7 @@ window.tally = new TallyWindowProvider({
     window.removeEventListener("message", fn, false),
   origin: window.location.origin,
 })
-window.ethereum = window.tally
+
+if (process.env.ENABLE_DEFAULT_WALLET === "true") {
```
wojciech-turek, line 20: Please use `features.ts` for feature flags.

wojciech-turek (changes requested):

src/window-provider.ts
greg-nagy: ehh, started makeing the change, because it made sense, when I realised that I can't do that here. 

`features.ts` is located in the background package. If I import from the background webpack will bundle the whole source of the background package into the window provider adding ~3.3Mb of js. The problem is that the `window-provider` is injected into every single website that a user visits if our extension is installed.

Thank you for the heads up though, and the existence of `features.ts` is pure bliss, as I spent a couple of minutes coursing why this feature flag does not work, until I realised that I need to check for the string `true`.

wojciech-turek: Thanks for that, moving the roadblock out of the way!

wojciech-turek (approved):

PR #617, opened 2021-12-14
Update to sign page to show more data

wojciech-turek:
We now have more data to fill in on the Sig Transaction page, we can now show the proper amount, destination and value.

This PR also introduces a new feature flag `ENABLE_EDIT_NETWORK_FEE` currently hardcoded to `false` to hide the `FeeSettingButton` since it's not fully ready and configured, instead I've decided to show the estimated fee based on the `maxFeePerGas` coming from the stored transaction request.

The user is now transferred to the activities tab of the currency used.

ezgif com-gif-maker (6)


henryboldi (commented):
I'm not sure what happened, but when I test this branch locally it seems that the activities tab has disappeared completely. Doesn't seem like an issue created here though

ui/components/Shared/SharedPanelSwitcher.tsx
```diff
@@ -23,7 +23,8 @@ export default function SharedPanelSwitcher(props: Props): ReactElement {
             {panelNames[0]}
           </button>
         </li>
-        <li>
+        {/* CURRENTlY UNUSED BUT STC */}
```
henryboldi, line 26: Ok, here's where the activities tab disappeared.

henryboldi (changes requested):

ui/components/Shared/SharedPanelSwitcher.tsx
Shadowfiend: Yeah… Let's be careful editing shared components, please :pray:

Shadowfiend (dismissed):
This is working, h/t @henryboldi for the activity list fix.

background/features/features.ts
```diff
@@ -2,3 +2,4 @@ export const HIDE_ADD_SEED = process.env.HIDE_ADD_SEED === "true"
 export const HIDE_SEND_BUTTON = process.env.HIDE_SEND_BUTTON === "true"
 export const HIDE_EARN_PAGE = process.env.HIDE_EARN_PAGE === "true"
 export const HIDE_SWAP = process.env.HIDE_SWAP === "true"
+export const ENABLE_EDIT_NETWORK_FEE = false
```
Shadowfiend, line 5: We should always add these with the associated env lookup, otherwise this is just another place to look for feature flags :)

Shadowfiend (approved):
Okay nooowww we're doing it.

ui/components/Shared/SharedPanelSwitcher.tsx
wojciech-turek: Aww man, all I wanted is to hide advanced tab on the sign page. Thanks for fixing that!

background/features/features.ts
wojciech-turek: @Shadowfiend What do you think about keeping everything here? If someone adds an env variable but forgets to add it to the defaults we may experience some bugs, with everything here we should be much safer?

PR #620, opened 2021-12-14
💸  Default wallet toggle

### LINKED ISSUE #625: [Bug] Unable to log into onboard.js apps
Build #1129

Tally no longer appears as a wallet option on yearn or curve (where it did previously!). In previous builds, curve displayed a Tally icon and yearn displayed a 'connected wallet' (or something) icon, which, when I clicked it, opened Tally.

yearn:
Screen Shot 2021-12-14 at 5 53 23 PM

curve:
Screen Shot 2021-12-14 at 5 53 18 PM

This isn't a huge deal ... but since saddle doesn't open on Firefox, it effectively means we can't connect to any dapps. Also ... just tested Saddle on Chrome and when I click to log in with 'Tally', MetaMask and coinbase open but not Tally. 

Screen Shot 2021-12-14 at 6 28 01 PM

### LINKED ISSUE #638: Add 'default wallet' to onboarding + settings
Onboarding - https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD?node-id=4747:124256#133319556
Settings - https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD?node-id=3589:170309#133319616

greg-nagy:
- implemented a migration logic for the preferences service
- added the a new default value to the defaults
- added the `defaultWallet` to the ui slice
  (set the config in the slice to be optional just to be safe)
- added the toggle switch for this in the ui
- wired up the communication flow back and forth: db <> preferences service <> ui <> settings toggle
- removed the process based feature flag
- created a private communication channel to send through the internal communication
- provider bridge asks background for config on initalization
- background notifies every connected `provider-bridge` if the defaultWalletValue changes
- added onboarding step about tally being the default wallet 

left a couple of TBD notes where it would be good to have a conversation
added @ts-expect-error where I could not solve the typing. I need help with those :)

CleanShot 2021-12-14 at 22 53 32

CleanShot 2021-12-15 at 16 02 20

Closes #638, closes #625


background/services/provider-bridge/index.ts
```diff
@@ -70,8 +70,10 @@ export default class ProviderBridgeService extends BaseService<Events> {
         port.onMessage.addListener((event) => {
           this.onMessageListener(port as Required<browser.Runtime.Port>, event)
         })
+        port.onDisconnect.addListener((port) => {
+          this.openPorts.filter((openPort) => openPort !== port)
```
Shadowfiend, line 74: Have to assign it otherwise it won't update. Other than that, this looks good.

greg-nagy: sent the update, thx

background/redux-slices/ui.ts
```diff
@@ -83,25 +75,16 @@ const uiSlice = createSlice({
       initializationLoadingTimeExpired: true,
     }),
     setDefaultWallet: (
-      immerState,
+      state,
```
greg-nagy, line 78: I thought we are using immerState through createSlice? 
Are we not? Or just to be safe?

background/redux-slices/ui.ts
```diff
@@ -115,6 +98,16 @@ export const {
 
 export default uiSlice.reducer
 
+// Async thunk to bubble the setNewDefaultWalletValue action from  store to emitter.
+export const setNewDefaultWalletValue = createBackgroundAsyncThunk(
+  "ui/setNewDefaultWalletValue",
+  async (defaultWallet: boolean, { dispatch }) => {
+    await emitter.emit("newDefaultWalletValue", defaultWallet)
+    // Once the default value has persisted, propagate to the store.
+    dispatch(uiSlice.actions.setDefaultWallet(defaultWallet))
```
greg-nagy, line 107: This is the most ingenious solution that I can think of for this problem! I am loving it a lot — and a bit ashamed that I did not think about it tbh ;) 

🧡🧡🧡

background/redux-slices/ui.ts
Shadowfiend: All state is wrapped in `immer`, but we can choose to use the mutable proxy demon spawn or treat it normally.

I'd like us to avoid using mutability unless we're doing some really deep updates. In those cases, I generally also drop the `immer` prefix, which I use as a warning to the reader that I'm about to do unholy things that will require confession later.

background/redux-slices/ui.ts
Shadowfiend: I almost thought I couldn't do it tbh. Remembered at the last minute that we get dispatch passed into the function in the thunk config param.

Shadowfiend (approved):
Go.

PR #623, opened 2021-12-14
🌊🧂 Sea Salt: Pull provider keys from GitHub secrets

mhluongo:
Use correct data provider keys in CI builds.

Let's use this opportunity to thank the pagan cryptography gods for [libsodium](https://github.com/jedisct1/libsodium/blob/master/AUTHORS), without which [GitHub secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) would not be possible.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,10 @@ jobs:
         node-version: "${{ steps.nvm.outputs.NVMRC }}"
         cache: "yarn"
     - run: yarn install --frozen-lockfile
-    - run: yarn build
+    - if: github.ref != 'refs/heads/main'
+      run: yarn build
+    - if: github.ref == 'refs/heads/main'
+      run: yarn build
       env:
         ALCHEMY_KEY: ${{ secrets.ALCHEMY_API_KEY }}
         BLOCKNATIVE_API_KEY,: ${{ secrets.BLOCKNATIVE_API_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -20,7 +20,13 @@ jobs:
...
         cache: "yarn"
     - run: yarn install --frozen-lockfile
-    - run: yarn build
+    - if: github.ref != 'refs/heads/main'
+      run: yarn build
+    - if: github.ref == 'refs/heads/main'
+      run: yarn build
+      env:
+        ALCHEMY_KEY: ${{ secrets.ALCHEMY_API_KEY }}
+        BLOCKNATIVE_API_KEY,: ${{ secrets.BLOCKNATIVE_API_KEY }}
```
Shadowfiend, line 29: Hmmm… This isn't the same condition we've been using before (see below, where we check for the `push` event), and this one is brittle to branch name changes (unlikely to change, I know)… We should also really use the step name to distinguish the main vs not build…

But let's ship it for now :shipit:

Shadowfiend (approved):

.github/workflows/build.yml
mhluongo: Yep, all of those things.

mhluongo: Note this `ref` check is their suggestion in the forums though, unless we want to move the branch up to make the whole job conditional... which I wanted to avoid.

PR #624, opened 2021-12-14
Fee Amalgamation: Move fee selection button into fee chooser component

Shadowfiend:
This makes adding a fee chooser with self-contained state a matter of invoking one component, instead of wiring up several bits of state between two components.

Also some React error fixes.


mhluongo (approved):
:rocket:

mhluongo:
No blocking comments, doing it

Issue #549, opened 2021-12-10
Use forms throughout the extension


mhluongo:
Tab order and enter-to-submit aren't working across the product. Using `<form>` where appropriate is a good baby step toward a more keyboard-friendly / accessible product.

PR #639, opened 2021-12-15
⌨️ Use forms

henryboldi:
Handles some of #549, but misses pages like send and swap.

Have you ever experienced the acute pain of hitting enter, being disappointed, and then clicking the submit button? Gone are those days.


ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -57,30 +57,38 @@ export default function KeyringSetPassword(): ReactElement {
...
+            onChange={handleInputChange(setPasswordConfirmation)}
+            errorMessage={passwordErrorMessage}
+          />
+        </div>
+        <SharedButton
+          type="primary"
+          size="large"
+          onClick={dispatchCreatePassword}
+          showLoadingOnClick={!passwordErrorMessage}
+          isFormSubmit
```
Shadowfiend, line 87: Buttons have a `type` attribute and a `disabled` attribute--no need to make up our own for either. 

I realize we've overloaded `type`, and now (we'll, in a follow up PR haha) is an excellent time to use a better name there and remove one more surprise point for someone unfamiliar with the codebase 💪

ui/components/Shared/SharedInput.tsx
```diff
@@ -45,7 +45,8 @@ export default function SharedInput(props: Props): ReactElement {
           input[type="number"] {
             -moz-appearance: textfield;
           }
-          .error {
+          .error,
+          .error:focus {
```
Shadowfiend, line 49: Just a note that inputs and their focus states look identical is not great. One of our other accessibility/usability issues right now is that our focus states are generally indistinguishable from unfocused states right now. 

Also a future note, since this is a consistency thing ofc.

Shadowfiend (approved):
This is glorious work and you should be proud of it 😁 

The only additional note I'll add is that components that are entirely forms,  even if they have warning content and such, should just use the form as their top-level container if possible. It can substitute for a vanilla div every day of the week and make the structure better for it.

Shadowfiend (approved):
Test drove, had to restart the continuous build to pick up the index.css change so burned some time on that 🙃  Let's go!

PR #643, opened 2021-12-15
⚖️ Alchemical Balance: Work around Alchemy API edge case pulling token balances

mhluongo:
Pulling from token balances from latest block height means we run the risk of hitting a node playing catch-up. Instead, look for asset transfers 100 blocks before the current block height.


Shadowfiend (approved):
I hate it. Let's do it.

PR #650, opened 2021-12-15
Default fee auto-select and sign page redirect

wojciech-turek:
Fix: When user navigates to a component with `NetworkFeesChooser` the `Regular` fee option will be auto-selected.

Example:
 * When user on the Send page selects `Express` as the fee option and goes to Sign and then rejects the signing he is returned to the Send page, but the fee option will be reverted back to `Regular`.
 
 FIx: Sign page should now show the maxFeePerGas from the transaction details.
 
 Fix: After signing the transaction users are redirected to the respective `SingleAsset` page.

### DIFF
--- a/ui/components/NetworkFees/NetworkFeesChooser.tsx
+++ b/ui/components/NetworkFees/NetworkFeesChooser.tsx
@@ -195,6 +195,17 @@ export default function NetworkFeesChooser({
   }, [updateGasOptions])
 
   useEffect(() => {
+    if (gasOptions.length > 0) {
+      onSelectFeeOption({
+        confidence: Number(gasOptions[activeFeeIndex].confidence),
+        price: gasOptions[activeFeeIndex].price,
+        maxFeePerGas: gasOptions[activeFeeIndex].maxFeePerGas,
+        maxPriorityFeePerGas: gasOptions[activeFeeIndex].maxPriorityFeePerGas,
+      })
+    }
+  }, [activeFeeIndex, onSelectFeeOption, gasOptions])
+
+  useEffect(() => {
     const selectedOptionIndex = gasOptions.findIndex(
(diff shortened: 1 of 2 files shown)


ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -194,6 +194,17 @@ export default function NetworkFeesChooser({
...
+  useEffect(() => {
+    if (gasOptions.length > 0) {
+      onSelectFeeOption({
+        confidence: Number(gasOptions[activeFeeIndex].confidence),
+        price: gasOptions[activeFeeIndex].price,
+        maxFeePerGas: gasOptions[activeFeeIndex].maxFeePerGas,
+        maxPriorityFeePerGas: gasOptions[activeFeeIndex].maxPriorityFeePerGas,
+      })
+    }
+  }, [activeFeeIndex, onSelectFeeOption, gasOptions])
```
Shadowfiend, line 206: Thank youuuuuu

ui/pages/SignTransaction.tsx
```diff
@@ -71,17 +69,16 @@ export default function SignTransaction(): ReactElement {
...
 
   useEffect(() => {
-    if (isTransactionSigning && isTransactionSigned) {
-      history.push("/")
-    }
-  }, [history, isTransactionSigned, isTransactionSigning])
-
-  useEffect(() => {
-    if (areKeyringsUnlocked && isTransactionSigned && signing) {
-      setSigning(false)
```
Shadowfiend, line 81: Oh goodness I ended up leaving both eh 😅

Nonetheless, I think we should still `setSigning(false)` here for completeness, even though future renders *should* reset this.

Shadowfiend (approved):
This should have been 2-3 commits!

That said, looks good let's do it.

PR #652, opened 2021-12-15
Safe env builds

### LINKED ISSUE #651: Transaction not sending/broadcasting
Tried to send ETH twice, each transaction is blocked in pending.
I got a hash but no information on etherscan. Assuming it's not being broadcasted.

Shadowfiend:
Few things here:
- Make builds error if the environment does not include variables
  listed in .env.example. Right now these are only API keys, as all
  other variables have defaults internally.
- Fix the blocknative API key env variable in our main builds.
- Give the main/non-main workflows step names that reflect what they do.
- Use the same conditional for main forking in the rest of the build
  workflow.

Fixes #651.


.github/workflows/build.yml
```diff
@@ -20,16 +20,17 @@ jobs:
         node-version: "${{ steps.nvm.outputs.NVMRC }}"
         cache: "yarn"
     - run: yarn install --frozen-lockfile
-    - if: github.ref != 'refs/heads/main'
+    - name: Build artifacts for verification
+      if: github.ref != 'refs/heads/main'
       run: yarn build
-    - if: github.ref == 'refs/heads/main'
+    - name: Build artifacts for upload
+      if: github.ref == 'refs/heads/main'
```
greg-nagy, line 27: It might be a good idea to build on every PR so it's easy to share in progress state with testers. 
But probably will not be important after release.

.env.example
```diff
@@ -0,0 +1,2 @@
+ALCHEMY_KEY=
```
greg-nagy, line 1: Why does it go in a separate file? Could this be in .env.defaults? My assumption is not, because otherwise it would be there. But why? :)

greg-nagy (approved):
let's go 🚀

Shadowfiend:
Hah. Caught myself for the dev build here. I will add dev keys for the dev builds.

.github/workflows/build.yml
Shadowfiend: Testers should be getting builds with contents that have seen review, tbh. We're not trying to add more reviewers with testing, just give access to bleeding edge stuff we expect will work.

.env.example
Shadowfiend: I thought about it, but I think they serve different purposes. We don't set defaults for these as they're secret (we do for Alchemy at the moment, but should probably stop now that we have proper secrets set up). We're also setting defaults in there for things that are optional.

greg-nagy (approved):
shoot

PR #653, opened 2021-12-15
Truncating and formatting 🚑

wojciech-turek:
Truncation emergency truck is here. 🚑

🦺  Standardise address truncation through the app.
🦺  Add decimal truncation function.
🦺  Fixing cutting last character of addresses in few places of the app.

Emergency truck fixed some minor issues while it was there:
 🚨  Remove unnecessary colon on Sign page.
 🚨  Update truncation on the backend.
 
 🚑 💨


background/redux-slices/activities.ts
```diff
@@ -27,7 +27,7 @@ const activitiesAdapter = createEntityAdapter<ActivityItem>({
 })
 
 function truncateAddress(address: string): string {
-  return `${address.slice(0, 6)}...${address.slice(37, 41)}`
+  return `${address.slice(0, 6)}...${address.slice(37, 42)}`
```
henryboldi, line 30:
```suggestion
  return `${address.slice(0, 6)}...${address.slice(-5)}`
```

greg-nagy, line 30: Of course it's 42 ...  :)

ui/components/SignTransaction/SignTransactionSignBlock.tsx
```diff
@@ -57,12 +57,12 @@ export default function SignTransactionSignBlock({
...
         {typeof transactionDetails.to === "undefined" ? (
           <>
-            <div className="label">Send to:</div>
+            <div className="label">Send to</div>
             <div className="send_to">Contract creation</div>
           </>
         ) : (
           <>
-            <div className="label">Send to:</div>
+            <div className="label">Send to</div>
```
henryboldi, line 65: We're removing the colons?

ui/utils/truncate.ts
```diff
@@ -0,0 +1,17 @@
...
+): string | number {
+  const valueString = value.toString()
+  const decimalIndex = valueString.indexOf(".")
+  if (decimalIndex !== -1) {
+    const integers = valueString.split(".")[0]
+    const decimals = valueString.split(".")[1]
+    return `${integers}.${decimals.substr(0, decimalLength)}`
+  }
+  return value
+}
```
henryboldi, line 17:
```suggestion
export function truncateDecimalAmount(
  value: number | string,
  decimalLength: number
): string {
  const valueString = value.toString()
  if (valueString.includes(".")) {
    const [integers, decimals] = valueString.split(".")
    return `${integers}.${decimals.substr(0, decimalLength)}`
  }
  return valueString
}
```

How about this?

ui/utils/truncate.ts
```diff
@@ -0,0 +1,17 @@
+export function truncateAddress(address: string): string {
+  return `${address.slice(0, 6)}...${address.slice(37, 42)}`
+}
+
+export function truncateDecimalAmount(
+  value: number | string,
```
greg-nagy, line 6: I would prepare this function to accept undefined / null, just in case. 

Not a blocker because it's handled when it's called.

ui/utils/truncate.ts
```diff
@@ -0,0 +1,17 @@
...
+  decimalLength: number
+): string | number {
+  const valueString = value.toString()
+  const decimalIndex = valueString.indexOf(".")
```
greg-nagy, line 10: It might be a good idea to check locale for decimal sign. 

I don't have too much info about how this works but I believe this can be affected with the locale settings of the browser. 

Not a blocker.

ui/utils/truncate.ts
```diff
@@ -0,0 +1,17 @@
+export function truncateAddress(address: string): string {
+  return `${address.slice(0, 6)}...${address.slice(37, 42)}`
```
henryboldi, line 2: Again, I'd go with
```suggestion
  return `${address.slice(0, 6)}...${address.slice(-5)}`
```

ui/utils/truncate.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export function truncateDecimalAmount(
+  value: number | string,
+  decimalLength: number
+): string | number {
+  const valueString = value.toString()
+  const decimalIndex = valueString.indexOf(".")
+  if (decimalIndex !== -1) {
+    const integers = valueString.split(".")[0]
```
greg-nagy, line 12: Is it 100% that only 1 `.` will be in the string? I thought there are some cases where both the thousands and the decimals are both use `.`.

background/redux-slices/activities.ts
```diff
@@ -27,7 +27,7 @@ const activitiesAdapter = createEntityAdapter<ActivityItem>({
 })
 
 function truncateAddress(address: string): string {
```
henryboldi, line 29: Why is this not using the shared `truncateAddress` function from the util you added?

ui/utils/truncate.ts
greg-nagy: I see that the props typing protects the only input and it's set for a default value a layer above, but this has been moved out of that context.

henryboldi (commented):
It'd be nice to use the util universally

ui/utils/truncate.ts
henryboldi: I'm under the impression that this function only handles `decimalValue` and not `localizedDecimalValue`

background/redux-slices/activities.ts
henryboldi: I personally feel this should be blocking because it partly defeats the purpose of the PR

henryboldi (dismissed):

background/redux-slices/activities.ts
Shadowfiend: I don't think it does—it moves us in the correct direction.

That said, I'm 100% unopposed to you stepping in and making this use the common utility  before merging :)

henryboldi: Fair, good point!

Sweet, I'll hop in to have this use the common util before merging

henryboldi:
I'd prefer an approval from someone else now, given my changes 😬

Shadowfiend (approved):
Yes sir.

ui/components/SignTransaction/SignTransactionSignBlock.tsx
wojciech-turek: Yes, consulted with Vlad on this.

wojciech-turek:
Thank you good people! @henryboldi @Shadowfiend

PR #657, opened 2021-12-15
Navi-gation: Add default names and avatars

### LINKED ISSUE #468: Avatars
We only need a handful, which will be autoassigned whenever a user creates an account

### LINKED ISSUE #582: [Bug]: Accounts menu cleanup
### Discord Discussion Link

_No response_

Chrome

-

### What did not work?

1 - "Read-only mode" => "Read-only"
2 - Are we missing an icon next to "imported"? Design @ https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD?node-id=3584:151367#132443544 cc @VladUXUI 

Screen Shot 2021-12-11 at 5 06 15 PM

### Version

_No response_

### Relevant log output

_No response_

Shadowfiend:
Because [Na'vi](https://en.wikipedia.org/wiki/Avatar_(2009_film)#Na'vi), hyuk hyuk.

Also fixes the imported icon. The fun has arrived.

-----

Fixes #582, fixes #468.


background/redux-slices/accounts.ts
```diff
@@ -66,23 +79,46 @@ export const initialState = {
...
 
-function newAccountData(address: HexString, network: Network): AccountData {
+function newAccountData(
+  address: HexString,
+  network: Network,
+  existingAccountsCount: number
+): AccountData {
+  const accountNameDefault =
+    availableDefaultNames[
+      // Don't reuse names unless we absolutely have to.
```
mhluongo, line 89: Is this comment accurate? We aren't tightly packing here, it's deterministic as a function of just the address, yeah?

mhluongo: Ah yep we're packing. Alrighty.

mhluongo (approved):
Appears to work well imported and read-only wallets, and doesn't break with ENS names with or without an avatar. Let's do this

PR #658, opened 2021-12-15
🌤 Change default wallet to true in .env.defaults

greg-nagy:
also added a fake blocknative api key because webpack did not work otherwise


greg-nagy:
good that we are calm and collected on release day ... <.< 😁

PR #666, opened 2021-12-15
🍫 Add a snackbar for all your yummy brief feedback needs

### LINKED ISSUE #630: Copy buttons should give user feedback that they copied
The only way to be sure the button actually did the thing is to paste elsewhere. Some affordance (eg button changing briefly from "Copy" to "Copied!") would go a long way.

henryboldi:
Can be set from anywhere using `dispatch(setSnackbarMessage("Copied!"))`

Snackbar

Closes #630

### DIFF
--- a/ui/components/Snackbar/Snackbar.tsx
+++ b/ui/components/Snackbar/Snackbar.tsx
@@ -41,6 +41,7 @@ export default function Snackbar(): ReactElement {
             transition: all 0.3s ease;
             opacity: 0;
             transform: translateY(10px);
+            pointer-events: none;
           }
           .open {
             opacity: 1;
--- a/ui/components/Snackbar/Snackbar.tsx
+++ b/ui/components/Snackbar/Snackbar.tsx
@@ -1,0 +1,53 @@
+import React, { ReactElement, useEffect } from "react"
+import { useDispatch } from "react-redux"
+import classNames from "classnames"
+import {
+  selectSnackbarMessage,
+  clearSnackbarMessage,
+} from "@tallyho/tally-background/redux-slices/ui"
+import { useBackgroundSelector } from "../../hooks"
+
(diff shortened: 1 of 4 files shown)


ui/components/Snackbar/Snackbar.tsx
```diff
@@ -0,0 +1,54 @@
...
+export default function Snackbar(): ReactElement {
+  const dispatch = useDispatch()
+  const snackbarMessage = useBackgroundSelector(selectSnackbarMessage)
+  const msTillDismiss = 5000
+
+  useEffect(() => {
+    setTimeout(() => {
+      dispatch(clearSnackbarMessage())
+    }, msTillDismiss)
+  }, [snackbarMessage, dispatch])
```
itsrachelfish, line 19: You're using a single timeout to clear all snackbar messages, if a 2nd snackbar message is displayed while the 1st is still active, the 2nd snackbar message will disappear abruptly when the 1st snackbar was originally set to disappear.

You should save the return value from `setTimeout` and call `clearTimeout`

itsrachelfish:
```suggestion
  const msTillDismiss = 5000
  let snackbarTimeout;

  useEffect(() => {
    if (snackbarTimeout) {
      clearTimeout(snackbarTimeout)
    }

    snackbarTimeout = setTimeout(() => {
      dispatch(clearSnackbarMessage())
    }, msTillDismiss)
  }, [snackbarTimeout, snackbarMessage, msTillDismiss, dispatch])
```

itsrachelfish (changes requested):
Tested and looks great! There's just one small edge case I ran into when multiple messages are dispatched

itsrachelfish (approved):
Tested locally, working as expected :tada:

PR #673, opened 2021-12-15
Defaultive fees: fix issues with dapp signing in some cases

Shadowfiend:
Should fix beachballs and a couple of other things.


mhluongo:
Screenshot from 2021-12-15 22-17-23

I get a beachball trying to deposit 0.04 ETH into [Alchemix](https://app.alchemix.fi/vault) from one of my test accounts. Unsure whether the tx was signed, definitely not broadcast

mhluongo:
... it turns out if I click that spinning button... the tx is signed and broadcast!

It appears it was a bad gas estimate though :thinking: 

https://etherscan.io/tx/0x81f9e85f135685e8cabdc570118260a78ba686c65aeaa2228946bdda95a94f8e

wojciech-turek:
@Shadowfiend 

This is nice PR and simplifies the selection of a default inside the `NetworkFeesChooser` but its re-introduces the 0 Gas fee problem.

Mainly because you removed the useEffect with `onSelectFeeOption` call which is responsible for selecting the default gas and passing it to the parent page.

The transaction hold now the `useState` default value if the fee is not manually selected, and the default is all zeros.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -144,22 +143,30 @@ export default function NetworkFeesChooser({
       }
     }
     if (estimatedFeesPerGas) {
-      const regular = estimatedFeesPerGas?.regular
-      const express = estimatedFeesPerGas?.express
-      const instant = estimatedFeesPerGas?.instant
+      const { regular, express, instant } = estimatedFeesPerGas ?? {}
```
wojciech-turek, line 146: I like this, I should use it more often.

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -144,22 +143,30 @@ export default function NetworkFeesChooser({
...
+          (estimate) =>
+            typeof estimate !== "bigint" &&
+            typeof estimate !== "undefined" &&
+            estimate.confidence === selectedGas?.confidence
+        )
+        const currentlySelectedFeeIndex =
+          selectedGasFeeIndex === -1 ? 0 : selectedGasFeeIndex
+        const currentlySelectedFee = basePrices[currentlySelectedFeeIndex]
+
+        const updatedGasOptions = basePrices.map((option) =>
```
wojciech-turek, line 163: One less useEffect hook and looks elegant. +1

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -188,35 +195,12 @@ export default function NetworkFeesChooser({
...
-  useEffect(() => {
-    if (gasOptions.length > 0) {
-      onSelectFeeOption({
-        confidence: Number(gasOptions[activeFeeIndex].confidence),
-        price: gasOptions[activeFeeIndex].price,
-        maxFeePerGas: gasOptions[activeFeeIndex].maxFeePerGas,
-        maxPriorityFeePerGas: gasOptions[activeFeeIndex].maxPriorityFeePerGas,
-      })
-    }
-  }, [activeFeeIndex, onSelectFeeOption, gasOptions])
```
wojciech-turek, line 206: Removing this means we never set any fee option unless user manually clicks on an option and save choice. Why did it cause issues with the balancer?

ui/pages/SignTransaction.tsx
```diff
@@ -129,21 +131,22 @@ export default function SignTransaction(): ReactElement {
...
-
     setSelectedEstimatedFeePerGas(estimate)
     const transaction = {
       ...transactionDetails,
       maxFeePerGas: estimate.maxFeePerGas,
       maxPriorityFeePerGas: estimate.maxPriorityFeePerGas,
       gasLimit: BigInt(gasLimit),
     }
-    dispatch(updateTransactionOptions(transaction))
+    await dispatch(updateTransactionOptions(transaction))
```
wojciech-turek, line 141: I don't think there's anything to await here. There is also a linter error thrown `'await' has no effect on the type of this expression.`

ui/pages/SignTransaction.tsx
```diff
@@ -129,21 +131,22 @@ export default function SignTransaction(): ReactElement {
...
   }
 
   const handleConfirm = async () => {
+    // FIXME Hackily handle the user not interacting with the fee selector for now.
+    if (transactionDetails.maxFeePerGas === 0n) {
+      await updateGasSettings(selectedEstimatedFeePerGas)
+    }
+
```
wojciech-turek, line 149: This will currently always run as transaction details come without selected network fees.

ui/pages/SignTransaction.tsx
Shadowfiend: There actually is. All our dispatches in the UI are async because they have to be transmitted to the background script. Typing doesn't reflect that and I haven't had a chance to figure out how to fix though.

ui/pages/SignTransaction.tsx
Shadowfiend: Yep, that was the intent! We need to do something similar in the send page temporarily.

ui/components/NetworkFees/NetworkFeesChooser.tsx
Shadowfiend: The useEffect was causing an async state change that retriggered it, leading to a rerender loop that evaded React's checking for such loops, and gradually froze the browser (probably due to memory growth, though did not confirm).

Shadowfiend: Of note, the intent was to (hackily) force an option on the calling page while we got to a place where we could more completely disentangle the relationships here.

ui/pages/SignTransaction.tsx
Shadowfiend: That said, I'm not sure I found this necessary. I don't recall why I added it. So if you're seeing it cause issues and I didn't leave myself enough info to justify keeping it, let's nuke it.

mhluongo (approved):
Testing looks good to me. dApps still aren't seeing when a signature has completed, but we can handle that in a follow-up.

PR #686, opened 2021-12-16
⛔ Contracts Not Welcome: Don't show contract interactions per asset

### LINKED ISSUE #676: Contract interactions are show under ETH activities (single-asset page)
Opening this for discussion of this behavior... it seems like a bug to me, especially if the value of a transaction is 0 ETH.

Introduced in #675

mhluongo:
We still show transfers, of course, or anything that interacts directly against an asset contract. Just want to avoid the ETH single asset page becoming a dumping ground for every uncategorized tx.

Closes #676

### DIFF
--- a/ui/pages/SingleAsset.tsx
+++ b/ui/pages/SingleAsset.tsx
@@ -33,22 +33,20 @@ export default function SingleAsset(): ReactElement {
         ) {
           return true
         }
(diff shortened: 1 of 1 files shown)


henryboldi (approved):
Tested, lookin good! 🙌

PR #687, opened 2021-12-16
⚔️ Definitely Errors: communicating errors

### LINKED ISSUE #685: Reject doesn't work when connecting to dapps
It just keeps opening up a new one.
image

greg-nagy:
Implemented EIP-1193 Standard error handling and let it bubble to the dApps.

Also implemented DAppRequest deny on closing the request window.

Restructured `provider-bridge-shared` packages so it does not have dependency cycles

Note: In case of reject the dApp can be implemented in a way that it triggers the request again. I did not debounce it for now. 
https://cln.sh/C9eL3V

Closes #685


provider-bridge-shared/eip-1193.ts
```diff
@@ -30,27 +32,43 @@ export const EIP1193_ERROR = {
...
-    .includes(code)
+export function isEIP1193ErrorCodeNumber(
+  code: unknown
+): code is EIP1193ErrorCodeNumbers {
+  return (
+    isNumber(code) &&
+    Object.values(EIP1193_ERROR_CODES)
+      .map((e) => e.code as number)
+      .includes(code)
+  )
```
Shadowfiend, line 64: A lot of type magic to implement this, but I don't think it actually needs it.

You can declare a const that does most of this:

```typescript
const EIP1193_ERROR_CODES = Object.values(EIP1193_ERROR_PAYLOADS).map(
  ({ code }) => code
)
```

This is already well-typed enough, and instead of this function you can just do `EIP1193_ERROR_CODES.includes(arg.code)`. You can also skip the `isNumber` below, since `includes` won't pass if the array contains numbers but the arg isn't a number.

Shadowfiend (approved):
Looks good, let's do it.

PR #693, opened 2021-12-16
🧦 Finders Keepers: fix chain related emits and type checks

greg-nagy:


greg-nagy:
It's enough if https://github.com/tallycash/extension/pull/701 get's merged

PR #698, opened 2021-12-16
Transactional Failure: Display failed transactions in the activity list

### LINKED ISSUE #587: [Bug]: Txn details page - missing gas
### Discord Discussion Link

_No response_

Chrome

View txn details

### What did not work?

Gas is populating as 'unknown' (this is an hour after adding account)

Screen Shot 2021-12-11 at 5 49 16 PM

### Version

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #678: Properly display failed transactions
As discovered testing a [recent PR](https://github.com/tallycash/extension/pull/673), we aren't showing failed transactions in activities! Here's what a failed tx currently looks like

Screenshot from 2021-12-15 22-26-50

Instead, it should look... failed, and show the gas that was expended rather than the transaction value which didn't do anything.

mhluongo:
Screenshot from 2021-12-16 16-19-00

Does what it says on the tin! Displays when transactions fail in the activity list.

Closes #678, closes #587.

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -592,6 +592,8 @@ export default class ChainService extends BaseService<Events> {
         } else if (transaction.blockHash) {
           // Get relevant block data.
           await this.getBlockData(transaction.network, transaction.blockHash)
+          // Retrieve gas used, status, etc
+          this.retrieveTransactionReceipt(transaction.network, transaction)
         }
       } catch (error) {
         logger.error(`Error retrieving transaction ${hash}`, error)
@@ -741,7 +743,7 @@ export default class ChainService extends BaseService<Events> {
    * the database and informing subscribers via the emitter.
    *
    * @param network the EVM network we're interested in
-   * @param txHash the hash of the unconfirmed transaction we're interested in
+   * @param transaction the unconfirmed transaction we're interested in
    */
   private async subscribeToTransactionConfirmation(
     network: EVMNetwork,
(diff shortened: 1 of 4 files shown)


background/services/chain/index.ts
```diff
@@ -592,6 +592,8 @@ export default class ChainService extends BaseService<Events> {
         } else if (transaction.blockHash) {
           // Get relevant block data.
           await this.getBlockData(transaction.network, transaction.blockHash)
+          // Retrieve gas used, status, etc
+          this.retrieveTransactionReceipt(transaction.network, transaction)
```
Shadowfiend, line 596: I hate how none of our things that save say that they do in their names, fwiw. Would prefer a `fetchAndSave` or something. It's systemic though, so not really for this PR.

background/services/chain/index.ts
```diff
@@ -741,7 +743,7 @@ export default class ChainService extends BaseService<Events> {
    * the database and informing subscribers via the emitter.
    *
    * @param network the EVM network we're interested in
-   * @param txHash the hash of the unconfirmed transaction we're interested in
+   * @param transaction the unconfirmed transaction we're interested in
```
Shadowfiend, line 746: :grimacing:

ui/components/Wallet/WalletActivityListItem.tsx
```diff
@@ -105,6 +105,11 @@ export default function WalletActivityListItem(props: Props): ReactElement {
               className={classNames("activity_icon", renderDetails.iconClass)}
             />
             {renderDetails.label}
+            {"status" in activity && activity.status !== 1 ? (
+              <div className="failed">Failed</div>
```
Shadowfiend, line 109: We should really add a `status` class for `failed` and `pending` and do `.status:before` in CSS, letting the `failed`/`pending` classes just apply the `color`.

We should probably also at some point give `ActivityItem` a more delineated `status` field that carries `pending`/`failed`/`succeeded`/whatever else in one field, populated in the selectors.

Shadowfiend (commented):
Code looks good, test driving now.

Shadowfiend (approved):
Does what it says on the tin indeed!

Shadowfiend:
Actually does more than it says on the tin, added another ticket close to the description 😏

ui/components/Wallet/WalletActivityListItem.tsx
mhluongo: From a sync discussion: we have a similar issue with `timestamp` and other stuff that's sort of transaction data, sort of "joined" network data (in these cases with the block and tx receipt). The choice here was to stay closer to the EVM with that `number`... but we'll want to reconsider `EVMTransaction` soon.

mhluongo: ... and yes, `ActivityItem` needs to lift heavier.

PR #694, opened 2021-12-16
More Bags to Grab: Fix a few additional minor but handy issues

Shadowfiend:
Fixes #684, fixes #655.

Issue #260, opened 2021-09-20
Flashbots support on transaction send


mhluongo:
Allow a user to opt in to Flashbots when broadcasting a transaction. This could be as simple as a checkbox + tooltip.

Relies on #259.

beemeeupnow:
can be closed, added via #3455

Issue #259, opened 2021-09-20
Alternative tx broadcast endpoint


mhluongo:
A variety of mempool and MEV-resistance tools need a custom broadcast endpoint. The first we've run into is [Flashbots](https://docs.flashbots.net/).

Modify the `ChainService.broadcastTransaction` call to accept a different endpoint, or somehow choose providers based on included options.

Refs #258

0xDaedalus:
Closed because we already default `eth_sendRawTransaction` to alchemy when an alchemy endpoint is available.  If we want to expand this to flashbots in the future we can open a new issue.

PR #699, opened 2021-12-16
More Bags to Grab Redux: A fix for a keyring status change and an enrichment approval parameter, oh and also a showstopper fix

### LINKED ISSUE #692: dApps aren't informed once a tx is signed
Currently they just hang. Some of the smarter ones (like Uniswap) will still pick up the tx after it's been mined.

Shadowfiend:
Fixes an issue introduced in #694 where the keyring status in a fresh extension install was unlocked rather than uninitialized. Also fixes an issue where transaction enrichment would fail for token approvals due to a misnamed parameter.

Finally, fixes #692.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -20,7 +20,7 @@ import {
 } from "./services"
 
 import { KeyringTypes } from "./types"
-import { EIP1559TransactionRequest } from "./networks"
+import { EIP1559TransactionRequest, SignedEVMTransaction } from "./networks"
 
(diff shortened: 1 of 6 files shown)


mhluongo:
Successful end-to-end approval on Uniswap. Well done.

Screenshot from 2021-12-16 16-44-43

mhluongo (approved):
Love it. Also going to unblock #259 and #260

PR #719, opened 2021-12-17
Correct fee value display

wojciech-turek:
Networks fees were showing too high in $$,because in order to work with bigInts the multipliers were converted into bigInts so 1.3 = 13n but after calculating the value was not divided by 10 so we were actually showing fees x13 not x1.3. This PR fixes that by adding a division by `10n`.


henryboldi (approved):
Nice!

works

PR #720, opened 2021-12-17
Alarming Balances: Update balances when transactions are confirmed

### LINKED ISSUE #707: After dapp interaction (trade or deposit), resulting token balances don't update in Tally
### Discord Discussion Link

_No response_

Chrome

-

### What did not work?

Trade on Sushi for DAI ... doesn't show up after 10 mins / 40+ confirmations

Screen Shot 2021-12-16 at 7 00 19 PM

I reinstall the app and it shows up

Log into Yearn and deposit DAI into a Vault and get yvDAI back ... still shows 40 DAI in Tally and not new yvDAI

Screen Shot 2021-12-16 at 7 08 47 PM

Screen Shot 2021-12-16 at 7 09 10 PM

I reinstall app and it shows up again

Screen Shot 2021-12-16 at 7 14 33 PM

### Version

_No response_

### Relevant log output

_No response_

Shadowfiend:
When transactions near the chain tip with a successful status are
seen, the indexing service now schedules a token balance lookup to
occur within 1 minute. This should ensure that any transactions that
affect balances update the balance in relatively short order, even if
we don't have enough insight into the transaction itself to locally
handle those updates.

~~Fixes~~ Refs #707 .

### DIFF
--- a/background/services/chain/utils.ts
+++ b/background/services/chain/utils.ts
@@ -12,6 +12,7 @@ import {
   SignedEVMTransaction,
   AnyEVMBlock,
   EIP1559TransactionRequest,
+  ConfirmedEVMTransaction,
 } from "../../networks"
 import { FungibleAsset } from "../../assets"
 import { getEthereumNetwork } from "../../lib/utils"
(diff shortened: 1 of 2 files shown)


background/services/chain/utils.ts
```diff
@@ -147,11 +148,16 @@ export function ethersTransactionFromSignedTransaction(
...
   receipt: EthersTransactionReceipt
-): AnyEVMTransaction {
+): ConfirmedEVMTransaction {
+  const gasUsed = receipt.gasUsed.toBigInt()
   return {
     ...transaction,
-    gasUsed: receipt.gasUsed.toBigInt(),
-    status: receipt.status,
+    gasUsed,
+    status:
```
mhluongo, line 156: Right! Good catch.

mhluongo (approved):
Code looks good and I don't see any regressions, but note... this won't actually fix #707 if the resulting token balance change is a token that Alchemy doesn't balance check by default. To fix that, we need either local contract inference, balance checking by token list, or more aggressive use of `getAssetTransfers` (which is impaired by chain tip data ingestion issues)

Shadowfiend:
> won't actually fix #707 if the resulting token balance change is a token that Alchemy doesn't balance check by default

Completely agree btw. This was meant to fix it for tokens the wallet is already monitoring, but not for new ones. Missed that 707 referred to both.

PR #722, opened 2021-12-17
Display fees in activities in Gwei 🐼

### LINKED ISSUE #717: Missing gas details
I've noticed several transactions that read (unknown) for gas price after confirmation in the activity details.

None of the transactions seem to have the Amount filled in.

One key detail that I feel is missing from the activity details page is the Transaction Fee. Would also be nice if we were able to convert these to the conventional Gwei instead of ETH.
missing gas price
missing ETH amount

wojciech-turek:
Display maxFee and price in Gwei instead of ETH, remove transformation of Gas used to different unit.

Remove "Activity" from single asset page.

Closes #717

### DIFF
--- a/background/redux-slices/utils/activity-utils.ts
+++ b/background/redux-slices/utils/activity-utils.ts
@@ -1,5 +1,5 @@
 import dayjs from "dayjs"
-import { convertToEth } from "../../lib/utils"
+import { convertToEth, weiToGwei } from "../../lib/utils"
 import { AnyEVMTransaction } from "../../networks"
 import { EnrichedEVMTransaction } from "../../services/enrichment"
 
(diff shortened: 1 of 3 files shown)


background/redux-slices/utils/activity-utils.ts
```diff
@@ -12,6 +12,15 @@ function ethTransformer(
   return `${convertToEth(value)} ETH`
 }
 
+function gweiTransformer(
+  value: string | number | bigint | null | undefined
```
henryboldi, line 16: Do we need to accept all of these types for the value param?

ui/pages/SingleAsset.tsx
```diff
@@ -121,7 +121,6 @@ export default function SingleAsset(): ReactElement {
           <div className="left">Asset is on: Arbitrum</div>
           <div className="right">Move to Ethereum</div>
         </div>
-        <div className="label_light standard_width_padded">Activity</div>
```
henryboldi, line 124: 💯

henryboldi (approved):
Lookin' good!

Gwei

PR #723, opened 2021-12-17
Import legacy keyring with BIP44 derivation path

7alip:
Allow users to provide custom derivation path when importing metamask account.

<image src="https://i.ibb.co/KhZfH2S/tally-path.gif" width="300" />


ui/components/Shared/SharedCheckbox.tsx
```diff
@@ -0,0 +1,86 @@
...
+        .checkbox {
+          display: flex;
+          align-items: center;
+          position: relative;
+          cursor: pointer;
+          font-size: 14px;
+          color: var(--green-60);
+          -webkit-user-select: none;
+          -moz-user-select: none;
+          -ms-user-select: none;
```
Shadowfiend, line 34: Believe styled-jsx takes care of this kind of expansion for us.

Shadowfiend (commented):
Gave this a super-quick skim, can't ship it this week but excited!

ui/hooks.ts
```diff
@@ -40,3 +40,27 @@ export const useAreKeyringsUnlocked = (redirectIfNot: boolean): boolean => {
...
+
+    document.addEventListener("mousedown", listener)
+    document.addEventListener("touchstart", listener)
+
+    return () => {
+      document.removeEventListener("mousedown", listener)
+      document.removeEventListener("touchstart", listener)
+    }
+  }, [ref, handler]) // Reload only if ref or handler changes
+}
```
Shadowfiend, line 66: Beautiful 🥲 Would love to use this instead of our extra-large button for the dApp connection close btw @henryboldi .

henryboldi: Yessss 👏

VladUXUI:
This is awesome 😍 
Will take a quick stab at the design early next week!

ui/pages/Onboarding/OnboardingDerivationPathSelect.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { ReactElement } from "react"
+import SharedSelect from "../../components/Shared/SharedSelect"
+
+type OptionType = {
+  value: string
+  label: string
+}
+const derivationPaths: OptionType[] = [
```
bguiz, line 8: for a future PR: I'd suggest that this list get refactored and be read in from a dedicated network configs file.

ui/hooks.ts
bguiz: 🙌

Shadowfiend:
What do y'all think about parking this behind a feature flag while the design piece settles out? Feels like a good way to avoid having to chase merges as we work on wrapping up that part.

mhluongo:
Seconded. I have a feeling we'll want this on a separate page shortly, but want to get the bulk of the review closed out now. @7alip @bguiz that work for you?

alepc253:
@mhluongo so what would be the next steps?

7alip:
Hey @mhluongo, I expected this PR to be considered when you start adding multi chain feature soon.
.

mhluongo:
Indeed, we just want to keep the main branch ship-able at all times and don't want to leave this PR open longer than necessary. Thus @Shadowfiend's feature flag idea — adding one gets this merged sooner rather than later.

Here's what I've done in the [other multi-network PR](bd0cc3e16f0d78dff7817041dfc7f14d3410dcc7).

Shadowfiend:
Yep, just to clarify what I'm proposing here:
- Adding an entry to [`background/features/features.ts`](https://github.com/tallycash/extension/blob/main/background/features/features.ts), probably `HIDE_IMPORT_DERIVATION_PATH`.
- Adding an entry to [`.env.defaults`](https://github.com/tallycash/extension/blob/main/.env.defaults) setting `HIDE_IMPORT_DERIVATION_PATH` to `true`.
- Rendering [the checkbox and select](https://github.com/tallycash/extension/pull/723/files#diff-6832ef2a8b409d32b90ac3dc4fd6908a103b9c59260905b5f4827aa5cf73a57cR132-R145) conditionally on that feature flag. I'm currently assuming that not rendering those two will leave the UI looking roughly the same as before this PR; if that's not the case, we can fork the file itself to `OnboardingImportMetamaskWithDerivation.tsx` and conditionally render it over the standard `OnboardingImportMetamask` [in `Popup.tsx`](https://github.com/tallycash/extension/blob/e65eaad0f6bc10bd513a0e4c8f27a48a96db39aa/ui/pages/Popup.tsx#L68).

This will let us wrap up review here, merge the PR, and continue to iterate within the repository as additional multi-chain stuff lands, while keeping the functionality out of the release build until we're ready to flip the switch on a few of these pieces together.

How does that sound?

7alip:
Thanks for the clear explanations @Shadowfiend. The only thing I changed in the existing UI is reducing the size of the image by 3/4 on the `OnboardingImportMetamask` page. I may conditionally change the sizes of the image like below:

```jsx
.metamask_onboarding_image {
  background: url("./images/illustration_import_seed@2x.png");
  background-size: cover;
  width: ${HIDE_IMPORT_DERIVATION_PATH ? "154px" : "205.3px"};
  height: ${HIDE_IMPORT_DERIVATION_PATH ? "144.75" : "193px"};
  margin-top: 27px;
  margin-bottom: 13px;
}
```

Shadowfiend:
Sounds glorious!

background/services/keyring/index.ts
```diff
@@ -290,7 +290,9 @@ export default class KeyringService extends BaseService<Events> {
       throw new Error("Invalid legacy mnemonic.")
     }
 
-    const newKeyring = new HDKeyring({ mnemonic })
+    const newKeyring = path
+      ? new HDKeyring({ mnemonic, path })
+      : new HDKeyring({ mnemonic })
```
7alip, line 295: `path` is optional but `new HDKeyring({ mnemonic, path })` was giving error when `path` is not provided (`undefined`).

mhluongo (approved):
Looks good to me. Let's do this!

bguiz:
> 



> What do y'all think about parking this behind a feature flag while the design piece settles out? Feels like a good way to avoid having to chase merges as we work on wrapping up that part.

Yes this sounds good, and glad this has been merged! 🎉 🎉 🎉 

I'd love to get the RSK community to beta test this ASAP, can we document exactly how to use the feature flags? (I'm not too clear myself, is this toggle-able at run time, or is this a build time/ compile time thing at the moment?)

PR #724, opened 2021-12-17
Rejectamundo: Proper support for transaction signing rejection

### LINKED ISSUE #690: Can't reject Sign Tx from DApps
Tested on multiple DApps, clicking on reject from sign tx doesn't do anything

Shadowfiend:
Also fixes some corner cases where an RPC request could fail
but the calling dApp would never find out about it.

Fixes #690.

### DIFF
--- a/background/services/internal-ethereum-provider/index.ts
+++ b/background/services/internal-ethereum-provider/index.ts
@@ -51,16 +51,27 @@ export default class InternalEthereumProviderService extends BaseService<Events>
       if (port.name === INTERNAL_PORT_NAME) {
         port.onMessage.addListener(async (event) => {
           logger.log(`internal: request payload: ${JSON.stringify(event)}`)
-          const response = {
-            id: event.id,
-            result: await this.routeSafeRPCRequest(
-              event.request.method,
-              event.request.params
-            ),
-          }
-          logger.log("internal response:", response)
+          try {
+            const response = {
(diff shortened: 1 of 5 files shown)


background/services/internal-ethereum-provider/index.ts
```diff
@@ -51,16 +52,27 @@ export default class InternalEthereumProviderService extends BaseService<Events>
...
+
+            port.postMessage(response)
+          } catch (error) {
+            logger.log("error processing request", event.id, error)
 
-          port.postMessage(response)
+            port.postMessage({
+              id: event.id,
+              result: new EIP1193Error(
+                EIP1193_ERROR_CODES.userRejectedRequest
```
greg-nagy, line 72: In this case the easiest thing would be just simply use `EIP1193_ERROR_CODES.userRejectedRequest` and not pack it into an error

The error class was necessary because linter inisited on having new Error instance being thrown. Here we use it as a payload so we can skip the whole create class convert to json part

not blocking though

Shadowfiend: Ah great point. Ok, will do some adjustments in another PR.

henryboldi (approved):
Works for both rejecting requests from dApps and our send page 🧙. The code looks good, but someone more familiar with this part of the codebase might have comments.

rejected

PR #725, opened 2021-12-17
v0.9.1: Polish + gold

Shadowfiend:
Includes a variety of UI fixes, support for rejecting dApp signatures,
and balance updates when transactions are submitted to catch any changes
from those transactions.


henryboldi (approved):
🚀

Issue #662, opened 2021-12-15
Switching accounts in Tally doesn't get passed to Dapp


VladUXUI:
Switching accounts doesn't seem to be registered by dapps.
Icould only test this with 1 import and 1 read me.

VladUXUI:
Tested with two imported accounts, the switch still can't be made.
I switch accounts in Tally, but DApp is still seeing my other account.

Shadowfiend:
Are we sure this behavior would be correct in the first place? Why should the dApp find out that I decided to look at my other account? What happens if I prefer to use account A in Alchemix and account B in Uniswap?

I realize that there's a piece of this that is “what other wallets do”, and that unless we support this behavior to some extent most dApps won't provide access to multiple addresses, but I'm wondering if the fundamental expectation is correct, and if we want to consider a better way, esp if one that places the decision-making in the user's hands in terms of switching accounts for a single dApp vs globally.

VladUXUI:
First to answer the questions:
>Are we sure this behavior would be correct in the first place? 

We are not, it definitely deserves some exploration. 

> Why should the dApp find out that I decided to look at my other account?

It shouldn't, you should have control and let that dapp know that you want to use another account.

Ideally the automatic switch will happen only if both of your accounts are already connected. And you should have the ability to switch between account on a specific dapp. (This means introducing a switch account functionality)

Users need a close correlation between what is connected to a dapp and what they see in their wallet. And that is because most defi activities they do are based on what they have in their wallet.

Quick flow to help illustrate:
1 -  I first connect Account A to Uniswap (Account A shown in wallet as well)
2 - I connect Account B to Alchemix (Account B shown in wallet as well)
3 - I go back to Uniswap (Account B shown in wallet, but Uniswap is showing me account A)

This adds a bit of a stress factor and some cognitive load because you always feel the need to double check that the account you are seeing in your wallet is the account you are connected with, and thus you know what you can do or can't in that DApp. Otherwise you will try to do transactions that are going to fail because of your understanding of current possibilities within that account. Prevention is always better than errors


> What happens if I prefer to use account A in Alchemix and account B in Uniswap?

You would need to switch accounts in wallet. Users focus on one task at a time, so if they are on Uniswap they focus on that, and the wallet should reflect what they are currently interacting with if they open the wallet <this is an important !IF.

PR #728, opened 2021-12-19
Quicksand Accounting: Improve handling of dApp connections with multiple accounts

### LINKED ISSUE #705: Dapp connections - Forcing connection to single read-only account

Shadowfiend:
The approach here roughly follows what I outlined in [this comment](https://github.com/tallycash/extension/issues/705#issuecomment-996413752).

In particular:
- The provider bridge only reports authorized addresses to dApps, not all
  wallet addresses. This bug could leak information.
- The provider bridge instantly rejects any signature requests for a
  transaction whose `from` address is not authorized. This occurs before
  the transaction transitions into the internal provider.
- The dApp connection icon in the header only appears if the current tab's
  origin is authorized to access the currently-selected account in the wallet.
- The sign transaction page header shows the account associated with the
  `from` address in the transaction being signed, rather than the currently-
  selected wallet account. For internal dApps and interactions like ETH send,
  these will be equivalent, but for dApp interactions they may not be, since
  the user may be connected with an account other than the one they last
  viewed in the wallet.

One additional tidbit here that followed naturally from the above: the user
may connect to a dApp with a read-only account, by design—this enables a
cross-dApp “explore” mode for users to explore dApps with cold accounts or
accounts they may be interested in learning more about. However, these accounts
naturally cannot sign. Previously, attempting to sign a transaction from such
an account would simply hang. Now, a message appears instead of the Sign button
letting the user know read-only accounts cannot sign. We should polish that up,
but it's better than the hang :)

--------

Fixes #705.

See #662, though this doesn't fix it—but I would argue this behavior is
overall preferable in some ways...

### DIFF
--- a/background/redux-slices/selectors/dappPermissionSelectors.ts
+++ b/background/redux-slices/selectors/dappPermissionSelectors.ts
@@ -1,6 +1,7 @@
 import { createSelector } from "@reduxjs/toolkit"
 import { RootState } from ".."
 import { DAppPermissionState } from "../dapp-permission"
+import { selectCurrentAccount } from "./uiSelectors"
 
 export const getProviderBridgeState = (state: RootState): DAppPermissionState =>
   state.dappPermission
@@ -26,5 +27,12 @@ export const selectCurrentPendingPermission = createSelector(
 
 export const selectAllowedPages = createSelector(
   getProviderBridgeState,
-  (slice: DAppPermissionState) => slice.allowedPages
+  selectCurrentAccount,
+  (slice, currentAccount) =>
+    Object.fromEntries(
+      Object.entries(slice.allowedPages).filter(
+        ([_, { accountAddress }]) =>
+          accountAddress.toLowerCase() === currentAccount.address.toLowerCase()
+      )
+    )
 )
--- a/background/redux-slices/selectors/dappPermissionSelectors.ts
+++ b/background/redux-slices/selectors/dappPermissionSelectors.ts
@@ -26,12 +26,14 @@ export const selectCurrentPendingPermission = createSelector(
 )
 
 export const selectAllowedPages = createSelector(
-  getProviderBridgeState,
+  (state: RootState) => getProviderBridgeState(state).allowedPages,
   selectCurrentAccount,
-  (slice, currentAccount) =>
+  (allowedPages, currentAccount) =>
+    // Decompose the origin -> permission mapping and leave only the origin ->
+    // permissions that reference the current account address.
     Object.fromEntries(
-      Object.entries(slice.allowedPages).filter(
-        ([_, { accountAddress }]) =>
+      Object.entries(allowedPages).filter(
+        ([, { accountAddress }]) =>
           accountAddress.toLowerCase() === currentAccount.address.toLowerCase()
       )
     )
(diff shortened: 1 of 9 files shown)


background/redux-slices/selectors/dappPermissionSelectors.ts
```diff
@@ -26,5 +27,12 @@ export const selectCurrentPendingPermission = createSelector(
 
 export const selectAllowedPages = createSelector(
   getProviderBridgeState,
-  (slice: DAppPermissionState) => slice.allowedPages
+  selectCurrentAccount,
+  (slice, currentAccount) =>
+    Object.fromEntries(
+      Object.entries(slice.allowedPages).filter(
+        ([_, { accountAddress }]) =>
```
greg-nagy, line 34: wow, this is really nice

greg-nagy: Object.values() would be more apropriate here

background/services/provider-bridge/index.ts
```diff
@@ -218,39 +224,46 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
   ): Promise<unknown> {
     try {
       switch (method) {
         case "eth_requestAccounts":
-          return await this.internalEthereumProviderService.routeSafeRPCRequest(
-            "eth_accounts",
-            params
-          )
+        case "eth_accounts":
+          return [enablingPermission.accountAddress]
```
greg-nagy, line 242: I believe this address will always belong to the first authorized account, esentially locking the dApp into the first account which is what the bug is about originally.

That is because the schema is defined like this currently
```
    this.version(1).stores({
      migrations: "++id,appliedAt",
      dAppPermissions: "&origin,faviconUrl,title,state,accountAddress",
    })
```

The `&` means the [index is unique](https://developer.mozilla.org/en-US/docs/Web/API/IDBIndex/unique) which results in the problem why I had to use that unholly migration in my PR. 

> unique, which if set to true means that the index will not be able to accept duplicate entries.

background/redux-slices/selectors/dappPermissionSelectors.ts
Shadowfiend: Unfortunately not if we want to keep the key association with the origin. The `entries`->`fromEntries` round-trip lets us keep the origin-to-permission request structure, while filtering for only the permissions that reference the current account address.

In a hypothetical future where we have `accountAddresses`, we can change this to be a `.includes` and maintain the same base behavior.

I can get rid of this linter warning though.

background/services/provider-bridge/index.ts
greg-nagy: And sorry for this bother, this is my fault because I forgot to update the index when added the record to store the address

greg-nagy (changes requested):
To my current understanding this PR does not solve the main underlying problem which is:
The dApp connection gets stuck to the first account address that was set when permission was granted. 

I would move this PR back to `Draft` and continue with #731 

I did not go through the transaction signing parts in depth but given they use the address stored in the permission they must be suffering from the same problem. 

There are merits in this PR though, around the transaction signing parts, which will be salvaged to the fullest 🏴‍☠️

CleanShot 2021-12-21 at 22 39 57

Shadowfiend:
This PR is missing a fix for a specific behavior, whereby the initial address a dApp receives permission to access is the only one it can be connected to in the future. That's addressed in #731, so closing this in favor of that PR.

Shadowfiend:
@greg-nagy please delete this branch once you've cherry-picked anything you might need!

PR #729, opened 2021-12-20
🐶 In top bar, replace dog name w/ shortened address

### LINKED ISSUE #714: Should first/last chars of address replace dog name in nav?
Screen Shot 2021-12-16 at 7 40 03 PM
Screen Shot 2021-12-16 at 7 39 55 PM

henryboldi:
Two **tiny** tweaks derived from real problems.

Currently there's a mental disconnect in the top nav wallet info provided, in
the dApp UI (shorted address) and our wallet (dog name), leaving users to
wonder. Basically, it's inconvenient to coordinate what's seen to make
decisions. In ~~buzzword~~ cognitive psychology terms, the cognitive load is too
high!
Screen Shot 407

So, for #714, we've brought back the shortened address in the top menu.

topmenu

But if you've got an ENS name, let's keep that because:
1) They're lovely.
2) They're what's expected. That's what dApps are displaying.

ENS

As for the code, this is resolved in literally one line change. Why? It's easy!
It's simple! But also, for real, once our account switcher names can be edited,
perhaps we'd like to bring back defaulting to the names. To be determined. It's
easy to revert.

Closes #714

<br/>

The other one-liner just fixes the send ETH address input width. The smaller
width originally accommodated the position of the paste button. The paste
button is no longer included. So gimme back the input space. Granted, a working
paste button would be much better. I believe we'd need to ask for more
permissions, though. Specifically clipboardRead. Is it worth getting access to
the whole clipboard for a paste button?

### DIFF
--- a/ui/components/Shared/SharedAssetInput.tsx
+++ b/ui/components/Shared/SharedAssetInput.tsx
@@ -262,7 +262,7 @@ export default function SharedAssetInput(
             box-sizing: border-box;
           }
           .token_input {
-            width: 204px;
+            width: 100%;
             height: 34px;
             font-size: 28px;
             font-weight: 500;
--- a/ui/components/Shared/SharedCurrentAccountInformation.tsx
+++ b/ui/components/Shared/SharedCurrentAccountInformation.tsx
@@ -14,7 +14,7 @@ export default function SharedCurrentAccountInformation(): ReactElement {
 
   return (
     <>
-      {name ?? shortenedAddress}
+      {name?.includes(".") ? name : shortenedAddress}
       <div className="avatar" />
       <style jsx>
         {`


wojciech-turek (approved):
Looks good and feels good.

VladUXUI:
Agreed!
1 tiny issue that derives from this is the Copy to clipboard when you hover over address/ens.
With an address we can show just the "copy to clipboard", with an ENS, we need to show the address + copy to clipoard.

 >But also, for real, once our account switcher names can be edited,perhaps we'd like to bring back defaulting to the names.
 
 Maybe this could be a global setting. Let users choose if they want to display the name or the address.

Good to keep paste button out, apparently it causes some confusion anyway.

Issue #366, opened 2021-10-28
dApp Connections v1


Shadowfiend:
This is a tracking issue for enabling dApp connections in the extension. 
Its implementation should be largely based on the [dApp Connections RFC](https://coda.io/d/Minimum-Viable-Launch_dznBnP8Wtkc/RFC-dApp-Connections_suxkn#_lueqf).

# v1.0

## 01.17 - 01.23 > 01.24 - 01.30
- [x] #842
- [x] #833 
- [x] #849

## Backlog
- [x] #486
- [x] Implement `wallet_watchAsset` RPC method (saddle integration update is required)
- [x] #504
- [x] `window.ethereum` protection: not writable, not deletable, object methods also protected 
- ~~Tally provider in ethers.js [based on the MM provider](https://github.com/ethers-io/ethers.js/blob/master/packages/experimental/src.ts/metamask-provider.ts)~~
- [x] web3modal integration PR
- [ ] Rate limit in `ProviderBridgeService` for better security
- [ ] `TallyWindowProvider` refactor: single event listener refactor (arrow function)
  - [x] arrow: https://github.com/tallycash/tally-extension/pull/440#discussion_r753485052
  - [x] single listener: https://github.com/tallycash/tally-extension/pull/440#discussion_r753509947
  - [ ] Map typing (or simple object)
- [x] `TallyWindowProvider` refactor:  custom handlers by payload 
  - https://github.com/tallycash/tally-extension/pull/440#discussion_r753504700
- [x] #466
- [ ]  threshold integration PR https://github.com/threshold-network/token-dashboard
- [ ] Handle not standardised RPC methods in `InternalEthereumProviderService` for compatibility
- [x] Move dApp connection responsibility from `CorePage` > `TopMenuConnectedDAppInfo`
  - https://github.com/tallycash/extension/pull/570#discussion_r767186696
  - https://github.com/tallycash/extension/pull/570#discussion_r767193266
- [ ] `PermissionRequest` type revise 
  - `faviconUrl` -> `faviconURL`
  - separate `PermissionRequest` from `PermissionResponse` 
     this implicitly implies that the current address is populated when the UI is shown and not in `main.ts`
- [ ] `provider-bridge-shared` typing simplification https://github.com/tallycash/extension/pull/687#discussion_r770738041
- [ ] combine the preference setting asyncThunks in ui slice into  `setPreference` async thunk that handles all type of preference change. Should work for single change and muiltiple
[...]

PR #731, opened 2021-12-20
🧘 Anicca: Accounts _will_ change

### LINKED ISSUE #662: Switching accounts in Tally doesn't get passed to Dapp

### LINKED ISSUE #705: Dapp connections - Forcing connection to single read-only account

greg-nagy:
https://en.wikipedia.org/wiki/Impermanence

> Impermanence, called anicca (Pāli) or anitya (Sanskrit), appears extensively in the Pali Canon[1] as one of the essential doctrines of Buddhism.[1][5][6] The doctrine asserts that all of conditioned existence, without exception, is "transient, evanescent, inconstant".[1] All temporal things, whether material or mental, are compounded objects in a continuous change of condition, subject to decline and destruction.[1][2] All physical and mental events are not metaphysically real. They are not constant or permanent; they come into being and dissolve.[7]

Closes #705, closes #662
Builds on #728 for transactions

# Purpose
Propagate the account changes to the dApps.

# Main decisions
- Moved the currentAccount into the background and made the background to be the source of truth. 
  - This information is essential in telling the dApp which account it has to use. The other solution would have been to make the background query the redux cache store which seemed like a horrible idea. 
- Fixed the primary key issue in the dApp service db with dexie migrations
  - It locked the dApp to the active address when it was given permission and could not be changed because of ther unique primary key
- Started to utilize dexie functionality for database migration and population
- Went with a data structure that has dApp origin + currentAddress as a composite key for permissions. 
  - This structure is easier to handle in most places imo. This can be changed in a way that the addresses is an array in the db. 
  - But the currentAddress will be necessary to inform the dApps
- Moved the preferences service away from appending behaviour to editing the current record. 
  - It did not seemed like a good idea to create a new version every time a user changes an account or switches the default wallet switch

# Dev testing 
(indexeddb migration)

- checkout current main and start
- remove extension and `Load unpacked`
- have multiple accounts in the wallet
- go to multiple dapps and grant permission for them 
- verify the permissions in background devtools
  - Application > IndexedDB > `tally/provider-bridge-service` > dAppPermissions
  - IndexedDB view does **NOT** refresh on change. You need to right click >`Refresh IndexedDB` if you expect a change.  
- checkout `anicca-accounts-will-change` and start
- reload extension
- verify the indexeddb has been migrated and the data has been kept
  -  Application > IndexedDB > `tally/provider-bridge-service`: refresh or reload the db 
- go to the dApps and check them if they still are connected
  - make sure that you have the same account selected that you granted permission for in previous step

# Demo 
CleanShot 2021-12-21 at 23 05 34


background/main.ts
```diff
@@ -581,6 +582,36 @@ export default class Main extends BaseService<never> {
...
+        } else {
+          // Update currentAddress in db if it's not set but it is in the store
+          // should run only one time
+          const { address } =
+            this.store.getState().ui.currentAccount.addressNetwork
+
+          if (address) {
+            await this.preferenceService.setCurrentAddress(address)
+          }
+        }
```
Shadowfiend, line 603: Do we need to do this? Given that it's dead code after the first run, it feels like we should do it once as part of main's startup, rather than once as part of an emitter pass.

background/redux-slices/ui.ts
```diff
@@ -123,6 +124,16 @@ export const setNewDefaultWalletValue = createBackgroundAsyncThunk(
   }
 )
 
+// TBD @Antonio: It would be good to have a consistent naming strategy
```
Shadowfiend, line 127: I the above thunk needs to be replaced with a single `setPreference`-type thunk that does all of this. We should emit `newPreferenceValue` or something like that, with preference name + preference value. Then that can be hooked up in one go to the preferences service.

The problem is… I don't know that the current account is a preference! Preferences persist between sessions and are changed rarely. The current account is meant to be easy to change, and to potentially change frequently. These are fundamentally different things IMO.

I think what we need to do here is hook the current address up to the bridge service (via `Main`). We can set the bridge service's view from redux at startup.

background/services/internal-ethereum-provider/index.ts
```diff
@@ -126,10 +130,11 @@ export default class InternalEthereumProviderService extends BaseService<Events>
...
       case "web3_sha3":
         return this.chainService.send(method, params)
-      case "eth_accounts": // This is a special method, because Alchemy provider DO support it, but always return null (because they do not store keys.)
-        return this.chainService
-          .getAccountsToTrack()
-          .then(([account]) => [account.address])
+      case "eth_accounts": {
+        // This is a special method, because Alchemy provider DO support it, but always return null (because they do not store keys.)
+        const currentAccount = await this.preferenceService.getCurrentAddress()
+        return [currentAccount]
```
Shadowfiend, line 136: I'm guessing this is part of the merge with my work you were talking about @greg-nagy , but I consider this the wrong place to do this. The internal Ethereum provider should provide full access, as it's not guarded by permissions. If other dApps have issues with many eth_accounts, bless them, we don't have to make our own dApps equally broken.

background/services/preferences/db.ts
```diff
@@ -1,43 +1,73 @@
...
+    // https://dexie.org/docs/Tutorial/Design#database-versioning
+    // I fully expect that I might need to revert all of this, but as per my current knowledge this seems to be a good idea
+    this.version(3).stores({
+      migrations: null, // If we use dexie built in migrations then we don't need to keep track of them manually
+      preferences: "++id", // removed all the unused indexes
+    })
+
+    // This is the old version for popuplate
+    // https://dexie.org/docs/Dexie/Dexie.on.populate-(old-version)
+    // The this does not behave according the new docs, but works
```
Shadowfiend, line 66: How does it differ from the new docs and how does it work compared to that expectation? Worried it seems to work in a new install, which is expected, but won't work in an existing install (but haven't had a chance to test yet).

background/services/preferences/index.ts
```diff
@@ -8,6 +8,7 @@ import BaseService from "../base"
 interface Events extends ServiceLifecycleEvents {
   preferencesChanges: Preferences
   initializeDefaultWallet: boolean
+  intilalizeCurrentAddress: string
```
Shadowfiend, line 11:
```suggestion
  initializeCurrentAddress: string
```

background/services/preferences/index.ts
```diff
@@ -54,7 +59,15 @@ export default class PreferenceService extends BaseService<Events> {
     return (await this.db.getLatestPreferences())?.defaultWallet
   }
 
-  async setDefaultWalletValue(newDefaultWalletValue: boolean): Promise<number> {
+  async setDefaultWalletValue(newDefaultWalletValue: boolean): Promise<void> {
```
Shadowfiend, line 62: Man I just spent 15 minutes trying to remember what `defaultWallet` was… We should really have named this flag `isDefaultWallet`, `setIsDefaultWallet`, etc.

background/services/provider-bridge/db.ts
```diff
@@ -19,15 +14,54 @@ function keyBy(
...
+      .upgrade((tx) => {
+        return tx
+          .table(tempTable)
+          .toArray()
+          .then((permissions) => tx.table(mainTable).bulkAdd(permissions))
+      })
+
+    this.version(5).stores({
+      [tempTable]: null,
+    })
```
Shadowfiend, line 64: I'm looking for a compelling reason why we would track one object per origin-address pair instead of a single origin object with an array of addresses and… The commit message for this change doesn't give me insight into *why* we want a compound index, and this migration makes me want to avoid the change at all costs :grimacing:

Why not migrate to make `accountAddresses` an array, and migrate the indexes to drop everything except origin? We can use a [multi-entry index](https://dexie.org/docs/MultiEntry-Index) for the addresses array if we're thinking we'll want to look permissions up by address, but fitting the record data to our expected shape will go a long way towards reducing plumbing work for future changes.

Lastly, we should either use `account` or `address` in the terminology here, but _not both_! Accounts are one concept, addresses another. I believe to be in line with `eth_accounts` and co, we should use `accounts`. If we end up supporting dApps on some future chain for which accounts have multiple addresses, and dApps should only access some addresses, I suspect we'll need to do some deeper thinking anyway.

provider-bridge-shared/runtime-typechecks.ts
```diff
@@ -78,5 +80,21 @@ export function isAllowedQueryParamPage(
 export function isTallyInternalCommunication(
   arg: unknown
 ): arg is TallyInternalCommunication {
+  return isObject(arg) && arg.id === "tallyHo"
```
Shadowfiend, line 83: Outside of this PR, now that we're out of crunch time, I would like us to slow down and completely reconsider this string-magic based approach. It feels very brittle, it's not materially more secure, and I'm still not entirely sure why we're doing it vs using more structured operators.

Shadowfiend (commented):
Left a bunch of comments, pending any changes related to merging the concepts in #728.

window-provider/index.ts
```diff
@@ -210,4 +212,11 @@ export default class TallyWindowProvider extends EventEmitter {
       this.transport.addEventListener(this.bridgeListeners.get(sendData.id))
     })
   }
+
+  handleAddressChange(address: string) {
+    if (this.selectedAddress !== address) {
+      this.selectedAddress = address
+      this.emit("accountsChanged", [this.selectedAddress])
```
Shadowfiend, line 219: Btw, given what you said about libraries using the first item here exclusively, I think what we should do is always return all available accounts, but make the first one always be the currently-selected one. That is, if the dApp is allowed to use accounts A, B, and C, and the user has B selected, `eth_accounts` should return `[B, A, C]`. Then if the user switches to C, `eth_accounts `should return `[C, A, B]` and `accountsChanged` should be emitted with value `[C, A, B]`. This is backwards-compatible with dApps that assume there is only one entry and pull that one entry, and forwards-compatible with the correct behavior (with the caveat that dApps that support multiple return values should be able to handle order changes).

henryboldi (dismissed):
Tested — works as expected! 👏

The only hiccup is with disconnecting. It doesn't display the disconnection in the dApp until refreshing. And then it appears to disconnect all accounts, not just the one you've clicked disconnect from. Overall, it's good though!

Leaving it to @mhluongo to review the Dexie stuff before merging

greg-nagy:
> The only hiccup is with disconnecting. It doesn't display the disconnection in the dApp until refreshing. 

I intentionally left this fix off this PR because this is already a big change and also serves as POC for how disconnect needs to communicate with dApps. Plan to have another PR with disconnect fix building on top of this today. 

> And then it appears to disconnect all accounts, not just the one you've clicked disconnect from. 

Thanks for flagging this 🧡  I wasn't aware that this was happening. 
When I implement disconnect properly it will fix this issue.

background/main.ts
```diff
@@ -581,6 +582,36 @@ export default class Main extends BaseService<never> {
       }
     )
 
+    this.preferenceService.emitter.on(
+      "intilalizeCurrentAddress",
```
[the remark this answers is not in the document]
greg-nagy: To me selecting the current account was similar than changeing hidedust or default wallet. 
I could envision a situation in the future where we will have a dropdown in the settings for selecting the default account or even default account per dApp. To me this problem is really similar to the permission settings (camera, microphone etc) in browsers per tab.

This could have also gone into the db of the provider bridge but that felt strange to have a table for a single value and the context did not really fit for me.

Or could have grabbed the currentAddress from redux on start but I think redux should depend on the background not the other way around.

background/main.ts
greg-nagy: It's important that we need this only if there is no address stored in the background. We could do this on main startup too, but then this logical check has to happen somewhere down the path because on startup we don't have any information whether the db has address or not. 

But agree that this is dead code after the first run. I would delete it in an upcoming release when we can be sure that it has run for everybody at least once.

background/redux-slices/ui.ts
greg-nagy: > I the above thunk needs to be replaced with a single setPreference-type thunk that does all of this. We should emit newPreferenceValue or something like that, with preference name + preference value. Then that can be hooked up in one go to the preferences service.

Thanks, that's exactly the idea I was looking for but could not find! Added this to https://github.com/tallycash/extension/issues/366

> The problem is… I don't know that the current account is a preference! Preferences persist between sessions and are changed rarely. The current account is meant to be easy to change, and to potentially change frequently. These are fundamentally different things IMO.

I think the concern you raise here is valid. It's not in the settings page and probably will be changed frequently.
But we will need a conclusion on this sooner than later, because more of these type will be coming. eg Network selector.

As I see we 2 paths to choose from:
1. Come up with a different name for this category and define it's behaviour and handling
2. Remove the destinction and treat both of them as preference.

I would go with 2. because to me the definition of a preference is: It's something that the user can change, and is persisted between sessions. I don't think that the frequency or the difficulty of the change should be involved in the definition.

> I think what we need to do here is hook the current address up to the bridge service (via Main). We can set the bridge service's view from redux at startup.

Before going down this path I think we should have a solution to store migration, because otherwise we will be cornered into the current structure of the store.

My other concern with this is complexity / length of steps the data has to travel / testability.

background/services/preferences/db.ts
greg-nagy: > Worried it seems to work in a new install, which is expected, but won't work in an existing install 

`populate` event is only fired on an empty database so just in case of a new install. 

For existing databases [I have updated the database in the realeased version through the `migrate` method.](https://github.com/tallycash/extension/blob/945b88396f1dd3fb9ea0e2bab54e1ba9c7f04b1c/background/services/preferences/db.ts#L76). Then recreated the same functionality in [version(2)](https://github.com/tallycash/extension/pull/731/files#diff-b16ed89668139a2b8d496b301498019d0c5e54a27fccbff1e5583d25a242a299R30) to showcase the difference and have that change reflected in the code. This can be deleted later.

I have tested with 
- remove extension
- checkout v.0.9.0, start
- check db
- change to this branch + restart
- reload extension
- check db 

It worked well.


> How does it differ from the new docs and how does it work compared to that expectation? 

https://dexie.org/docs/Dexie/Dexie.on.populate

> API changed since 0.9.8 but is still backward compatible. To view the old documentation, see Dexie.on.populate (old version). The difference is that there is no more need to provide a Transaction instance as argument to the callback. But for backward compatibility we still do.

In the new version it would look like this: 

```
    this.on("populate", function () {
      this.preferences.add(DEFAULT_PREFERENCES)
    })
```

Old version:

background/services/preferences/index.ts
greg-nagy: added here https://github.com/tallycash/extension/issues/366

background/services/internal-ethereum-provider/index.ts
greg-nagy: added https://github.com/tallycash/extension/issues/366

background/services/provider-bridge/db.ts
greg-nagy: > why we want a compound index

Mainly for performance and code ergonomics. 
- This data structure can be used both in redux and background. 
- The most often asked question is: "Can I as a dApp do this with X address?".  With this structure it's a matter of checking a key. 
- The data in this structure can be represented in the array format easily.
- IndexedDB is doing the heavy lifting for us this way

Any time I tried designing your solution it felt to be a lot more complex and the only pro I saw was not needing this migration. 
- The data or usage is different in the ui and in the background
  - Has to transform the data between transfering from background to redux
  - Or alternatively `.includes()` has to be used 
- The address array has to be manipulated on every permission check to put the current address to index 0
- Some other extra complexities around data handling, but those are not relevant enough in this question

> Lastly, we should either use account or address in the terminology here, 

Thanks, I needed this clarification! Recorded in #366

provider-bridge-shared/runtime-typechecks.ts
greg-nagy: I would like to spend some time on this redesign. This was the best I could come up with at the moment tbh. 
I use this to signal the `window-provider` that this is a communication coming from the background not the dApp.

window-provider/index.ts
greg-nagy: This can be implemented, but since last time I have found another limitation. 

- If I have `A` and `B` accounts in my wallet
  - `B` is connected to the site, `A` is not
  -  I choose `A` in my wallet as a current account 
  - then the dApp will show my balance and use my `B` account
- If my current account is `C` and I have `[C,B,A]` connected
  - I disconnect from `C`
  - the dApp will jump to `B`

Having said all that I recorded this on #366

greg-nagy:
There was a lot of discussion on this PR about data structure for the permissions and what should we return to the `eth_accounts` and `eth_requestAccounts` queries. 

This current solution is not what Antonio wants on the longterm for sure. I have put all the arguments for my approach in the conversation. I won't argue for my solution anymore after this, because all the reasons are there and it's entirely possible that I don't see something important here. 

What's important for me is how to proceed:
- Merge this PR then move forward with this as a base
- Refactor this PR to be aligned with Antonio's approach

I am ok with either one. I can make Antonio's approach work and can fully accept it.

greg-nagy:
I am in the middle of rebuilding #728 on top of this branch and holy moly. Other than the philosophical disagreement on data structures the teamwork was spot on! These 2 PRs almost fit like a glove. 

Will update this comment when finished. Surprises are bound to come when you are the most hopeful. :)

greg-nagy:
> And then it appears to disconnect all accounts, not just the one you've clicked disconnect from.

@henryboldi this should fix it #744

mhluongo (dismissed):
I want to have a bigger discussion about data flow and app-synchronous state... and I'm also concerned past me had a very clever idea with that append-only preference table that's been lost to time. The fact that we're storing the "current address" (whatever that means) in preferences at all is a little odd.

All that said — this is good work! I've tested from a few wallet states to confirm preferences don't get overridden. Merge o'clock

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -278,14 +278,26 @@ export const selectAccountTotalsByCategory = createSelector(
...
 )
 
+function findAccountTotal(
+  categorizedAccountTotals: CategorizedAccountTotals,
+  accountAddress: string
+): AccountTotal | undefined {
+  return Object.values(categorizedAccountTotals)
+    .flat()
+    .find(
+      ({ address }) => address.toLowerCase() === accountAddress.toLowerCase()
```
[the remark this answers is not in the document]
mhluongo: Not blocking, but worth a follow-up IMO... otherwise we'll have a repeat `truncate`-Gate :D

mhluongo (approved):
:checkered_flag:

PR #736, opened 2021-12-21
🕵️ Update unlock and set password pages

henryboldi:
This PR includes,

- [x] An updated unlock screen for both the extension and dApp related popups
     - [x] A "cancel tx" button
- [x] An updated set password page per Vlad's design
- [x] A hook for determining if the window is a dApp related popup

Unlock wallet

Unlock popup

Set password


ui/components/Keyring/KeyringUnlock.tsx
```diff
@@ -1,15 +1,24 @@
...
 import SharedInput from "../Shared/SharedInput"
 import titleStyle from "../Onboarding/titleStyle"
 
 export default function KeyringUnlock(): ReactElement {
   const [password, setPassword] = useState("")
   const [errorMessage, setErrorMessage] = useState("")
-  const history = useHistory()
+  const isPopup = useIsPopup()
+  const history: {
+    entries?: { pathname: string }[]
```
Shadowfiend, line 19: Note that we won't be able to rely on this when we bump the react-router version.

ui/components/Keyring/KeyringUnlock.tsx
```diff
@@ -30,11 +39,29 @@ export default function KeyringUnlock(): ReactElement {
...
 
+  const handleReject = async () => {
+    await dispatch(rejectTransactionSignature())
+  }
+
   return (
     <section>
-      <div className="full_logo" />
+      {isPopup && (
+        <div className="back_button_wrap">
```
Shadowfiend, line 49: Feels weird to me to call this `back_button_wrap` when back buttons are always on the left and this is fixed on the right, fwiw.

ui/components/Keyring/KeyringUnlock.tsx
```diff
@@ -49,23 +76,46 @@ export default function KeyringUnlock(): ReactElement {
...
           }
+          .subtitle {
+            width: 321px;
+            text-align: center;
+          }
+          .back_button_wrap {
+            width: 100%;
+            display: flex;
+            justify-content: flex-end;
+            opacity: 0.7;
```
Shadowfiend, line 107: Why is this semitransparent? 🤔 Are we trying to achieve a color, and if so should we use `rgba` instead? opacity feels a bit heavy-handed for some reason.

ui/components/Keyring/KeyringUnlock.tsx
```diff
@@ -30,11 +39,29 @@ export default function KeyringUnlock(): ReactElement {
...
     <section>
-      <div className="full_logo" />
+      {isPopup && (
+        <div className="back_button_wrap">
+          <SharedButton
+            type="tertiaryWhite"
+            size="small"
+            onClick={handleReject}
+          >
+            Cancel tx
```
Shadowfiend, line 55: So, I kind of intentionally avoided forking the unlock page based on popup vs not. Unlock shows for dApps that are and are internal and external, and it's not clear the UI should distinguish in terms of available actions between the two. For example, why would “Cancel tx” show up for a dApp popup, but not for an internal dApp unlock? The user can close the popup at any time with the native close button, and I believe we can implement logic in the provider bridge to detect window close and reject any still-open requests that window was trying to service for the dApp, making “close popup” equivalent to “cancel transaction”. Do we need this button explicitly in this case?

Additionally, the unlock screen will very soon show up for things that *aren't* transactions, and *are* from external dApps---for example `eth_signTypedData`/Sign in with Ethereum. Unlock is in a sense an interjection in the keyring-related flows, rather than an explicit part of any them.

Note also that the unlock itself doesn't communicate anything to the dApp, it's only the next page (signing) that does.

Anyway, rolling out that there is some complexity here—I realize at the moment we're just implementing to the designs, and I missed my opportunity to float this over there, so let's land as-is but consider the complexity of trying to provide a cancel button for every action vs allowing the popup close to cancel the action for external dApps.

ui/pages/SignTransaction.tsx
```diff
@@ -299,6 +299,13 @@ export default function SignTransaction(): ReactElement {
           }
         `}
       </style>
+      <style jsx global>
+        {`
+          body {
+            height: 100%;
+          }
```
Shadowfiend, line 306: Is this not needed elsewhere? Why do we end up needing to special-casing these popups? And does this affect when they show up as internal screens?

ui/public/popup.html
```diff
@@ -2,6 +2,10 @@
   <head>
     <meta charset="utf-8" />
     <title>Tally</title>
+    <link rel="preload" as="font" href="./fonts/QuincyCF-Medium.woff2" />
+    <link rel="preload" as="font" href="./fonts/segment-semibold.woff2" />
+    <link rel="preload" as="font" href="./fonts/segment-regular.woff2" />
+    <link rel="preload" as="image" href="./images/dark_forest@2x.png" />
```
Shadowfiend, line 8: Interesting. Do we materially gain something here even though these files are all local?

Shadowfiend (dismissed):
Left a few notes. Going to give this a test drive + land.

ui/components/Keyring/KeyringUnlock.tsx
henryboldi: 100%

ui/components/Keyring/KeyringUnlock.tsx
henryboldi: I both didn't feel great about adding a new button, but neither did I feel good about adding another option to `SharedButton` that emulated the design 😬. Ended up with this opacity lipstick over the `SharedButton`. I'm not happy about it.

ui/pages/SignTransaction.tsx
henryboldi: In `index.css` body is set to `600px`. So on a popup page it ends up scrolling. Without body generally set to `600px`, the extension window becomes very short

ui/public/popup.html
henryboldi: What particularly annoyed me was the `dark_forest` image flashing in after a second. Especially given that the dark forest image is essential to how the layout looks — two color blocks w/ the forest acting as the top border of the bottom color. This seems to resolve that flash since now the image arrives before React renders the page

ui/components/Keyring/KeyringUnlock.tsx
Shadowfiend: Haha makes a lot of sense actually. Something to worry about later for sure.

ui/pages/SignTransaction.tsx
Shadowfiend: I wonder if the solution then is to set body to 600px in `Popup.tsx` using that shiny new `useIsPopup` hook? :grin:

… Though worth noting, of course, that `Popup.tsx` implies all instances of our UI `isPopup` :grimacing: So we've got a bit of a terminology clash. Perhaps `useIsPopup` -> `useIsDappPopup` or something of the sort?

ui/public/popup.html
Shadowfiend: Fascinating. Makes perfect sense to me 🤷

ui/pages/SignTransaction.tsx
henryboldi: > I wonder if the solution then is to set body to 600px in Popup.tsx using that shiny new useIsPopup hook? 😁

I gave this a go! Granted, feels a bit wonky how I've done this,

```
      {isPopup && (
        <style jsx global>
          {`
            body {
              height: 100%;
            }
          `}
        </style>
      )}
 ```
 Whatcha think?

henryboldi: Crap, one sec 😅

Shadowfiend: Interestingly I was thinking `height: 100%` would be the default and `height: 600px` the exception, but I think this is equally reasonable.

Shadowfiend:
Tested, let's go!

Shadowfiend (approved):

PR #737, opened 2021-12-21
No interaction fee

wojciech-turek:
The way we handled network fees quickly became difficult to be used in many parts of the app.
In this PR I am moving some of the logic and calculations from the UI side to the backend, so the results
we get on different pages are more unified. The main objective was to make it more about the type of
the network fee selected, i.e. `regular`, `express` or `instant`, and it is now the prime way of checking 
how much the user prefers to pay for the transaction.

Now when we decide to change the multipliers of the fees they can be done through the constants file rather directly
in the component where it is used.

Because the value is saved in Redux state, currently **the choice of the fee type is persisted even if user changes page or the extension is hidden.**
The above is subject to change, as we may want to reset the choice when user leaves.


wojciech-turek:
@itsrachelfish removed code passed to network fees chooser from the swap component, caused errors in the terminal and if this PR lands it will be unnecessary,

ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -43,29 +47,44 @@ export default function NetworkFeesChooser({
...
   }
 
   const handleSelectGasOption = (index: number) => {
-    onSelectFeeOption({
-      confidence: Number(gasOptions[index].confidence),
-      price: gasOptions[index].price,
-      maxFeePerGas: gasOptions[index].maxFeePerGas,
-      maxPriorityFeePerGas: gasOptions[index].maxPriorityFeePerGas,
-    })
+    setActiveFeeIndex(index)
```
henryboldi, line 67: Why not use the gas choice dispatched in `saveUserGasChoice`?

henryboldi (commented):
This is a great upgrade! 🤩👏

I'll give it a test drive before approval + merge.

ui/components/NetworkFees/NetworkFeesChooser.tsx
wojciech-turek: The reason is that since we have a save button, I want to allow users to make a visual effect on click on a fee but do not automatically select it unless users "Saves" the choice.

henryboldi: Ah, makes sense!

henryboldi (approved):
I don't notice any breaking changes here. However, at some point the `pending` transaction feature appears to have been broken. After sending ETH to myself,
send
Nothing appeared. A couple minutes later the transaction did show up.

PR #738, opened 2021-12-21
Add KeeperDAO Limit Orders

0xDaedalus:
This PR adds the ability to place gasless limit orders using the KeeperDAO Hiding Game.

Swap Page with "Swap Assets" Active:
image

Swap page with "Limit Order" Active:
image

Transaction Settings on Swap Page:
image

Limit Order Review Page:
image

Transaction Settings On Review Page:
image

There are some outstanding issues that I'm not sure of the best way to move forward with - they are outlined in in-line comments below.


ui/pages/Limit.tsx
```diff
@@ -0,0 +1,249 @@
...
+      >
+        <LimitQuote />
+      </SharedSlideUpMenu>
+      <div className="form">
+        <div className="form_input">
+          <SharedAssetInput
+            assets={limit.tokens}
+            defaultToken={limit.sellToken}
+            controlledToken={limit.sellToken}
+            maxBalance={Infinity}
```
BazFi, line 119: I am unable to get account balances to load while developing - I think this is maybe because I don't have valid alchemy / blocknative keys?  They were loading for me prior to the 0.9.0 release so I assume its some sort of misconfiguration on my part.

ui/pages/Limit.tsx
```diff
@@ -0,0 +1,249 @@
...
+            onAssetSelect={fromAssetSelected}
+            onAmountChange={fromAmountChanged("limit")}
+            amount={limit.sellAmount}
+            footer={
+              limit.sellToken ? `1 ${limit.sellToken?.symbol} = $XXXX` : ""
```
BazFi, line 124: Possibly tied to not being able to load account balances -  I don't see a `prices` array tied to any asset in the `assets` reducer state.

ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -24,17 +25,22 @@ function SelectTokenMenuContent(
...
         </div>
       </div>
       <div className="divider" />
       <ul>
-        {assets.map((asset) => {
+        {assets.map((asset, i) => {
           return (
             <SharedAssetItem
-              key={asset.metadata?.coinGeckoID || asset.symbol}
+              key={i}
```
BazFi, line 43: This was breaking in dev - I think there are assets that share the same symbol and do not have a coingecko id in our list of assets.  changed to `i` for now to make it work.

ui/pages/Limit.tsx
mhluongo: We should include dev keys in the repo for anyone to use.. but yeah, it'd be good to have local keys as the public keys can get DoS'd, etc.

Shadowfiend: Repo does have an alchemy dev key, so it should work... Blocknative needs to be filled in by the dev (I need to flag this in the readme) as the non-paid plan is extremely frugal in terms of rate limits. It's only used for gas prices atm though.

BazFi: Awesome - after rebasing latest commits in looks like things are working - I must have grabbed a bad commit when i checkout out my branch.

ui/pages/Limit.tsx
BazFi: Resolved 👍

ui/pages/Limit.tsx
BazFi: Resolved 👍

mhluongo:
Very excited about this! @Shadowfiend and @itsrachelfish and going to take point here to get this merged and planned for a release, but taking a light look in the meantime

background/lib/utils/index.ts
```diff
@@ -106,3 +122,96 @@ export function getEthereumNetwork(): EVMNetwork {
 export function truncateAddress(address: string): string {
   return `${address.slice(0, 6)}...${address.slice(-5)}`
 }
+
+const augmentLimitOrder = async (
```
mhluongo, line 126: We should put KeeperDAO related utils in one place. I'd recommend `background/lib/keeper-dao.ts`

background/lib/utils/index.ts
```diff
@@ -1,8 +1,24 @@
...
+  ROPSTEN,
+  RINKEBY,
+  GOERLI,
+  KOVAN,
+  ZEROEX_DOMAIN_DEFAULTS,
+} from "../../constants"
+
+import * as ethjs from "ethereumjs-util"
+import { ethers } from "ethers"
+import axios from "axios"
```
mhluongo, line 21: `axios` is a transitive dependency, we should avoid it or discuss bringing it into `package.json`

background/lib/utils/index.ts
```diff
@@ -1,8 +1,24 @@
...
+import {
+  ETHEREUM,
+  ROPSTEN,
+  RINKEBY,
+  GOERLI,
+  KOVAN,
+  ZEROEX_DOMAIN_DEFAULTS,
+} from "../../constants"
+
+import * as ethjs from "ethereumjs-util"
```
mhluongo, line 19: Prefer `ethers` utilities in the codebase rather than the proliferation of other libraries. Also, `ethereumjs-util` is transitive... so we shouldn't rely on it. There's not much appetite for making it a named dependency

background/lib/utils/index.ts
```diff
@@ -106,3 +122,96 @@ export function getEthereumNetwork(): EVMNetwork {
...
+  }
+
+  const signed = await keyring.signTypedData(
+    normalizedLimitOrder.maker,
+    domain,
+    types,
+    normalizedLimitOrder as unknown as Record<string, unknown>
+  )
+
+  const vals = ethjs.fromRpcSig(signed)
```
mhluongo, line 198: This should be replaced by `splitSignature` from `ethers`. Also a good spot to destructure

```
const { r, s, v } = ...
```

background/lib/utils/index.ts
```diff
@@ -106,3 +122,96 @@ export function getEthereumNetwork(): EVMNetwork {
...
+  taker: string
+  pool: string
+}
+interface KeeperDAOInfoResponse {
+  result: {
+    orderDetails: KeeperDAOOrderDetails
+  }
+}
+
+export const signKeeperDaoLimitOrder = async (
```
mhluongo, line 157: I'd like to see `signTypedData()` exposed by the `KeyringService`, then this decomposed into a "prepare order" and "execute order" steps

background/main.ts
```diff
@@ -387,6 +394,27 @@ export default class Main extends BaseService<never> {
...
+      "signAndSendLimitOrder",
+      async (transaction: BaseLimitOrder) => {
+        logger.log("Got Limit Order Request!")
+        try {
+          const signature = await this.keyringService.signLimitOrder(
+            transaction
+          )
+
+          const res = await axios.post(
+            "https://hidingbook.keeperdao.com/api/v1/orders",
```
mhluongo, line 407: This should be in a service — `main.ts` should only be coordinating between Redux and services rather than a biz logic grab-bag

background/services/keyring/index.ts
```diff
@@ -303,6 +313,29 @@ export default class KeyringService extends BaseService<Events> {
     return newKeyring.id
   }
 
+  signLimitOrder = async (
```
mhluongo, line 316: Re-jiggering this to be a generic `signTypedData` would be :ok_hand:

background/types.ts
```diff
@@ -50,3 +50,46 @@ export enum KeyringTypes {
   metamaskMnemonic = "mnemonic#metamask",
   singleSECP = "single#secp256k1",
 }
+
+export interface BaseLimitOrder {
```
mhluongo, line 54: Another candidate for a KeeperDAO-specific module

mhluongo (commented):
Left a few thoughts on the background code layout and dependencies, but didn't dig into the frontend much. Really promising.

background/lib/utils/index.ts
0xDaedalus: ✅

background/lib/utils/index.ts
0xDaedalus: I'm a big fan of the typesafety and error messages `axios`  has - but I think adding it is out of the scope of this PR.  I've removed it for now.

background/lib/utils/index.ts
0xDaedalus: Was unaware of that method - ty.

background/lib/utils/index.ts
0xDaedalus: ✅

background/services/keyring/index.ts
0xDaedalus: ✅

background/types.ts
0xDaedalus: Moved to background/lib/keeper-dao.ts

background/main.ts
0xDaedalus: moved to keeper-dao.ts

background/lib/utils/index.ts
0xDaedalus: ✅

0xDaedalus:
Temporarily sent back to "draft" as there have been quite a few architectural changes since this PR was made.  Still on the roadmap but on the backburner for now.

tallysam:
Is this still relevant?

Shadowfiend:
Closing this as it's quite outdated. If and when limit orders return to the roadmap, we can still reference the PR and restore the branch if needed.

PR #743, opened 2021-12-22
Quicksand Accounting Remastered: Improve handling of dApp connections with multiple accounts 

### LINKED ISSUE #705: Dapp connections - Forcing connection to single read-only account

greg-nagy:
This is essentially the same PR as #728 and solves the transaction parts,
but rebased on top of #731 which solves the underlying dapp selection issue.

From the original PR written by @Shadowfiend 

---

The approach here roughly follows what I outlined in [this comment](https://github.com/tallycash/extension/issues/705#issuecomment-996413752).

In particular:
- The provider bridge only reports authorized addresses to dApps, not all
  wallet addresses. This bug could leak information.
- The provider bridge instantly rejects any signature requests for a
  transaction whose `from` address is not authorized. This occurs before
  the transaction transitions into the internal provider.
- The dApp connection icon in the header only appears if the current tab's
  origin is authorized to access the currently-selected account in the wallet.
- The sign transaction page header shows the account associated with the
  `from` address in the transaction being signed, rather than the currently-
  selected wallet account. For internal dApps and interactions like ETH send,
  these will be equivalent, but for dApp interactions they may not be, since
  the user may be connected with an account other than the one they last
  viewed in the wallet.

One additional tidbit here that followed naturally from the above: the user
may connect to a dApp with a read-only account, by design—this enables a
cross-dApp “explore” mode for users to explore dApps with cold accounts or
accounts they may be interested in learning more about. However, these accounts
naturally cannot sign. Previously, attempting to sign a transaction from such
an account would simply hang. Now, a message appears instead of the Sign button
letting the user know read-only accounts cannot sign. We should polish that up,
but it's better than the hang :)

### DIFF
--- a/ui/components/Shared/SharedCurrentAccountInformation.tsx
+++ b/ui/components/Shared/SharedCurrentAccountInformation.tsx
@@ -1,17 +1,16 @@
 import React, { ReactElement } from "react"
 
-import { selectCurrentAccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
-import { useBackgroundSelector } from "../../hooks"
+type Props = {
+  shortenedAddress: string
+  name: string | undefined
+  avatarURL: string | undefined
+}
 
-export default function SharedCurrentAccountInformation(): ReactElement {
-  const currentAccountTotal = useBackgroundSelector(selectCurrentAccountTotal)
-
-  if (typeof currentAccountTotal === "undefined") {
-    return <></>
-  }
-
-  const { shortenedAddress, name, avatarURL } = currentAccountTotal
-
+export default function SharedCurrentAccountInformation({
+  shortenedAddress,
+  name,
+  avatarURL,
+}: Props): ReactElement {
   return (
     <>
       {name?.includes(".") ? name : shortenedAddress}
--- a/ui/components/SignTransaction/SignTransactionNetworkAccountInfoTopBar.tsx
+++ b/ui/components/SignTransaction/SignTransactionNetworkAccountInfoTopBar.tsx
@@ -1,7 +1,26 @@
+import { HexString } from "@tallyho/tally-background/types"
+import { getAccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
 import React, { ReactElement } from "react"
+import { useBackgroundSelector } from "../../hooks"
 import SharedCurrentAccountInformation from "../Shared/SharedCurrentAccountInformation"
 
-export default function SignTransactionNetworkAccountInfoTopBar(): ReactElement {
+type Props = {
+  transactionSenderAddress: HexString
+}
+
+export default function SignTransactionNetworkAccountInfoTopBar({
+  transactionSenderAddress,
+}: Props): ReactElement {
+  const accountTotal = useBackgroundSelector((state) =>
+    getAccountTotal(state, transactionSenderAddress)
+  )
+
+  if (typeof accountTotal === "undefined") {
+    return <></>
+  }
+
+  const { shortenedAddress, name, avatarURL } = accountTotal
+
   return (
     <div className="top_bar_wrap standard_width">
       <div className="row_part">
@@ -9,7 +28,11 @@ export default function SignTransactionNetworkAccountInfoTopBar(): ReactElement 
         <span className="network_name">Arbitrum</span>
       </div>
       <div className="row_part">
-        <SharedCurrentAccountInformation />
+        <SharedCurrentAccountInformation
+          shortenedAddress={shortenedAddress}
+          name={name}
+          avatarURL={avatarURL}
+        />
       </div>
       <style jsx>
         {`
--- a/ui/components/SignTransaction/SignTransactionNetworkAccountInfoTopBar.tsx
+++ b/ui/components/SignTransaction/SignTransactionNetworkAccountInfoTopBar.tsx
@@ -1,20 +1,14 @@
-import { HexString } from "@tallyho/tally-background/types"
-import { getAccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
+import { AccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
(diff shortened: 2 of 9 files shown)


ui/components/Shared/SharedCurrentAccountInformation.tsx
```diff
@@ -1,17 +1,16 @@
 import React, { ReactElement } from "react"
 
-import { selectCurrentAccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
-import { useBackgroundSelector } from "../../hooks"
-
-export default function SharedCurrentAccountInformation(): ReactElement {
-  const currentAccountTotal = useBackgroundSelector(selectCurrentAccountTotal)
```
henryboldi, line 7: If we're no longer getting the selected account in this component and accepting props instead, `SharedCurrentAccountInformation.tsx` becomes a misleading name, no?

ui/components/SignTransaction/SignTransactionNetworkAccountInfoTopBar.tsx
```diff
@@ -1,15 +1,32 @@
+import { AccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
 import React, { ReactElement } from "react"
```
henryboldi, line 2: Tiny nit but in almost all files `.tsx` files we import from `react` first.

ui/components/TopMenu/TopMenuProfileButton.tsx
```diff
@@ -1,13 +1,27 @@
...
 }): ReactElement {
+  const { shortenedAddress, name, avatarURL } =
+    useBackgroundSelector(selectCurrentAccountTotal) ?? {}
+
   const { onClick } = props
+
   return (
     <button type="button" onClick={onClick}>
-      <SharedCurrentAccountInformation />
+      {typeof shortenedAddress === "undefined" ? (
```
henryboldi, line 16: Just for my own understanding, in what circumstance would `shortenedAddress === "undefined"`?

ui/pages/SignTransaction.tsx
```diff
@@ -64,6 +66,12 @@ export default function SignTransaction(): ReactElement {
   )
   const transactionDetails = useBackgroundSelector(selectTransactionData)
 
+  const signerAccountTotal = useBackgroundSelector((state) =>
+    typeof transactionDetails === "undefined"
+      ? undefined
+      : getAccountTotal(state, transactionDetails.from)
```
henryboldi, line 72: Would this eventually be useful to reuse in other components? It might be worth adding as a selector in `redux-slices`

ui/pages/SignTransaction.tsx
```diff
@@ -127,7 +135,10 @@ export default function SignTransaction(): ReactElement {
     return <></>
   }
 
-  if (typeof transactionDetails === "undefined") {
+  if (
+    typeof transactionDetails === "undefined" ||
+    typeof signerAccountTotal === "undefined"
+  ) {
     // TODO Some sort of unexpected state error if we end up here... Or do we
     // go back in history? That won't work for dApp popovers though.
```
henryboldi, line 143: > That won't work for dApp popovers though.

I've added a hook called `useIsDappPopup()` that might be useful here

ui/pages/SignTransaction.tsx
```diff
@@ -229,15 +242,19 @@ export default function SignTransaction(): ReactElement {
...
-          type="primary"
-          iconSize="large"
-          size="large"
-          onClick={handleConfirm}
-          showLoadingOnClick
-        >
-          {signContent[signType].confirmButtonText}
-        </SharedButton>
+        {signerAccountTotal.accountType === AccountType.Imported ? (
+          <SharedButton
```
henryboldi, line 246: 😁

ui/components/TopMenu/TopMenuProfileButton.tsx
greg-nagy: If `categorizedAccountTotals` does not have the `currentAccount`. Looking at that selector it should never happen, so I assume it's mainly because of the find signature and fail safety.

```
function findAccountTotal(
  categorizedAccountTotals: CategorizedAccountTotals,
  accountAddress: string
): AccountTotal | undefined {
  return Object.values(categorizedAccountTotals)
    .flat()
    .find(
      ({ address }) => address.toLowerCase() === accountAddress.toLowerCase()
    )
}
```

ui/pages/SignTransaction.tsx
greg-nagy: I am not sure if it will tbh.

ui/components/Shared/SharedCurrentAccountInformation.tsx
greg-nagy: How so? What is the nameing convention you use?

henryboldi: Well, it's now not always true that this component shows the current account. I'd just change the name to `SharedAccountInformation.tsx`

ui/components/SignTransaction/SignTransactionNetworkAccountInfoTopBar.tsx
Shadowfiend: Let's not enforce this unless we can find an eslint rule that does it (may well exist). We already have some import ordering stuff enforced there, and handling stylistic nits like that in review is a drain on everyone's energy and our commit history 😁

henryboldi: Got it

ui/components/Shared/SharedCurrentAccountInformation.tsx
Shadowfiend: `SharedAccountInformation` probably, since it's independent of the current one.

PR #747, opened 2021-12-22
✨ Fancy input labels

henryboldi:
Of note, I added a `useRunOnFirstRender` hook. I've added it because the
`useEffect` approach of passing `[]` hits snag here — the functions need to be
passed, which breaks it running only on first render.

- `SharedInput` now can accept a default value
- The animated label can adapt to different backgrounds, like in a slide up
menu
- The label will turn red on an error

ENS

### DIFF
--- a/ui/components/Keyring/KeyringSetPassword.tsx
+++ b/ui/components/Keyring/KeyringSetPassword.tsx
@@ -66,7 +66,7 @@ export default function KeyringSetPassword(): ReactElement {
         <div className="input_wrap">
           <SharedInput
             type="password"
-            placeholder="Password"
+            label="Password"
             onChange={handleInputChange(setPassword)}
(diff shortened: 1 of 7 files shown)


ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -66,15 +66,15 @@ export default function KeyringSetPassword(): ReactElement {
...
+            label="Password"
             onChange={handleInputChange(setPassword)}
             errorMessage={passwordErrorMessage}
           />
         </div>
         <div className="input_wrap repeat_password_wrap">
           <SharedInput
             type="password"
-            placeholder="Repeat Password"
+            label="Repeat Password"
```
mhluongo, line 77: :raised_hands:

mhluongo (approved):
LGTM!

PR #748, opened 2021-12-22
Alchemical Misadventures: Loosen the Alchemy API validation to allow errors

### LINKED ISSUE #746: Don't fail an Alchemy balance check if one token looks off
Some balance checks are returning incredibly large string response, eg the yv1INCH vault. We should...

- [x] Not fail if only a single balance check looks off
- [x] Truncate all balance check responses to 32 bytes, discarding the rest

mhluongo:
A couple bad token contracts were preventing us from balance checking the entire token list. This should handle it.

Closes #746


henryboldi (approved):
Thumbs up

PR #751, opened 2021-12-23
Fix a bug where low balance ETH would be hidden

wojciech-turek:
We always want to show ETH balance regardless of its amount.


background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -172,7 +172,8 @@ export const selectCurrentAccountBalances = createSelector(
           typeof assetAmount.mainCurrencyAmount === "undefined"
             ? true
             : assetAmount.mainCurrencyAmount > userValueDustThreshold
-        const isPresent = assetAmount.decimalAmount > 0
+        const isPresent =
+          assetAmount.decimalAmount > 0 || assetAmount.asset.symbol === "ETH"
```
Shadowfiend, line 176: May want to mark this with a TODO that we'll want to change it to be based on the currently-selected network's base asset.

mhluongo (approved):
:foot:  :bug:

PR #752, opened 2021-12-23
Pull updated ETH balance after transaction

wojciech-turek:
This PR fixes the ETH balance update once a transaction is completed.

Users should see the new balance within one minute from the transaction confirmation.

### DIFF
--- a/background/lib/alchemy.ts
+++ b/background/lib/alchemy.ts
@@ -174,17 +174,6 @@ const isValidAlchemyTokenBalanceResponse = jtdValidatorFor(
   alchemyTokenBalanceJTD
 )
 
-export async function getMainNetworkCurrencyBalance(
-  provider: AlchemyProvider | AlchemyWebSocketProvider,
-  address: HexString
-): Promise<{ amount: bigint }> {
-  const json: HexString = await provider.send("eth_getBalance", [
-    address,
-    "latest",
-  ])
-  return { amount: BigInt(json) }
-}
-
 /**
  * Use Alchemy's getTokenBalances call to get balances for a particular address.
  *
--- a/background/lib/alchemy.ts
+++ b/background/lib/alchemy.ts
@@ -174,6 +174,17 @@ const isValidAlchemyTokenBalanceResponse = jtdValidatorFor(
   alchemyTokenBalanceJTD
 )
 
+export async function getMainNetworkCurrencyBalance(
+  provider: AlchemyProvider | AlchemyWebSocketProvider,
+  address: HexString
+): Promise<{ amount: bigint }> {
+  const json: HexString = await provider.send("eth_getBalance", [
+    address,
+    "latest",
+  ])
+  return { amount: BigInt(json) }
+}
+
 /**
  * Use Alchemy's getTokenBalances call to get balances for a particular address.
  *
--- a/background/services/indexing/index.ts
+++ b/background/services/indexing/index.ts
@@ -270,11 +270,7 @@ export default class IndexingService extends BaseService<Events> {
           address: transaction.from.toLowerCase(),
           network: getEthereumNetwork(),
         }
-        const balance = await this.chainService.getLatestBaseAccountBalance(
-          addressNetwork
-        )
-        this.emitter.emit("accountBalance", balance)
-        await this.db.addBalances([balance])
+        await this.chainService.getLatestBaseAccountBalance(addressNetwork)
       }
     })
   }
--- a/background/services/indexing/index.ts
+++ b/background/services/indexing/index.ts
@@ -265,7 +265,10 @@ export default class IndexingService extends BaseService<Events> {
       ) {
         this.scheduledTokenRefresh = true
       }
-      if ("status" in transaction && (transaction.status === 1 || 0)) {
+      if (
+        "status" in transaction &&
+        (transaction.status === 1 || transaction.status === 0)
+      ) {
         const addressNetwork = {
           address: transaction.from.toLowerCase(),
           network: getEthereumNetwork(),
--- a/background/services/indexing/index.ts
+++ b/background/services/indexing/index.ts
@@ -13,11 +13,7 @@ import {
 } from "../../assets"
 import { BTC, ETH, FIAT_CURRENCIES, USD } from "../../constants"
 import { getBalances as getAssetBalances } from "../../lib/erc20"
-import {
-  getMainNetworkCurrencyBalance,
-  getTokenBalances,
-  getTokenMetadata,
-} from "../../lib/alchemy"
+import { getTokenBalances, getTokenMetadata } from "../../lib/alchemy"
 import { getPrices, getEthereumTokenPrices } from "../../lib/prices"
 import {
   fetchAndValidateTokenList,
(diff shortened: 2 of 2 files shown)


background/lib/alchemy.ts
```diff
@@ -174,6 +174,17 @@ const isValidAlchemyTokenBalanceResponse = jtdValidatorFor(
   alchemyTokenBalanceJTD
 )
 
+export async function getMainNetworkCurrencyBalance(
```
mhluongo, line 177: This isn't Alchemy specific, no need for it to go in here.

background/lib/alchemy.ts
```diff
@@ -174,6 +174,17 @@ const isValidAlchemyTokenBalanceResponse = jtdValidatorFor(
...
+  provider: AlchemyProvider | AlchemyWebSocketProvider,
+  address: HexString
+): Promise<{ amount: bigint }> {
+  const json: HexString = await provider.send("eth_getBalance", [
+    address,
+    "latest",
+  ])
+  return { amount: BigInt(json) }
```
mhluongo, line 185: Missing error handling, etc, etc. Instead of rolling your own, use `ChainService.getLatestBaseAccountBalance(addressNetwork)`.

background/services/indexing/index.ts
```diff
@@ -13,7 +13,11 @@ import {
 } from "../../assets"
 import { BTC, ETH, FIAT_CURRENCIES, USD } from "../../constants"
 import { getBalances as getAssetBalances } from "../../lib/erc20"
-import { getTokenBalances, getTokenMetadata } from "../../lib/alchemy"
+import {
+  getMainNetworkCurrencyBalance,
```
mhluongo, line 17: This idea of a "main" network is problematic, and works against us going multi-network.

background/services/indexing/index.ts
```diff
@@ -508,6 +512,23 @@ export default class IndexingService extends BaseService<Events> {
     const activeAssetsToTrack = assetsToTrack.filter(
       (t) => t.homeNetwork.chainID === getEthereumNetwork().chainID
     )
+    const accounts = await this.chainService.getAccountsToTrack()
+    const mainCurrencyBalance = await getMainNetworkCurrencyBalance(
```
mhluongo, line 516: Instead of doing this periodically, we should only do a balance check once we've sent or received a transaction.

mhluongo (changes requested):
I think you're approaching this in a way that fights the current architecture.

background/services/indexing/index.ts
```diff
@@ -265,6 +265,17 @@ export default class IndexingService extends BaseService<Events> {
       ) {
         this.scheduledTokenRefresh = true
       }
+      if ("status" in transaction && (transaction.status === 1 || 0)) {
```
mhluongo, line 268: Pretty sure this conditional is the same as

```
if ("status" in transaction && transaction.status === 1)
```

because of operator precedence... is that what you intended?

background/services/indexing/index.ts
```diff
@@ -265,6 +265,17 @@ export default class IndexingService extends BaseService<Events> {
...
+        const addressNetwork = {
+          address: transaction.from.toLowerCase(),
```
mhluongo, line 270: `lib.utils.normalizeEVMAddress`

background/lib/alchemy.ts
wojciech-turek: Removed altogether, there is another function already existing we can use for that. Thanks for pointing out :)

background/services/indexing/index.ts
```diff
@@ -265,6 +265,17 @@ export default class IndexingService extends BaseService<Events> {
...
+          network: getEthereumNetwork(),
+        }
+        const balance = await this.chainService.getLatestBaseAccountBalance(
+          addressNetwork
+        )
+        this.emitter.emit("accountBalance", balance)
```
mhluongo, line 276: The chain service already does this, no need to emit and save from both service. I think you can replace 273 - 278 with just

```
await this.chainService.getLatestBaseAccountBalance(
    addressNetwork
)
```

background/services/indexing/index.ts
wojciech-turek: My thinking was its a main currency of a given network but I've probably phrased it poorly.

background/services/indexing/index.ts
mhluongo: Maybe? It's not actually clear to me when / whether we normalize, which is a problem.

mhluongo (changes requested):
So sleek! Couple comments to address but like the direction :)

background/services/indexing/index.ts
wojciech-turek: Moved it out to a on("transaction") handler, very good point, I was thinking of ETH as a token initially.

background/services/indexing/index.ts
wojciech-turek: Isn't that a case that when transaction fails and has status 0, the user might still incur a fee for gas spent? With that condition reduced to checking only status of `1` it won't run when status is 0 and therefore won't update the amount while it should?

background/services/indexing/index.ts
wojciech-turek: I am not sure either, I saw `.toLowerCase()` being used in another case and followed the same pattern.

background/services/indexing/index.ts
wojciech-turek: Right, I missed this. Thank you!

background/services/indexing/index.ts
mhluongo: Oh I didn't know your intent, I'm just saying that `x === 1 || 0` is that same as `(x === 1) || 0`.

If you want to check if it's either, you need `x == 1 || x === 0`.

background/services/indexing/index.ts
mhluongo: Not blocking rn

background/services/indexing/index.ts
wojciech-turek: Updated, good catch, my brain took a shortcut which wasn't a shortcut 🙈

mhluongo (approved):
:rocket:

background/services/indexing/index.ts
```diff
@@ -265,6 +265,16 @@ export default class IndexingService extends BaseService<Events> {
       ) {
         this.scheduledTokenRefresh = true
       }
+      if (
+        "status" in transaction &&
+        (transaction.status === 1 || transaction.status === 0)
```
Shadowfiend, line 270: Should we have bundled this in with the scheduled refresh were already doing for tokens? Should that periodic check also check the ETH balance?

wojciech-turek: We discussed that with @matt on fd and since it is likely that every tx will have an impact on ETH balance we should refresh it after every transaction failed or successful.

mhluongo: I still think the ERC-20 check makes sense FWIW, just thought it deserved a bit more discussion on approach... especially as we figure out how "smart" we want the wallet to be. Eg if a tx is a known ERC-20 transfer, we can stick with checking 1 contract

PR #754, opened 2021-12-24
support for unstoppable domains

mul1sh:
Will fix https://github.com/unstoppabledomains/gitcoin-bounties/issues/3


background/services/name/index.ts
```diff
@@ -1,3 +1,4 @@
+import { Resolution } from "@unstoppabledomains/resolution"
```
Shadowfiend, line 1: Let's alias this to `UnstoppableDomainsResolution` since this file is meant to hit more than one name service and this name is by default generic to the point of losing local info when reading consuming code.

background/services/name/index.ts
```diff
@@ -155,20 +156,41 @@ export default class NameService extends BaseService<Events> {
...
+    if (
+      !name.match(/.*\.eth$/) ||
+      !name.match(/.*\.crypto$/) ||
+      !name.match(/.*\.coin$/) ||
+      !name.match(/.*\.wallet$/) ||
+      !name.match(/.*\.bitcoin$/) ||
+      !name.match(/.*\.888$/) ||
+      !name.match(/.*\.nft$/) ||
+      !name.match(/.*\.dao$/) ||
+      !name.match(/.*\.zil$/)
```
Shadowfiend, line 170: Thoughts on adjusting to something like:
- a const top-level array with supported domains
- a `name.slice(name.lastIndexOf("."))` to extract the last domain component
- an array `.includes` test

Even better would be abstracting one level out and having a set of recognizer+resolver functions that are tried in turn and each support one name service.

Shadowfiend (commented):
Excited to see this! Left a couple of small thoughts on what's currently in place.

background/services/name/index.ts
mul1sh: I've gone with the array option, it was much easier for me to implement 😅

mul1sh:
@Shadowfiend I'm getting this error after I merged the changes in main to my local branch.
Screenshot 2022-01-17 at 8 05 31 PM

Any idea as to what's causing this?

mul1sh:
Will open a better final PR tomorrow morning :)

PR #756, opened 2021-12-25
🔧  MacGyver Serialise: Fix redux devtools by extending with our BigInt format

greg-nagy:
CleanShot 2021-12-25 at 11 51 38

Redux devtools hung after a while and was unusable for me. It happened because it did not know how to desierialise our bigint format, even though we had enhancers and whatnot.

The error pointed to [this line of code](https://github.com/kolodny/jsan/blob/master/lib/index.js#L33), which failed when it had to deal with our bigint format. When this was enhanced with our custom serialiser, redux devtools started working.

Tried to fix the issue but it was deep in the stack:
background > `remote-redux-devtools` > `redux-devtools-core` > `remotedev-serialize` > `jsan`. The problem seems to be that `redux-devtools-core` does not propagate our serialiser enhancement here. Can continue going down the rabbit hole to find a good fix but instead changed directions.

It's a simple patch, which fixes our problem. There is this package https://www.npmjs.com/package/patch-package which let's you do that easily. It will apply the patch on every install. (One thing to note is that I DID not install the `postinstall-postinstall` because it seemed that we don't need it and did not want to add dependency unless required)

### DIFF
--- a/patches/jsan+3.1.13.patch
+++ b/patches/jsan+3.1.13.patch
@@ -1,0 +1,18 @@
+diff --git a/node_modules/jsan/lib/index.js b/node_modules/jsan/lib/index.js
+index c32b85b..dcdde87 100644
+--- a/node_modules/jsan/lib/index.js
++++ b/node_modules/jsan/lib/index.js
+@@ -30,7 +30,12 @@ exports.stringify = function stringify(value, replacer, space, _options) {
+ 
+   var decycled = cycle.decycle(value, options, replacer);
+   if (arguments.length === 1) {
+-    return JSON.stringify(decycled);
++    return JSON.stringify(decycled, (_, value) => {
++      if (typeof value === "bigint") {
++        return { B_I_G_I_N_T: value.toString() }
++      }
++      return value
++    })
+   } else {
+     // decycle already handles when replacer is a function.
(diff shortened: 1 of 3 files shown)


patches/jsan+3.1.13.patch
```diff
@@ -0,0 +1,18 @@
+diff --git a/node_modules/jsan/lib/index.js b/node_modules/jsan/lib/index.js
+index c32b85b..dcdde87 100644
+--- a/node_modules/jsan/lib/index.js
++++ b/node_modules/jsan/lib/index.js
+@@ -30,7 +30,12 @@ exports.stringify = function stringify(value, replacer, space, _options) {
+ 
+   var decycled = cycle.decycle(value, options, replacer);
+   if (arguments.length === 1) {
+-    return JSON.stringify(decycled);
++    return JSON.stringify(decycled, (_, value) => {
```
mhluongo, line 10: Nice. It might make sense longer-term to maintain our own fork or contribute back upstream, but this is a clever fix in the interim.

mhluongo: cc @Shadowfiend to get yours eyes on this later

mhluongo (approved):
:raised_hands:

PR #759, opened 2021-12-27
✨ What in animation tarnation

henryboldi:
Have you ever wondered what a world would be like with animated page transitions in Tally? You and me are alike. It's as if a tranquil sunset washes over each of your clicks in the UI.

To make this possible we've rejiggered some things,
- `CorePage` is wrapped in `Popup.tsx` rather than again and again in pages
- `TopMenu` acts independently from core page
- Added the `react-transition-group` package (5,438,627 weekly downloads)
- Created a config file for routes in `utils/routes.tsx`
- The bulk of the animation logic is handled in `pageTransition.ts`

These animations feature a couple details,
- Top bar stays still when appropriate
- Slides in different directions based on a couple factors,
        - What the clicked tab bar icon's location is relative to the current tab
        - Back or forward in history
        

https://user-images.githubusercontent.com/1918798/147542986-bc07f0bd-7bfc-4d9c-b702-deefce488192.mp4


ui/components/TabBar/TabBarIcon.tsx
```diff
@@ -1,35 +1,79 @@
...
+          .active span {
+            opacity: 1;
+          }
+
+          .tab_bar_icon_wrap:hover,
+          .active {
+            transform: translateY(-8px);
+          }
+          .active span {
+            color: #d08e39;
```
wojciech-turek: Is this custom or can we use a variable?

ui/pages/DAppConnectRequest.tsx
```diff
@@ -110,65 +109,65 @@ export default function DAppConnectRequest(): ReactElement {
...
           display: flex;
           flex-direction: column;
           height: 100vh;
+          margin: 0 auto;
         }
         .page {
           background-color: var(--green-95);
           height: 100vh;
           width: 100vw;
+          z-index: 1000;
```
wojciech-turek, line 170: Its easy to get lost in the z-indexes, we should make a proper structure for it, like X is for modals so i know I need X+1 for something absolute to show over a modal.

ui/pages/Earn.tsx
```diff
@@ -80,26 +79,24 @@ function EarnCard() {
 export default function Earn(): ReactElement {
   return (
     <>
-      <CorePage>
```
wojciech-turek, line 83: Seems we are ditching `CorePage` use in a lot of places, was its only purpose before to display the topBar and apply some container styles?

ui/pages/Popup.tsx
```diff
@@ -50,66 +51,83 @@ function transformLocation(inputLocation: Location): Location {
...
+                                  hasTabBar={hasTabBar}
+                                  hasTopBar={hasTopBar}
+                                >
+                                  <Component location={transformedLocation} />
+                                </CorePage>
+                              </Route>
+                            )
+                          }
+                        )}
+                      </Switch>
```
wojciech-turek, line 119: Me gusta <3

wojciech-turek (commented):
Loving this PR, left some small notes and questions, nothing blocking.

Fantastic job!

ui/pages/Earn.tsx
henryboldi: Pretty much. `CorePage` also contains the `SnackBar` component and community edition label. Previously it also handled displaying a variety of slide up menus, but now those have been moved to `TopMenu`. So `CorePage` is a lot less chunky and, making use of `routes.tsx`, is wrapping all of the pages via `Popup.tsx`

ui/pages/DAppConnectRequest.tsx
henryboldi: Good point! I've got the bad habit of just putting a high number 😅

mhluongo (approved):
Approving based on @wojciech-turek's comments offline. Polish is nice 💅

Issue #726, opened 2021-12-18
Properly display infinite token approvals


mhluongo:
Compliant "infinite" ERC-20 approvals use `type(uint256).max - 1` as a special value to denote an infinite approval. We should display "Infinite" in the approval signature and wallet activity pages instead of including that ridiculously long number :)

- [x] Display infinite approval in activity list.
- [x] Display properly-styled infinite approval in sign transaction page (see https://github.com/tallycash/extension/issues/726#issuecomment-999431175). Note: should branch off of https://github.com/tallycash/extension/pull/1053 if it is not yet merged.

VladUXUI:
![image](https://user-images.githubusercontent.com/80846886/146733472-dc7c8ead-36ed-4b46-96ea-11f287f3a24e.png)
Do we feel like infinite should be marked as an error? maybe a exclamation mark in front or written with red?

VladUXUI:
![image](https://user-images.githubusercontent.com/80846886/147071331-fc37976c-274f-4063-801b-983690ab3966.png)
Hopefully red is enough, adding an icon just messes things up.
![image](https://user-images.githubusercontent.com/80846886/147071369-0cabb1a1-fc29-45e8-8253-1f2d0091387e.png)
This tooltip could probably deserve a re-write

VladUXUI:
I feel like i want to add an option to only approve the amount the current transaction is using.
This means that editing won't be done inline, but would open a modal/pop-up.

What do you think?

mhluongo:
> I feel like i want to add an option to only approve the amount the current transaction is using.

This sounds great, but it'd be really, *really* hard to implement right now... dApps usually send each transaction separately, so while the wallet is signing the approval tx we don't know what the token amount will need to be. Also for some transactions, the only way to know the token amount needed is to run it.

For now, custom amounts inline is probably the best we can swing. For `permit()` and other signature-based approval schemes, though, we might be able to do better (eg Swap and Earn).

Shadowfiend:
@VladUXUI also missing here is how these render in the activity list.

VladUXUI:
![image](https://user-images.githubusercontent.com/80846886/151924284-546292d5-900a-409b-b75e-f884bf1a2bad.png)
Let's spell it out for consistency sake

Shadowfiend:
Actually I may have been too trigger-happy here. I believe there's a missing piece here, which is to highlight the spend amount in red in the sign page if this is an infinite approval.

[...]

PR #760, opened 2021-12-27
Approve token spend page

### LINKED ISSUE #727: Token approvals should have editable amounts on the signature page

wojciech-turek:
When users are approving token spend of an ERC-20 they should see a specific page and be able to change the approval amount.

https://user-images.githubusercontent.com/61434499/148082553-f3c706a8-282c-4fb7-bce2-97cb1c2ae1dc.mov

See #726 
Closes #727


ui/components/SignTransaction/SignTransactionApproveSpendAssetBlock.tsx
```diff
@@ -1,8 +1,67 @@
...
+  const dispatch = useDispatch()
+  const [changing, setChanging] = useState(false)
+  const accountData = useBackgroundSelector(selectCurrentAccountBalances)
+
+  const { assetAmounts } = accountData ?? {
+    assetAmounts: [],
+  }
+
+  // ERC-20 the approval amount will be in the range of 74-138 in form of a 32 bytes hex string
+  const approveAmount = transactionDetails?.input?.substring(74, 138)
```
mhluongo, line 29: I think we should properly parse this with the ABI, ala `EnrichmentService`... but not blocking.

ui/pages/SignTransaction.tsx
```diff
@@ -47,22 +47,26 @@ export default function SignTransaction(): ReactElement {
   const history = useHistory()
   const dispatch = useBackgroundDispatch()
   const location = useLocation<SignLocationState | undefined>()
+  const transactionDetails = useBackgroundSelector(selectTransactionData)
   const { assetSymbol, amount, to, value, signType } = location.state ?? {
-    signType: SignType.Sign,
+    // If tx input begins with bytecode that corresponds to ERC-20 approve
+    // we load the approve page
+    signType: transactionDetails?.input?.startsWith("0x095ea7b")
```
mhluongo, line 54: Ditto

mhluongo (changes requested):
The code looks mergeable, but the approval selector isn't coming up for dashboard.keep.network :grimacing:

mhluongo:
I couldn't test on Uniswap (this wallet already had an approval) so I tried the Keep dashboard. Results below

Screenshot from 2021-12-28 10-58-20

ui/components/SignTransaction/SignTransactionApproveSpendAssetBlock.tsx
```diff
@@ -1,8 +1,57 @@
...
+  const infiniteApproval =
+    approvalLimit ===
+    "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
+
+  const asset = assetAmounts.find((el) =>
+    "contractAddress" in el.asset
+      ? el.asset.contractAddress === transactionDetails.to
+      : undefined
+  )
+  const hexedAmount = ethers.utils.hexlify(2)
```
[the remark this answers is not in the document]
wojciech-turek: I will create that in this PR

ui/components/SignTransaction/SignTransactionApproveSpendAssetBlock.tsx
wojciech-turek: You are right, I am not 100% how to wire it up at this point though.

background/lib/utils/index.ts
```diff
@@ -106,3 +106,28 @@ export function getEthereumNetwork(): EVMNetwork {
...
+  return Number(BigInt(hexValue) / 10n ** BigInt(tokenDecimals)).toString()
+}
+
+export const numberTo32BytesHex = (value: string, decimals: number): string => {
+  const withDecimals = BigInt(value) * 10n ** BigInt(decimals)
+  const hex = utils.hexlify(withDecimals)
+  return utils.hexZeroPad(hex, 32).split("0x")[1]
+}
+
+export const verifyIsInfiniteAmount = (amount: string): boolean => {
```
mhluongo, line 128: This name is both too specific and not enough! Perhaps `isInfiniteERC20Amount` or `isMaxUint256`?

background/services/indexing/index.ts
```diff
@@ -387,6 +387,22 @@ export default class IndexingService extends BaseService<Events> {
     }
   }
 
+  getTokenMetadata = async (
```
mhluongo, line 390: I'd recommend

```typescript
async getTokenMetadata(contractAddress: HexString, network: EVMNetwork)
```

and error out if the network isn't what you want, or ignore it and leave a TODO. Multi-network is open in another PR, don't want to keep making this mistake.

background/services/indexing/index.ts
```diff
@@ -387,6 +387,22 @@ export default class IndexingService extends BaseService<Events> {
...
+    contractAddress: string
+  ): Promise<SmartContractFungibleAsset | undefined> => {
+    const provider = this.chainService.pollingProviders.ethereum
+    // pull metadata from Alchemy
+    const customAsset =
+      (await getTokenMetadata(provider, contractAddress)) || undefined
```
mhluongo, line 396: What if we already have the data in the DB? Shouldn't we check their first?

Either paradigm works, but would be good in docs

ui/components/SignTransaction/SignTransactionApproveSpendAssetBlock.tsx
```diff
@@ -1,28 +1,120 @@
...
+    setChanging(!changing)
+    if (changing && transactionDetails && asset?.decimals) {
+      const updatedInput = `${transactionDetails.input?.substring(
+        0,
+        74
+      )}${numberTo32BytesHex(
+        approvalLimit,
+        asset?.decimals
+      )}${transactionDetails.input?.substring(
+        138,
```
mhluongo, line 63: Can we pull this out to a util function? Fine to leave it in the same file (for now), but it's tough to understand tx parsing + rendering intermingled.

ui/components/SignTransaction/SignTransactionApproveSpendAssetBlock.tsx
```diff
@@ -1,28 +1,120 @@
...
+          <SharedAssetIcon size="large" symbol={asset?.symbol ?? ""} />
         </div>
       </div>
-      <span className="site">Uniswap</span>
-      <span className="spending_label">Spend KEEP tokens</span>
+      <span className="site">Smart Contract Interaction</span>
+      <span className="spending_label">
+        {asset?.symbol ? (
+          `Spend ${
+            asset?.symbol ?? truncateAddress(transactionDetails.to || "")
```
mhluongo, line 84: Putting on my Antonio hat, I think you want `??`

ui/pages/SignTransaction.tsx
```diff
@@ -50,23 +50,29 @@ export default function SignTransaction({
 
   const history = useHistory()
   const dispatch = useBackgroundDispatch()
-
-  const { assetSymbol, amount, to, value, signType } = location?.state ?? {
-    signType: SignType.Sign,
+  const transactionDetails = useBackgroundSelector(selectTransactionData)
+  const approvalCodes = ["0x095ea7b3", "0xcae9ca51"] // approve, approveAndCall
```
mhluongo, line 54: This should also be a util IMO. It'll make refactoring later simpler.

mhluongo (commented):
Left a few thoughts, haven't tried the latest code locally though

ui/components/SignTransaction/SignTransactionApproveSpendAssetBlock.tsx
Shadowfiend: I'm going to say… It *is* blocking. We should not be doing byte twiddling. Our previous foray, which we let through as not blocking, led to a lot of unexpected issues and some half fixes, including one I did that was completely broken. In general: we have the tools to do this right and should wield them.

I expect the preferred _flow_ here is going to be to pipe through the enrichment service; however, we don't need to solve that full flow here to use the correct parsing strategy. The short version here is we can move [`parseERC20Tx`](https://github.com/tallycash/extension/blob/main/background/services/enrichment/index.ts#L80-L88) to `lib/erc20.ts`, then use it here. That parse will give us an object whose `name` property should be `"approve"` (we should check that) and which should also have a property `args: { spender: string, value: BigNumber }`. The `value` can then be compared against [`ethers.constants.MaxUint256`](https://docs.ethers.io/v5/api/utils/constants/#constants-MaxUint256).

To rewrite the transaction, we can use the `ERC20_INTERFACE` exported by `lib/erc20.ts` to do `ERC20_INTERFACE.encodeFunctionData("approve", { spender: ..., value: updatedValue })`.

Shadowfiend (changes requested):
Just one… Blocker :grimacing:

background/services/indexing/index.ts
wojciech-turek: This whole thing is gone with the newest iteration.

ui/pages/SignTransaction.tsx
wojciech-turek: It doesn't belong there anyway
Alt Text

ui/components/SignTransaction/SignTransactionApproveSpendAssetBlock.tsx
wojciech-turek: This part was removed and replaced with use of ERC20_INTERFACE, but I know it was ugly :(

mhluongo:
Testing this on the [Keep dashboard](https://dashboard.keep.network/0x4495c5bf67ba43aa279f41c90083bbbfcf46e643/coverage-pools/deposit) and I'm still not seeing an approval page.

Screenshot from 2022-01-14 08-58-07

mhluongo:
Testing on Uniswap and Chrome as the Keep dash appears to use `approveAndCall`. Running into unrelated bugs. @wojciech-turek can you rebase or merge `main`?

Shadowfiend (approved):
I think we're good to go here, but haven't had a chance to test drive. Once that's done, merge is a go, and the remaining comments here can be handled separately.

Shadowfiend:
Changed closed to see for #726 as it also covers rendering infinite approvals correctly in the transaction list.

mhluongo:
Screenshot from 2022-01-19 15-29-30
Screenshot from 2022-01-19 15-30-08
Screenshot from 2022-01-19 15-30-46

PR #761, opened 2021-12-27
Moar Accounts Plz: Support additional address derivation in KeyringService

### LINKED ISSUE #763: Derive new addresses per mnemonic / HD wallet

mhluongo:
Backend component for the "Add Address" feature, allowing users to derive new addresses from a recovery phrase.

- [x] Add `KeyringService.getAddresses(keyringID)`
- [x] Add `KeyringService.deriveAddress()` and Redux wiring
- [x] Test both

Ran into and squashed a small bug in `hd-keyring` along the way.

Closes #763

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -497,6 +497,10 @@ export default class Main extends BaseService<never> {
       await this.keyringService.unlock(password)
     })
 
+    keyringSliceEmitter.on("deriveAddress", async (keyringID) => {
+      await this.keyringService.deriveAddress(keyringID)
+    })
+
     keyringSliceEmitter.on("generateNewKeyring", async () => {
       // TODO move unlocking to a reasonable place in the initialization flow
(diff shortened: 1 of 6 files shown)


wojciech-turek (approved):
This looks good to me, I just wonder about `this.requireUnlocked()`. I would probably forget to add it if I was doing that task so my question is when it is necessary to add it?

mhluongo:
> I would probably forget to add it if I was doing that task so my question is when it is necessary to add it?

Any time we are handling key material that needs to be unencrypted first. There's an argument that we don't strictly need this for `getKeyrings()` but we clear the private service keyring state... :shrug: . For `deriveAddress()` it's necessary because address derivation of these particular keyrings requires key material access... though in some crypto systems that isn't the case (eg with a Bitcoin `xpub`), there's not a common `xpub` equivalent for Ethereum.

PR #764, opened 2021-12-28
🛀🏾  No No No: fix new read-only accounts from appearing

greg-nagy:
[Cleveland Brown-No no no no NO!](https://www.youtube.com/watch?v=iabC7-9YUG4)

Modified `updateAccountBalance` reducer in a way that it can not register new account.
Can modify only accounts that exists in the store.

The bug presented when a new account was imported through a seed phrase.
After a while new read-only accounts started to appear like they were imported manually,
but without user interaction. These accounts all sent funding to the current account.

The call stack went like this:
```
/redux-slices/accounts.ts
	updateAccountBalance
chainservice
    getLatestBaseAccountBalance
indexing service
    this.chainService.emitter.on("transaction",
chain service
    saveTransaction
    retrieveTransactionReceipt
    handleQueuedTransactionAlarm
```


henryboldi:
Fuck yes!

henryboldi (approved):
It appears to work 👏

Issue #56, opened 2021-05-26
Investigate performance + extension size impact of bundling + minification


Shadowfiend:
Since extensions are loaded locally rather than requiring network round-trips and bandwidth, I haven't been able to find strong evidence either way on whether bundling and minification provide a significant runtime impact. The downloaded extension itself is gzipped, so a lot of the file size improvement of minification _most likely_ already comes from that compression process.

The goal here is to understand:
- What the loading performance impact is of unminified code.
- What the loading performance impact is of unbundled code.
- What the loading performance impact is of bundled source maps, and whether browsers pick these up easily.
- What the extension size impact is of unminified code.
- What the extension size impact is of unbundled code.
- What the extension size impact is of bundled source maps.

These can then be synthesized with any other concerns to make a decision about the best way to move forward on minification and bundling.

## But Why?

One of the most developer-hostile things modern web technology does is obfuscate code. Source maps help, but require additional work and are often not included in production bundles. Tally being an open source project, we should encourage exploration and tinkering with the underlying source---the easiest way to do that is to give users and developers an already-usable code base with the shipped extension.

0xDaedalus:
Closed b/c stale - and this does not seem to currently be a concern of the project.

PR #770, opened 2021-12-29
🔥🦊 Fox Kit: Minify the Firefox production build

mhluongo:
The Mozilla add-on store has an automated policy preventing JS files over 4 MB. Because we use webpack, they already require a full extension source code submission every version bump... leading me to believe that minification, including mangling and compression, is the right approach for Firefox in production versus other options like chunking.

To test this PR thoroughly, make sure...
* `yarn start` works as expected
* `yarn start --config-name chrome` works as expected
* `yarn start --config-name firefox` works as expected
* `yarn build --config-name firefox` produces a zip with a `ui.js` that's smaller than `yarn build --config-name chrome`
* The productions builds behave as expected

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -92,18 +92,6 @@ const baseConfig: Configuration = {
     }) as unknown as WebpackPluginInstance,
   ],
   optimization: {
-    minimizer: [
-      new TerserPlugin({
-        terserOptions: {
-          mangle: false,
-          compress: false,
-          output: {
-            beautify: true,
-            indent_level: 2, // eslint-disable-line camelcase
-          },
-        },
-      }),
-    ],
     splitChunks: { automaticNameDelimiter: "-" },
   },
 }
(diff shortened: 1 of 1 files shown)


henryboldi (approved):
I followed your testing steps, and it works as expected in all of the circumstances! ✅

Firefox build → ui.js 2,401,771 bytes (2.4 MB on disk)
Chrome build → ui.js 4,382,279 bytes (4.4 MB on disk)

Out of curiosity, why not mangle and compress the Chrome build too?

webpack.config.ts
```diff
@@ -122,13 +110,44 @@ const modeConfigs: {
...
+            mangle: false,
+            compress: false,
+            output: {
+              beautify: true,
+              indent_level: 2, // eslint-disable-line camelcase
+            },
+          },
+        }),
+      ],
+    },
```
Shadowfiend, line 126: Because this is merged into the base config, you could have left this bit there and only specified the production override. As is, using a mode other than development or production, should we ever start to do so, won't run through Terser (which might be just fine tbh). 

Not really much of an issue, but wanted to shine a light on the merge of config options that happens behind the scenes.

Shadowfiend:
> Out of curiosity, why not mangle and compress the Chrome build too?

Mangling makes the extension less discoverable and hackable for the curious dev. That ain't our jam 😁  Often this type of compression doesn't have *huge* impact on the final ZIP because source code is plain text, and ZIP compressed plain text very efficiently. 

That said, #56 remains open to turn the "often"s and "in many cases"es into concrete answers for Tally specifically.

PR #775, opened 2021-12-29
⛽ Patent Cazeline Oil: Fall back to local gas estimates

### LINKED ISSUE #735: Fall back to local gas price estimation if BlockNative is unavailable

mhluongo:
> "Gasoline" is an English word that denotes fuel for automobiles. The term is thought to have been influenced by the trademark "Cazeline" or "Gazeline", named after the surname of British publisher, coffee merchant, and social campaigner John Cassell. On 27 November 1862, Cassell placed an advertisement in The Times of London:
> > The Patent Cazeline Oil, safe, economical, and brilliant … possesses all the requisites which have so long been desired as a means of powerful artificial light.[10]
> 
> This is the earliest occurrence of the word to have been found.

— [Wikipedia](https://en.wikipedia.org/wiki/Gasoline#Etymology)

Fall back on gas estimates based on Alchemy or another provider, both on unsupported chains and in case of a BlockNative failure.

As we aren't multi-network quite yet, the best way to test this is to...
* make sure gas prices are still working with a BlockNative API key
* make sure they work after misspelling the API key

Closes #735


background/services/chain/index.ts
```diff
@@ -406,11 +391,12 @@ export default class ChainService extends BaseService<Events> {
    * Write block prices to IndexedDB so we have them for later
    */
   async pollBlockPrices(): Promise<void> {
-    // Immediately fetch the current block prices when this function gets called
-    if (this.blocknative) {
-      const blockPrices = await this.blocknative?.getBlockPrices()
-      this.emitter.emit("blockPrices", blockPrices)
-    }
+    await Promise.allSettled(
```
henryboldi, line 394: This no longer immediately fetches the current block prices, which could be a blocking regression

henryboldi (changes requested):
This no longer immediately fetches the current block prices, which, in my opinion, is a blocking regression. Users would have a decent chance of seeing an empty gas selector screen.

`this.subscribedNetworks` is an empty array when `pollBlockPrices` initially runs, because `subscribeToNewHeads` happens after

mhluongo:
Easy fix, thanks for digging in @henryboldi. One minute

mhluongo:
That last commit should have fixed it @henryboldi!

henryboldi (approved):
Awesome! Works as expected 😁

<br/>

Anyway, should `runAtStart` be `false` here now? (line 151)
```
 blockPrices: {
        runAtStart: true,
        schedule: {
          periodInMinutes:
            Number(process.env.BLOCKNATIVE_POLLING_FREQUENCY ?? "120") / 60,
        },
   ...
 ```

henryboldi (approved):
Excellent!

PR #782, opened 2021-12-30
v0.10.1

mhluongo:
"Add account", Firefox build and visual improvements, a stronger account switcher, and more resilient gas price estimation!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tallyho/tally-extension",
   "private": true,
-  "version": "0.10.0",
+  "version": "0.10.1",
   "description": "Tally, the community owned and operated Web3 wallet.",
   "main": "index.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tallyho/tally-extension",
   "private": true,
-  "version": "0.10.0",
+  "version": "0.10.1",
```
henryboldi, line 4: Could we add this version to `BUG.yml`?

Took a stab at making that automatic here → https://github.com/tallycash/extension/pull/783

henryboldi (dismissed):

package.json
mhluongo: Yep, but let's handle both after this PR? Want to get this to extension stores

greg-nagy (approved):
Let's ship this! 🛳

PR #784, opened 2021-12-31
Redirect to approve page when approving

wojciech-turek:
Instead of showing a sign page when users are approving token spend we should show the approve page.

This PR verifies the transaction input data and if it begins with `approve` or `approveAndCall` it will show the below approval page. This page will soon be updated with token, amount and a possibility to manually adjust the approve amout.

image


mhluongo:
At a bare minimum we need to either show the address of the asset and the address of what's being given a spending allowance, or the full tx data... otherwise this becomes an attack vector. Eg any dApp can request any asset to get infinite approval to an EOA. Though I think a shortcut is fine, it should be primarily aesthetic.

PR #798, opened 2022-01-04
🙈 Fix styling issue hiding addresses in account switcher

henryboldi:
Read-only back in biz. My bad!

here it is, a lovely read-only


Shadowfiend (dismissed):
Concept ack. Can't test it right now, would love it if someone else can test + merge. Otherwise will try and hit it when I get a chance.

mhluongo (approved):
Much improved.

PR #800, opened 2022-01-05
Fees components divorce 💔

wojciech-turek:
This PR separates the `NetworkFeesChooser` from the `FeeSettingsButton` and removes the `SharedSlideUp` wrapper component over the fees so that they can be reused in different areas of the app that not necessarily want to show them in a slide up. 

The `FeeSettingButton` has removed background as it was clarified with @Vlad the idea for the background was completely different than what was implemented. We are leaving that for next iteration. The button now has a hover effect.

The NetworkSettingsSelect (previously known as NetworkSettingsOptions) is now much more simpler and dumber, it takes estimated fees and transforms them in selectable options. Once the user clicks save the selected fee data that has `maxFeePerGas` and `maxPriorityFeePerGas` is transported to the parent component while the selected feeType is being stored in redux.

The next iteration of work over network fees should cover #810.


ui/components/NetworkFees/NetworkFeesChooser.tsx
```diff
@@ -54,14 +58,18 @@ export default function NetworkFeesChooser({
...
     setActiveFeeIndex(gasOptions.findIndex((el) => el.name === selectedFeeType))
-    setFeeModalOpen(false)
-  }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [feeModalOpen])
+
+  useEffect(() => {
+    if (setSelectedFeeInGwei) {
+      setSelectedFeeInGwei(gasOptions?.[activeFeeIndex]?.estimatedGwei ?? "")
```
Shadowfiend, line 69: This should be eg `onSelectedFeeInGweiChange`, right? I think of "set" as inbound ("set my X to Y") and "on" as outbound ("my X is now Y").

Shadowfiend (commented):
Tiny quick note, had to move to something else before going too far.

itsrachelfish (dismissed):
This is a step in the right direction, but you are still making too many assumptions about how this will be included in other components.

Ideally I'd like a fee chooser component that *only* includes fee selection, without a title or save button. For example, I need the save button on the swap page to save both the fee and the slippage tolerance, not just the fee selection. 

fee-selection-component

Shadowfiend (commented):
Okay, digging deep, I have a bigger thing here.

Why fire this when the active fee index changes? It feels like this should be fired when the user saves their changes, and not before. It also feels like the responsibility for doing something with those changes should be the caller's; that is, the redux dispatches should be handled by whoever creates the network fees chooser.

The same goes for whether the fee modal is open. The fees chooser should be strictly focused on presenting fees, selecting fees, and communicating that back out to a calling component. It should only depend on data that is required to populate its fee selector.

I would argue that *even the selected fee type should come from the caller*. Anything related to the user's selection should be passed into the chooser. Anything related to managing other components should be outside of here.

What this would look like (paying zero attention to our style, sorry, I'm typing fast):

```
const estimatedFeesPerGas = useSelector(...)
const [gasLimit, setGasLimit] = useState(...)

const networkSettingsSaved = (networkSettings) => {
  setGasLimit(networkSettings.gasLimit)
  // setFeeType should handle updating the transaction options
  // if needed; this may mean setFeeType should be an async thunk!
  dispatch(setFeeType(networkSettings.feeType))
  // Could be the first thing, not sure what the best order is
  // if any.
  setNetworkSettingsModalOpen(false)
}
...
<section class="fee-settings">
  <FeeSettingsButton
    currentFeeSelected={selectedFeeInGwei}
    openModal={() => setNetworkSettingsModalOpen(true)} />

  <SharedSlideUpMenu size="custom" isOpen={networkSettingsModalOpen} ...>
    <NetworkSettingsChooser
      networkSettings={{ estimatedFeesPerGas, gasLimit }}
      onNetworkSettingsSave={saveSettings} />
  </SharedSlideUpMenu>
</section>
```

A few changes here:
- `NetworkFeesChooser` -> `NetworkSettingsChooser`. The chooser is already for more than just fees (gas limit!), so let's name it as such and give ourselves a name that can contain more than just a fee choice.
- I originaly suggested `setSelectedFeeInGwei` -> `onSelectedFeeInGweiChange`, but this replaces it with `onNetworkSettingsSave`. This last iteration focuses on communicating the change in selection once, when the save button is hit, and makes it clear that the selection isn't final for the caller until that happens. The selection may have changed several times, but it's when it saves that we communicate it out.
- A more complex component like the one @itsrachelfish needs for swap could be e.g. `SwapSettingsChooser`, and would have a similarly one-input one-output interface; these use JS objects to bundle up the actual data they're presenting to the user and expose a limited, clear, data-focused, and unambiguous interface to the component that uses them. The calling component can then handle the wiring to any backend state, while the settings component can focus on bundling, presenting, and updating data for the user.
- In the `SwapSettingsChooser` case, we would factor out everything except for the save button from `NetworkFeesChooser` (now `NetworkSettingsChooser`) to a common component with a `networkSettings`/`onNetworkSettingsChange` interface, which would then be usable from `NetworkSettingsChooser` and `SwapSettingsChooser`:

```
export function SwapSettingsChooser(props...): ReactElement {
  const [networkSettings, setNetworkSettings] = useState(props.networkSettings)
  // ...
  const saveSettings = () => {
    props.onSwapSettingsSave.?({
      networkSettings,
      slippageTolerance,
    })
  }

  <form onSubmit={saveSettings}>
    <fieldset class="slippage-tolerance">
      <legend>Slippage Tolerance</legend>
      <input type="number"> // or whatever
    </fieldset>
    <fieldset class="network-fees">
      <legend>Network Settings<legend>
      <NetworkSettingsChooser
        networkSettings={networkSettings}
        onNetworkSettingsSave={setNetworkSettings} />
    </fieldset>

    <button type="submit">Save</button>
  </form>
}

// ....consumed as....

<section class="swap-settings">
  <header>
    <h3>Transaction settings</h3>
    <SharedButton icon="settings" onClick={() => setSwapSettingsModalOpen(true)} />
  </header>

  <ul>
    <li>Slippage tolerance ...</li>
    <li>Estimated network fee ...</li>
    <li>Tally fee for the DAO ...</li>
  </ul>

  <SharedSlideUpMenu size="custom" isOpen={swapSettingsModalOpen} ...>
    <SwapSettingsChooser
      swapSettings={{ networkSettings, slippageTolerance }}
      onSwapSettingsSave={saveSettings} />
  </SharedSlideUpMenu>
</section>
```
-------
I'm not marking this as a change request because it may be large enough that we should do it separately from the changes here, but I feel reasonably strongly that this is the correct end state for this component setup.

ui/components/NetworkFees/NetworkSettingsChooser.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <div className="fees standard_width">
+          <div className="title">Network Fees</div>
+          <div className="divider">
+            <div className="divider-background" />
+            <div
+              className="divider-cover"
+              style={{ left: -384 + (384 - timeRemaining * (384 / 120)) }}
+            />
+          </div>
+          <NetworkSettingsOptions
```
Shadowfiend, line 65: Non-blocking:

I would call this `NetworkSettingsSelect`, mentally mapping to the HTML `select` element. “Options” implies listing, but not necessarily selection (e.g. in `select` elements, the `option` elements provide the choices, but not the actual selection).

ui/components/NetworkFees/NetworkSettingsOptions.tsx
```diff
@@ -0,0 +1,260 @@
...
+      if (
+        typeof instant !== "undefined" &&
+        typeof express !== "undefined" &&
+        typeof regular !== "undefined"
+      ) {
+        const basePrices = [regular, express, instant]
+
+        const updatedGasOptions = basePrices.map((option) =>
+          formatBlockEstimate(option)
+        )
```
Shadowfiend, line 132: Non-blocking:

I would lift all of this out of the component, to end up with something like:

```typescript
    const updatedGasOptions = formatGasEstimates(estimatedFeesPerGas)

    if (typeof updatedGasOptions !== "undefined") {
      const selectedGasFeeIndex = ...
      const currentlySelectedFeeIndex = ...

      set, set
    }
  })
```

And where `formatGasEstimates` would return `undefined` or the full list of prices it wants to present.

I would also suggest that the gas estimates should render even if we're missing one of instant, regular, or express. Let's just show what we have—unless our styling would break, in which case we should *still* show what we have, but we can punt until we can figure out how to adapt the styles someday in the probably-very-distant future haha.

ui/components/NetworkFees/NetworkSettingsOptions.tsx
```diff
@@ -0,0 +1,260 @@
...
+        const currentlySelectedFeeIndex =
+          selectedGasFeeIndex === -1 ? 0 : selectedGasFeeIndex
+
+        setGasOptions(updatedGasOptions)
+        setActiveFeeIndex(currentlySelectedFeeIndex)
+        setGasOptions(updatedGasOptions)
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Shadowfiend, line 144: Non-blocking: same question as earlier re: exhaustiveness.

ui/components/NetworkFees/NetworkSettingsOptions.tsx
```diff
@@ -0,0 +1,260 @@
...
+        )
+        const selectedGasFeeIndex = updatedGasOptions.findIndex(
+          (el) => el.name === selectedFeeType
+        )
+        const currentlySelectedFeeIndex =
+          selectedGasFeeIndex === -1 ? 0 : selectedGasFeeIndex
+
+        setGasOptions(updatedGasOptions)
+        setActiveFeeIndex(currentlySelectedFeeIndex)
+        setGasOptions(updatedGasOptions)
```
Shadowfiend, line 141: Just to check: this is a typo, and we only need to run `setGasOptions` once, right?

ui/pages/Send.tsx
```diff
@@ -143,11 +155,26 @@ export default function Send(): ReactElement {
...
-          <NetworkFeesChooser
-            estimatedFeesPerGas={estimatedFeesPerGas}
-            gasLimit={gasLimit}
-            setGasLimit={setGasLimit}
-          />
+          <SharedSlideUpMenu
+            size="custom"
+            isOpen={networkSettingsModalOpen}
+            close={() => setNetworkSettingsModalOpen(false)}
+            customSize={`${3 * 56 + 320}px`}
```
Shadowfiend, line 162: Non-blocking:

Lot of magic numbers here. What are we trying to do?

ui/pages/SignTransaction.tsx
```diff
@@ -178,26 +194,26 @@ export default function SignTransaction({
...
+                estimatedFeesPerGas,
+                gasLimit,
+              }}
+              onNetworkSettingsSave={networkSettingsSaved}
+            />
+          </SharedSlideUpMenu>
+          <span className="detail_item">
+            Estimated network fee
+            <FeeSettingsButton
+              openModal={() => setNetworkSettingsModalOpen(true)}
```
Shadowfiend, line 214: Non-blocking:

Worth noting, given the repetition here, that we can now create a very simple component that handles the simple fee selection flow. We should probably wait until we've done it in 3 places first though.

ui/components/NetworkFees/FeeSettingsButton.tsx
```diff
@@ -1,35 +1,39 @@
...
 }
 
 export default function FeeSettingsButton({
   openModal,
-  currentFeeSelected,
-  minFee,
-  maxFee,
 }: FeeSettingsButtonProps): ReactElement {
+  const estimatedFeesPerGas = useBackgroundSelector(selectEstimatedFeesPerGas)
+  const selectedFeeType = useBackgroundSelector(selectFeeType)
```
Shadowfiend, line 18: Non-blocking:

I don't think this needs to be in redux. I think we can track it as a `useState` in the parent component, and use redux only to resolve the fee type to specific estimated fees per gas. Whether we pass a prop to this component that is `estimatedFeesPerGas`, or pass `selectedFeeType` as a prop, I think could go either way.

Shadowfiend (commented):
Going to let @itsrachelfish make the final call here. I skimmed somewhat, as I assume a lot of the code is moved around rather than rewritten. The overall structure looks sound to me, assuming it looks good to others as well.

Left a bunch of ways to tighten things up, but I don't think we need to block anything on them. The only one I'm curious about is the repeated `setGasOptions` in `NetworkSettingsOptions`---but I think that's nonblocking as well, as even if it's a mistake I don't think it harms anything.

Not hitting approve because *I didn't test this*! @itsrachelfish please give it a test whirl before merging :pray:

background/redux-slices/transaction-construction.ts
```diff
@@ -22,6 +22,15 @@ const enum TransactionConstructionStatus {
   Signed = "signed",
 }
 
+export type NetworkFeeSetting = {
+  feeType: string
+  gasLimit: string
+  values: {
+    maxFeePerGas: bigint
+    maxPriorityFeePerGas: bigint
+  }
```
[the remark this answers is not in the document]
wojciech-turek: We definitely can, I have added a new selector for that, passing the values from a component might be just a faster way to consume the data without fetching it from redux.

background/redux-slices/transaction-construction.ts
```diff
@@ -22,6 +22,15 @@ const enum TransactionConstructionStatus {
   Signed = "signed",
 }
 
+export type NetworkFeeSetting = {
+  feeType: string
```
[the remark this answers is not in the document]
wojciech-turek: The `NetworkFeeSetting` also includes the gasLimit and the values, naming it `NetworkFeeTypeChosen` could suggest it only contains the type? I won't argue if you feel strong about the names here.

background/redux-slices/transaction-construction.ts
```diff
@@ -120,18 +129,35 @@ const transactionSlice = createSlice({
...
+            feeTypeSelected: NetworkFeeTypeChosen.Instant,
+          }
+
+        default:
+          break
+      }
+      return {
+        ...state,
+        feeTypeSelected: NetworkFeeTypeChosen.Express,
+      }
```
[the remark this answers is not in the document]
wojciech-turek: It is more elegant I agree, will make a commit with suggested change.

ui/components/NetworkFees/FeeSettingsButton.tsx
```diff
@@ -1,35 +1,39 @@
...
 
 interface FeeSettingsButtonProps {
   openModal: () => void
-  currentFeeSelected: string
-  minFee: number
-  maxFee: number
 }
 
 export default function FeeSettingsButton({
   openModal,
```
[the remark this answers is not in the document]
wojciech-turek: 💯

ui/components/NetworkFees/FeeSettingsButton.tsx
```diff
@@ -1,35 +1,39 @@
...
+    if (estimatedFeesPerGas !== undefined && selectedFeeType !== undefined) {
+      const estimatedGweiAmount = weiToGwei(
+        (estimatedFeesPerGas?.baseFeePerGas *
+          ESTIMATED_FEE_MULTIPLIERS_BY_TYPE[selectedFeeType.toString()]) /
+          10n
+      ).split(".")[0]
+
+      setEstimatedGweiPrice(estimatedGweiAmount)
+    }
+  }, [estimatedFeesPerGas, selectedFeeType])
```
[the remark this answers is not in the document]
wojciech-turek: Applied the above, thanks 🙏

ui/components/NetworkFees/NetworkSettingsOptions.tsx
wojciech-turek: That would be a good suggestion, with that we should also get ready for more than 3 options. The design includes an option called `Suggested by DApp` which we are not showing. I will make a new issue for it seems we won't have time for that now.

ui/components/NetworkFees/NetworkSettingsOptions.tsx
wojciech-turek: Just once, good catch, thanks!

ui/pages/Send.tsx
wojciech-turek: Umm 🤔  I know its not sustainable. Its basically numberOfOptions * optionHeight + other space needed. I will change it to a single value so its less confusing.

ui/components/NetworkFees/FeeSettingsButton.tsx
wojciech-turek: You're most likely right I made that structure based on your earlier suggestion that we should be keeping the type selected in redux, but I can see how that might have changed considered how we changed the structure since then.

ui/components/NetworkFees/NetworkSettingsOptions.tsx
wojciech-turek: I don't remember after this restructure what was the point of it but I think i didnt want the updateGasOptions function to run once user selected the fee option. Removed in this iteration.

ui/components/NetworkFees/FeeSettingsButton.tsx
Shadowfiend: Sorry, I think I muddied the waters here a bit because I was going fast.

We should track the fee type the user has selected in redux. However, that is the fee type *after the user hits save*. Before the user hits save, the fee type is not truly “selected”, because the user can hit the X to cancel and not change the selection in redux.

For this button in particular, I meant that the button could receive the information as a prop, rather than reading it from redux. The parent component would read from redux and update redux when the network settings were saved.

Does that clarify things?

wojciech-turek: Maybe I misunderstood your intent too.

We do only save the selected fee in redux once the user hits the save button, but I was missing on the state refresh when user closes the modal with X button. I have followed your advice and used the selected fee type from redux as default value for local useState and it seems to be working great.

Thanks!

ui/pages/Send.tsx
```diff
@@ -159,20 +159,21 @@ export default function Send(): ReactElement {
             size="custom"
             isOpen={networkSettingsModalOpen}
             close={() => setNetworkSettingsModalOpen(false)}
-            customSize={`${3 * 56 + 320}px`}
+            customSize="488px"
```
Shadowfiend, line 162: To follow up on my previous comment here: I actually liked the explicitness of breaking out the math, I would just have preferred we have some constant names (e.g., 3 * OPTION_HEIGHT + PADDING or whatever). Ideally we would give the slide-up menu the ability to figure out its own size, maybe with a `fitToContents` or something—but it's possible that interacts poorly with the transition.

ui/components/NetworkFees/NetworkSettingsChooser.tsx
```diff
@@ -62,12 +66,17 @@ export default function NetworkSettingsChooser({
...
-            setCustomGasLimit={setCustomGasLimit}
-            onSelectNetworkSetting={setSelectedSetting}
-          />
+          {visible ? (
+            <NetworkSettingsSelect
+              estimatedFeesPerGas={estimatedFeesPerGas}
+              gasLimit={customGasLimit}
+              setCustomGasLimit={setCustomGasLimit}
+              onSelectNetworkSetting={setSelectedSetting}
+              selectedFeeType={selectedFeeType}
```
Shadowfiend, line 75: This still isn't quite the goal I was going for. `NetworkSettingsSelect`

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -0,0 +1,267 @@
...
+  maxFeePerGas: bigint
+  maxPriorityFeePerGas: bigint
+}
+
+export default function NetworkSettingsSelect({
+  estimatedFeesPerGas,
+  gasLimit,
+  setCustomGasLimit,
+  onSelectNetworkSetting,
+  selectedFeeType,
```
Shadowfiend, line 46: I realized I missed something important when I suggested the rename here, which is that this component isn't just the select, it's the select and the gas limit setting...

I think this still needs some refinement both on interface and component breakdown, but let's punt for now.

Shadowfiend (approved):
Test drove, left a few final remarks, but this is ready to rock.

PR #801, opened 2022-01-05
📝 Create a recovery phrase/full account in Tally onboarding

henryboldi:
Want to create a new recovery phrase inside of Tally? Let's do it.

There are some controversial changes to note,
 - 12 words, not 24
 - `generateNewKeyring` doesn't save the keyring. Saving becomes the job of `importKeyring`. This way, we don't have to do the dance of deleting a key if the user doesn't complete the verify flow.

### Add wallet

The previous design we had implemented only had room from two options. So I've brought from this design from the archives with three options.

add wallet

### Write down phrase

Let's encourage actually writing this down rather than putting it inside of the clipboard. Clipboards can be easily accessed by other programs or malicious software.
save

### Verify a phrase

Did you really write it down?
verify

### 🎉 Successfully repeated back the phrase
congrats


background/redux-slices/keyrings.ts
```diff
@@ -83,16 +88,20 @@ const keyringsSlice = createSlice({
...
     },
+    setKeyringToVerify: (state, { payload }) => ({
+      ...state,
+      keyringToVerify: payload,
+    }),
   },
   extraReducers: (builder) => {
     builder
-      .addCase(importLegacyKeyring.pending, (state) => {
+      .addCase(importKeyring.pending, (state) => {
```
greg-nagy, line 98: I have ran into situations where this method proved to be unreliable. 
But that was around dApp permission request so it was probably the result of a window closing.

The solution was to dispatch sync actions from the async thunk. Not sure if it applies here, but good thing to have in mind.

background/services/keyring/index.ts
```diff
@@ -255,41 +255,29 @@ export default class KeyringService extends BaseService<Events> {
   ): Promise<{ id: string; mnemonic: string[] }> {
     this.requireUnlocked()
 
-    if (type !== KeyringTypes.mnemonicBIP39S256) {
+    if (type !== KeyringTypes.mnemonicBIP39S128) {
```
greg-nagy, line 258: Poor Johnny got a memory downgrade ... :D

> The Johnny Mnemonic character is a data trafficker who has undergone cybernetic surgery to have a data storage system implanted in his head.

https://williamgibson.fandom.com/wiki/Johnny_Mnemonic

background/services/keyring/index.ts
```diff
@@ -255,41 +255,29 @@ export default class KeyringService extends BaseService<Events> {
...
     }
 
-    const newKeyring = new HDKeyring({ strength: 256 })
-    this.#keyrings.push(newKeyring)
-    const [address] = newKeyring.addAddressesSync(1)
-    await this.persistKeyrings()
-
-    this.emitter.emit("address", address)
-    this.emitKeyrings()
+    const newKeyring = new HDKeyring({ strength: 128 })
```
greg-nagy, line 264: I am loving that you are deleting/modifying background services :)

greg-nagy: `getKeyrings()` returns `mnemonicBIP39S256`. Not sure if it's an issue, but the method is used in `emitKeyrings` and the `keyrings` method is propagated to the store.

greg-nagy: Might be a good time to increase our leverage for stability opportunities ...  eg. write test for it :)

ui/components/Shared/SharedButton.tsx
```diff
@@ -54,9 +54,11 @@ export default function SharedButton(props: Props): ReactElement {
     return <Redirect push to={linkTo} />
   }
 
-  function handleClick(e: React.MouseEvent<HTMLButtonElement, MouseEvent>) {
+  async function handleClick(
+    e: React.MouseEvent<HTMLButtonElement, MouseEvent>
+  ) {
     setIsClicked(true)
-    onClick?.(e)
+    await onClick?.(e)
```
greg-nagy, line 61: Why did you need this here? To wait for the dispatches In SuccessMessage? 
This worries me, if it's really the direction we need?

ui/pages/Onboarding/OnboardingVerifySeed.tsx
```diff
@@ -45,56 +71,81 @@ function SuccessMessage() {
...
+              onClick={() => {
+                handleRemove(item)
+              }}
               icon="close"
-              isDisabled
-            >{`${index + 1} - cat`}</SharedButton>
+            >{`${index + 1} - ${item}`}</SharedButton>
           </li>
         ))}
+        <KeepInView />
```
greg-nagy, line 127: Did you intentionally shift out of the screen the first part of the entered seed phrase while there is a lot of space underneth? 

I wanted to check back if I entered my mnemonics in the correct order, but could not do it. I think I would prefer if I could.

greg-nagy (changes requested):
I love, that you edited the keyring service in such a bald way! :) 

- I alway ran into the incorrect order, which I believe is a bug and can be related to the getKeyrings() method still using 256 mnemonics
- It was strange that on verify step the entered seeds are running out of the window and I could not verify what I have entered 
- the async onClick felt really weird. Can we avoid doing that somehow? 

Overall I am loving this PR :) 
I never really understood why we postponed this for so long.

ui/pages/Onboarding/OnboardingVerifySeed.tsx
henryboldi: You can scroll up, no?

ui/components/Shared/SharedButton.tsx
henryboldi: Yes, to wait for the dispatches in `SuccessMessage `. What worries you about this? Do you have a different solution in mind?

background/redux-slices/keyrings.ts
```diff
@@ -83,16 +88,20 @@ const keyringsSlice = createSlice({
...
         return {
           ...state,
           importing: "pending",
         }
       })
-      .addCase(importLegacyKeyring.fulfilled, (state) => {
+      .addCase(importKeyring.fulfilled, (state) => {
```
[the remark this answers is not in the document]
henryboldi: Ah good point! Thank you

background/main.ts
```diff
@@ -617,17 +618,19 @@ export default class Main extends BaseService<never> {
...
     })
 
-    keyringSliceEmitter.on(
-      "importLegacyKeyring",
-      async ({ mnemonic, path }) => {
-        await this.keyringService.importLegacyKeyring(mnemonic, path)
-      }
-    )
+    keyringSliceEmitter.on("importKeyring", async ({ mnemonic, path }) => {
+      await this.keyringService.importKeyring(mnemonic, path)
```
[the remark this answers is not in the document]
mhluongo: The original intent was to only import 12-word phrases, and always generate 24-word for new recovery phrases.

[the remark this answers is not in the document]
henryboldi: The original intention behind creating `importLegacyKeyring` was to import MetaMask "legacy" 12 word keys. But given #723 I don't think that's the case anymore.

cc @Shadowfiend

ui/pages/Onboarding/OnboardingVerifySeed.tsx
greg-nagy: I can — now that you told me that I can. I don't have a scrollbar and there is a lot of space under the confirm button. 
It even occured to me to scroll, but I guess I did not hit the area so I could not, when tried.

ui/components/Shared/SharedButton.tsx
greg-nagy: If you move the cleanup logic into the fullfilled reducer then you don't need 2 dispatches and I think you can skip the await part. 

For me using promises in a click handler felt strange. And also it's kind of a big change in a widely used component only for a single use case.

Generally async logic should live in async thunks or the main.ts in my view. Patching things together in the view feels out of place.

henryboldi: Makes sense! Thank you. I'll update this

background/main.ts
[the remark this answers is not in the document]
greg-nagy: So this 12-word phrase is not legacy after all? Ot this event/method will be responsible for importing all the keyrings? 

Removing the `Legacy` part signaled to me an increase in responsibility which might not have been the original idea. 
So I thought it worth a question.

background/services/keyring/index.ts
henryboldi: It returns `mnemonicBIP39S256` in the `type` attribute? If we've been importing legacy keyrings then this type has been mostly wrong all along, no?

ui/pages/Onboarding/OnboardingAddWallet.tsx
```diff
@@ -1,141 +1,153 @@
...
+import { generateNewKeyring } from "@tallyho/tally-background/redux-slices/keyrings"
+import { useBackgroundDispatch, useAreKeyringsUnlocked } from "../../hooks"
 import SharedBackButton from "../../components/Shared/SharedBackButton"
 import SharedButton from "../../components/Shared/SharedButton"
 
-export default function OnboardingAddWallet(): ReactElement {
+export default function OnboardingStartTheHunt(): ReactElement {
+  const dispatch = useBackgroundDispatch()
+
+  useAreKeyringsUnlocked(!HIDE_CREATE_PHRASE && true)
```
greg-nagy, line 14: The `&& true` does not seem to be necessary here.

greg-nagy (approved):
it worked for me this time and I don't see any blocking issue, so let's do this 🎸

background/main.ts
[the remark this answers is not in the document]
Shadowfiend: > The original intention behind creating importLegacyKeyring was to import MetaMask "legacy" 12 word keys. But given #723 I don't think that's the case anymore.

I don't think that's true, fwiw. 723 didn't change anything about the kind of key we're importing, just what we're deriving from it.

> Removing the Legacy part signaled to me an increase in responsibility which might not have been the original idea.

I think this is probably still true. The underlying question is whether we want the same method to import “legacy“ keys and our own keys. Mostly the argument here is a validation one: if we use different methods for the two, validating different reqs is straightforward (e.g., 24 vs 12 words). We _should_ validate those requirements in the keyring service _in addition to_ any UI validation we do—this is the standard principle of having the UI do user-facing validation, and letting the sensitive systems in the backend be locally responsible for data-shape validation.

From that perspective, if we move generated keyrings to 24-word, we should make sure that we can declare intent (e.g. by passing a keyring type alongside the mnemonic), so that we can validate that the given mnemonic and desired keyring type agree.

PR #805, opened 2022-01-06
🪳 Fix for precompiled validation: refactor uniswap token list validation to use precompiled validators

### LINKED ISSUE #797: Improve developer ergonomics for precompiled validator usage
# Background

In #796 we moved to precompile the validator functions so we can remove the `unsafe-eval` CSP policy but it happened in a bit of a rush. 

# Purpose 
The usage of ajv is not really developer-friendly, and it would be good to be able to use it more conveniently.

## Outline for the solution

- move all the schema definitions into a separate folder
- write code that imports all of them and
  - adds them to an ajv instance
  - compiles the validators in esm modules
  - writes the compiled output into a file
- maybe
  - generates type guard functions and exports them, so they can be used throughout the codebase
  - keep the schema files next to the implementation code

# References
- ajv standalon usage docs https://ajv.js.org/standalone.html
- ajv standalone esm code generation issue: https://github.com/ajv-validator/ajv/issues/1523
- ajv standalon example thread: https://github.com/ajv-validator/ajv/issues/1831
- ajv esm compile docs https://ajv.js.org/guide/environments.html#cjs-vs-esm-exports
- this project uses ajv in a similar manner described above [gajus/table](https://github.com/gajus/table)
  - https://github.com/gajus/table/blob/master/package.json#L73 - compile npm script 
  - https://github.com/gajus/table/blob/master/src/types/generated/validators.d.ts - typing

greg-nagy:
## Rest of the commits - generator script based approach
- wrote a generator typescript file, that can generate the validators 
  - this imports the typscript schema definitions and uses that for validator generation
- updated the validators and simplified their typings utilising `.d.ts` files
- removed the unused code
- removed the duplicated json files

## First commit - ajv-cli approach solution
- added nohoist config to package.json > workspaces so these packages are kept in the local `node_modules`
- extended the json schema generation with
  - the uniswap tokenlist json schema
  - usage of ajv-formats
- created the wrapper code for `isValidUniswapTokenListResponse()`
- refactored the `tokenList.ts` to use the new generated validator `isValidUniswapTokenListResponse()`

This was tricky, because
- we use json schema published in @uniswap/token-list
- the schema uses date format validators which require ajv-formats
- ajv-cli and ajv-formats needs to be in the same `node_modules` for the npm script to work
- package.json is the root for the npm scripts so it's not possible to use `../` in a path to leave the folder of the current package
- yarn workspaces hoist packages and try to put as much as it can into the root `node_modules` https://classic.yarnpkg.com/blog/2018/02/15/nohoist/

Closes #797


Shadowfiend:
This is a great example of one commit that should be many commits, one per bullet :wink:

> added nohoist config to package.json > workspaces so these packages are kept in the local node_modules

Also need to better understand why we want this. Is the goal here to avoid webpack pulling it in or what?

greg-nagy:
> This is a great example of one commit that should be many commits, one per bullet 😉
> 
> > added nohoist config to package.json > workspaces so these packages are kept in the local node_modules
> 
> Also need to better understand why we want this. Is the goal here to avoid webpack pulling it in or what?

Your wish is my command! :D 

This whole setup bothered me a lot more than it should have, so I went ahead and finished the idea in #797

README.md
```diff
@@ -93,8 +93,10 @@ $ ./scripts/macos-setup.sh
 If you need to create or update a validation function then:
 
 - You need to write the schema in the `.ts` file to have correct typing.
-- And you also need to create a json equivalent of that schema in `/background/lib/validate/...` folder
+- Add the new schema with the validator function name to the `generate-validators.ts` file
+- You need to update the `json-validators.d.ts` or `json-validators.d.ts` files with the typing definition
```
wojciech-turek, line 97: Seems like a typo duplicate.

wojciech-turek (approved):

wojciech-turek (approved):

PR #818, opened 2022-01-11
🗃 Preserve location when reopening extension

henryboldi:
This also preserves the location's `state` attribute. For example, when the token clicked is passed to the single asset page.

After discovering `Router`'s `initialEntries` prop I was able to massively simplify the approach here.

⚠️ If a user navigates to a page they can't exit, there's no escape


wojciech-turek (dismissed):
I have found a breaking bug. To replicate:

Go to Send page => Select data => Go to Sign Page =>  click away to close extension => go back to extension and click Reject

We are left with a blank screen, if i close and reopen the send page shows correctly

ui/pages/Popup.tsx
```diff
@@ -24,9 +24,16 @@ const pagePreferences = Object.fromEntries(
   ])
 )
 
+const initialEntries = JSON.parse(
+  (localStorage.getItem("entries") !== "undefined" &&
```
greg-nagy, line 28: Not blocking but using `??` would have been simpler here.

ui/pages/Popup.tsx
```diff
@@ -54,6 +61,39 @@ export default function Popup({ store }: { store: Store }): ReactElement {
...
+        "/keyring/"
+      )
+
+    // Initial extension load takes two renders because of setting
+    // animation control states. `initialEntries` needs to be a reversed
+    // version of route history entities. Without avoiding the initial load,
+    // entries will keep reversing.
+    if (renderCount.current > 1 && isNotOnKeyringRelatedPage) {
+      const entries = routeHistoryEntities
+        .reduce((agg: Partial<Location>[], entity) => {
```
greg-nagy, line 80: hm, did not know that agg also works. Thanks

greg-nagy (approved):
Love how easy this important feature was 

🥷

PR #820, opened 2022-01-12
Move swap provider integration to background script

Shadowfiend:
As advertised. To do this, the internal provider has to be moved away
from using ports, which can only operate between extension contexts,
to using a shim that can operate inside the same extension context (in
this case, the background script). Everything still pipes through the
Ethers Web3Provider abstraction, but the internals use simple Emittery
events and method calls.

Most likely the right home for the provider abstraction is in the service
rather than in the redux-slices utils, but right now the goal is to
achieve something that works, then immediatley clean it up a bit.


background/redux-slices/0x-swap.ts
```diff
@@ -165,6 +171,49 @@ export const fetchSwapQuote = createBackgroundAsyncThunk(
...
+      pendingTransactionPromises.push(
+        signer.sendTransaction(approvalTransactionData)
+      )
+    }
+
+    pendingTransactionPromises.push(
+      signer.sendTransaction(quote as TransactionRequest)
+    )
+
+    await Promise.all(pendingTransactionPromises)
```
itsrachelfish, line 213: Could this lead to a race condition? How do you ensure the allowance transaction gets mined before the swap transaction?

Shadowfiend: Because we send them to the provider one after the other, they get separate nonces. The provider moves through messages sequentially. However, the logic for the nonces isn't there yet. That and handling navigation are the two missing pieces here.

itsrachelfish (approved):

background/redux-slices/0x-swap.ts
```diff
@@ -165,6 +171,49 @@ export const fetchSwapQuote = createBackgroundAsyncThunk(
...
+
+    const pendingTransactionPromises: Promise<TransactionResponse>[] = []
+
+    const existingAllowance: BigNumber =
+      await assetContract.callStatic.allowance(
+        await signer.getAddress(),
+        quote.allowanceTarget
+      )
+
+    if (existingAllowance.lt(quote.sellAmount)) {
```
greg-nagy, line 195: What should happen if this is not true? Should we signal it or communicate to the user?

PR #821, opened 2022-01-12
📸 Fix tab bar icon sometimes flashing background color

henryboldi:
I've never seen this flicker in production yet, but it pops up sometimes in development, so we might as well patch it!


wojciech-turek (approved):

wojciech-turek:
Lookin' good!

PR #822, opened 2022-01-12
Nonces and queues: More direct nonce management

Shadowfiend:
Nonce management only in this branch now. The lifecycle still needs some
tweaks to invalidate nonces that don't get a signature.

Queuing transactions is moving to a separate PR.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -467,6 +467,7 @@ export default class Main extends BaseService<never> {
           this.store.dispatch(signed(signedTx))
         } catch (exception) {
           logger.error("Error signing transaction; releasing nonce", exception)
+          this.chainService.releaseEVMTransactionNonce(transactionWithNonce)
         }
       }
     )
--- a/background/main.ts
+++ b/background/main.ts
@@ -420,50 +420,23 @@ export default class Main extends BaseService<never> {
     )
 
     transactionConstructionSliceEmitter.on("updateOptions", async (options) => {
-      // TODO Deal with pending transactions.
-      const resolvedNonce =
-        await this.chainService.pollingProviders.ethereum.getTransactionCount(
-          options.from,
-          "latest"
+      const { transactionRequest: populatedRequest, gasEstimationError } =
+        await this.chainService.populatePartialEVMTransactionRequest(
+          getEthereumNetwork(),
+          options
         )
 
-      // Basic transaction construction based on the provided options, with extra data from the chain service
-      const transaction: EIP1559TransactionRequest = {
-        from: options.from,
-        to: options.to,
-        value: options.value ?? 0n,
-        gasLimit: options.gasLimit ?? 0n,
-        maxFeePerGas: options.maxFeePerGas ?? 0n,
-        maxPriorityFeePerGas: options.maxPriorityFeePerGas ?? 0n,
-        input: options.input ?? null,
-        type: 2 as const,
-        chainID: "1",
-        nonce: resolvedNonce,
-      }
-
-      try {
-        // We use estimateGasLimit only if user did not specify the gas explicitly or it was set below minimum
-        if (
-          typeof options.gasLimit === "undefined" ||
-          options.gasLimit < 21000n
-        ) {
-          transaction.gasLimit = await this.chainService.estimateGasLimit(
-            getEthereumNetwork(),
-            transaction
-          )
-        }
-        // TODO If the user does specify gas explicitly, test for success.
-
+      if (typeof gasEstimationError === "undefined") {
         this.store.dispatch(
           transactionRequest({
-            transactionRequest: transaction,
+            transactionRequest: populatedRequest,
             transactionLikelyFails: false,
           })
         )
-      } catch (error) {
+      } else {
         this.store.dispatch(
           transactionRequest({
-            transactionRequest: transaction,
+            transactionRequest: populatedRequest,
             transactionLikelyFails: true,
           })
         )
--- a/background/main.ts
+++ b/background/main.ts
@@ -62,6 +62,7 @@ import {
   emitter as providerBridgeSliceEmitter,
   initializeAllowedPages,
 } from "./redux-slices/dapp-permission"
+import logger from "./lib/logger"
 
 // This sanitizer runs on store and action data before serializing for remote
 // redux devtools. The goal is to end up with an object that is directly
@@ -452,12 +453,21 @@ export default class Main extends BaseService<never> {
 
     transactionConstructionSliceEmitter.on(
       "requestSignature",
-      async (transaction: EIP1559TransactionRequest) => {
-        const signedTx = await this.keyringService.signTransaction(
-          transaction.from,
-          transaction
-        )
-        this.store.dispatch(signed(signedTx))
+      async (
+        transaction: EIP1559TransactionRequest & { nonce: number | undefined }
+      ) => {
+        const transactionWithNonce =
+          await this.chainService.populateEVMTransactionNonce(transaction)
+
+        try {
+          const signedTx = await this.keyringService.signTransaction(
+            transaction.from,
+            transactionWithNonce
+          )
+          this.store.dispatch(signed(signedTx))
+        } catch (exception) {
+          logger.error("Error signing transaction; releasing nonce", exception)
+        }
       }
     )
 
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -37,7 +37,7 @@ import {
   ethersTransactionFromSignedTransaction,
   transactionFromEthersTransaction,
 } from "./utils"
-import { getEthereumNetwork } from "../../lib/utils"
+import { getEthereumNetwork, normalizeEVMAddress } from "../../lib/utils"
 
 // We can't use destructuring because webpack has to replace all instances of
 // `process.env` variables in the bundled output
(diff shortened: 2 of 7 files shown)


background/services/chain/index.ts
```diff
@@ -707,14 +800,31 @@ export default class ChainService extends BaseService<Events> {
...
+            typeof addressNetwork.network.chainID !== "undefined" &&
+            typeof this.evmChainNoncesByAddress[
+              addressNetwork.network.chainID
+            ]?.[transaction.from.toLowerCase()] !== "undefined" &&
+            this.evmChainNoncesByAddress[addressNetwork.network.chainID]?.[
+              transaction.from.toLowerCase()
+            ] <= transaction.nonce
+          ) {
+            this.evmChainNoncesByAddress[addressNetwork.network.chainID][
+              transaction.from.toLowerCase()
```
mhluongo, line 824: `normalizeEVMAddress` plz and ty :)

background/services/chain/index.ts
```diff
@@ -707,14 +800,31 @@ export default class ChainService extends BaseService<Events> {
...
+            getEthereumNetwork()
           )
+
+          // If this is an EVM chain, we're tracking the from address's
+          // nonce, and the pending transaction has a higher nonce, update our
+          // view of it. This helps reduce the number of times when a
+          // transaction submitted outside of this wallet causes this wallet to
+          // produce bad transactions with reused nonces.
+          if (
+            typeof addressNetwork.network.chainID !== "undefined" &&
```
mhluongo, line 815: This is difficult to parse and something we need elsewhere. Can we use the typeguard from the multi-network PR or another broken out here?

background/services/chain/index.ts
```diff
@@ -211,6 +216,94 @@ export default class ChainService extends BaseService<Events> {
...
+    }
+
+    // TODO Deal with multi-network.
+    const networkNonce =
+      await this.pollingProviders.ethereum.getTransactionCount(
+        address,
+        "latest"
+      )
+
+    this.evmChainNoncesByAddress[network.chainID] ??= {}
```
mhluongo, line 301: It'd be nice to know if these are the nonces that we've seen, the nonces that have been confirmed, or...? Feels like a better var name is in order

background/services/chain/index.ts
```diff
@@ -111,6 +112,10 @@ export default class ChainService extends BaseService<Events> {
     provider: AlchemyWebSocketProvider
   }[]
 
+  private evmChainNoncesByAddress: {
```
mhluongo, line 115: Ser please my family needs docs

mhluongo (changes requested):
Concept ACK, open implementation question

background/services/chain/index.ts
Shadowfiend: Doesn't everything :)

background/services/chain/index.ts
Shadowfiend: (Note that this filter is _more type-safe than a type guard_. Type guards are assumed to be internally correct, this is verified by the compiler to filter the type union as needed.)

mhluongo: I just meant: is there not a better way to do this that we can reuse? But it's not blocking at all

Shadowfiend: Unfortunately the type guard is the way that's reusable, but it's also less type-safe. Probably the right long-term solution though.

background/services/chain/index.ts
Shadowfiend: Light docs + FIXME added.

background/services/chain/index.ts
Shadowfiend: Tried to clarify this.

background/services/chain/index.ts
Shadowfiend: Donezo.

Shadowfiend:
Need to refine this on top of the base swap branch, but the changes will remain more or less as-is. Tackling the lifecycle stuff tomorrow.

Shadowfiend:
Need to do a little commit juicing, but this is the expected final state here.

Shadowfiend:
Flipping this to ready for review.

Shadowfiend:
Lol CI disagreed 😢

background/main.ts
```diff
@@ -479,12 +453,22 @@ export default class Main extends BaseService<never> {
...
+
+        try {
+          const signedTx = await this.keyringService.signTransaction(
+            transaction.from,
+            transactionWithNonce
+          )
+          this.store.dispatch(signed(signedTx))
+        } catch (exception) {
+          logger.error("Error signing transaction; releasing nonce", exception)
+          this.chainService.releaseEVMTransactionNonce(transactionWithNonce)
```
mhluongo, line 470: Woo we've got locks

background/services/chain/index.ts
```diff
@@ -111,6 +112,16 @@ export default class ChainService extends BaseService<Events> {
     provider: AlchemyWebSocketProvider
   }[]
 
+  /**
+   * For each chain id, track an address's last seen nonce. The tracked nonce
+   * should generally not be allocated to a new transaction, nor should any
+   * nonces that precede it, unless the intent is deliberately to replace an
+   * unconfirmed transaction sharing the same nonce.
+   */
+  private evmChainLastSeenNoncesByNormalizedAddress: {
```
mhluongo, line 121: :pray: thank you for making this multi-network :heart:

background/services/chain/index.ts
```diff
@@ -211,6 +222,163 @@ export default class ChainService extends BaseService<Events> {
...
+      transactionRequest.gasLimit = estimatedGasLimit
+    }
+
+    return { transactionRequest, gasEstimationError }
+  }
+
+  /**
+   * Populates the nonce for the passed EIP1559TransactionRequest, provided
+   * that it is not yet populated. This process generates a new nonce based on
+   * the known on-chain nonce state of the service, attempting to ensure that
```
mhluongo, line 297: Might be good to call out more specifically that this ignores the mempool?

background/services/chain/index.ts
```diff
@@ -211,6 +222,163 @@ export default class ChainService extends BaseService<Events> {
...
+   * the nonce will be unique and an increase by 1 over any other confirmed or
+   * pending nonces in the mempool.
```
mhluongo, line 299: Ah. This contradicts the "on-chain" bit earlier.

background/services/chain/index.ts
```diff
@@ -211,6 +222,163 @@ export default class ChainService extends BaseService<Events> {
...
+    // and continuously; if the address has a pending transaction floating
+    // around with a nonce that is not an increase by one over previous
+    // transactions, this approach will allocate more nonces that won't mine.
+    // FIXME Double-check the getTransactionCount-based nonce to make sure
+    // FIXME using its precedent would result in a replacement transaction
+    // FIXME error or a nonce reuse error.
+    // TODO Deal with multi-network.
+    this.evmChainLastSeenNoncesByNormalizedAddress[chainID][
+      normalizedAddress
+    ] ??=
```
mhluongo, line 328: so fancy

background/services/chain/index.ts
```diff
@@ -211,6 +222,163 @@ export default class ChainService extends BaseService<Events> {
...
+    const { chainID, nonce } = transactionRequest
+    const normalizedAddress = normalizeEVMAddress(transactionRequest.from)
+    const lastSeenNonce =
+      this.evmChainLastSeenNoncesByNormalizedAddress[chainID][normalizedAddress]
+
+    // TODO Currently this assumes that the only place this nonce could have
+    // TODO been used is this service; however, another wallet or service
+    // TODO could have broadcast a transaction with this same nonce, in which
+    // TODO case the nonce release shouldn't take effect! This should be a
+    // TODO relatively rare edge case, but we should handle it at some point.
```
mhluongo, line 365: Way less rare when hardware wallet support goes live I'd bet.

mhluongo: Thinking about it, I'm most concerned about edge cases where the extension is closed to switch wallets or leave, etc. Any concerns around the lask of persistence here?

mhluongo (approved):
Appears to work well locally. I ran into an issue with broadcasting an ETH payment to my own wallet, but it's a separate bug.

mhluongo:
Opened a new bug for a nonce-related issue I ran into here... but IMO it's out of the scope of this PR.

https://github.com/tallycash/extension/issues/858

background/services/chain/index.ts
Shadowfiend: It doesn't! We update the seen nonces in the `filteredNewFullPendingTransactionsSubscriptionID` event watcher, so pending transactions external to the wallet also update this state.

There's an interesting question here on whether that should be the only way to increment the nonce (with releasing being the only way to decrease it).

mhluongo: > the known on-chain nonce state of the service

My bone to pick then :D

background/services/chain/index.ts
Shadowfiend: Curious what you mean re: hardware wallet support here.

Currently the only time we actively allocate a nonce is right before internal signing occurs (i.e., after the user has hit the approve button). For the keyring this means the nonce is immediately consumed, though broadcast may be delayed if it's for a dApp using `eth_signTransaction`. For hardware wallets we should probably time the nonce out and throw an exception, releasing the nonce at that point.

Obviously this still leaves edge cases, and there's a question of e.g. what we should do if we're currently signing a nonce and receive a pending transaction with the same nonce, but that seems even less likely with hardware signing since generating a nonce conflict would requires a synchronous action from the same hardware wallet in a different place.

Ultimately we'll need to handle, in the UI, errors like `replacement transaction underpriced` and `nonce too low`, either by displaying the relevant state or by prompting the user to re-sign or both. Lot of complexity behind that one.

Issue #793, opened 2022-01-03
PRs from other forks fail CI, even if the CI run is approved


mhluongo:
We need fallback for a couple environment variables like `ALCHEMY_KEY`.

See eg https://github.com/tallycash/extension/pull/786

PR #828, opened 2022-01-13
Show version, git commit and date in settings page

### LINKED ISSUE #806: Include the manifest version and git commit hash in the settings page
This is a small change that would make bug-tracking and support much easier.

cc @VladUXUI @mr-michael @Shadowfiend

cairomassimo:
Fixes #806.

NB: version is taken from `git describe`, not from manifest.

Screenshot from 2022-01-13 19-17-22

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "@tallyho/tally-ui": "0.0.1",
     "@tallyho/window-provider": "0.0.1",
     "buffer": "^6.0.3",
+    "git-revision-webpack-plugin": "^5.0.0",
     "react": "^17.0.2",
     "webext-options-sync": "^2.0.1",
     "webextension-polyfill": "^0.7.0"
--- a/ui/pages/Menu.tsx
+++ b/ui/pages/Menu.tsx
@@ -102,6 +102,10 @@ export default function Menu(): ReactElement {
           >
             Give feedback!
           </SharedButton>
+          <div className="version">
+            Version: {process.env.VERSION ?? `<unknown>`} (
+            {process.env.GIT_COMMIT_DATE?.slice(0, "YYYY-MM-DD".length)})
+          </div>
         </div>
       </section>
       <style jsx>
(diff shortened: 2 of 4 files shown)


mhluongo (approved):
@Shadowfiend might have nits so I'll leave it to him to review and merge, but strong concept ACK!

mhluongo:
#793 raises its ugly head again :flushed:

henryboldi:
> #793 raises its ugly head again 😳

Oy vey 🤦

mhluongo:
Whelp, we now have an answer to the biggest process issues for contributors IMO. Checking CI locally then merging

mhluongo:
The version string could maybe use a little more margin, and click to copy would be really nice. Either way, merged! @VladUXUI I'll let you sound off on any visual tweaks

VladUXUI:
image
[Figma Link](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=5495%3A135410)

Added a bit of space and bigger size.

Quick questions here
1 - Is it useful to have "109-g9dc16dp" Bit of extra cognitive/visual load that is probably not that important to most?
2- Also, can we change the date format? because of global differences having the month abbreviated is good ux. 
Otherwise some users will read month as day and day as month depending on were you are from.

cairomassimo:
> Quick questions here 1 - Is it useful to have "109-g9dc16dp" Bit of extra cognitive/visual load that is probably not that important to most?

That second part only appears on unreleased/unofficial versions. Otherwise it's just `v0.10.1`.

> 2- Also, can we change the date format? because of global differences having the month abbreviated is good ux. Otherwise some users will read month as day and day as month depending on were you are from.

Makes sense. I used a technical format (YYYY-MM-DD), but the purpose is just to let the user know how recent it is, so it should be changed.

To do:
- [ ] Update UI
- [ ] Fix date format
- [ ] Change version command to `git describe --tags` (otherwise some versions are skipped)
- [ ] Ensure version is computed correctly in CI?

PR #844, opened 2022-01-18
Added LedgerService with address derivation and transaction signing and SiginingService skeleton

### LINKED ISSUE #838: Ledger service skeleton setup
- [ ] create Ledger service
- [ ] create Signing Service

### LINKED ISSUE #839: Implement webusb connection in the `LedgerService`
- [ ] connection handling
- [ ] get device info
- [ ] get addresses
- [ ] sign transaction
- [ ] sign typed data

### LINKED ISSUE #918: Ledger communication: initial setup with WebUSB

fernya:
First PR draft, don't expect ground-breaking stuff right now!

Closes #918, closes #838, closes #839


background/services/ledger/index.ts
```diff
@@ -0,0 +1,53 @@
...
+ * The main purpose for this service/layer is
+ *
+ * The responsibility of this service is 2 fold.
+ * - xxx
+ */
+export default class LedgerService extends BaseService<Events> {
+  static create: ServiceCreatorFunction<
+    Events,
+    LedgerService,
+    [] // we don't know our final dependencies
```
Shadowfiend, line 36: Fwiw—this is more likely to be a dependee than a dependent. Keyring service is a decent, though probably incomplete, guide for the API here.

package.json
```diff
@@ -69,6 +69,7 @@
     "@types/react-router-dom": "^5.3.1",
     "@types/remote-redux-devtools": "^0.5.5",
     "@types/terser-webpack-plugin": "^5.0.3",
+    "@types/w3c-web-usb": "^1.0.5",
```
greg-nagy, line 72: This should be in the background package.json I think.

greg-nagy, line 72: This should be in the background package.json I think.

greg-nagy (changes requested):

background/services/ledger/index.ts
```diff
@@ -42,24 +52,73 @@ type Events = ServiceLifecycleEvents & {
...
-    //   await this.db.getAllPermission()
-    // )
+    logger.info("LedgerService::internalStartService")
+
+    navigator.usb.addEventListener(
+      "connect",
+      async (event: USBConnectionEvent) => {
+        // how to make it removable?
+        if (
+          Object.keys(SupportedLedgerPids).includes(
```
fernya, line 88: Early return?

background/services/ledger/index.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+  connected: { id: string; type: LedgerType }
+  disconnected: { id: string; type: LedgerType }
+  address: { ledgerID: string; derivationPath: string; address: HexString }
+  signedTransaction: SignedEVMTransaction
+}
+
+const UnknownLedgerId = "unrecognizable"
+
+const idGeneratorPath = "44'/60'/0'/0/0"
```
mhluongo, line 44: We should probably call this derivation path to be consistent internally and w other wallets

background/services/ledger/index.ts
```diff
@@ -0,0 +1,185 @@
...
+      break
+    default:
+      extensionDeviceType = LedgerType.UNKNOWN
+  }
+
+  if (extensionDeviceType === LedgerType.UNKNOWN) {
+    return [UnknownLedgerId, extensionDeviceType]
+  }
+
+  const address = await deriveAddressOnLedger(idGeneratorPath, eth)
```
mhluongo, line 81: Ahh I see. No way to ID a device but by it's address derivation?

background/services/ledger/index.ts
fernya: Then it'll be idDerivationPath to distinguish it from other derivation paths.

background/services/ledger/index.ts
fernya: From the depths of the internet it seems so, there are people who inject custom uuid to the devices but that should be a big privacy no-no!

> Yes i got to know about this from ledger support team, there are certain security measures due to which unique identifier are not attached to these devices. however i needed unique identifier for some specific reasons in the project, so i stored uuid in the device via APDU commands and mapped the devices against that uuid. It is working fine for me as of now.

mhluongo: I certainly don't want to make fingerprinting easier for anyone else, just for consenting users. I assume that APDU approach is per device rather than per device / application? If the latter, the privacy concern might not be so bad. If the former... :grimacing: 

At the end of the day we only need a device that can sign for a particular account, but if we can tell people they're using the wrong device... or that the device has bee tampered with?... earlier in the process *without* chipping it so anyone who connects knows the user used Tally, this could be a big UX win.

background/services/ledger/index.ts
```diff
@@ -0,0 +1,334 @@
...
+  address: { ledgerID: string; derivationPath: string; address: HexString }
+  signedTransaction: SignedEVMTransaction
+}
+
+const UnknownLedgerId = "unrecognizable"
+
+export const idDerviationPath = "44'/60'/0'/0/0"
+
+async function requireAvailableLedger() {
+  const devs = await navigator.usb.getDevices()
```
greg-nagy, line 56: I also like short variable names, but let's use readable names, please :)

background/services/ledger/index.ts
```diff
@@ -0,0 +1,334 @@
...
+
+  if (devs.length === 0) {
+    throw new Error("No available USB devices to use!")
+  }
+}
+
+async function deriveAddressOnLedger(path: string, eth: Eth) {
+  const a = await eth.getAddress(path)
```
greg-nagy, line 64: old habbits die hard :)

greg-nagy (commented):
There are bits and pieces which will require some polish, but overall looks decent so far.

background/services/ledger/index.ts
greg-nagy: > At the end of the day we only need a device that can sign for a particular account, but if we can tell people they're using the wrong device

We will be able to do this using the first address as an id.

package.json
fernya: Done.

background/services/ledger/index.ts
fernya: Alrighty!

fernya:
Shall I make it non-draft or wait for Massimo's branch to get merged?

mhluongo:
> Shall I make it non-draft or wait for Massimo's branch to get merged?

If it's ready to review for merge into `main`, undraft away! The only extra caveat is that we keep `main` releasable, but AFAIK this won't impact a user until there's a frontend?

background/redux-slices/ledger.ts
```diff
@@ -138,36 +142,18 @@ export const { ledgerReset, addLedgerAccount } = ledgerSlice.actions
 
 export default ledgerSlice.reducer
 
-let fakeDeviceNonce = 0
-
-async function doConnectFake() {
-  await new Promise((resolve) => {
-    setTimeout(resolve, 500)
+async function doConnectReal() {
```
greg-nagy, line 145: You don't need this function. You can emmit in the async thunk
See [setNewSelectedAccount](https://github.com/tallycash/extension/blob/8fb4db8ed7836ae082dd4a80b6b53eac80f55330/background/redux-slices/ui.ts#L105) and `newSelectedAccount`

`emit()` will return a promise and you can wait for that. 

See the [wiring in main.ts](https://github.com/tallycash/extension/blob/d070f46b7bf109e08a75d057954169537b2196a4/background/main.ts#L842)

background/redux-slices/ledger.ts
```diff
@@ -138,36 +142,18 @@ export const { ledgerReset, addLedgerAccount } = ledgerSlice.actions
...
+  return new Promise<string>((resolve, reject) => {
+    emitter.emit("connectLedger", { resolve, reject })
   })
-
-  fakeDeviceNonce += 1
-  return `fake-device-id-${fakeDeviceNonce}`
 }
 
 async function doFetchAddressReal(path: string) {
```
greg-nagy, line 151: Same goes here as for the `doConnectReal()`

background/redux-slices/ledger.ts
```diff
@@ -189,7 +175,7 @@ async function doFetchBalanceFake(address: string) {
 export const connectLedger = createBackgroundAsyncThunk(
   "ledger/connectLedger",
   async (unused, { dispatch }) => {
-    const deviceID = await doConnectFake()
+    const deviceID = await doConnectReal()
```
greg-nagy, line 178: emit comes here

background/redux-slices/ledger.ts
```diff
@@ -202,7 +188,7 @@ export const fetchAddress = createBackgroundAsyncThunk(
     { dispatch }
   ) => {
     dispatch(ledgerSlice.actions.setFetchingAddress({ deviceID, path }))
-    const address = await doFetchAddressFake(path)
+    const address = await doFetchAddressReal(path)
```
greg-nagy, line 191: emit comes here

background/services/ledger/db.ts
```diff
@@ -0,0 +1,39 @@
...
+  ledgerId: string
+  address: HexString
+  path: string
+}
+
+export class LedgerDatabase extends Dexie {
+  private accounts!: Dexie.Table<LedgerAccount, number>
+
+  constructor() {
+    super("tally/ledgers")
```
greg-nagy, line 15: I would use `tally/ledger` or `tally/ledger-service` as a name.

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+import { ETH } from "../../constants"
+import { getEthereumNetwork } from "../../lib/utils"
+
+enum LedgerType {
+  UNKNOWN,
+  LEDGER_NANO_S,
+  LEDGER_NANO_X,
+}
+
+const LedgerTypeStr = Object.values(LedgerType)
```
greg-nagy, line 30: It's better not to use abbrevations

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+
+  return [address, extensionDeviceType]
+}
+
+function signatureToString(signature: {
+  v: number
+  s: string
+  r: string
+}): string {
+  let v: string | number = signature.v - 27
```
greg-nagy, line 111: This is some hardcore numerology :) 

Is it possible to do this in a way that is easier to understand? Or if not, then please add some comments here

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+ * connected Ledgers' derived identifiers and make show an unified interface
+ * to the most common operation (ie. signing)
+ * - xxx
+ */
+export default class LedgerService extends BaseService<Events> {
+  #currentLedgerId = UnknownLedgerId
+
+  static create: ServiceCreatorFunction<Events, LedgerService, []> =
+    async () => {
+      return new this(await getOrCreateDB(), undefined)
```
greg-nagy, line 135: You can remove the undefined from here and the constructor parameter list. 

The transport can a normal class property.

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+
+enum LedgerType {
+  UNKNOWN,
+  LEDGER_NANO_S,
+  LEDGER_NANO_X,
+}
+
+const LedgerTypeStr = Object.values(LedgerType)
+
+const admittedProductIds = [0x1015, 0x4015]
```
greg-nagy, line 32: What are these? I assume they are the ids for ledger nano s and nano x. If so please use a data structure that is easier to follow.

greg-nagy (changes requested):
Partial pass on the PR, but want to save state.

background/services/ledger/index.ts
fernya: Already fixed :)

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+
+    const eth = new Eth(this.transport)
+
+    const [id, type] = await generateLedgerId(this.transport, eth)
+
+    const ethVersion = (await eth.getAppConfiguration()).version
+
+    this.#currentLedgerId = `${LedgerTypeStr[type]}_${id}`
+
+    this.emitter.emit("connected", { id: this.#currentLedgerId, type })
```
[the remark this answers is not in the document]
greg-nagy: I think "saving seen ledgers and storing them" responsibility is missing from this function.

Or do you want to put that into the `SigningService`?

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+
+  async saveAddress(path: HexString, address: string): Promise<void> {
+    await this.db.addAccount({ ledgerId: this.#currentLedgerId, path, address })
+  }
+
+  async signTransaction(
+    address: HexString,
+    transactionRequest: EIP1559TransactionRequest
+  ): Promise<SignedEVMTransaction> {
+    requireAvailableLedger()
```
greg-nagy, line 251: This should also use the attached device and the created transport

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+
+    try {
+      if (!this.transport) {
+        throw new Error("Uninitialized transport!")
+      }
+
+      const ethersTx =
+        ethersTransactionRequestFromEIP1559TransactionRequest(
```
greg-nagy, line 259: What a gorgeous and easy to remember name ... :)

But this is why we have IntelliSense ...

background/services/ledger/db.ts
```diff
@@ -0,0 +1,39 @@
...
+    })
+  }
+
+  async addAccount(account: LedgerAccount): Promise<void> {
+    await this.accounts.add(account)
+  }
+
+  async getAccountByAddress(address: HexString): Promise<LedgerAccount | null> {
+    return (
+      (await this.accounts.where("address").equals(address).first()) ?? null
```
greg-nagy, line 28: Based on how this is used this should be just 

`return this.accounts.where("address").equals(address).first()`

It's generally a nice idea to use `null` to say explicitly `we haven't found anything` but only if it's used consistently as a common understanding through the codebase. Generally we use the dexie api in other parts of the codebase so that should be fine here too.

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+      }
+
+      const eth = new Eth(this.transport)
+      const signature = await eth.signTransaction(
+        accountData.path,
+        serializedTx,
+        null
+      )
+
+      const alteredSig = {
```
greg-nagy, line 281: Why do we need this?

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    throw new Error("Transaction signing is unsuccessful!")
+  }
+
+  async signTypedData(
+    typedData: EIP712TypedData,
+    account: HexString
+  ): Promise<string> {
+    requireAvailableLedger()
```
greg-nagy, line 349: Use the currently attached device, please.

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+      return signatureToString(signature)
+    } catch (error) {
+      throw new Error("Signing data failed")
+    }
+
+    throw new Error("Typed data signing is unsuccessful!")
+  }
+
+  async signMessage(address: string, message: string): Promise<string> {
+    requireAvailableLedger()
```
greg-nagy, line 385: Use the currently attached device, please.

background/services/ledger/index.ts
fernya: Fixed.

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+      const signature = await eth.signPersonalMessage(address, message)
+      this.emitter.emit("signedData", signatureToString(signature))
+      return signatureToString(signature)
+    } catch (error) {
+      throw new Error("Signing data failed")
+    }
+
+    throw new Error("Typed data signing is unsuccessful!")
+  }
+}
```
greg-nagy, line 403: I believe we will need a couple more signing methods, but not sure tbh. 

In the UI we have these:
```
  Sign = "sign",
  SignSwap = "sign-swap",
  SignSpend = "sign-spend",
  SignTransfer = "sign-transfer",
```

Not sure how these maps to ledger interactions though.

background/services/signing/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  ) {
+    super()
+  }
+
+  protected async internalStartService(): Promise<void> {
+    await super.internalStartService() // Not needed, but better to stick to the patterns
+  }
+
+  async deriveAddress(accountID: string): Promise<HexString> {
+    this.deriveAddress = this.deriveAddress.bind(this)
```
greg-nagy, line 41: I think you won't need these binds. But this seems to be placeholder code anyway.

greg-nagy (changes requested):
Good amount of feedback, but want to say that this is a huge lift!

This is good work, but has the initial ramp up patterns.

background/main.ts
```diff
@@ -357,6 +376,8 @@ export default class Main extends BaseService<never> {
       this.nameService.startService(),
       this.internalEthereumProviderService.startService(),
       this.providerBridgeService.startService(),
+      this.ledgerService.startService(),
```
greg-nagy, line 379: Please put the initialization of these services and all their usage in the `main.ts` behind the `HIDE_IMPORT_LEDGER` feature flag from the environment variables. 

* service create
* Main class 
  * create dependency list
  * constructor
* internal start
* internal stop 
* initializeRedux

This is uncharted territory, so if you run into any roadblocks please let me know.

greg-nagy (changes requested):
Put the draft service initialization logic behind a feature flag please.

background/services/ledger/index.ts
fernya: Done.

background/services/ledger/index.ts
fernya: Restructured the whole thing into a more descriptive data structure (LedgerProductDatabase) and a tester function (TestedProductId) to decide the same thing.

background/main.ts
```diff
@@ -587,6 +611,36 @@ export default class Main extends BaseService<never> {
     )
   }
 
+  async connectLedgerService(): Promise<void> {
+    ledgerSliceEmitter.on("importLedgerAccounts", async (accounts) => {
+      for (let i = 0; i < accounts.length; i += 1) {
```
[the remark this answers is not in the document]
cairomassimo: > You could use `for (const {path, address} of accounts) { ...`

I wanted to at some point, but TSC complains about missing `--downlevelIteration` option. Should we set it? I'm not 100% sure about its implications though...

background/redux-slices/ledger.ts
cairomassimo: Agree. And it can be made even simpler once/if #911 is merged ;)

background/redux-slices/ledger.ts
cairomassimo: I don't think we can do it here, though. We need a result (the address) and `emit()` returns `Promise<void>`. (I assume. How could it know from _which_ listener to get the result? What if there are none? What determines the return type?) Definitely simpler via #911

background/services/ledger/db.ts
fernya: Done.

background/main.ts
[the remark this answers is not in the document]
greg-nagy: I think we should not need to set `--downlevelIteration` https://www.typescriptlang.org/tsconfig#downlevelIteration, because our target is esnext and that plays a role only when compiling to ES5.

I get some really weird eslint error when I try to use it `iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations.eslintno-restricted-syntax`

Have no idea what's the cause of this. I assume some eslint inherited rules. Tbh I would not touch it as the eslint rules are shared through thesis. Does not really worth the effort.

background/redux-slices/ledger.ts
fernya: Tbh I can't tell the difference between this and the one made on Line 151. We need the return value at both stuff (ledgerId and address respectively).

background/redux-slices/ledger.ts
fernya: Tbh I can't tell the difference between this and the one made on Line 145. We need the return value at both stuff (ledgerId and address respectively).

background/services/signing/index.ts
fernya: Yes, this is needed only for successfully commit the placeholders. Otherwise linter complains that the member never use 'this'.

background/main.ts
fernya: Done.

background/redux-slices/ledger.ts
fernya: I'll skip this in this round.

background/redux-slices/ledger.ts
fernya: I'll skip this in this round.

background/redux-slices/ledger.ts
fernya: I'll skip this right now.

background/redux-slices/ledger.ts
fernya: I'll skip this right now.

background/main.ts
```diff
@@ -323,7 +332,17 @@ export default class Main extends BaseService<never> {
      * the communication coming from dApps according to EIP-1193 and some tribal
      * knowledge.
      */
-    private providerBridgeService: ProviderBridgeService
+    private providerBridgeService: ProviderBridgeService,
+
+    /**
+     * A promise TODO
```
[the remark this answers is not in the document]
fernya: Done :)

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+      metadata: { ethereumVersion: ethVersion },
+    })
+  }
+
+  async #onUSBConnect(event: USBConnectionEvent): Promise<void> {
+    if (!admittedProductIds.includes(event.device.productId)) {
+      return
+    }
+
+    this.onConnection(event.device.productId)
```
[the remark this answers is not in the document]
fernya: We may get the productId of the connected devices from two different places.
1) navigator.usb "connect" event -> when the device was already onboarded and connected to the computer - it comes from the event descriptor
2) prepopulated list of paired devices -> when the device is newly onboarded - it comes the first paired device received from navigator.usb.getDevices()

It seems logical to me to create a wrapped function to the one that takes the productId as number and does the heavy-lifting, and wrap it with functions which extract the productId from the appropriate place.

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+      type: LedgerType.LEDGER_NANO_S,
+    })
+
+    this.#currentLedgerId = UnknownLedgerId
+  }
+
+  protected async internalStartService(): Promise<void> {
+    await super.internalStartService() // Not needed, but better to stick to the patterns
+
+    navigator.usb.addEventListener("connect", this.#onUSBConnect)
```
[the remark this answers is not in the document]
fernya: :)

background/services/ledger/db.ts
```diff
@@ -0,0 +1,36 @@
+import Dexie from "dexie"
+import { HexString } from "../../types"
+
+export interface LedgerAccount {
+  ledgerId: string
```
cairomassimo, line 5: The `ID` part is usually all-caps in the rest of the codebase, so I think we should be consistent here.

Also, I used the name `deviceID` for this, which I think is a bit clearer.

background/services/signing/index.ts
```diff
@@ -0,0 +1,71 @@
...
+type Events = ServiceLifecycleEvents
+
+/**
+ * The SigningService is responsible for
+ *
+ * The main purpose for this service/layer is
+ *
+ * The responsibility of this service is 2 fold.
+ * - xxx
+ */
```
cairomassimo, line 19: Stale comment.

background/services/signing/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    ledgerService: LedgerService
+  ) {
+    super()
+  }
+
+  protected async internalStartService(): Promise<void> {
+    await super.internalStartService() // Not needed, but better to stick to the patterns
+  }
+
+  async deriveAddress(accountID: string): Promise<HexString> {
```
cairomassimo, line 40: I don't understand the purpose of this method. I assume it was in the RFC, but I'd rather not add it until it's implemented (even better if we don't add it until it's actually used for something).

It's not clear to me the meaning of `accountID` (e.g., for ledgers, an account is created from a pair `deviceID,derivationPath`)

background/services/ledger/index.ts
```diff
@@ -0,0 +1,403 @@
...
+  async connectLedger(): Promise<string> {
+    const devArray = await navigator.usb.getDevices()
+
+    this.onConnection(devArray[0].productId)
+
+    return this.#currentLedgerId
+  }
+
+  async deriveAddress(accountID: string): Promise<HexString> {
+    requireAvailableLedger()
```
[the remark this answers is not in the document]
fernya: A'ight, changed logic in all relevant places!

background/services/signing/index.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    throw new Error("Unimplemented")
+  }
+
+  async signMessage(address: string, message: string): Promise<string> {
+    this.signMessage = this.signMessage.bind(this)
+
+    throw new Error("Unimplemented")
+  }
+}
```
cairomassimo, line 71: Is there an implementation of this?
It's ok for me if it's coming in a different PR, but even better if we do it here

cairomassimo (commented):
Some minor comments :)

background/services/ledger/index.ts
fernya: I think these could be mapped to these three in LedgerService.

background/services/ledger/index.ts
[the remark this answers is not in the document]
fernya: Saving seen ledgers (LedgerService::saveAddress) is triggered by the UI and done in background/main.ts when importLedgerAccounts event is handled.

The distinction is not made here I'll add it.

background/services/ledger/index.ts
fernya: I had to convert the received signature format to the one which is compatible with the serialization functions ('v' field is string in the former and number in the latter).

I don't know any better than this.

background/services/ledger/index.ts
fernya: Built on current foundations! :)

background/services/signing/index.ts
fernya: No, it's a skeleton as yet (in this PR)

greg-nagy (approved):
This should be good to go 👷‍♂️

PR #846, opened 2022-01-18
Add signing typed data to keyrings

wojciech-turek:
Adds a keyring function that will allow signing typed data, the request will come from the ui through main.

### DIFF
--- a/background/services/keyring/index.ts
+++ b/background/services/keyring/index.ts
@@ -11,7 +11,7 @@ import {
   encryptVault,
   SaltedKey,
 } from "./encryption"
-import { HexString, KeyringTypes, UNIXTime } from "../../types"
+import { HexString, KeyringTypes, TypedData, UNIXTime } from "../../types"
 import { EIP1559TransactionRequest, SignedEVMTransaction } from "../../networks"
 import BaseService from "../base"
 import { ETH, MINUTE } from "../../constants"
@@ -32,6 +32,7 @@ interface Events extends ServiceLifecycleEvents {
   address: string
   // TODO message was signed
   signedTx: SignedEVMTransaction
+  signedData: string
 }
 
 /*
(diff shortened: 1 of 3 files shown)


background/services/keyring/index.ts
```diff
@@ -331,6 +332,21 @@ export default class KeyringService extends BaseService<Events> {
     return newAddress
   }
 
+  /**
+   * Find keyring associated with an account.
+   *
+   * @param account - the account desired to search the keyring for.
+   */
+  findKeyring = async (account: HexString): Promise<HDKeyring> => {
```
mhluongo, line 340: Two things here

1. Why are we avoiding normal async method syntax?
2. This should be private using `#` We don't want key material to leave the service outside a very tightly defined API.

mhluongo (changes requested):

background/services/keyring/index.ts
Shadowfiend: To be a little more precise here, this should just be:

```suggestion
  async #findKeyring(account: HexString): Promise<HDKeyring> {
```

background/services/keyring/index.ts
```diff
@@ -397,6 +408,27 @@ export default class KeyringService extends BaseService<Events> {
     return signedTx
   }
 
+  async signTypedData({
```
Shadowfiend, line 411: Let's give this some docs, particularly calling out the use of the latest `signTypedData` version defined in EIP-712, and link to https://eips.ethereum.org/EIPS/eip-712 .

background/types.ts
```diff
@@ -50,3 +52,17 @@ export enum KeyringTypes {
...
 }
+
+export type EIP712DomainType = {
+  name?: string
+  version?: string
+  chainId?: number
+  verifyingContract?: HexString
+}
+
+export type TypedData = {
```
Shadowfiend, line 63: Let's call this EIP712TypedData. There's a lot of room for other typed data across chains and such, so we should try to be specific if possible.

background/services/keyring/index.ts
wojciech-turek: Thanks, I instinctively write arrow functions, changed it to be consistent with other function declarations as suggested.

Shadowfiend (approved):
Looks good, let's roll :guitar:

PR #847, opened 2022-01-19
🔄 Swap UI tasks

### LINKED ISSUE #829: Swap UI Tasks
There are a few UI tweaks that would help clean up the swap flow tremendously! 

## Insufficient balance warning always displayed when selecting swap assets
insufficient-balance-swap

## Search / filtering tokens
The filter doesn't actually change which assets are displayed.
swap-search

## Icons should match exchange route
Right now the exchange route only has a single hardcoded Uniswap icon. Fortunately 99% of the time our swaps are going through uniswap!
uniswap-icon

## Loading indicator when waiting for quote
It takes a couple seconds to load the quote from the 0x API so it would be nice if the "Get final quote" button could display a loading state.
get-final-quote

## Loading indicator for asset amounts
We need to refactor the swap redux slice to use the quote API for everything instead of the now deprecated prices API. However this will add a couple seconds of loading time whenever a user changes the amount they want to swap. It would be nice if we could display a loading indicator where the asset amount input field is.

henryboldi:
These basically closes #829 aside from parts which will require a more significant refactor, like "We need to refactor the swap redux slice to use the quote API"

### DIFF
--- a/ui/components/Shared/SharedAssetInput.tsx
+++ b/ui/components/Shared/SharedAssetInput.tsx
@@ -15,6 +15,7 @@ function SelectAssetMenuContent(
   props: SelectAssetMenuContentProps
 ): ReactElement {
   const { setSelectedAssetAndClose, assets } = props
+  const [searchTerm, setSearchTerm] = useState("")
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -136,7 +142,7 @@ interface SharedAssetInputProps {
   label: string
   defaultAsset: Asset
   amount: string
-  maxBalance: number
+  maxBalance: number | boolean
```
Shadowfiend, line 145: Not blocking: can I suggest we instead make `maxBalance` optional? Having a number or a boolean as a type is rather confusing, especially since the intent is for the boolean to only ever be `false`.

ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -187,7 +193,8 @@ export default function SharedAssetInput(
 
   const getErrorMessage = (givenAmount: string): string | undefined => {
     return (!isTypeDestination && maxBalance >= Number(givenAmount)) ||
```
Shadowfiend, line 195: Also not blocking: similarly, it seems like the code flow here is “if this is not a type destination, check the max balance”. It feels like it would be cleaner to separate “should we check max balance” (~`const shouldCheckMaxBalance = typeof maxBalance !== "undefined" && !isTypeDestination`) from “is there sufficient balance” (`shouldCheckMaxBalance && maxBalance > givenAmount ...`).

Finally, want to remind that we're in the usual pickle of parsing an amount that could be unrepresentable by `Number` as a `Number`; we should be using [`parseToFixedPointNumber`](https://github.com/tallycash/extension/blob/main/background/lib/fixed-point.ts#L139-L151) and accepting a [`FixedPointNumber`](https://github.com/tallycash/extension/blob/main/background/lib/fixed-point.ts#L1-L21) as the `maxBalance` value, though we can continue to punt that for now.

Shadowfiend (approved):
Changes look good here. Rather than testing this PR separately, I'm just going to land it on the main swap branch so we can QA the whole thing.

ui/components/Shared/SharedAssetInput.tsx
henryboldi: Good point. Actually, `maxBalance` is replaced in #856

PR #856, opened 2022-01-20
💞 Make max balance reusable in shared asset input

henryboldi:
Previously the parent page, like Swap or Send, would have to handle its own max balance control. Now the max balance control can just be flipped on whenever it's needed.


mhluongo:
Are you still wanting to partially fix #858 here @henryboldi?

henryboldi:
@mhluongo Yes!

Shadowfiend (commented):
Hmmm... I'm a little wary of this change, and generally of using selectors in shared components. These *should* be fed by the caller, because imo they should be maximally flexible and their performance should be a direct function of their inputs. 

Another approach that might work here is to accept an `AssetAmount` array as an alternative to an `AnyAsset` array and, when the array has `AssetAmount`s, the amount is used as a max (if the max control is enabled). This would make the input instantly usable with balances, but avoid the need for doing additional store lookups. 

Thoughts?

henryboldi:
Nice idea, I'll try that! Thank you @Shadowfiend

henryboldi:
I will revisit this idea after the dust settles on swap

Shadowfiend:
Quick prototype on the typing front here:

```

type AnyAssetWithOptionalAmount<T extends AnyAsset> = { asset: T; amount?: bigint }

type SharedAssetInputProps<
  AssetType extends AnyAsset,
  AssetAmountType extends AnyAssetAmount<AssetType> = never
> = {
  isTypeDestination: boolean
  label: string
  defaultAsset: AssetType
  amount: string
  maxBalance: number | boolean
  isAssetOptionsLocked: boolean
  disableDropdown: boolean
  isDisabled?: boolean
  onAssetSelect: (asset: AssetType) => void
  onAmountChange: (value: string, errorMessage: string | undefined) => void
  onSendToAddressChange: (value: string) => void
} & ({
  assets: AssetType[]
    }
  | { assetAmounts: AssetAmountType[] }
)

export default function SharedAssetInput<
  AssetType extends AnyAsset,
  AssetAmountType extends AnyAssetAmount<AssetType> = never
>(
  props: SharedAssetInputProps<AssetType, AssetAmountType>
): ReactElement {
  const {
    isTypeDestination,
    label,
    defaultAsset,
    amount,
    maxBalance,
    isAssetOptionsLocked,
    disableDropdown,
    isDisabled,
    onAssetSelect,
    onAmountChange,
    onSendToAddressChange,
  } = props

  const assetAmounts: AnyAssetWithOptionalAmount<AssetType>[] = "assets" in props ? props.assets.map((asset) => ({ asset })) : props.assetAmounts
```

`AnyAssetWithOptionalAmount` can then be used in helpers like `SelectAssetMenuContent` instead of a bare `AnyAsset` subtyp, while preserving the `AnyAsset` subtype throughout.

Issue #860, opened 2022-01-20
Can't approve token for swaps


mhluongo:
Just tried to swap some KEEP to renZEC. Results below

![Screenshot from 2022-01-20 18-08-06](https://user-images.githubusercontent.com/427505/150436938-11614183-b2e4-41d9-a780-ceb147454737.png)
![Screenshot from 2022-01-20 18-09-37](https://user-images.githubusercontent.com/427505/150436949-7a6f66cf-a1f3-430c-bd48-0b7b41e9735d.png)
![Screenshot from 2022-01-20 18-10-05](https://user-images.githubusercontent.com/427505/150436966-069a0d01-beef-48fb-ad48-4aa1cd2d0d00.png)

While I couldn't get a screenshot fast enough, it appears the "Swap assets" page flashes at the bottom before the extension blanks. There aren't any obviously relevant errors in the frontend or background console, which surprised me tbh

Tested against `swaps-qa-build` on Chromium 97.0.4692.71 / Ubuntu

0xDaedalus:
Interesting - uniswap v3 swaps seem to be fine - but uniswap v2 result in a blank screen

0xDaedalus:
Digging into ethers.js I see that https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/json-rpc-provider.ts#L220 always throws an error - are we using the wrong function here?

Shadowfiend:
Not the wrong function… But one that is unexpectedly not supported by default.

PR #861, opened 2022-01-21
Use EnrichmentService To Derive Additional Data About Tx Requests

### LINKED ISSUE #848: Use EnrichmentService to derive additional data about transaction requests
In #760, we special-case checking for token approval to determine whether to display a more specific signing page with limit adjustments. This special-casing misses a few more specific situations, like `approveAndCall`, and generally isn't a good path forward for any other custom signing pages (for example, richer support for swap signing on various DEXes).

Instead, a good path forward is to build on the already-budding [`EnrichmentService`](https://github.com/tallycash/extension/blob/main/background/services/enrichment/index.ts), which is designed to enrich pending and confirmed transactions with additional metadata derived from inspecting the transaction. Currently, this service [watches for `ChainService` `transaction` events](https://github.com/tallycash/extension/blob/84778351bf1936e68aa6923e87e9bf23cc25f80a/background/services/enrichment/index.ts#L130-L135), [attempts to derive additional metadata about the transaction](https://github.com/tallycash/extension/blob/84778351bf1936e68aa6923e87e9bf23cc25f80a/background/services/enrichment/index.ts#L138-L246), and [emits `enrichedEVMTransaction` with an additional `annotation` field that carries any additional derived metadata](https://github.com/tallycash/extension/blob/84778351bf1936e68aa6923e87e9bf23cc25f80a/background/services/enrichment/index.ts#L252-L260).

To scalably look up the kind of signing metadata token approvals currently use, the `EnrichmentService` should start watching the `InternalEthereumProvider`'s emitter for `transactionSignatureRequest`, attempt to derive an annotation in the same way that transactions do[^1], and emit an `enrichedEVMTransactionSignatureRequest` event. In turn, `Main` should rewire its [`transactionSignatureRequest`](https://github.com/tallycash/extension/blob/main/background/main.ts#L637-L671) event handler to instead trigger on `enrichedEVMTransactionSignatureRequest` from the `EnrichmentService`. Finally, the `Sign` page can use the `annotation` data to tailor the rendered page as relevant.

One additional feature that the sign page providers that isn't relevant to pending and confirmed transactions is the ability to customize the transaction's inputs. This issue doesn't cover refactoring that component; instead, the sign page should continue to manage this in a way custom to its particular implementation (for example, for the approval page, directly interacting with the ERC20 ABI to rewrite the approval transaction data based on edited form inputs).

[^1]: One notable difference is in the type of partial transaction requests vs full transactions. In particular, transactions can have `null`, `0x`, or data for their `input` field, while partial transaction requests can additional have an `undefined` `input` field.

0xDaedalus:
This PR adds functionality to enrichment service that attempts to add information about the assets included in a transaction as well as sender & receiver addresses.  Enrichment Service now listens to the `transactionSignatureRequest` event and emits an `enrichedEVMTransactionSignatureRequest` event which passes the transaction along with any available additional data to the transaction-construction redux store.


0xDaedalus:
Cannot do a full e2e test until #860 is resolved

background/services/enrichment/index.ts
```diff
@@ -137,7 +154,11 @@ export default class EnrichmentService extends BaseService<Events> {
         timestamp: Date.now(),
         type: "contract-deployment",
       }
-    } else if (transaction.input === null || transaction.input === "0x") {
+    } else if (
+      transaction.input === null ||
+      transaction.input === "0x" ||
+      transaction.input === undefined
```
mhluongo, line 160: You probably want `typeof(tx.input) === "undefined"` here

0xDaedalus:
Turns out the blocking bug was in my code - unrelated to #860 - although manifesting in a very similar way.

0xDaedalus:
https://user-images.githubusercontent.com/94649004/150457040-d77cfd62-532a-40b7-876e-1ec610fd3080.mov

background/redux-slices/transaction-construction.ts
```diff
@@ -38,7 +44,10 @@ export enum NetworkFeeTypeChosen {
 }
 export type TransactionConstruction = {
   status: TransactionConstructionStatus
-  transactionRequest?: EIP1559TransactionRequest
+  // @TODO Check if this can still be both types
```
0xDaedalus, line 47: not sure of the best way to check this but I assume that with this PR all transactions will now go through `EnrichmentService` yes?

mhluongo:
Initial nits:

1. Would be nice to have a real PR description as that's what ends up in the merge commit.
2. Commit message discipline is low. Not worth blocking merge, but want to make sure the rationale for changes is documented. @Shadowfiend takes this to the most extreme, and it's wonderful — here's his commit history from [the recent nonces PR](https://github.com/tallycash/extension/pull/822/commits). You can read it as a development log.

Screenshot from 2022-01-21 16-39-41

mhluongo:
@wojciech-turek I'm reviewing rn, but would like to give you the final :+1: to merge as you were just working on this.

mhluongo (approved):
Concept ACK, and works great in a local test. Leaving @wojciech-turek to take this one home

mhluongo:
Oh! We also need signed commits to merge @0xDaedalus

wojciech-turek (approved):
This looks good and works very well when we're interacting with dapps. One question I have is, should we enrich the transactions details when we build the transaction ourselves? Currently when we go through Send => Sign the transaction is not enriched, I guess we can work around it by passing necessary data from Send page. What do you think @Shadowfiend ?

0xDaedalus:
> Initial nits:
> 
> 1. Would be nice to have a real PR description as that's what ends up in the merge commit.
> 2. No need to prefix the branch name with the issue number. We try to keep branches shorter lived, deleting them after merge. Ideally you won't have more local branches at a given time than the number of people on the team.
> 3. [Commit message discipline is low](https://cbea.ms/git-commit/). Not worth blocking merge, but want to make sure the rationale for changes is documented. @Shadowfiend takes this to the most extreme, and it's wonderful — here's his commit history from [the recent nonces PR](https://github.com/tallycash/extension/pull/822/commits). You can read it as a development log.

I love nits!  Thank you!  I'm coming from a system of 1 commit per PR, and maximum 60 chars per commit message so please bear with me as I work on shaking that awful pattern.  I did not that PR description goes into the merge commit - will edit this one and make sure to have descriptive ones going forard.

0xDaedalus:
Signed and described ✅

Shadowfiend:
> maximum 60 chars per commit message

I literally died, sorry folks you're doing the rest of this without me ☠️

Shadowfiend:
> One question I have is, should we enrich the transactions details when we build the transaction ourselves? Currently when we go through Send => Sign the transaction is not enriched

We should! I don't think it needs to block here, but let's file an issue. ERC-20 transfers will depend on it, since it's what will populate the sign page more correctly; however, since all sends are ETH sends right now the base signature page is actually the right thing to show.

wojciech-turek (approved):
Looks good, leaving the internal transaction construction to go through the enrichment service for next iteration.

Shadowfiend:
> This looks good and works very well when we're interacting with dapps.

Btw the reason this is the case is that we're not actually using these annotations anywhere yet!

We need to start piping better typing + annotation info through the sign subcomponents. I'm probably going to tackle some of this in the swaps work.

PR #873, opened 2022-01-21
Separation of signing

wojciech-turek:
Since signing transactions and signing data have only few common points it made sense to separate them for the sake of better readability and in case those components need to be expanded they will be able to grow independently.

New route was added under `/signData`.


background/services/provider-bridge/index.ts
```diff
@@ -267,6 +267,23 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
+    params: unknown[],
+    popupPromise: Promise<browser.Windows.Window>
+  ): Promise<unknown> {
+    const response = await this.internalEthereumProviderService
+      .routeSafeRPCRequest(method, params)
+      .finally(async () => {
+        // Close the popup once we're done submitting.
+        const popup = await popupPromise
+        if (typeof popup.id !== "undefined") {
+          browser.windows.remove(popup.id)
```
mhluongo, line 281: Aside: It feels like we should have a better way to manage popups by now :thinking:

background/services/provider-bridge/index.ts
```diff
@@ -278,34 +295,42 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
+          // We are monsters and aren't breaking a method out quite yet.
+          // eslint-disable-next-line no-case-declarations
+          const transactionRequest = params[0] as EthersTransactionRequest
+          // eslint-disable-next-line no-case-declarations
+          const signTransactionPopupPromise =
+            ProviderBridgeService.showExtensionPopup(
+              AllowedQueryParamPage.signTransaction
+            )
+
+          if (transactionRequest.from === enablingPermission.accountAddress) {
```
mhluongo, line 328: You should use `normalizeEVMAddress` here

provider-bridge-shared/runtime-typechecks.ts
```diff
@@ -64,6 +64,7 @@ export function isPortResponseEvent(arg: unknown): arg is PortResponseEvent {
 export const AllowedQueryParamPage = {
   signTransaction: "/signTransaction",
   dappPermission: "/dapp-permission",
+  signData: "/signData",
```
mhluongo, line 67: WhatTypeOf-casing_areWeUsingHERE?

mhluongo (commented):
The background wiring stuff looks good to merge, but I'm not familiar enough with routing to opine there. @henryboldi @Shadowfiend WDYT?

background/services/provider-bridge/index.ts
wojciech-turek: I will take it a step further and use `sameEVMAddress`, good callout.

Shadowfiend (approved):
Looks good to me high-level. Can't test drive to make sure signing still works today, but if someone can confirm let's merge.

greg-nagy (approved):
This works and the code looks good. Haven't gone too deep though. 

Let's merge this 🙈

PR #874, opened 2022-01-21
No intent to merge: Swaps qa build

Shadowfiend:
Big bundle of changes here. Big pieces:
- Much better asset deduplication. Requires a reset of the assets slice, which is NOT covered in this PR.
- Swap numbers based on pairwise quotes. Updates numbers when assets or amounts are changed. Disables an asset input when its value is being loaded as the other side of a swap quote; currently does not show loading feedback beyond disabling the input.
- Updated approval flow. Likely buggy.
- RFQ-T on the final swap. Doesn't have an affiliate address set yet.

This rewrites a lot of the redux and component aspects of swaps as I worked to make state management simple so we can get this out. Likely some state management needs to move back to redux, but deciding what goes where was out of scope.

Other tidbits:
- Shared asset input has a full disabled state, used during loading.
- Disabling things in shared asset input done HTML-natively rather than with pointer event shutdown.
- Shared asset input allows searching by contract address as suggested by the placeholder.
- When the shared asset input is brought up, the search field is auto-focused for text input. Note that we're still missing auto-selection of single matches as well as dismiss-with-enter and other keyboard navigation tidbits.
- `SharedSlideUpMenu` now only renders its children if it is being brought up.
- Single-asset page swap link points to a version of the swap page with an unchangeable sell asset, similar to how the single-asset page's send link points to a version of the send page with an unchangeable send asset.


Shadowfiend:
Since swaps are now enabled by default on `main` (#967), this branch is no longer needed!

PR #732, opened 2021-12-20
🏞️ Rovin' Kovan

mhluongo:
Moving Kovan testnet support to be accessible via the network selector. This implies making a ton of code multi-network aware, though still short of production support as in #5 or #7.

- [x] Support multiple networks in `ChainService` 
- [x] Support multiple networks in `IndexingService`
- [x] Support multiple networks in `KeyringService` 
- [x] Kovan-specific gas price estimates
- [x] Allow new networks to be added for tracking via Redux
- [ ] Track down and nuke all instances of `getEthereumNetwork()` in the background script
- [x] Connect the network selector to Redux
  - [x] Track all accounts by `AddressNetwork` rather than just address
  - [ ] Switch the dApp provider network (#733)

What this PR *won't* do....
* Add all accounts to track on new network whenever network is changed. This was originally a goal, but in retrospect we only want to track accounts that we've seen the user visit on a particular network to save resources
* Include any sort of multi-network asset or asset transfer/historical transaction support.
* Update token lists to track tokens on Kovan
* Pull prices for assets on other networks
* #776

The user-facing bits here are feature-flagged.

PR #878, opened 2022-01-23
Asset-ic Acid: Improve asset deduplication and merging

Shadowfiend:
This is the second PR extracted from the swaps flow rework, again
rebased against main for holistic review. This one is focused on
a cleaner set of available assets, which feeds the swap buy list
in the new quote-based flow.

The outcome still isn't perfect here, and to update existing installs
will require clearing and repopulating the asset store from
`getCachedAssets`. Nonetheless, this should be a large improvement,
removing a whole swath of duplicates and irrelevant assets that were
slipping through before.

Also packaged here are a few type-level tidbits that are useful
external to the assets pipeline.

------

One thing to look at carefully: right now the new `mergeAssets` helper
merges assets by symbol if a contract address is unavailable. This
may not be the correct behavior, so please check me if that's the case
and we can drop that.

### DIFF
--- a/background/redux-slices/assets.ts
+++ b/background/redux-slices/assets.ts
@@ -1,5 +1,6 @@
 import { createSelector, createSlice } from "@reduxjs/toolkit"
 import { AnyAsset, PricePoint } from "../assets"
+import { normalizeEVMAddress } from "../lib/utils"
 
 type SingleAssetState = AnyAsset & {
   prices: PricePoint[]
@@ -162,7 +163,8 @@ const assetsSlice = createSlice({
                 "homeNetwork" in a &&
                 "contractAddress" in a &&
                 a.homeNetwork.name === asset.homeNetwork.name &&
-                a.contractAddress === asset.contractAddress) ||
+                normalizeEVMAddress(a.contractAddress) ===
+                  normalizeEVMAddress(asset.contractAddress)) ||
               asset.name === a.name
           )
           // if there aren't duplicates, add the asset
--- a/background/redux-slices/utils.ts
+++ b/background/redux-slices/utils.ts
@@ -151,3 +151,17 @@ export function createBackgroundAsyncThunk<
 
   return webextActionCreator
 }
+
+/**
+ * Utility type to extract the fulfillment type of an async thunk. Useful when
+ * wanting to declare something as "the type that this thunk will return once
+ * it completes".
+ */
+export type AsyncThunkFulfillmentType<T> = T extends Pick<
+  // We don't really need the other two inferred values.
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  AsyncThunk<infer Returned, infer _1, infer _2>,
+  "fulfilled"
+>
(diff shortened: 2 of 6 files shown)


background/redux-slices/utils.ts
```diff
@@ -119,7 +120,14 @@ export function createBackgroundAsyncThunk<
   // Use reduxtools' createAsyncThunk to build the infrastructure.
   const baseThunkActionCreator = createAsyncThunk(
     typePrefix,
-    payloadCreator,
+    async (...args: Parameters<typeof payloadCreator>) => {
+      try {
+        return await payloadCreator(...args)
+      } catch (error) {
+        logger.error(error)
+        throw error
```
mhluongo, line 128: It's the little things, ty

background/redux-slices/utils.ts
```diff
@@ -151,3 +151,17 @@ export function createBackgroundAsyncThunk<
 
   return webextActionCreator
 }
+
+/**
+ * Utility type to extract the fulfillment type of an async thunk. Useful when
+ * wanting to declare something as "the type that this thunk will return once
+ * it completes".
+ */
+export type AsyncThunkFulfillmentType<T> = T extends Pick<
```
mhluongo, line 160: Might need you to explain this black magic

background/services/indexing/index.ts
```diff
@@ -169,9 +169,10 @@ export default class IndexingService extends BaseService<Events> {
     const tokenListPrefs =
       await this.preferenceService.getTokenListPreferences()
     const tokenLists = await this.db.getLatestTokenLists(tokenListPrefs.urls)
+
     return baseAssets
       .concat(customAssets)
-      .concat(networkAssetsFromLists(tokenLists))
+      .concat(networkAssetsFromLists(getEthereumNetwork(), tokenLists))
```
mhluongo: Wish this were coming from an argument, it feels like we're working against #732 whenever we use `getEthereumNetwork()` deep in a service

background/redux-slices/assets.ts
```diff
@@ -162,7 +163,8 @@ const assetsSlice = createSlice({
                 "homeNetwork" in a &&
                 "contractAddress" in a &&
                 a.homeNetwork.name === asset.homeNetwork.name &&
-                a.contractAddress === asset.contractAddress) ||
+                normalizeEVMAddress(a.contractAddress) ===
```
mhluongo, line 166: :tada:

mhluongo (approved):
It feels like you rewrote some of the stuff in Redux, but... maybe that's okay, and it should be separate? Regardless though I think we're good to merge

mhluongo:
NB: there's no asset slice migration here, and I'm assuming that's going to happen later.

Shadowfiend:
> It feels like you rewrote some of the stuff in Redux, but... maybe that's okay, and it should be separate?

Yeah I think the actual merging should be, but I want to see about pulling the similarity score calculation up somewhere for shared use.

PR #880, opened 2022-01-24
Quoted Late: Rework swaps to use quotes for all pricing

Shadowfiend:
This is the last PR from extracted from the swaps flow rework, and the one
primarily focused on the swap flow itself. The commits here are in the
[owl-drawing](http://knowyourmeme.com/memes/how-to-draw-an-owl) category, unfortunately, due to the onion layers peeled back to
get all of this behaving the way it needed to with quoting in the mix.

High-level, the original way swaps worked is that when an asset was selected, an
API call was used to look up all prices between that asset and its respective
pairs. Then, any time the other asset in the pair was changed, the amount of
that asset was calculated using that price.

Two issues came up here:

- The prices that were being reported were strange, possibly due to pool size or
  other issues; e.g. the price for KEEP in USDC didn't really make sense. The
  final quote would provide a correct price, but the mass prices endpoint
  didn't work.
- 0x has deprecated their prices endpoint, because it's difficult to maintain on
  their end (no public reference discussion to link to here, just informal
  comms from their team).

Additionally, the existing swap implementation used the 0x `/tokens` endpoint to
pull token data that 0x supported. The issue there is that 0x has deprecated
that endpoint as well, as it is difficult to maintain a full list. Instead,
they recommend using a standard token list and checking for swap availability
by attempting to quote a swap.

As such, this PR reworks swaps to always fetch the pairwise quote for the
current pair and amount, and to not use the 0x token information for the UI. A
few architectural changes occurred along the way:

- The swap slice only stores the final quote; otherwise, quotes are fired
  through a background async thunk but the quote is simply returned to the UI.
- Token approval must occur *before* the final quote is requested. This ensures
  the time between the final quote is requested and when the associated
  transaction is signed and broadcast can be minimized, since quotes are not
  held forever by the 0x providers.
- Available tokens to buy are sourced from the extension's full asset list.

A couple of other tidbits were added as well:

- Swap quotes update in either direction; that is, if the buy asset or amount is
  changed by the user, the sell asset amount is updated to reflect a swap
  buying that amount of the buy asset. If the sell asset or amount is changed,
  the buy asset amount is updated to reflect a swap selling that amount of the
  sell asset.
- All known assets in the extension are listed as buy assets.
- RFQ-T is used (with `intentOnFilling` and `takerAddress` set) when a final
  quote is requested.

----

Note that a big chunk of related work was done to clean up asset data and tee up
some of the interaction requirements for the shared asset input; these are
targeted directly to main and captured in their own PRs at #875 and #878.


Shadowfiend:
Not as clean as I expected—for this code to build I'm going to have to merge main after the other PRs land due to some usage of the generics in there.

greg-nagy (dismissed):
Just a quick initial look, but this code is not compiling :( 
Seems to be typing issues.

Shadowfiend:
> for this code to build I'm going to have to merge main after the other PRs land due to some usage of the generics in there 

👆 that's what I meant with this but. The swaps QA branch has everything merged, but I'll fix this up now that the other PRs are in.

Shadowfiend:
Sorry for the rebase here… I really messed up the branch management on this one 🙈

background/lib/validate/jtd-validators.d.ts
```diff
@@ -31,3 +31,15 @@ export const isValidSwapPriceResponse: ValidateFunction<
...
+ * Helper type that can extract the concrete TypeScript type that a JTD
+ * validation function is checking for. This allows using and aliasing the
+ * JTD-inferred types without redefining them in TypeScript.
+ *
+ * Sample usage:
+ * ```
+ * type ZrxQuote = ValidatedType<typeof isValidSwapQuoteResponse>
+ * ```
+ */
+export type ValidatedType<T> = T extends ValidateFunction<infer V> ? V : never
```
greg-nagy, line 45: Oh, nice 🤩

I did not think about using JTD types directly like this, but makes total sense!

Note to self: check back on this, when I found the usage. Originally the theory was that validator functions are acting as a type guard and that should be sufficient. Let's confirm, deny or correct that theory.

background/redux-slices/0x-swap.ts
```diff
@@ -1,178 +1,68 @@
...
-        `sellAmount=${sellAmount}`
+    const assetContract = new ethers.Contract(
+      assetContractAddress,
+      ERC20_ABI,
+      signer
     )
+    const approvalTransactionData =
+      await assetContract.populateTransaction.approve(
+        approvalTarget,
+        ethers.constants.MaxUint256.sub(1) // infinite approval :(
```
greg-nagy: This reminds me of my electric engineering times and bit hacking...

background/lib/validate/jtd-validators.d.ts
greg-nagy: Answer to self: typing information is required for slice creation.

greg-nagy: Some explanation for the future: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html

background/redux-slices/0x-swap.ts
```diff
@@ -252,147 +128,97 @@ const swapSlice = createSlice({
...
+    const tradeAddress = await signer.getAddress()
+
+    const tradeAmount = utils.parseUnits(
+      "buyAmount" in amount ? amount.buyAmount : amount.sellAmount,
+      "buyAmount" in amount
+        ? assets.buyAsset.decimals
+        : assets.sellAsset.decimals
+    )
+
+    // When available, use smart contract addresses.
```
greg-nagy: Why is it better to use a contract address than a symbol?

ui/pages/Swap.tsx
```diff
@@ -1,197 +1,287 @@
...
+    if (changeSource === "sell") {
+      setBuyAmountLoading(true)
+    } else {
+      setSellAmountLoading(true)
+    }
+
+    latestQuoteRequest.current = quoteRequest
+    const { quote, needsApproval: quoteNeedsApproval } = ((await dispatch(
+      fetchSwapData(quoteRequest)
+    )) as unknown as AsyncThunkFulfillmentType<typeof fetchSwapData>) ?? {
```
greg-nagy: It would be good to somehow use generic for the return type here.

background/redux-slices/0x-swap.ts
```diff
@@ -217,8 +152,82 @@ export const fetchSwapData = createBackgroundAsyncThunk(
...
+
+    const assetContract = new ethers.Contract(
+      assetContractAddress,
+      ERC20_ABI,
+      signer
+    )
+    const approvalTransactionData =
+      await assetContract.populateTransaction.approve(
+        approvalTarget,
+        ethers.constants.MaxUint256 // infinite approval :(
```
greg-nagy, line 186: Why do we not need the -1 here?

background/redux-slices/0x-swap.ts
Shadowfiend: This is also wrong... And I'm nervous about it as I thought I fixed it. Going to give the diff a once-over to see if anything else got missed.

background/redux-slices/0x-swap.ts
Shadowfiend: Symbols are not necessarily unique; see https://github.com/tallycash/extension/pull/875/commits/f671cc8007787fb08fed00f586fd86d3ec7cbf5d for an example.

ui/pages/Swap.tsx
```diff
@@ -317,35 +346,45 @@ export default function Swap(): ReactElement {
...
-                  showLoadingOnClick={!confirmationMenu}
-                >
-                  Get final quote
-                </SharedButton>
-              )}
+              {
+                // Would welcome an alternative here---this is pretty gnarly.
+                // eslint-disable-next-line no-nested-ternary
+                needsApproval ? (
+                  isApprovalInProgress ? (
```
greg-nagy, line 353: The nicest solution would be creating a wrapper for `isApprovalInProgress` and use `needsApproval ? ... : ...`  inside that component. 

But if you just switch the order of the checks it makes the flow easier to reason about. 

But honestly, I think this is good as is. We can refactor it next time when we touch it, and find it hard to understand.

ui/pages/Swap.tsx
Shadowfiend: The issue here is that the `dispatch` function isn't getting augmented with the async thunk middleware return type. It's not entirely clear why that's the case, and I haven't had a chance to dig—it looks like redux toolkit's type structure *should* permit it. 

However, there's a world where `useBackgroundDispatch` integrates some magic to handle this.

ui/components/Swap/SwapQuote.tsx
```diff
@@ -20,7 +20,7 @@ type Props = {
   finalQuote: ZrxQuote
 }
 
-export default function SwapQoute({
+export default function SwapQuote({
```
greg-nagy, line 23: Mitsuketa! 

https://www.wordhippo.com/what-is/the-meaning-of/japanese-word-7e917c05cc585eb42cf78a32ca29c9b23ecf5443.html

:)

greg-nagy (approved):
The code looks good, but when I tried to test the functionality I got a "TypeError: Could not serialize message." or some async thunk "invalid decimal value error"

CleanShot 2022-01-25 at 14 24 58

I trust you that, you have tested this, and will go on an approve with the assumption that there must be some detail that I am not aware of and we are behind a feature flag. 

I also coult not get to the final swap page using the developer account. Couldn't find any pair that would work.

CleanShot 2022-01-25 at 14 21 19

ui/components/Swap/SwapQuote.tsx
```diff
@@ -1,60 +1,52 @@
...
-          ),
-          buyAmount: utils.formatUnits(
-            state.swap.quote.buyAmount,
-            state.swap.buyAsset?.decimals
-          ),
-          sources: state.swap.quote.sources.filter((source) => {
-            if (parseFloat(source.proportion) > 0) {
-              return true
-            }
+export default function SwapQoute({
```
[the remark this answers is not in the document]
greg-nagy: Fixed in a later commit.

background/redux-slices/0x-swap.ts
greg-nagy: You did, I just went commit by commit.

background/redux-slices/0x-swap.ts
greg-nagy: As you mentioned in a previous comment, this is how it should look now.

background/redux-slices/0x-swap.ts
Shadowfiend: Ah okay, thanks haha. I thought I was taking crazy pills :grin:

background/redux-slices/0x-swap.ts
```diff
@@ -195,6 +129,7 @@ export const fetchSwapData = createBackgroundAsyncThunk(
         isValidSwapQuoteResponse.errors
       )
 
+      logger.log("Booyak")
```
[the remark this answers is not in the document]
Shadowfiend: Nooooooo

Shadowfiend:
> when I tried to test the functionality I got a "TypeError: Could not serialize message." or some async thunk "invalid decimal value error"

This is me not providing testing instructions heh. You have to fill in an amount on one side of the swap for the swap to quote and work properly.

greg-nagy (approved):
I tried to test it, but the flow is always stuck for me. 

I have tried all the combination that I could think of but still nothing.
This is after a reintall of the extension.

I still see the async thunk error when I try to change the amount.

I leave this with an approve so because this is the end of the day for me and I don't want to block QA and this is behind a feature flag.


CleanShot 2022-01-25 at 19 57 31

Issue #843, opened 2022-01-18
Update derivation path #723 to Vlad's design


henryboldi:
#723

On the Onboarding page in UI - Dark Figma
<img width="807" alt="Derivation path" src="https://user-images.githubusercontent.com/1918798/149979890-36d38a5e-cfab-4dfe-b605-641b1b6bf919.png">

mhluongo:
@7alip if there's any chance you want to snipe this, I believe it's the last thing we need to unflag #723 now that #819 is merged! No worries if you don't have cycles, I just know it's going to take a bit otherwise

7alip:
Hey @mhluongo, It looks amazing! Of course I would like to do it with great pleasure.

7alip:
Is it possible to view the Figma design or maybe copy of the OnboardingPage?

VladUXUI:
Hello @7alip , i added the derivation path screens in this figma.
Do let me know if you have any questions about it, or if you need any other screens.
https://www.figma.com/file/EiXpBpNz2HcoeFJftgC5oc/?node-id=332%3A3772

mhluongo:
Closed by #884

PR #884, opened 2022-01-24
Update Custom Derivation Path Design

7alip:
tally dpath

Updated custom derivation path design based on #843 
May affect #893 since `SharedSelect` design has changed

- [ ] Save custom path added by user to browser db
- [ ] No action added for `How do I find the recovery phrase?`
- [ ] Component tests
- [ ] Input Validations & Error messages
- [x] SharedModal portal
- [x] SharedSelect component


ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
```diff
@@ -1,45 +1,447 @@
...
+  const handleChangeCustomPath = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setCustomPath({
+      ...customPath,
+      [e.target.name]: e.target.value,
+      isReset: false,
+    })
+  }
 
   return (
-    <SharedSelect
```
henryboldi, line 38: Would it be possible to keep `SharedSelect`? This will be a useful component for most of our dropdowns :)

ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
```diff
@@ -1,45 +1,447 @@
...
-      onChange={onChange}
-    />
+    <>
+      <div
+        className={classNames("modal", {
+          [`open step_${customModalStep}`]: customModalStep > 0,
+        })}
+      >
+        <div className="modal_overlay" />
+        <div className="modal_content">
```
henryboldi, line 107: How about creating a `SharedModal` component for this?

ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
```diff
@@ -1,45 +1,447 @@
...
-      placeholder="Select derivation path"
-      onChange={onChange}
-    />
+    <>
+      <div
+        className={classNames("modal", {
+          [`open step_${customModalStep}`]: customModalStep > 0,
+        })}
+      >
+        <div className="modal_overlay" />
```
henryboldi, line 106: Not blocked but the hook you added, `useOnClickOutside`, could be useful here

henryboldi (changes requested):
Lookin good! 🔥👏

Still, I'd love to keep `SharedSelect` since it'll be used for most of our dropdowns and for consistency. What do you think?

henryboldi:
I'd also like to note that `SharedSelector` is being used in @cairomassimo's PR https://github.com/tallycash/extension/pull/893

ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
7alip: Alright! I had considered it just as a proposal as I didn't know about Vlad's design. Let's keep it then.

7alip: Do you mean, should I change the `SharedSelect` component style according to @VladUXUI's design for derivation path.
Or SharedSelect will remain as it is, and derivation path select has its own uniq design?

henryboldi: I believe the updated design here should apply to `SharedSelect` in general. So could `OnboardingDerivationPathSelect` use `SharedSelect`?

ui/components/Shared/SharedModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  onClose,
+  isOpen,
+  minHeight,
+  closeOnOverlayClick = true,
+}: SharedModalProps): ReactElement {
+  const ref = useRef(null)
+
+  useOnClickOutside(ref, closeOnOverlayClick ? onClose : () => {})
+
+  return createPortal(
```
henryboldi, line 29: Very nice! 🤌

henryboldi:
@7alip Thank you so much for the fantastic contribution! 💪

I'll test this later today and hopefully we can merge :)

henryboldi (approved):
Phenomenal work man!

I want to note that there's not much room for the error message. Granted, I don't think this has been accounted for in our design yet, cc @VladUXUI 

error space

henryboldi:
@7alip, could you sign all the commits please? 🙏

VladUXUI:
That is true, my bad for not leaving enough room for error messages.

Best case scenario would be to increase the space if helper/error text is visible (making the page scrollable)

Another solution here is to increase the spacing of the bottom elements, not the biggest fan of how squeezed everything is, but should be good for a while till we add 24w seed.

Updated Figma with the second solution if first one ain't feasible at the moment.

VladUXUI:
Also, this is looking really nice @7alip!

7alip:
> @7alip, could you sign all the commits please? 🙏

@henryboldi, I couldn't sign the previous commits. I have rebased and committed them by signing with gpg/yubikey, but nothing changed. Is there any alternative solution?

henryboldi:
>  I have rebased and committed them by signing with gpg/yubikey, but nothing changed.

@7alip, I think you'll need to force push after rebasing? And a command like `git rebase -i --root --exec 'git commit --amend --no-edit --no-verify -S'` should work 🤞

7alip:
> That is true, my bad for not leaving enough room for error messages.
> 
> Best case scenario would be to increase the space if helper/error text is visible (making the page scrollable)
> 
> Another solution here is to increase the spacing of the bottom elements, not the biggest fan of how squeezed everything is, but should be good for a while till we add 24w seed.
> 
> Updated Figma with the second solution if first one ain't feasible at the moment.

@VladUXUI , hope it's what you expect.
tally

henryboldi (approved):
Thank you for signing the commits. Lookin great! ❤️

greg-nagy:
Created a PR that should solve this build issue: #938

cairomassimo:
> I'd also like to note that `SharedSelector` is being used in @cairomassimo's PR #893

FYI, that was mostly a dummy, I'm now changing the code to use `OnboardingDerivationPathSelectAlt` instead.

henryboldi:
@7alip, thank you for your patience here! 🙇 We're trying to solve the external build error issue soon. cc @greg

Shadowfiend:
Once more with feeling.

PR #892, opened 2022-01-25
🙅 Only ask to create a password when needed

henryboldi:
## What
A fix to not require creating a password unless explicitly importing or creating a keyring.

## Why
With recent changes, navigating to the "add wallet" page, which has an option for read-only, asks the user to create a password. Therefore it asks the user to create a password even they're just trying to test an address. This wasn't the design or security plan. The intent was basically to only request passwords when interacting with the keyring. This PR fixes the issue.

## How
First, I removed the password guard (`useAreKeyringsUnlocked(true)`) on the add wallet page. Then created an "interstitial page" which handles triggering generating a phrase. So the add wallet page can simply link to the interstitial page, which:
1. Requests the user to create a password if none exists
2. Asks for a password if it already exists
3. Dispatches generate keyring
4. Redirects to the verify seed flow when all these conditions are met

## Testing

1. Try loading a read-only account and confirm it doesn't ask for a password
2. Try creating a phrase and see if it asks you to create a password, and check that it continues to the verify phrase flow

## Demo
create password when needed


mhluongo (approved):
Excellent PR description and code, tested paths locally, looking good!

PR #893, opened 2022-01-25
Add UI to import accounts from Ledger (wired to Redux but not actual services)

### LINKED ISSUE #850: build + wire "Before we get started" screen

### LINKED ISSUE #851: build + wire "Congratulations!" screen

### LINKED ISSUE #852: build + wire "Select Ledger accounts" screen

cairomassimo:
Closes #850, closes #851, closes #852.


background/redux-slices/ledger-import.ts
```diff
@@ -0,0 +1,234 @@
...
+  importLedgerAccounts: Array<{
+    path: string
+    address: string
+  }>
+}
+
+export const emitter = new Emittery<Events>()
+
+export const initialState: LedgerImportState = {
+  nonce: 0,
```
greg-nagy, line 41: Instead of nonce we should keep track of ledgers by `ledgerID`

background/redux-slices/ledger-import.ts
```diff
@@ -0,0 +1,234 @@
...
+  connected: false,
+  parentPath: null,
+  accounts: [],
```
greg-nagy, line 44: Maybe use object instead of an array, that could clean up the logic.

background/redux-slices/ledger-import.ts
```diff
@@ -0,0 +1,234 @@
...
+  }
+  importLedgerAccounts: Array<{
+    path: string
+    address: string
+  }>
+}
+
+export const emitter = new Emittery<Events>()
+
+export const initialState: LedgerImportState = {
```
greg-nagy, line 40: Change this data structure to be able to handle multiple ledgers. A main object with `ledgerID` as key, and all the ledger info here should be populated under that key. 

As a next step `nonce` could be used as `ledgerID`, then we can swap that to proper `ledgerID`

background/redux-slices/ledger-import.ts
```diff
@@ -0,0 +1,234 @@
+import { createSlice } from "@reduxjs/toolkit"
```
greg-nagy, line 1: We should rename this to `ledger.ts` and the name of the slice should also be `ledger`

greg-nagy (changes requested):
First changes, that we talked through in the PR call.

ui/pages/Ledger/Ledger.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }, [dispatch])
+
+  return (
+    <TabContainer>
+      {!done && !connected && !connecting && (
+        <LedgerPrepareScreen
+          onContinue={() => {
+            setConnecting(true)
+            navigator.usb.requestDevice({ filters: [] }).finally(() => {
+              /* Allow some time to react to clicks before changing the UI. */
```
henryboldi, line 27: Why? The additional 100ms isn't so perceptible

ui/components/Ledger/LedgerContinueButton.tsx
```diff
@@ -0,0 +1,50 @@
+import React from "react"
+
+export default function LedgerContinueButton({
```
henryboldi, line 3: If the only thing that's unique about this button is that its full width to its parent, is it worth creating a new component vs. `SharedButton`?

ui/components/Ledger/LedgerContinueButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+      </button>
+      <style jsx>{`
+        .button {
+          background: var(--trophy-gold);
+          border-radius: 4px;
+          margin: 1rem 0;
+          padding: 0.5rem 1rem;
+          text-align: center;
+          font-family: Segment;
+          font-style: normal;
```
henryboldi, line 30: `font-style: normal;` is the default

ui/components/Ledger/LedgerImportDoneScreen.tsx
```diff
@@ -0,0 +1,44 @@
...
+            You can open Tally now.
+          </>
+        }
+        subHeading="Selected accounts were succesfully connected."
+      >
+        <div className="button-container">
+          <SharedButton
+            size="medium"
+            icon="close"
+            iconPosition="left"
```
henryboldi, line 27: Based on the design,

```suggestion
            iconPosition="right"
```

henryboldi (dismissed):

ui/components/Ledger/LedgerContinueButton.tsx
cairomassimo: Good point. I didn't do it just yet, for the following reason.

I think it's not great when a shared component has (too many) props that control its style. IMHO it would be better if we could just style it from outside, with custom CSS and/or reusing CSS fragments. (This is especially true for shared buttons, which in most cases are just shared CSS, but partially applies also to other components such as `SharedSelect`.) However, I'm not sure what is the correct way to do this with the `<style jsx>` syntax...

For the moment, I could just patch `SharedButton` and add a new prop (e.g. `fullWidth`) for this use case.
Otherwise, we could also start sharing the CSS fragments for buttons (how?) and moving towards my approach, if you like it, or perhaps allow custom CSS to be applied to shared components from outside (how?) which is a kind of compromise. Let me know!

cairomassimo:
Thanks for the great comments @greg-nagy and @henryboldi !
I should have addressed all of them (except the one comment above), but another review pass is probably due.

ui/pages/Ledger/Ledger.tsx
```diff
@@ -0,0 +1,85 @@
...
+import LedgerPrepare from "./LedgerPrepare"
+
+export default function Ledger(): ReactElement {
+  const [phase, setPhase] = useState<
+    "0-prepare" | "1-request" | "2-connect" | "3-done"
+  >("0-prepare")
+  const [deviceID, setDeviceID] = useState<string | null>(null)
+
+  const devices = useBackgroundSelector((state) => state.ledger.devices)
+  const device = deviceID === null ? null : devices[deviceID] ?? null
```
greg-nagy, line 22: The ternary check is not necessary here.

ui/pages/Tab.tsx
```diff
@@ -0,0 +1,39 @@
...
+import TabNotFound from "./TabNotFound"
+
+/**
+ * Entry point for UI shown in browser tabs.
+ */
+export default function Tab({ store }: { store: Store }): ReactElement {
+  return (
+    <>
+      <Provider store={store}>
+        {/* HashRouter seems the only choice supporting safe page reloads. */}
```
greg-nagy, line 15: ❗️

greg-nagy (approved):
This PR should be good to go. Very well done. 

Not like a steak, because that's horible when well done. 

🥩

ui/components/Ledger/LedgerContinueButton.tsx
henryboldi: How about adding a `style` prop to `SharedButton` which can handle these edge cases?

ui/public/index.css
```diff
@@ -37,16 +37,13 @@
 }
 
 body {
-  width: 384px;
-  height: 594px;
```
henryboldi, line 41: I believe removing `width` and `height` here is causing a flash of a smaller size when opening the extension. Not a huge deal but worth noting.

Issue added here → https://github.com/tallycash/extension/issues/924

PR #897, opened 2022-01-25
Bar the queue: reference draft for transaction queueing

Shadowfiend:
This is incomplete, and based on some older swaps code, I do not expect to merge it though some of the code may see reuse. Main work of interest is in the third and second to last commits: 

https://github.com/tallycash/extension/pull/897/files/f1f55300d1c9859d399c6ea837bb9d1328785639..d20cc9c6b2289772660f6b40dddff6dc64c8901e


Shadowfiend:
We'll be going down a full RFB path with this once @VladUXUI gets a chance to write up the UX side of things, so I'm closing this PR. It may still be referenced in the RFB if there's useful lessons to be taken from it.

Issue #834, opened 2022-01-16
Funds (small amount) not showing up


Naxsun:
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/932187466373406750

### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

User was testing the wallet by sending 0.001 ETH to an address imported in Tally.

### What did not work?

ETH balance as well as $  Balance not showing up (should be around 3$).

https://etherscan.io/tx/0x502d48f91d942086c2997728fb72f02b73ae359b063228fa6a6937288dead6be

chrome version: 96.0.4664.110

![image](https://user-images.githubusercontent.com/71384922/149660999-0735ad53-2783-493b-9be9-5ed955145260.png)
https://discord.com/channels/808358975287722045/888500174685614090/932187466373406750

### Version

Community Edition v0.10.1

Naxsun:
Added to https://github.com/tallycash/extension/issues/813

PR #898, opened 2022-01-26
Automatically determine # of decimals to display for fungible asset balances

0xDaedalus:
This PR allows us to use an assets price to guess how many decimal places it should display for a given asset.  The heuristic is very simple - for each power of 10 over `100` show an extra decimal place - while still defaulting to 2 decimal places for assets that we may not have prices for (and a minimum of 2 decimal places per asset).  This allows us to show relevant balances both for stablecoins and high price-per-token assets like YFI and ETH.

image

image

This addresses the issue brought up in #834


background/services/indexing/index.ts
```diff
@@ -426,9 +426,14 @@ export default class IndexingService extends BaseService<Events> {
 
     // Filter all assets based on the currently selected network
     const activeAssetsToTrack = assetsToTrack.filter(
-      (t) => t.homeNetwork.chainID === getEthereumNetwork().chainID
+      (t) =>
+        t.symbol === "ETH" ||
+        t.homeNetwork.chainID === getEthereumNetwork().chainID
```
Shadowfiend, line 431: While you're in here, do you mind switching `t` to `asset`, or using destructuring? Must have missed this in my hit last week.

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -164,7 +170,11 @@ export const selectCurrentAccountBalances = createSelector(
           return enrichedAssetAmount
         }
 
-        return enrichAssetAmountWithDecimalValues(assetAmount, desiredDecimals)
+        return enrichAssetAmountWithDecimalValues(
+          assetAmount,
+          desiredDecimals,
+          assetPricePoint
+        )
```
Shadowfiend, line 177: Not sure how this happened, but I think the code for `selectCurrentAccountBalances ` and `selectAccountAndTimestampedActivities ` is both complex and identical, so we probably want to extract a helper in this file…

If we do that, I wonder if we could make the decimals decision-making in here, rather than doing it in `enrichAssetAmountWithDecimalValues`. It feels like the latter shouldn't be concerned with the main currency at all, and passing the price point in expands its list of concerns and incidental complexity considerably.

In particular, I'm thinking of something like this:

```typescript
        if (assetPricePoint) {
          const mainCurrencyEnrichedAssetAmount = 
            enrichAssetAmountWithMainCurrencyValues(
              assetItem,
              assetPricePoint,
              desiredDecimals
            )

          // Heuristically add decimal places to high-unit-price assets, `
          // 1 decimal place per order of magnitude in the unit price; e.g.
          // a if USD is the main currency and the asset unit price is $100,
          // 2 decimal points, $1000, 3 decimal points, $10000, 4 decimal
          // points, etc. `desiredDecimals` is treated as the minimum, and
          // order of magnitude is rounded up (e.g. $2000 = >3 orders of
          // magnitude, so 4 decimal points).
          const decimalValuePlaces = Math.min(
            // Using ?? 0, safely handle cases where no main currency is
            // available.
            Math.ceil(
              Math.log10(mainCurrencyEnrichedAssetAmount.unitPrice ?? 0)
            ),
            desiredDecimals
          )

          const fullyEnrichedAssetAmount = enrichAssetAmountWithDecimalValues(
            mainCurrencyEnrichedAssetAmount,
            decimalValuePlaces
          )

          if (typeof fullyEnrichedAssetAmount.mainCurrencyAmount !== "undefined") {
            totalMainCurrencyAmount ??= 0 // initialize if needed
            totalMainCurrencyAmount += fullyEnrichedAssetAmount.mainCurrencyAmount
          }

          return fullyEnrichedAssetAmount
        }
```

A few things I like better about this approach:
- The main currency enrichment already gets us the main currency unit price as a `number`, so we can use it directly to run the heuristic.
- The decimals logic is separate from the enrichment, and in a place where we can take user preferences into account directly later.
- The logic of “if we have an asset price point do X, otherwise do Y” is not repeated in the decimal value enricher; instead, we make one set of decisions right here regarding that.

The main knock on it is if we start to use similar logic outside of this file; however, if we see this develop, I think we can extract a joint main currency + decimal value enrichment utility to asset-utils for use everywhere that does this bit.

Thoughts?

Shadowfiend (commented):
Big fan of the heuristic approach here. Suggested a slight rearrangement of *where we do it* for consideration.

background/redux-slices/selectors/accountsSelectors.ts
0xDaedalus: Love it - i noticed the similarities as well and I think a helper function will be a big win. ✅

0xDaedalus:
Code Review Addressed - sorry for the force push - I think I clicked the "Merge" button on github a few days ago and also merged in main via cli over the weekend.  Will make a mental note and avoid in the future.

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -28,6 +31,88 @@ const mainCurrencySymbol = "USD"
...
+        const mainCurrencyEnrichedAssetAmount =
+          enrichAssetAmountWithMainCurrencyValues(
+            assetAmount,
+            assetPricePoint,
+            desiredDecimals
+          )
+
+        // Heuristically add decimal places to high-unit-price assets, `
+        // 1 decimal place per order of magnitude in the unit price; e.g.
+        // a if USD is the main currency and the asset unit price is $100,
```
Shadowfiend, line 68:
```suggestion
        // if USD is the main currency and the asset unit price is $100,
```

Shadowfiend (approved):
Left a couple of additional nits, but this is good to ship. Will take it for a test drive soon and merge.

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -28,6 +31,88 @@ const mainCurrencySymbol = "USD"
...
+        // magnitude, so 4 decimal points).
+        const decimalValuePlaces = Math.max(
+          // Using ?? 0, safely handle cases where no main currency is
+          // available.
+          Math.ceil(Math.log10(mainCurrencyEnrichedAssetAmount.unitPrice ?? 0)),
+          desiredDecimals
+        )
+
+        logger.info("unitPrice: ", mainCurrencyEnrichedAssetAmount.unitPrice)
+        logger.info("decimal value places: ", decimalValuePlaces)
```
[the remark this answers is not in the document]
mhluongo: `logger.debug()` definitely seems more appropriate... though I'll also call out that I don't love the side-effecting logging *in* the selector.

0xDaedalus:
> Left a couple of additional nits, but this is good to ship. Will take it for a test drive soon and merge.

All good nits - will get them in shortly

background/redux-slices/selectors/accountsSelectors.ts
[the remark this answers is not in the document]
0xDaedalus: I agree with you both - maybe we dont even need these?  I'll remove for now.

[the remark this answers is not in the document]
Shadowfiend: For in-the-moment debugging/analysis (rather than stuff I want in the actual app), I still use `console`, fwiw (usually `console.warn` for visibility/filtering). The linter pre-commit hook yells at me if I leave them in, which is, in this case, a [plus to avoid](https://github.com/tallycash/extension/pull/880#discussion_r791686328)  :grimacing:

Shadowfiend:
Works as advertised! Checked out an account with BTC derivatives as well and saw that sweet sweet extra precision:
 
Screen Shot with 1024.00067 WBTC displayed

(0xf9930a9d65cc57d024cf9149ae67e66c7a77e167 pulled from Etherscan.)

Also confirmed the appropriate precision shows up on the overview as well:

Screen Shot showing 4 decimals of precision on ETH, 2 decimals on various smaller assets, 2 decimals on WETH, and 5 decimals on BTC-derived assets

Notably here, WETH here shows as 0.25 because the number is exactly 0.25 (i.e., 0.2500) and we don't show extra precision if we don't need it, which is 👍

0xDaedalus:
Hmmmnn - do we want to add a heuristic for notional prices as well?

image

Shadowfiend:
Eep, missed that. 

That would def be nice for completeness, esp if it's straightforward.

PR #895, opened 2022-01-25
Sign with Ledgers

cairomassimo:
Implementing [Figma](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=5242%3A123793)

PR #823, opened 2022-01-12
Sign typed data

wojciech-turek:
Signing typed data flow.

Signing slice -> async thunk -> emit event -> main.ts listening -> keyring sign typed data

PR #857, opened 2022-01-20
Typed data signing

wojciech-turek:
This PR will allow signing typed data according to [EIP-712](https://eips.ethereum.org/EIPS/eip-712) standard, one of the use cases is to provide gasless 'approval' for token spend with  [EIP-2612](https://eips.ethereum.org/EIPS/eip-2612).

New `signing` redux slice has been introduced for the purpose of UI integration.
New SignTransaction sub page has been created to display to users what are they signing.

Example flow:

https://user-images.githubusercontent.com/61434499/150401600-8a8ece1f-3237-41d3-8377-1fc89655132f.mov

Closes #823

PR #899, opened 2022-01-26
Refactor sign transaction UI

cairomassimo:
Some refactoring seems necessary to push forward signing with Ledger (#895).
Doing it in a separate PR to address here conflicts with #823 and #873 (BTW #873 may become redundant after this).


ui/pages/SignTransaction.tsx
```diff
@@ -83,10 +69,10 @@ export default function SignTransaction({
       : getAccountTotal(state, transactionDetails.from)
   )
 
-  const [gasLimit, setGasLimit] = useState("")
-  const estimatedFeesPerGas = useBackgroundSelector(selectEstimatedFeesPerGas)
+  const areKeyringsUnlocked = useAreKeyringsUnlocked(
+    signerAccountTotal?.signingMethod?.type === "keyring"
+  )
```
Shadowfiend, line 74: Just want to flag that this will probably ultimately end up in its own component that is only rendered if the signing account is a local keyring. The intent is for keyring unlock to move to after the "sign" but is clicked. This will let the keyring flow more closely match the hardware wallet flow.

That said, obviously this doesn't need to happen in this PR :)

ui/components/SignTransaction/SignTransactionDetailPanel.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  if (transactionDetails === undefined) return <></>
+
+  return (
+    <div className="detail_items_wrap standard_width_padded">
+      <SharedSlideUpMenu
+        size="custom"
+        isOpen={networkSettingsModalOpen}
+        close={() => setNetworkSettingsModalOpen(false)}
+        customSize={`${3 * 56 + 320}px`}
```
greg-nagy, line 32: This is a bit of magic here

greg-nagy (approved):
This should be good to go. 

It should not cause any conflicts with #857 — thought hopefully. 🙈

PR #900, opened 2022-01-26
Revert "🗃 Preserve location when reopening extension"

greg-nagy:
Reverts tallycash/extension#818

It's more dangerous than I thought.


mhluongo (approved):
Thank you! Let's get that last PR re-opened, sorry for the churn @henryboldi

PR #901, opened 2022-01-26
🗃 Preserve location when reopening extension

henryboldi:
## What
Preserves the location when reopening Tally. This also preserves the location's state attribute. For example, when the token clicked is passed to the single asset page. This saving routes feature is feature flagged. I've also added an error boundary to prevent a user from getting stuck on a page in the event of a crash.

## Why
It's easy to unintentionally close the extension when opening Etherscan, copying an address, opening a dApp, etc. And it's frustrating to lose your progress on whatever you've been doing. This change lets you open Tally right where you left off.

## How
Uses `Router`'s `initialEntries` prop to pass an initial history. On route changes, the entries are saved to the UI slice. Each route has a `persistOnClose` defined in the routes config file to avoid saving dangerous locations. Generally, it's safe to save if a page has a tab bar. If there isn't a tab bar, it's hard to guarantee that the user has a way of escaping that page. In the event of a crash, I've added `react-error-boundary` and an `ErrorFallback` page which provides a link back to home. `react-error-boundary` has 1.5 million weekly downloads and was motivated by React's own `Error Boundaries` not supporting functional components.

## Test
Turn on `PERSIST_UI_LOCATION` and then click around — close and reopen the extension and see you get stuck on a page.

## Demo
persist


Shadowfiend:
> ⚠️ If a user navigates to a page they can't exit, there's no escape

Eep! Any thoughts on how we might mitigate this? Our routing is too stringly-typed here and having a known risk of completely disabling someone's extension makes me very nervous. A sub-concern of this is if the extension navigates to a page that doesn't exist or somehow has broken state, which could lead to no escape.

ui/pages/Popup.tsx
```diff
@@ -54,6 +61,39 @@ export default function Popup({ store }: { store: Store }): ReactElement {
...
+    // animation control states. `initialEntries` needs to be a reversed
+    // version of route history entities. Without avoiding the initial load,
+    // entries will keep reversing.
+    if (renderCount.current > 1 && isNotOnKeyringRelatedPage) {
+      const entries = routeHistoryEntities
+        .reduce((agg: Partial<Location>[], entity) => {
+          const { ...entityCopy } = entity as Partial<Location>
+          delete entityCopy.hash
+          delete entityCopy.key
+          agg.push(entityCopy)
```
cairomassimo, line 84: Not blocking: we can avoid the `delete` operator by destructuring. E.g.

```suggestion
          const { key: {}, hash: {}, ...entry } = entity as Partial<Location>
          agg.push(entry)
```

ui/pages/Popup.tsx
```diff
@@ -54,6 +61,39 @@ export default function Popup({ store }: { store: Store }): ReactElement {
...
   const [showTabBar, setShowTabBar] = useState(true)
+  const renderCount = useRef(0)
+
+  function saveHistoryEntries(routeHistoryEntities: Location[]) {
+    const isNotOnKeyringRelatedPage =
+      routeHistoryEntities[routeHistoryEntities.length - 1].pathname !==
+        "/signTransaction" &&
+      !routeHistoryEntities[routeHistoryEntities.length - 1].pathname.includes(
+        "/keyring/"
+      )
```
cairomassimo, line 72: This is a necessary but ugly hack. In the future, I think we should:

1. add a nonce/ID to the `signTransaction` route. If the ID does not match a pending transaction in the state, it should automatically call `history.goBack()` or similar,
2. replace `/keyring/` routes with modal-like components (e.g. `if (needsKeyrings && keyringsLocked) return <UnlockKeyrings />`)

ui/pages/Popup.tsx
```diff
@@ -54,6 +61,39 @@ export default function Popup({ store }: { store: Store }): ReactElement {
...
+        .reduce((agg: Partial<Location>[], entity) => {
+          const { ...entityCopy } = entity as Partial<Location>
+          delete entityCopy.hash
+          delete entityCopy.key
+          agg.push(entityCopy)
+          return agg
+        }, [])
+        .reverse()
+
+      localStorage.setItem(
```
cairomassimo, line 89: Use of `localStorage` may be the right choice here, but it scares me a bit anyway since we already have state scattered across extension storage area (`runtime.storage.*`), IndexedDB (Dexie), current URL / navigation history, in-memory state (`useState`) and persisted Redux (wherever that is)....

cairomassimo (commented):
It looks hacky, but I think that in order to make it cleaner we need a bunch of work all over the place, so definitely out-of-scope for this PR.

ui/pages/Popup.tsx
```diff
@@ -62,14 +102,16 @@ export default function Popup({ store }: { store: Store }): ReactElement {
...
+              // @ts-expect-error TODO: fix the typing
+              saveHistoryEntries(routeProps?.history?.entries)
+
               const normalizedPathname =
                 transformedLocation.pathname !== "/wallet"
-                  ? routeProps.location.pathname
+                  ? transformedLocation.pathname
                   : "/"
 
               setAnimationConditions(
```
cairomassimo, line 117: This seems a side-effect during rendering, which is a big no-no in React.
From https://reactrouter.com/docs/en/v6/api#uselocation, we should do this instead:

```
  const location = useLocation();

  useEffect(() => {
    // ...
  }, [location]);
```

ui/pages/Popup.tsx
Shadowfiend: Agree here. Is this not storable in redux?

henryboldi: Good point! Thank you. How about in the UI slice?

Shadowfiend: Seems quite reasonable!

Shadowfiend:
> Whitelist pages that should keep their state — eg onboarding.

An explicit whitelist for routes that are saved and routes that persist in-memory state both sound excellent here. In addition to easing concerns about being stuck in a weird place, I suspect it'll prove useful in other ways as well as we run into folks' real-life expectations/preferences.

greg-nagy (approved):
Lovely PR description, with all the important details, well considered approach. Awesome PR! 
Glad that I got to review it!

🙇‍♂️

ui/routes/routes.tsx
```diff
@@ -20,126 +20,157 @@ import DAppPermissionRequest from "../pages/DAppConnectRequest"
 import KeyringUnlock from "../components/Keyring/KeyringUnlock"
 import KeyringSetPassword from "../components/Keyring/KeyringSetPassword"
 
-const pageList = [
+interface PageList {
```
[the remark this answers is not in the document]
Shadowfiend: `type =`!

PR #909, opened 2022-01-28
Signing fixes

wojciech-turek:
When restructuring the SignTransaction page we have accidentally omitted Approval Page.

In order to avoid future unexpected outcomes from deleting type from types to be signed we now destructure it instead.

Removed eslint disable from capitalization function.


cairomassimo (approved):
I don't see any issue with the code.
Thanks for the fix!

PR #910, opened 2022-01-28
Add "Awaiting hardware signature" screen

cairomassimo:
Shown when a transaction is being signed with a Ledger.
Does not fully match UI specs just yet (reusing existing info blocks).
Screenshot from 2022-01-28 10-46-14


ui/pages/SignTransaction.tsx
```diff
@@ -79,14 +79,13 @@ export default function SignTransaction({
     return undefined
   })
 
-  const areKeyringsUnlocked = useAreKeyringsUnlocked(
-    signerAccountTotal?.signingMethod?.type === "keyring"
-  )
+  const needsKeyrings = signerAccountTotal?.signingMethod?.type === "keyring"
+  const canSign = useAreKeyringsUnlocked(needsKeyrings) || !needsKeyrings
```
greg-nagy, line 83: This nameing here does not feel right. I mean a ledger can also sign. 

The `useAreKeyringsUnlocked` will redirect if the keyring is locked. If it's not of `keyring` type then it shouldn't do that.

ui/pages/SignTransaction.tsx
```diff
@@ -133,12 +132,16 @@ export default function SignTransaction({
...
+  const isWaitingForHardware =
+    signerAccountTotal?.signingMethod?.type === "ledger" && isTransactionSigning
+
   switch (signType) {
     case SignType.SignSwap:
       return (
         <SignTransactionContainer
           signerAccountTotal={signerAccountTotal}
           title="Swap assets"
+          isWaitingForHardware={isWaitingForHardware}
```
greg-nagy, line 144: What is the reason behind this API change?

For the keyring service we have multiple signing screens, but with ledger we force this to be a single screen. 
What is the reason for this difference?

greg-nagy (dismissed):
The API change does not feel right, but I might just need more context to understand.

ui/pages/SignTransaction.tsx
cairomassimo: It's a bit backwards: it should be `const canSign = !needsKeyrings || useAreKeyringsUnlocked(needsKeyrings) ` but it's not allowed because `useAreKeyringsUnlocked` is a hook and must be called unconditionally.

I could invert the logic and write:
```suggestion
  const waitingForKeyrings = !useAreKeyringsUnlocked(needsKeyrings) && needsKeyrings
```

but a more long-term solution would be to move the call `useAreKeyringsUnlocked` to another component that's only rendered when `signerAccountTotal.signingMethod.type === 'keyrings'`

ui/pages/SignTransaction.tsx
cairomassimo: > For the keyring service we have multiple signing screens

You mean the screen to unlock the keyrings? Right now it's handled separately, but I agree that in the long term they should be handled at the same level (i.e., when I approve a transaction, the UI proceeds to the actual signing step which depends on the signing method).

It's definitely a cut corner, but I don't know how to make it right without some refactoring of keyrings. So perhaps in another PR?

cairomassimo: Leaving this one as is. Needs a refactoring but right now it's premature.

greg-nagy: This should be ok for now.

> they should be handled at the same level (i.e., when I approve a transaction, the UI proceeds to the actual signing step which depends on the signing method).

I agree with this, and I think the solution here will be to move the signer-specific concerns a level up, to the `SignTransaction.tsx`. There we should decide what to show based on the `SignType` __AND__ the signing method.
It would also make sense to refactor `SignTransactionContainer` into `KeyringSignContainer` and create a `LedgerSignContainer`. This would leave us in a happy place when we will add the next type of signer.

What do you think @cairomassimo? Does it sound good? 
If so I think doing this refactor should happen in #921

cairomassimo: Yes, sounds right!

ui/pages/SignTransaction.tsx
greg-nagy: Blimey! Let's do this, when we get to this task `update unlock behavior for keyring service to happen after the user clicks "Sign"` https://github.com/tallycash/extension/issues/496

greg-nagy (approved):
https://imgflip.com/i/63dlzd

PR #911, opened 2022-01-28
Allow invoking services API methods from thunks directly

cairomassimo:
This PR introduces a mechanism for async thunks to invoke service API methods directly.

Using this mechanism, we can:

- eliminate many events fired by thunks, where the emitter expects the event to be handled in a certain way (implicit dependencies), and replace them with direct API calls (explicit dependencies),
- perform full round-trip async communication between UI and services, mediated by Redux.


background/main.ts
```diff
@@ -386,6 +386,34 @@ export default class Main extends BaseService<never> {
...
+          network: nameNetwork.network,
+        }
+        await this.chainService.addAccountToTrack(addressNetwork)
+        this.store.dispatch(loadAccount(address))
+        this.store.dispatch(setNewSelectedAccount(addressNetwork))
+      } else {
+        throw new Error("Name not found")
+      }
+    } catch (error) {
+      throw new Error(
```
cairomassimo, line 411: This errors now propagate to thunks, whereas before they would have been eaten by Emittery and ignored.
Should not be an issue, but a second opinion is valuable here.

Shadowfiend: This is Good™, even if it's different. We may see some breakage if there are transient network issues though—would love to intentionally fail here and make sure that the UI continues to work normally/doesn't get stuck (other than presumably not adding the account).

background/redux-slices/utils.ts
```diff
@@ -7,6 +7,10 @@ import {
   createAsyncThunk,
 } from "@reduxjs/toolkit"
 import logger from "../lib/logger"
+// FIXME: this utility file should not depend on actual services.
+// Creating a properly typed version of `createBackgroundAsyncThunk`
+// elsewhere has proven quite hard, so we are hardwiring typing here.
+import type Main from "../main"
```
Shadowfiend, line 13: This is actually two-layered: this file shouldn't depend on the service proper, and the slices should ideally declare an interface with just the external methods they expect to call (not unlike what they were doing with events, just more flexible).

That said, I think we can lump that whole thing in with the work to separate redux slices into a distinct package that we've been discussing for a while. Having this just be `Main` for now is dirty but I agree the win is worth it. I do have some thoughts on how to handle the typing (in particular, leveraging the fact that we're already changing the returned action creator signature to pass the extra parameter as its own distinct argument, opening the door for the caller to declare the expected type and have it be inferred as a type parameter), but haven't had a chance to experiment with them.

Shadowfiend (approved):
Noting here that there's some disagreement about pursuing this approach, but we're going to give it a fair shake and see if things get nasty in the future.

PR #912, opened 2022-01-28
Fix Edge Case When Adding Pricepoint for "0th" Asset

### LINKED ISSUE #813: Account balances and prices incorrect
### Discord Discussion Link

na

Chrome
Brave

Add 1 account through seed and 1 address in read-only.

### What did not work?

The account that is added through the seed has 0 balance and activity. This is shown correctly, except it still says: "Tally will populate your historical activity over time; this may take an hour or more for accounts that have been active for a long time. For new accounts, new activity will show up here."

Account has been imported > 12 hours ago.

The address that I added in read only mode has 0.1269 ETH in it. This is rounded down by the wallet to 0.12. However the wallet $ balance is shown as 0$. I expect it doesn't get an ETH price feed.
Screenshot 2022-01-09 at 10 14 40

In the overview page I see both addresses, the total ETH balance (rounded down again) is shown correctly, but there's no ETH price loaded --> 0 $ balance.

Screenshot 2022-01-09 at 10 15 16

Also for this read-only address the account activity is not loaded yet although the account was added >12 hours ago.

Update: others users with the same issues: https://discord.com/channels/808358975287722045/888500174685614090/929674738157096960
https://discord.com/channels/808358975287722045/888500174685614090/929785515824259115
https://discord.com/channels/808358975287722045/888500174685614090/932187466373406750

### Version

Community Edition v0.10.1

### Relevant log output

_No response_

0xDaedalus:
This PR adds an explicit null check that fixes a bug where the 1st asset in the asset list was not having its prices populated.  Since we explicitly prepend the assets array with ETH and BTC this would always result in ETH not having its prices populated.

This was a fun one - I went down the rabbit hole into indexDb and alchemy.ts before coming back out. 🐰

Closes #813

### DIFF
--- a/background/redux-slices/assets.ts
+++ b/background/redux-slices/assets.ts
@@ -189,7 +189,7 @@ const assetsSlice = createSlice({
         const index = findClosestAsset(pricedAsset, [
           ...immerState,
         ] as AnyAsset[])
-        if (index) {
+        if (index !== null) {
           // append to longer-running prices
           const prices = prunePrices(
             [...immerState[index].prices].concat([pricePoint])


background/redux-slices/assets.ts
```diff
@@ -189,7 +189,7 @@ const assetsSlice = createSlice({
         const index = findClosestAsset(pricedAsset, [
           ...immerState,
         ] as AnyAsset[])
-        if (index) {
+        if (index !== null) {
```
mhluongo, line 192: Whoops, good catch! This explains a lot :)

mhluongo (approved):
:rocket:

background/redux-slices/assets.ts
Shadowfiend: Yiiiikesssss. Turning on those TypeScript linter rules real soon.

PR #923, opened 2022-01-31
:monocle_face: Basic telemetry service for tracking extension storage usage

### LINKED ISSUE #216: Periodically log `IndexedDB` and extension storage size
We don't have a direction on telemetry longer-term yet, and we certainly don't have performance metrics to track in CI... but in the meantime a good consistent reminder of the storage space we're using will help build intuition during initial development.

H/t @itsrachelfish for making sure this stays on our radar.

itsrachelfish:
Resolves #216 

This PR adds a barebones service which is used to display extension usage information. The size of IndexedDB and Redux is displayed immediately on startup and once per hour after that.

As discussed in #216, this PR also adds a "debug" log level to the logger library and changes the behavior of the logger library to use native console functions (`console.warn`, `console.error`) instead of `console.log` for everything. 

extension-usage


Shadowfiend (approved):
Effin' glorious.

Going to test-drive this, but the logger changes alone are making me googly-eyed.

Epic parting shot @itsrachelfish !!

PR #929, opened 2022-02-01
Smooth Liminal: Improve transitions for snackbar and shared slide-up, plus bonus tidbits

### LINKED ISSUE #726: Properly display infinite token approvals

### LINKED ISSUE #888: Fix janky shared slide up menu
The `SharedAssetInput` component has a janky animation. @henryboldi proposed a fix in the original PR

> That solution sounds good to me, but using a ref for the timeout. The `Snackbar` does something similar, although remains a victim to the same problem:

```js
  const msTillDismiss = 2500

  const snackbarTimeout = useRef<ReturnType<typeof setTimeout>>()

  const clearSnackbarTimeout = () => {
    if (snackbarTimeout.current) {
      clearTimeout(snackbarTimeout.current)
    }
  }

  useEffect(() => {
    clearSnackbarTimeout()

    snackbarTimeout.current = setTimeout(() => {
      dispatch(clearSnackbarMessage())
    }, msTillDismiss)
  }, [snackbarTimeout, snackbarMessage, msTillDismiss, dispatch])

  useEffect(() => {
    window.onblur = () => {
      clearSnackbarTimeout()
      dispatch(clearSnackbarMessage())
    }
  }, [dispatch])
 ```

_Originally posted by @henryboldi in https://github.com/tallycash/extension/pull/875#discussion_r790995747_

### PR DESCRIPTION
**Pull request by Shadowfiend**
This PR introduces a new hook, `useDelayContentChange`, that allows
access to a delayed view to a piece of content (text, element, group
of elements—anything, really). It works by taking the content, a
condition, and a delay in ms. When the condition is `false`, the
content is returned unchanged. When the condition is `true`, the
content is fixed until the delay elapses.

This is used in the snackbar and `SharedSlideUpMenu` to delay the
hiding of content until after hiding transitions complete in the UI.

Also bundled here:
- Infinite approvals now render as such in the activity list.
- The snackbar fully conforms to @VladUXUI's designs. Thanks to the
  delayed content change, we can now use auto-width instead of having
  the snackbar width be fixed! Also a couple of tweaks to colors.
- The snackbar now adjusts its distance to the bottom of the extension
  depending on whether the tab bar is visible. The way this is handled
  is frankly rather hacky, but it does the trick for now.

#### Snackbar

The snackbar can be tested by going to the Receive page and clicking
the address for copying. The snackbar on a tab bar page can be tested
by inserting this into a component that appears normally (for example,
`WalletActivityList`):
```
useEffect(() => dispatch(setSnackbarMessage("Transaction received")), [])
```

You can also expose a function to the console to test the snackbar at
will by putting this in `ui/index.ts`:

```
  // @ts-expect-error boop
  window.snackbar = (message) => { backgroundStore.dispatch(setSnackbarMessage(message)) }
```

You can then call `snackbar("Transaction received")` in the console
to see the snackbar. This is useful for testing how it behaves when
the message is updated before it hides (the 2.5s hiding timeout should
reset) and, using `setTimeout`, how it behaves when the message is
updated mid-hiding transition (the snackbar should transition back to
visible for 2.5s, and the text should not disappear).

#### Slide-up menu

Open and interact with any slide-up menus, including the account
selector and the asset selector on the swap page.

#### Infinite approvals

Two ways to do this. (1) create a new infinite approval. (2) use the
default testing account (`0x208e94d5661a73360d9387d3ca169e5c130090cd`)
and then modify `ChainService` like so:

```
192 ⋮    │-    this.transactionsToRetrieve = { ethereum: [] }
    ⋮192 │+    this.transactionsToRetrieve = {
    ⋮193 │+      ethereum: [
    ⋮194 │+        "0x57f8c31c33d44e3dddab3ed162723480420c439a3f8dc8199f44bf01ed6fde50",
    ⋮195 │+      ],
    ⋮196 │+    }
```

After 1 minute or so, the activity list should include a December
14th transaction that is an infinite approval.

Closes #888, closes #726 .


Shadowfiend:
(Whoops, missed a push here.)

henryboldi (changes requested):
Awesome!! 🔥🧙  

Although I've got a couple of comments/questions first, which hopefully aren't too persnickety:

1) I don't quite understand the reason behind `setShouldOffsetByTabBar`. Is the position of a floating snackbar worth changing depending on the tab bar? The positioning with this update looks a bit strange anyway. Guessing that's because the tab bar is visible behind the slide up menu.

snackbar position

2) Either way, `TAB_BAR_TAG_CLASS` seems unnecessary since the routes config already determines if the tab bar is being displayed. It seems a bit convoluted to keep track of that in two ways. And if the `TAB_BAR_TAG_CLASS` is being exported, why not also export `TABBAR_OFFSET` from the `TabBar` and use that in its CSS?

Granted, I know you mentioned that,
> The way this is handled is frankly rather hacky, but it does the trick for now.

Anyway, there aren't critical breaking changes, so we can merge this if you'd like. It's working smoothly! And the comments for the new ` useDelayContentChange` hook are superb.

VladUXUI:
Just a quick note here, the general rule for snack-bars is that they should always be in the same place.
In the bottom of the page, this is the least intrusive place (for most use cases), it also creates less surprises for the users.

Not something that needs to be fixed atm, but good to keep in mind

Shadowfiend:
@VladUXUI maybe I misunderstood wrong and we can remove this work chunk entirely! My understanding from Figma was that if the tab bar was visible the top of the tab bar should count as the "bottom" of the page, and when it wasn't, the bottom of the extension popup should.

VladUXUI:
Sorry for the confusion in the Figma. But yeah, the snack-bar should always be in same place.
In design that is 72x from the bottom

Shadowfiend:
Just to confirm the final outcome here with 72px offset, on receive page and page with tab bar:

Screen Shot 2022-02-02 at 11 28 50

Screen Shot 2022-02-02 at 11 31 34

Very happy I could nuke that tab bar detection stuff, it was nasty :laughing:

henryboldi (approved):
Honestly I prefer the previous style and positioning of `SnackBar`, but this matches the design better. Working nicely and the code is great 👏😍

PR #932, opened 2022-02-02
Update "Awaiting hardware wallet signature" screen according to UI specs

### LINKED ISSUE #921: Sign with Ledger: implement "Awaiting hardware wallet signature" screen according to specs
[Figma](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=6062%3A153814)

Some work done in #910.

cairomassimo:
Closes #921.

Includes an extensive refactoring of `SignTransaction`. The kind reviewer is expected to test that this does not break existing behavior with transaction signing (for the various supported types of transactions). Sorry about that!

This does not yes show estimated gas fee: better to wait after #922 is merged. Make sure to create a new issue about this, if this PR is merged (before #922) without implementing the missing functionality.

Screenshot from 2022-02-02 15-29-48

## The refactoring

We now have a `SignTransactionInfoProvider` component which encapsulates all the `SignType`-specific work, and provides it to its `children`. In particular, it renders some `SignType`-specific nodes (title, info block, textual info block, confirm button label), which can be displayed at discretion of its `children`.

This was done to decouple `SignType`-specific code from `SigningMethod`-specific code, which in turn allows to support new signing flows (including Ledger) more easily, while still allowing `SignType`-specific components to maintain and share a state across the entire signing flow (e.g., the resolution of addresses into names). In fact, `SignTransactionInfoProvider` is implemented in terms of  `SignType`-specific components, such as `SignTransactionTransferInfoProvider`, which can maintain their own state.

The next step would be to encapsulate `SigningMethod`-specific code into separate components (e.g., `SignTransactionWithKeyring`, `SignTransactionWithLedger`), but it does not affect this PR so it's not done here.


cairomassimo:
@VladUXUI there is a UI spec mismatch in the "To:" line (I'm using a shared button of type `tertiaryGray`). Which do you prefer:

- leave it as is
- make only the icon clickable (using a shared button), and match the styling specs for the text (easy),
- add a new type of shared button for this use case (or some other technical solution) to make both the address and the icon clickable while exactly matching the specs.

Thanks!

VladUXUI:
Could you please tell me what size is that icon? it should be 12, but it looks smaller

ui/components/SignTransaction/SignTransactionInfoProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const parsedTx = parseERC20Tx(transactionDetails?.input ?? "")
+  const isApproveTx = parsedTx?.name === "approve"
+
+  const getSignType = () => {
+    if (isApproveTx) {
+      return SignType.SignSpend
+    }
+    return SignType.Sign
+  }
```
0xDaedalus, line 46: Do you think it makes sense to extract these lines into a function that takes `parsedTx` as an argument and then returns `SignType`?  That way we could a)define the function outside of this component - and b) potentially reuse it elsewhere down the line.

cairomassimo: Good suggestion, thanks!
I changed it, and the code looks cleaner to me now.

cairomassimo:
> Could you please tell me what size is that icon? it should be 12px, but it looks smaller

@VladUXUI It's exactly 12x12 pixels (to my surprise - it looks small to me as well). Maybe looks smaller because it's faint?

VladUXUI:
I think the issue is that we are using the wrong icon here.
External account has two icons. a 24px one and 12 px one. I think we are using the 24px one and resizing it.
https://github.com/tallycash/extension/blob/main/ui/public/images/external%402x.png

This is the icon that should be used
external-L

cairomassimo:
> This is the icon that should be used

I fixed it, thanks!
The icon is shown without colors though, I assume it's a setting of the shared button.

Screenshot from 2022-02-03 18-39-19

ui/components/SignTransaction/SignTransactionSignInfoProvider.tsx
```diff
@@ -0,0 +1,153 @@
...
+  return (
+    <SignTransactionBaseInfoProvider
+      title="Sign Transaction"
+      infoBlock={
+        <div className="sign_block">
+          <div className="container">
+            {typeof transactionDetails.to === "undefined" ? (
+              <>
+                <div className="label">Send to</div>
+                <div className="send_to">Contract creation</div>
```
wojciech-turek, line 56: If its a contract creation should we say "Send to" ?

ui/components/SignTransaction/SignTransactionSignInfoProvider.tsx
```diff
@@ -0,0 +1,153 @@
...
+              </>
+            ) : (
+              <>
+                <div className="label">Send to</div>
+                <div className="send_to">
+                  {transactionDetails.to.slice(0, 6)}...
+                  {transactionDetails.to.slice(-4)}
```
wojciech-turek, line 63: Please use shared function `truncateAddress`

ui/components/SignTransaction/SignTransactionSwapAssetInfoProvider.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <div className="asset_items_wrap">
+            <div className="asset_item">
+              <SharedAssetIcon size="small" />
+              <span className="asset_name">ETH</span>
+            </div>
+            <div className="icon_switch" />
+            <div className="asset_item">
+              <span className="asset_name">ETH</span>
+              <SharedAssetIcon size="small" />
+            </div>
```
wojciech-turek, line 30: Is this some placeholder for the swaps sign transaction?

ui/components/SignTransaction/SignTransactionTransferInfoProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+}): ReactElement {
+  return (
+    <SignTransactionBaseInfoProvider
+      title="Sign Transfer"
+      confirmButtonLabel="Sign"
+      infoBlock={
+        <div className="sign_block">
+          <div className="container">
+            <div className="label">Send to</div>
+            <div className="send_to">{truncateAddress(destination)}</div>
```
wojciech-turek, line 33: The className is `send_to`... (continued in the following comment)

ui/components/SignTransaction/SignTransactionTransferInfoProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+                opacity: 60%;
+                background-color: var(--green-120);
+              }
+              .container {
+                display: flex;
+                margin: 20px 0;
+                flex-direction: column;
+                align-items: center;
+              }
+              .send-to {
```
wojciech-turek, line 76: ... but the style is `send-to`

ui/components/TransactionDetail/TransactionDetailItemLong.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { ReactElement, ReactNode } from "react"
+
+export default function TransactionDetailItemLong({
```
wojciech-turek, line 3: This component and `TransactionDetailItemShort` share a lot of code, maybe they can be merged together and just a different prop should be passed indicating whether its long or short?

wojciech-turek (commented):
Left some small suggestions and comments but overall the code looks good.

I find it more complicated and harder to understand with this approach, but it works as intended.

I have tested signing send transaction, dapp swapp sign, typed data sign and approve. All of those render correctly and with correct data.

Looking forward to merging this after small suggestions are resolved.

ui/components/SignTransaction/SignTransactionSwapAssetInfoProvider.tsx
cairomassimo: It is, I suppose. The code was just moved from `SignTransactionSwapAssetBlock.tsx` (now deleted).

Shadowfiend (commented):
My overall take here is that we're doing a lot of stuff in components that we could be doing in bare functions, and we're doing a lot of stuff in individual props that we could be doing in packaged objects of related data. The result is a set of logic that's much harder to follow than I think it could be.

Here's the fundamental problem I see the `SignTransaction` component needs to solve:

```
-------------------
| common header   |
||---------------||
|| signer header ||
||   (optional)  ||
||---------------||
|                 |
| common body     |
|                 |
|-----------------|
|  signer footer  |
|-----------------|
```

Where the signer header has the Ledger connection information.

The common body in turn is in charge of presenting the transaction data, independent of signer.

The thing that is tripping us up and turning this into a convoluted journey is the requirement Ledger has of displaying a sign screen that shows the data as it would present on the Ledger. I would argue that this screen is signer-specific, and has little place in the above structure. Instead, it's a modal—different than the others we have, but a modal nonetheless. It blocks all interaction with the wallet, and it's specific to the current interaction with the Ledger. It shares some components and styling, but is semantically separate from the rest of the signing structure. I think if we accept some light code duplication and make that piece its own component, we're likely to get an extreme simplification of control flow that will make the tradeoff worth it.

I've got a more detailed idea of how to do this in a way that's cleaner and easier to follow, but no time to write it up all the way right now. In the meantime, I'm cool with this landing as-is to keep Ledger work moving.

cairomassimo:
@Shadowfiend, @wojciech-turek thanks for the review!

@Shadowfiend, I share the sentiment that this could be simplified further.
Right now the main issue is not so much shared code as shared _state_, such as here:
https://github.com/tallycash/extension/blob/c902b51cfb7ed29955f90764e7804ed4df488dff/ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx#L31-L34

(actually, it _will be_ shared later on, when we show the estimated gas fee in the "Awaiting..." screen)

This state could be moved to Redux at some point, making it much easier to simplify the code in this PR. Here I'm mostly moving around existing code, changing as little as possible...

@wojciech-turek I think most of your comments apply to existing code that was moved around. I'll double-check that, and either fix them (if I manage to test the result) or leave them as they are and add a GH issue.

This PR is not blocking for Ledger support, I'll come back to it when other stuff has settled...

Shadowfiend:
I think we can package the shared state up along clear domain lines, not unlike what we've done with the network settings selector.

That said, all of this *does* exist in Redux, as the transaction construction data—because updating this value must update the transaction data, or it won't take effect. What hasn't been done yet is setting up a clean pipeline from transaction data -> transaction screen -> editing approval -> transaction data update/enrichment -> transaction screen. This is largely because the transaction construction slice has evolved haphazardly in the last 3-4 months, starting before a lot of the current components like enrichment were in place, and needs to be refactored a bit (the direction in #897 being part of that) to reflect the true intended use and data flow in the current codebase/feature set.

Anyway… More thinking + iteration to be done on this component for sure. Happy to wait on it if you think that's the right call, but want to be clear that I'm also happy to move the ball forward instead of seeking perfection here.

ui/components/SignTransaction/SignTransactionTransferInfoProvider.tsx
cairomassimo: Fixed!

ui/components/SignTransaction/SignTransactionSignInfoProvider.tsx
cairomassimo: Issue is not new. Deferred to #955.

ui/components/SignTransaction/SignTransactionSignInfoProvider.tsx
cairomassimo: Issue is not new. Deferred to #955.

cairomassimo:
I think I addressed the blocking comments @wojciech-turek, @Shadowfiend. Let's merge it if it looks good to you.

greg-nagy (approved):
Seems to be good to go 🤘

Could not test without the ledger wiring, so the QA there will include this part as well. 

Didn't see anything that should be a blocker amond the existing reviews.

PR #935, opened 2022-02-02
Introduce working SigningService to wrap the chaos on the bottom legs

### LINKED ISSUE #919: Add SigningService on top of LedgerService and KeyringService
Based on the corresponding [RFC](https://coda.io/d/Minimum-Viable-Launch_dznBnP8Wtkc/RFC-Hardware-wallet-signers_susPg#_lugiK) and discussions on Flowdock.

fernya:
Closes #919 

Keep this short just to have the possibility to integrate with the UI!

# Goals
* If the feature flag is disabled then the old code path shall be used.
* If the feature flag is enabled then the `deriveAddress` and `signTransaction` calls should be routed through `SignatureService`

# Key changes in this PR
 * The nonce acquisition/release is moved from the main wiring functions to the `SigningService`
 * This leads to the fact that transactions routed through `LedgerService::signTransaction` are finally initialized properly
 * A proposed routing between sub-services is introduced:
   * Calls using addresses: listening to sub-services emitted "address" events and keeping track which ones should be called
   * Calls using accountIDs: FIXME prefix the original accounts with unique IDs (see SignerType) and decide based on those.
   *


background/services/signing/index.ts
```diff
@@ -37,19 +54,64 @@ export default class SigningService extends BaseService<Events> {
...
+    }
 
-    throw new Error("Unimplemented")
+    if (signerID.startsWith("keyring-")) {
+      return this.keyringService.deriveAddress(
+        signerID.substring("keyring-".length)
+      )
+    }
+
+    throw new Error(`Unknown signerID: ${signerID}`)
```
cairomassimo, line 70: Here we are hiding type information into a custom string format. It should be either: `signerID: { type: 'keyring', ... } | { type: 'ledger', deviceID: string, path: string }`, or not having this method at all and relying on keyring/ledger specific implementations instead.

However, it is not a problem with this PR specifically (so not blocking, from my point of view), but with the overall design:
_Should we abstract away the specific mechanism for address derivation, when the UI interaction is completely different between the two mechanisms that we have?_ IMO, it's no good. @Shadowfiend?

background/services/signing/index.ts
```diff
@@ -37,19 +54,64 @@ export default class SigningService extends BaseService<Events> {
...
+          throw new Error(
+            `Unknown address (${address}) or handler (${actualHandler}) provided!`
+          )
+      }
+    } catch (exception) {
+      logger.error("Error signing transaction; releasing nonce", exception)
+      this.chainService.releaseEVMTransactionNonce(transactionWithNonce)
+    }
+
+    throw new Error("Impossible to reach")
```
cairomassimo, line 110:
```suggestion
    } finally {
      this.chainService.releaseEVMTransactionNonce(transactionWithNonce)
    }
```

No point in catching the original exception and re-throw our own.
I assume the logging is done elsewhere anyway. If not, then it should be added higher up in the call hirearchy.

background/services/signing/index.ts
```diff
@@ -37,19 +54,64 @@ export default class SigningService extends BaseService<Events> {
...
+        case "keyring":
+          return await this.keyringService.signTransaction(
+            transactionRequest.from,
+            transactionWithNonce
+          )
+        default:
+          throw new Error(
+            `Unknown address (${address}) or handler (${actualHandler}) provided!`
+          )
+      }
```
cairomassimo, line 104: (Not blocking) I would extract this block into its own method `signTransactionWithNonce`, to avoid deep nesting and the multiple `return await` constructs (only needed because we are in a `try` block, otherwise just `return` would be fine).

background/main.ts
```diff
@@ -631,6 +647,16 @@ export default class Main extends BaseService<never> {
...
   }
 
+  async connectSigningService(): Promise<void> {
+    this.keyringService.emitter.on("address", (address) =>
+      this.signingService.addTrackedAddress(address, "keyring")
+    )
+
+    this.ledgerService.emitter.on("address", (obj) =>
+      this.signingService.addTrackedAddress(obj.address, "ledger")
+    )
```
cairomassimo, line 657: By using desctructuring you can avoid the `obj` variable

```suggestion
    this.ledgerService.emitter.on("address", ({ address }) =>
      this.signingService.addTrackedAddress(address, "ledger")
    )
```

Otherwise, choose either: a more specific name (e.g. `addressData`, `account`, etc., depending of the name of the type) or a better generic name (e.g. `event`).

background/main.ts
```diff
@@ -694,7 +721,12 @@ export default class Main extends BaseService<never> {
     })
 
     keyringSliceEmitter.on("deriveAddress", async (keyringID) => {
-      await this.keyringService.deriveAddress(keyringID)
+      if (HIDE_IMPORT_LEDGER) {
+        await this.keyringService.deriveAddress(keyringID)
+      } else {
+        const signerID = `keyring-${keyringID}`
+        await this.signingService.deriveAddress(signerID)
+      }
```
cairomassimo, line 729: I think we should use `signingService` anyway here (i.e., without the conditioning on the feature flag).
If it breaks something, we have the chance to test it and fix it now. (We are not in code freeze yet.)

cairomassimo (changes requested):
Some minor stylistic suggestions.

Overall I'm not sure about the abstraction provided by `SigningService`, but let's not block this PR. (Anyway, my concern would be better addressed when everything in Ledger-land is merged)

background/main.ts
fernya: Fixed, thanks!

background/services/signing/index.ts
fernya: Restructured, thanks!

background/services/signing/index.ts
```diff
@@ -18,17 +33,24 @@ type Events = ServiceLifecycleEvents
  * - xxx
  */
 export default class SigningService extends BaseService<Events> {
+  addressHandlers: AddressHandler[] = []
```
greg-nagy, line 36: This could be a private variable.

greg-nagy (approved):
🚢

PR #938, opened 2022-02-03
🎬 Quick Start: Add default API keys for Alchemy and BlockNative

### LINKED ISSUE #793: PRs from other forks fail CI, even if the CI run is approved

greg-nagy:
# What

Added default API keys for BlockNative and Alchemy in the `.env.defaults` file.

# Why

- To remove the necessity of creating own API keys to build the project
- To solve the CI build issue in case PRs coming from forked repos

# How

- Created a new app for alchemy for this key
  - Alchemy does not have daily request limits, only rate-limiting. So multiple devs using the same key should not affect each other
  - They have generous compute unit/month limits which should be more than enough. If it becomes an issue we can change it. 
- Created a new registration with my email address for this API key

Closes  #793


mhluongo (approved):
Yep yep  yep 👍

PR #939, opened 2022-02-03
🐧 Fix CI build for PRs from forked repos

greg-nagy:
Set default API key values in CI for PR builds.


Shadowfiend (dismissed):
This doesn't seem right -- https://www.npmjs.com/package/dotenv-webpack says nothing about only working on certain platforms, and the build would have failed due to the [`safe` setting](https://github.com/tallycash/extension/blob/main/webpack.config.ts#L67) if the env vars weren't set.

What about this wasn't working? (Of note: that info should have been in the PR description and commit message :wink:)

greg-nagy:
Correct, my bad. This fix does not make sense at first. I came up with it through trial and error. Let me share how it went:

- We have this PR https://github.com/tallycash/extension/pull/884 which is not getting merged because the CI build is failing 
- This was the error: https://github.com/tallycash/extension/runs/5046647551?check_suite_focus=true - error message
- Fresh clone, `yarn install`, `yarn build`
```
Failed to load ./.env.
[webpack-cli] Error: Missing environment variable: BLOCKNATIVE_API_KEY
    at /Users/greg/tally/ci-test/node_modules/dotenv-webpack/dist/index.js:90:17
    at Array.forEach (<anonymous>)
    at Dotenv.gatherVariables (/Users/greg/tally/ci-test/node_modules/dotenv-webpack/dist/index.js:85:30)
    at Dotenv.apply (/Users/greg/tally/ci-test/node_modules/dotenv-webpack/dist/index.js:63:28)
    at createCompiler (/Users/greg/tally/ci-test/node_modules/webpack/lib/webpack.js:74:12)
    at /Users/greg/tally/ci-test/node_modules/webpack/lib/webpack.js:44:48
    at Array.map (<anonymous>)
    at createMultiCompiler (/Users/greg/tally/ci-test/node_modules/webpack/lib/webpack.js:44:33)
    at create (/Users/greg/tally/ci-test/node_modules/webpack/lib/webpack.js:118:16)
    at webpack (/Users/greg/tally/ci-test/node_modules/webpack/lib/webpack.js:135:47)
```
- Played with the `safe` settings locally, and turning it off fixed the problem, but I was still curious and wanted to keep the flag `true` because it makes sense in dev time
- Kept the `safe: true` and added the API keys to the .env.default. And it worked (on my mac).
  - haveing a throwaway API key setup as default would be good anyway because it would make it easier to start working on 
  the project
  - keeping the `safe:true` is also a win 
  - worked on my mac
- Then this PR happened https://github.com/tallycash/extension/pull/938
- With the new PR merged I got the same error: https://github.com/tallycash/extension/runs/5052314520?check_suite_focus=true
- So I was, fudge, it works locally, does not work in the CI. What are my options: 
  - Let the CI block all outside PRs from being merged OR do the merge manually
  - Set `safe:false` (and revert the API key changes)
  - or add a single line to the CI which 
    - solves the problem that is: there is no `.env` file in the CI context
    - lets us have default API keys 
    
I think this current setup is best for all aspects. Did check the pricing and limitations for the 2 products and our usages, and I think it would not cause us any trouble. If it does at one point its effect will be minimal. It will affect only those who are using the commonly shared keys. And it will be an easy fix because we just need to create another key.

Shadowfiend:
So the actual issue here is tucked away in a collapsed section heh:

```
yarn build
shell: /usr/bin/bash -e {0}
env:
  ALCHEMY_KEY: 
  BLOCKNATIVE_API_KEY: 
yarn run v1.22.17
$ webpack --mode=production
Failed to load ./.env.
[webpack-cli] Error: Missing environment variable: ALCHEMY_KEY
```

From [the docs](https://github.com/mrsteele/dotenv-webpack#properties):

```
      systemvars: true, // load all the predefined 'process.env' variables which will trump anything local per dotenv specs.
```

The handling inside dotenv of systemvars is:

```
        var value = Object.prototype.hasOwnProperty.call(vars, key) ? vars[key] : env[key];
```

So I think what's happening here is that we're deliberately setting the env vars in the actual environment to be blank, which is not what we want.

I think to actually fix this, we need to add a fallback in the build in the [two](https://github.com/tallycash/extension/blob/main/.github/workflows/build.yml#L27-L28) [places](https://github.com/tallycash/extension/blob/main/.github/workflows/build.yml#L33-L34) where the keys are resolved. We can use `${{ secrets.VARIABLE || "default-value" }}` to do this.

greg-nagy:
As it turns out they also have a flag `allowEmptyValues` which results in this behavior. My expectation was that if the env variable is '' then it will use the value from the `.env.defaults`. Now that I think about it, it actually makes a lot of sense. 

Thanks.

Appreciate the links and the thoughts.

greg-nagy:
Now I have set the same default keys for both places but this will result in the default keys baked into our latest build on main after an external PR. 

Based on our workflow it's not likely that before a release the last merge is a merge from an outside PR and it's better not to put our prod keys in the source code.

Shadowfiend:
Latest build on main should always run with our secrets, only the PR builds should use the alternates (as far as I understand it).

Shadowfiend:
Actually I'm dumb. We should only use the alternates in PR builds. The main build should fail if secrets are unavailable or blank. Sorry.

Shadowfiend:
Maybe we also add a comment in the PR section indicating that the defaults are for external PRs?

greg-nagy:
I think this should work like that.

greg-nagy:
Since we are here: I would upload the build artifacts for every PR so it's easier to QA. What do you think about it?

Shadowfiend:
It makes me uncomfortable but I can't really articulate it, so I'm going with your take 😉

.github/workflows/build.yml
```diff
@@ -24,8 +24,14 @@ jobs:
...
-        ALCHEMY_KEY: ${{ secrets.DEV_ALCHEMY_API_KEY }}
-        BLOCKNATIVE_API_KEY: ${{ secrets.DEV_BLOCKNATIVE_API_KEY }}
+        ALCHEMY_KEY: ${{ secrets.DEV_ALCHEMY_API_KEY || "oV1Rtjh61hGa97X2MTqMY9kEUcpxP-6K" }}
+        BLOCKNATIVE_API_KEY: ${{ secrets.DEV_BLOCKNATIVE_API_KEY || "f60816ff-da02-463f-87a6-67a09c6d53fa" }}
+    - name: Upload build on PRs
+      if: github.ref != 'refs/heads/main'
+      uses: actions/upload-artifact@v2
+      with:
+        name: pr-extension-builds
+        path: dist/*.zip
```
Shadowfiend, line 34: I think we'll probably be able to just unify the stuff since we're doing it for both builds. Something for a later PR.

Shadowfiend (dismissed):
Let's go 💪

Shadowfiend:
Bah. Single quote strings. Who does that...

Shadowfiend (approved):

PR #941, opened 2022-02-03
Max asset input

wojciech-turek:
Move the max balance selection from the Send page to the shared asset input component.
Made that an optional parameter so in case the input is a destination type we do not show it.

The balance will be re-fetched on asset change if the input allows it.


Shadowfiend:
Closed in favor of #943.

PR #943, opened 2022-02-04
Asset Amounty Input: Alternate approach to SharedAssetInput max/balance info

Shadowfiend:
This PR is a bigger rework than I originally embarked on. It
implements balance info on the shared asset input, as well as
in the asset selector, and includes a max button that can be
disabled if desired (used in swaps for buy assets).

There were a couple of tangential changes included here:

- The fixed-point library's string handling is much more robust.
- `SharedAssetInput`'s “now for something completely different” option,
  `isTypeDestination`, is gone in favor of custom styling in the Send
  component.

The primary changes to `SharedAssetInput` are one big commit
(should really have been at least 2-3, but time ran out). Here's
the relevant commit message, which covers the details:

```
SharedAssetInput now takes an assetsAndAmounts prop, which is of a new
type, AnyAssetWithOptionalAmount (defined in SharedAssetItem for now).
This type can contain either an asset, or an asset with associated
amount and localized decimal information.

This then cascades into a few additional bits of functionality:

- SharedAssetItem now displays the associated amount in the dropdown if
  available.
- SharedAssetInput now displays the associated amount for the selected
  asset (if any) as a balance.
- SharedAssetInput now displays, by default, a "Max" button to set the
  input amount to the associated amount for the selected asset.

Additionally, the SharedAssetInput now parses the inputted amount using
fixed-point parsing and compares the fixed-point amount to the max
balance using fixed-point comparisons instead of potentially-lossy
JS number parsing.
```

----

Opening as a draft to decide between this approach and the one in #941.

If we stick with 941, I'll drop the big change commit so we can still
land the remaining tidbits in this PR.

### DIFF
--- a/background/lib/fixed-point.ts
+++ b/background/lib/fixed-point.ts
@@ -170,17 +170,33 @@ export function parseToFixedPointNumber(
   }
 }
 
-export function fixedPointNumberToString({
-  amount,
-  decimals,
-}: FixedPointNumber): string {
+/**
+ * Converts a fixed point number with a bigint amount and a decimals field
+ * indicating the orders of magnitude in `amount` behind the decimal point into
+ * a string in US decimal format (no thousands separators, . for the decimal
+ * separator).
+ *
+ * Note that for UI usage, it is _usually_ a better strategy to use
+ * ``fromFixedPoint`` to convert to a JavaScript number and then use an
+ * Intl.NumberFormat to format the number unless that is not an available
+ * option or precision is critical. This function currently does _not_ respect
+ * localization settings.
+ */
+export function fixedPointNumberToString(
+  { amount, decimals }: FixedPointNumber,
+  trimTrailingZeros = true
+): string {
   const undecimaledAmount = amount.toString()
   const preDecimalLength = undecimaledAmount.length - decimals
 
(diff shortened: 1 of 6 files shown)


wojciech-turek:
Some notes: 

- I second to the idea of dropping the `isTypeDestination` from SharedAssetInput I was thinking about it, but considered it out of scope for #941. 
- The units conversion were a big pain for me and using the shared conversion functions are imo they way to go however at the moment they seem to be broken for amounts that are full ints like `13.0000...`. There's an error when selecting max balance for those.
image
- Not for this PR but we can probably make a separate component for destination input in case we need it in other places, but I guess we can leave that out until that moment comes.

Shadowfiend:
> Not for this PR but we can probably make a separate component for destination input in case we need it in other places, but I guess we can leave that out until that moment comes.

Yeah, I think the right thing to do is wait until it's used in multiple places. Right now I'm not sure we'll have a lot of places that take a direct address input in a context that appears alongside an asset selector input—unsure though.

wojciech-turek (dismissed):
I was looking for silver, but found gold instead.

wojciech-turek (approved):
Like I said, gold.

PR #944, opened 2022-02-04
Fix logic error in `SharedSlideUpMenu`

cairomassimo:
Error was due to a faulty merge, and caused incorrect
slide-up animations when signing with Ledger.


greg-nagy (approved):
🐶

greg-nagy (approved):
grrr

PR #946, opened 2022-02-04
Background earn

wojciech-turek:
In order to interact with Earn vault I have reacted a separate redux slice that will handle all the interaction with the vaults and approvalTarget contracts.

TODOs

- [ ] get the r, s, v values from earnSlice when we call vault deposit
- [ ] add selector to check if specific vault is approved in the frontend
- [ ] add selector to check if a vault is eligible to call permit instead of approve
- [ ] check allowance of the approvalTarget on input change in the deposit amount
- [ ] check if permit is still valid (unsusre if a permit that has its deadline still valid can be reused multiple times?)
- [ ] do a test of the approval flow using any erc-20 token (button state change after approval tx is mined)

### DIFF
--- a/background/lib/vault.ts
+++ b/background/lib/vault.ts
@@ -1,5 +1,4 @@
-// eslint-disable-next-line import/prefer-default-export
-export const VAULT_ABI = [
+const VAULT_ABI = [
   {
     inputs: [
       {
@@ -992,3 +991,5 @@ export const VAULT_ABI = [
     type: "function",
   },
 ]
+
+export default VAULT_ABI
--- a/background/lib/vault.ts
+++ b/background/lib/vault.ts
@@ -1,16 +1,51 @@
 // eslint-disable-next-line import/prefer-default-export
 export const VAULT_ABI = [
   {
-    anonymous: false,
     inputs: [
       {
-        indexed: false,
         internalType: "address",
-        name: "token",
+        name: "vault_",
         type: "address",
       },
       {
-        indexed: false,
+        internalType: "address",
+        name: "want_",
+        type: "address",
+      },
+      {
+        internalType: "uint256",
+        name: "targetManagementFee_",
+        type: "uint256",
+      },
+      {
+        internalType: "address",
+        name: "approvalTarget_",
+        type: "address",
+      },
+      {
+        internalType: "uint256",
+        name: "rewardDuration_",
+        type: "uint256",
+      },
+      {
+        internalType: "address",
+        name: "rewardToken_",
+        type: "address",
+      },
+      {
+        internalType: "address",
+        name: "feeRecipient_",
+        type: "address",
+      },
+    ],
+    stateMutability: "nonpayable",
+    type: "constructor",
+  },
+  {
+    anonymous: false,
+    inputs: [
+      {
+        indexed: true,
         internalType: "address",
         name: "recipient",
         type: "address",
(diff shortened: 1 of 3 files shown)


greg-nagy:
Lookin' hot 🔥

background/redux-slices/earn.ts
```diff
@@ -0,0 +1,255 @@
...
+  getProvider,
+  getSignerAddress,
+} from "./utils/contract-utils"
+
+export type EarnState = {
+  signature: {
+    r: string
+    s: string
+    v: number
+  }
```
cairomassimo, line 26: Don't we have a shared type for this? If not, we probably should add one.

background/redux-slices/earn.ts
```diff
@@ -0,0 +1,255 @@
...
+        amount,
+        signerAddress,
+        signerAddress,
+        amount,
+        (await provider.getBlock(provider.getBlockNumber())).timestamp +
+          3 * HOUR
+
+        // r
+        // s
+        // v
```
cairomassimo: What's the meaning of these comments? If they are TODOs please say so :)

background/redux-slices/earn.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+)
+
+const earnSlice = createSlice({
+  name: "earn",
+  initialState,
+  reducers: {
+    saveSignature: (
+      state,
+      { payload: { r, s, v } }: { payload: { r: string; s: string; v: number } }
```
cairomassimo, line 152: Could re-use a shared type for the signature here

background/redux-slices/earn.ts
```diff
@@ -0,0 +1,255 @@
...
+      if (
+        typeof tx.r !== "undefined" &&
+        typeof tx.v !== "undefined" &&
+        typeof tx.s !== "undefined"
+      ) {
+        const signature = {
+          r: tx.r,
+          s: tx.s,
+          v: tx.v,
+        }
```
cairomassimo, line 211: Minor: I this destructuring here would be cleaner:
```ts
const { r, v, s } = tx;

if (typeof r !== "undefined" && ...) {
  earnSlice.actions.saveSignature({ r, v, s })
}
```

background/lib/vault.ts
```diff
@@ -0,0 +1,995 @@
+const VAULT_ABI = [
```
cairomassimo, line 1: Unless it's obvious to everyone else, I would add a comment here explaining how this file was obtained/generated. An URL would be enough.

background/redux-slices/earn.ts
```diff
@@ -0,0 +1,327 @@
...
+    saveSignature: (
+      state,
+      { payload: { r, s, v } }: { payload: { r: string; s: string; v: number } }
+    ) => ({
+      ...state,
+      signature: {
+        r,
+        s,
+        v,
+      },
```
cairomassimo, line 159: _Really_ minor: it looks better on a single line (and Prettier should allow it)
```suggestion
      signature: { r, s, v },
```

cairomassimo (commented):
Just some minor (and non-blocking) code-style comments!

background/redux-slices/earn.ts
cairomassimo: (I think they were replaced in a later commit, so scratch this comment)

background/redux-slices/earn.ts
wojciech-turek: 100%

background/redux-slices/earn.ts
wojciech-turek: Yess, few lines of code less, thanks!

background/redux-slices/earn.ts
wojciech-turek: It does look better! :)

greg-nagy (approved):
I don't see any blocker so let's ship this 🚀

> activities tab for the moment to see it pending and after its done go back to the vault page

This didn't work for me, but it's not a blocker.

greg-nagy (approved):
🚀

PR #947, opened 2022-02-04
🆘 Revert ledger sign transaction container buttons

henryboldi:
I believe this is where that Ledger info popped up on the swap sign page

@Shadowfiend

### DIFF
--- a/ui/components/SignTransaction/SignTransactionContainer.tsx
+++ b/ui/components/SignTransaction/SignTransactionContainer.tsx
@@ -63,30 +63,17 @@ export default function SignTransactionContainer({
             >
               Reject
             </SharedButton>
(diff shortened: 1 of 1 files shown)


henryboldi:
Well, so there's a build error 😬

greg-nagy:
@henryboldi https://github.com/tallycash/extension/pull/944 This is the PR that trips you over

Not sure what is a good fix here.

henryboldi:
Reverting both didn't do the trick. I can dig in after the community call. I'll probably add better use of the feature flag instead of reverting

greg-nagy:
Always fall forward! :)

Shadowfiend (approved):
Looks good :thumbsup:

PR #948, opened 2022-02-05
Exploratory: Previewing ERC-20 Token Transactions on Smart Contract Interaction

0xDaedalus:
This is a very naive implementation of how we can preview ERC-20 tokens  to be sent when interacting with Uniswap per suggestion from @wojciech-turek.

Currently when we interact with uniswap (and probably any dex) the "Spend Amount" shown on our Sign Transaction page is `0` for any asset except ETH.  For Uniswap specifically - this happens because the swap is a `contract-interaction` whose inputs we do not know how to parse.

### At a high level - what is going on is

1. Transaction Signature Request comes in.
2. We attempt to parse the input data using the uniswap v3 abi to see if it is a uniswap v3 contract function
3. If it is - we assume that is a call to `multicall` - and attempt to parse the arguments to `multicall`.
4. If one of the functions called by `multicall` is `exactInputSingle` or `exactOutputSingle` - assume the interaction is a swap - and enrich enrich transaction annotation with amount being swapped.

The implementation as-is has a number of limitations.
• It only works with assets that actually have a pool - swaps that require multiple-hops will not be parsed correctly.
• It is tightly coupled to uniswap (by virtue of relying on the uniswap abi) - I'm not sure if there is a more general solution possible but I imagine the input is un-parseable without relying on the abi.
• It assumes that the inital transaction is always a `multicall` transaction - this may be because I am testing on a wallet that has no assets approved on uniswap.
• This pattern of parsing contract-interaction amounts likely relies on an integration-per-dapp - this is not ideal.
• Many shortcuts were taken with typing / linter rules

image

If this is the direction we want to head in - I am happy to push forward and take this PR into a mergeable state.  

This PR is brought to you by virtue of accidentally drinking 4 shots of espresso assumed to be one cup of coffee ☕ ☕

### DIFF
--- a/background/lib/uniswap.ts
+++ b/background/lib/uniswap.ts
@@ -589,7 +589,6 @@ export async function parseUniswapTx(
       return undefined
     }
   } catch (err) {
-    console.log(err)
     return undefined
   }
   return undefined
--- a/background/lib/uniswap.ts
+++ b/background/lib/uniswap.ts
@@ -1,0 +1,596 @@
+import { AlchemyProvider } from "@ethersproject/providers"
+import { BigNumber, ethers, logger } from "ethers"
+import { getNetwork } from "@ethersproject/networks"
+import { getTokenMetadata } from "./alchemy"
+import { getEthereumNetwork } from "./utils"
+import { SmartContractFungibleAsset } from "../assets"
+import { HexString } from "../types"
+
+const UNISWAP_ABI = [
+  {
+    inputs: [
+      { internalType: "address", name: "_factoryV2", type: "address" },
+      { internalType: "address", name: "factoryV3", type: "address" },
+      { internalType: "address", name: "_positionManager", type: "address" },
+      { internalType: "address", name: "_WETH9", type: "address" },
+    ],
+    stateMutability: "nonpayable",
+    type: "constructor",
+  },
+  {
+    inputs: [],
+    name: "WETH9",
+    outputs: [{ internalType: "address", name: "", type: "address" }],
+    stateMutability: "view",
+    type: "function",
+  },
+  {
+    inputs: [{ internalType: "address", name: "token", type: "address" }],
+    name: "approveMax",
+    outputs: [],
+    stateMutability: "payable",
+    type: "function",
+  },
+  {
+    inputs: [{ internalType: "address", name: "token", type: "address" }],
+    name: "approveMaxMinusOne",
+    outputs: [],
+    stateMutability: "payable",
+    type: "function",
+  },
+  {
+    inputs: [{ internalType: "address", name: "token", type: "address" }],
+    name: "approveZeroThenMax",
+    outputs: [],
+    stateMutability: "payable",
+    type: "function",
+  },
+  {
+    inputs: [{ internalType: "address", name: "token", type: "address" }],
+    name: "approveZeroThenMaxMinusOne",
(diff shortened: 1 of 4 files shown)


mhluongo:
Comments on your comments!

> I'm not sure if there is a more general solution possible but I imagine the input is un-parseable without relying on the abi.

There are more than one! A couple approaches to expand on this

* We can plug in to a variety of ABI databases we've been slowly putting together. Here's one [of my favorites](https://github.com/tintinweb/smart-contract-sanctuary).
* We can track contracts per network by address, and grow metadata over those over time. This is a little handwavey, but the basic idea is to crowdsource annotations beyond just ABIs.
* We can use a service to do [EVM tracing of transactions](https://geth.ethereum.org/docs/dapp/tracing). The idea here is to break down the contract call into sub-calls, then do inference at each level. Of course, the problem is we need to be able to do these *before* sending a transaction... which is why we've been asking for a partner to solve this! And BlockNative has, exposing a [tx preview API](https://www.blocknative.com/blog/ethereum-transaction-preview-api) that will give us the most info about a tx possible.

background/lib/uniswap.ts
```diff
@@ -0,0 +1,595 @@
+import { AlchemyProvider } from "@ethersproject/providers"
+import { BigNumber, ethers, logger } from "ethers"
+import { getNetwork } from "@ethersproject/networks"
+import { getTokenMetadata } from "./alchemy"
+import { getEthereumNetwork } from "./utils"
+import { SmartContractFungibleAsset } from "../assets"
+import { HexString } from "../types"
+
+const UNISWAP_ABI = [
```
mhluongo, line 9: Any early thoughts on how we scale ABI inclusion?

background/lib/uniswap.ts
```diff
@@ -0,0 +1,595 @@
...
+    type: "function",
+  },
+  { stateMutability: "payable", type: "receive" },
+]
+
+const UNISWAP_INTERFACE = new ethers.utils.Interface(UNISWAP_ABI)
+
+const ALCHEMY_KEY = process.env.ALCHEMY_KEY // eslint-disable-line prefer-destructuring
+
+const alchemyProvider = new AlchemyProvider(
```
mhluongo, line 500: Let's stop doing this, and instead either take an `AlchemyProvider` as a param or get it from the `ChainService`.

It starts getting absolutely impossible when we go multi-network. Give [Rovin' Kovan](https://github.com/tallycash/extension/pull/732) a read to see what I mean

background/services/enrichment/index.ts
```diff
@@ -241,6 +243,21 @@ export default class EnrichmentService extends BaseService<Events> {
         }
       } else {
         // Fall back on a standard contract interaction.
+
+        // If we are interacting with Uniswap
+        if (transaction.to === "0x68b3465833fb72a70ecdf485e0e4c7bd8665fc45") {
```
mhluongo, line 248: We should also check the network here

mhluongo (commented):
Exploratory code gets exploratory comments!

mhluongo:
@0xDaedalus should we nix this, or do you still have plans for the draft?

Shadowfiend:
Closing this for now. There's some promising stuff here, but ultimately we expect it to be superseded by a possibly-server-side simulations approach, rather than this ABI-introspection approach.

PR #956, opened 2022-02-08
Disable `ledger` slice with feature flag

### LINKED ISSUE #934: Mark the ledger slice as not release ready
> But the redux will be tricky I think. I would probably go with a draft naming convention for the slice DRAFTledger and when we release rename it and create a clean-up migration. But that’s really a good solution. But if put the reducers behind a feature flag everything that depends on that will break.

This is the best idea I have had so far, but this is a really bad idea. Open to better solutions.

cairomassimo:
Closes #934.

Solution: include the `ledger` reducer, as part of the call to `combineReducers`, only when Ledger feature is enabled.

Interestingly, this results in **no typing errors** whatsoever.
This is due to the fact that the Redux state for a slice is always allowed be `undefined` (as it is at the very beginning), and Redux Toolkit takes care of initializing it to `initialState` if necessary when a reducer is invoked.

Currently, no actions should ever be dispatched to this slice, when it is disabled, since Ledgers cannot be imported and the Ledger service is not started when the feature flag is disabled.

If an action is dispatched nonetheless, the runtime behavior is currently unknown. However, it should never result in a persisted change to the Redux state, since the root reducer returned by `combinedReducers` (which is now unaware of the concept of slices) has no way of introducing the `ledger` slice by itself.


cairomassimo:
BTW, this approach was used in #946 as well:
https://github.com/tallycash/extension/blob/691e0277d30506896757eca035f6461aa755c8e0/background/redux-slices/index.ts#L28

greg-nagy (approved):
Happy that this worked! 

If we do the same trick with the UI then the selectors will cause compilation error. 
`const snackbarMessage = useBackgroundSelector(selectSnackbarMessage)` in the `Snackbar.tsx. 

The common pattern when the problem arises is when the `useBackgroundSelector()` gets an inherited selector fn `createSelector`

But we don't need to solve that problem now.

greg-nagy:
> BTW, this approach was used in https://github.com/tallycash/extension/pull/946 as well:

🙌

Yeah, let's stick with this pattern for now and see when we hit a roadblock

greg-nagy:
> If an action is dispatched nonetheless, the runtime behavior is currently unknown. 

The action will probably be dispatched because it's exported from the return value of the `createSlice` so it will exist and not throw an error, but will not trigger any change to the state because there is no reducer who is listening for this action. Probably will be seen by middlewares but that should not cause any error.

Issue #774, opened 2021-12-29
Don't allow a transaction to be signed if an EVM address with mixed caps doesn't pass the checksum


mhluongo:
![Screenshot from 2021-12-29 11-13-08](https://user-images.githubusercontent.com/427505/147682245-f44d8dbd-4814-4d28-a779-6d0d57a2bd54.png)

Mixed-case hex addresses imply following [EIP-55](https://eips.ethereum.org/EIPS/eip-55), which can help prevent users from sending funds to the wrong address. If the address doesn't check out, the form shouldn't validate.

If an address is all lower-case or all upper-case, a warning should be displayed below the address input... maybe with a link to EtherScan to help users check what's up.

mhluongo:
Note this goes for sending ETH, ERC-20s, and any transaction construction really... including if a dApp built the tx.

SJQuality:
It looks like in this field we are only checking the length of the address, not actually validating if it's a legit address or not. Well, unless my taking a legit address and randomly changing characters is somehow always a legit address.

0xDaedalus:
While this is objectively good for security - while working on the RSK integration we've determined that this is not necessarily good for UX - since plenty of website (etherscan included) present users with and allow them to copy addresses that do not necessarily pass the checksum.

PR #961, opened 2022-02-08
EIP-55 & EIP-1191 Checksum Validation

7alip:
tally-checksum

# Info

Related #774 

> This Draft PR is for discussing about the way of implementing different checksum validations.

User must be prevented from sending funds to the wrong address. In the case of providing all lower-case or all upper-case, user should be warned.

# Problem

As mentioned earlier that RSK network has [EIP-1191](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1191.md) checksum, which is different than Ethereum. Since EIP-19191 is not supported by Ethers, I'd like to figure out the best approach to support it.

# Proposal

- Set checksum type in Network configuration (default is EIP-55 for EVM type Networks).
- Check address checksum validation with a custom function based on the checksum type of the selected network
- For non EIP-55 addresses, if the address passes the validation, create the transaction request with lower-case address.


Shadowfiend:
An RSK-only version of this was implemented @ #2526, I believe based on the work behind this PR.

We still need to implement EIP-1191/55 verification for Ethereum addresses, but that'll be a separate thread.

PR #963, opened 2022-02-09
Show real ETH balances when importing from Ledger

cairomassimo:
Replace fake ETH balances with real balance, when importing accounts from a Ledger.

**Implementation.** Keeps existing Redux slice and thunk. A method `getAccountEthBalanceUncached` is added to the `main` service, which, in turn, invokes `chainService.pollingProviders.ethereum.getBalance`. The thunk is changed to use this new method instead of the fake one, and the fake one is removed.

Screenshot from 2022-02-09 11-16-28

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -547,6 +547,12 @@ export default class Main extends BaseService<never> {
     return this.ledgerService.refreshConnectedLedger()
   }
 
+  async getAccountEthBalanceUncached(address: string): Promise<bigint> {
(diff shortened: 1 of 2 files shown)


mhluongo:
One more nit: let's use `kebab-case` for branch names rather than `snake_case` please cc @fernya.

Yes, I've decided today I'm the PR police :sweat_smile:

greg-nagy (approved):
🚀

Issue #925, opened 2022-02-01
Add target for Ledger-related help links "I need help" and "I don't see my device"


cairomassimo:
Ledger-related help links are dead links at the moment.

SJQuality:
Seen here: https://www.awesomescreenshot.com/video/9096544?key=621290272c5eb7b046119379998fa06c

0xDaedalus:
No longer an issue ✅

PR #965, opened 2022-02-09
Do not merge! Ledger QA build

### LINKED ISSUE #952: Provide QA build of the Ledger-enabled wallet, fix most annoying bugs around the happy path

fernya:
Closes #952 

# What
Tally now can use your Ledger device for address derivation and transactions signing. Just add a new wallet following "Connect to a Ledger" link. Only transaction signing flow is tested.

# Why
Everybody loves to keep their assets in the most secure way available. With the communication provided to Ledger devices Tally can be used to easily manage these assets.

# How
## Backend
- Introduced LedgerService and SigningService
- SigningService routes address derivation and transaction signing requests to LedgerService and KeyringService

## UI
- New tab UI for Ledger onboarding
- Accounts are integrated into the current UI elements

# Test
## Onboarding of a Ledger
1. Install the QA version of extension
2. Click “Add wallet” on the bottom of account list
3. Follow the instructions appearing on the new tab
4. **Follow them!**
5. The selected accounts should be available for use

## Coexistence with other accounts
1. Onboard a new Tally or MetaMask wallet
2. Check if onboarded Ledger wallets and the ones just added are all available from the address selector
3. Check if they are under separate blocks in the UI ("Full access" vs. "Full access via Ledger")

## Signing transaction
1. Select an address in "Full access via Ledger" block
2. Click on Send
3. Send a transaction to an address of your wish
4. You should check that this (last) transaction changes state in the transaction history to "sent"

# Demo
https://user-images.githubusercontent.com/499307/153201469-7b14c51b-0fd1-43ac-981e-f7f02cff6c2d.mov


mhluongo:
Quick comments from the demo video:

* it'd be excellent to see an address show up before it's ETH balance has loaded cc @cairomassimo 
* can we open the extension popup again from the tab? Clicking "open extension" in the tab rather than "close tab" or clicking the extension would feel really natural

Going to start lightly testing on my end.

mhluongo:
Screenshot from 2022-02-09 14-22-27

This is what happens when I try to connect but haven't plugged any devices in. Would be nice for it to auto-refresh / use the same page to share the error.

mhluongo:
Screenshot from 2022-02-09 14-27-57

On Chromium `98.0.4758.80 (Official Build) snap (64-bit)` on Ubuntu, I can't get a Nano X to connect. I've debugged it against Ledger Live and fixed an initial USB permissions issues with Chromium, but still no dice. After selecting it I see this

Screenshot from 2022-02-09 14-28-08

Apologies for the lack of error messages, but I'm not seeing anything in the local log... I'll try it again with the background script console visible

mhluongo:
Aha! Here's what happened.

The page says to open the Ethereum app on the Ledger. If I do, then I wait for more than a few seconds, the page pushes the Ledger back to the manager app. If I click "connect" while the device is on the manager app, I get an error. If I open the Ethereum app and *quickly* connect to that device in the browser, it goes through.

Also notable that the ETH amounts for the addresses are very wrong. Selected an address and adding it seems to work though :tada: 

I'll try testing signatures later.

mhluongo:
Aha, just saw https://github.com/tallycash/extension/pull/963

greg-nagy:
> it'd be excellent to see an address show up before it's ETH balance has loaded cc @cairomassimo

I think that's what happens, is it not? What do I don't get? 

> can we open the extension popup again from the tab? Clicking "open extension" in the tab rather than "close tab" or clicking the extension would feel really natural

Sadly no. This is a security boundary in the chrome. The extension UI can only be opened from the browser icon on actual user action. Possible workarounds is opening a new window as with dApp permissions or opening a new tab with the extension.

cairomassimo:
> > it'd be excellent to see an address show up before it's ETH balance has loaded cc @cairomassimo
> 
> I think that's what happens, is it not? What do I don't get?

@mhluongo @greg-nagy That's correct. We _do_ show addresses _before_ the balance is loaded. (The balance is fake, but #963 fixes it.)

cairomassimo:
> This is what I see if I click "I don't see my device". Love the path on there, lol. Probably need to add something here or unlink that bit before prod though.

@mhluongo #925

fernya:
> Aha! Here's what happened.
> 
> The page says to open the Ethereum app on the Ledger. If I do, then I wait for more than a few seconds, the page pushes the Ledger back to the manager app. If I click "connect" while the device is on the manager app, I get an error. If I open the Ethereum app and _quickly_ connect to that device in the browser, it goes through.
> 
> Also notable that the ETH amounts for the addresses are very wrong. Selected an address and adding it seems to work though 🎉
> 
> I'll try testing signatures later.

Weird stuff, I'll check it under Linux, I can't see this under MacOS. Did it just drop you to manager or also requested the PIN?

mhluongo:
Just manager, no PIN. This was with a Nano X.

cairomassimo:
I have rebased this branch on (an old snapshot of) `main`, as it previously contained duplicated commits that were rebased/cherry-picked. I can recover the old version if needed.

fernya:
> I have rebased this branch on (an old snapshot of) `main`, as it previously contained duplicated commits that were rebased/cherry-picked. I can recover the old version if needed. Now this shows the correct diff `HIDE_IMPORT_LEDGER={true => false}` (@fernya please double-check that you didn't make other relevant changes.)

Looking awesome, thanks @cairomassimo !

fernya:
This is active on main so we can drop it.

PR #967, opened 2022-02-10
💃 Let's Swap: Show Swaps in the wallet by default

mhluongo:
Time to do some community QA! Turning on swaps by default means we can do an interim release to test swaps, fix bugs, then follow up with a bigger public push.


greg-nagy (approved):
I am not really happy with the coding style here tbh. 
The separation of concerns are not well defined and redundant ... 

or it's just a feature flag 

who knows ... 

🚚

PR #969, opened 2022-02-11
Minty Fresh: Fix a few outstanding bugs in swaps

### LINKED ISSUE #937: Snackbar is covering clickable elements even when its not visible
### Discord Discussion Link

_No response_

Chrome

Interact with a button.

### What did not work?

The hidden wrap is blocking user interaction in that area as it covers the button.

image

### Version

_No response_

### Relevant log output

_No response_

Shadowfiend:
This hits a few issues observed during final QA of swaps:

- Closes #937 .
- Reworks handling of transaction state so that the swap transition
  to transaction signing works correctly; when transaction state
  clearing became more aggressive, it left the sign transaction page
  in a state where it might decide it was seeing a rejected transaction
  and redirect back to the caller---not great!
- Rework selected asset handling in `SharedAssetInput` so that the
  caller provides the currently-selected asset at all times (instead of
  providing the starting default). This allowed the swap flip button to
  correctly update the selected assets in the UI.
- Added sorting for assets in `SharedAssetInput`. Sorting without a
  search term sorts alphabetically by symbol, while sorting with a search
  term sorts alphabetically but bubbles to the top any search term matches
  that match at the beginning of the symbol. This bubbles close-to-exact
  matches up where they're easier to get to, and tends to push down
  LP tokens and similar derivative tokens that include the base symbol
  in their name. Sorting is locale-aware based on the current JS locale
  at all times.

### Testing

For the snackbar issue, the Get Final Quote button and Approve button on
swaps should be clickable across the button; on main, it should have dead
space in the middle.

For the transaction signing issue, try to execute a swap. If the sign
transaction screen comes up every time, you win! Rejection should also
work, Send signing should work, and dApp signing should work.

For the `SharedAssetInput` changes, look at swaps and make sure that,
after selecting two assets and doing a quote, the flip button in the
middle flips both amount and assets. The swap-to list should show the
sorting behavior noted above. Finally, make sure that the other uses
of `SharedAssetInput` continue to work as expected with the sorting
behavior; these include Earn (with feature flag) and the unchangeable
asset input in the Send flow (which is hardcoded to ETH).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -23,7 +23,7 @@ import {
 } from "./services"
 
 import { EIP712TypedData, HexString, KeyringTypes } from "./types"
-import { EIP1559TransactionRequest, SignedEVMTransaction } from "./networks"
+import { SignedEVMTransaction } from "./networks"
 import { AddressNetwork, NameNetwork } from "./accounts"
 
 import rootReducer from "./redux-slices"
@@ -56,9 +56,9 @@ import {
   transactionRequest,
   signed,
   updateTransactionOptions,
-  broadcastOnSign,
   clearTransactionState,
   selectDefaultNetworkFeeSettings,
+  TransactionConstructionStatus,
 } from "./redux-slices/transaction-construction"
 import { allAliases } from "./redux-slices/utils"
 import {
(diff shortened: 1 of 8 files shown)


mhluongo (approved):
Very happy with this. Asset sorting heuristics are certainly going to be revisited, but damn, swaps are feeling great.

PR #976, opened 2022-02-11
Tally at the Gates: Implement 0x gated API support, plus sundry fixes for swaps

Shadowfiend:
NOTE: I've temporarily included a commit here that enables the 0x API
key on PR builds; we shouldn't do this in general, and I'll be
dropping the commit before undrafting.

Fixes:
- Release nonces when broadcast fails, ensure nonces don't drop below
  the transaction count on-chain, and fix an off-by-one issue when updating
  seen nonces based on transaction confirmation.
- Dropped transactions are marked as such in the activity list.
- Activity items are sorted by nonce when block height is unavailable and
  two activities are on the same network, and nonces are displayed in activity
  details.
- Logs include timestamps.
- Swap approval should show the waiting state correctly until the approval
  transaction mines, and should correctly requote once approval has mined.
- Some fixes around adding 10% to the gas limit were completely unnecessary.
  These are gone now.


background/constants/index.ts
```diff
@@ -23,5 +23,8 @@ export const MINUTE = 60 * SECOND
 export const HOUR = 60 * MINUTE
 export const DAY = 24 * HOUR
 
+export const COMMUNITY_MULTISIG_ADDRESS =
+  "0x99b36fDbC582D113aF36A21EBa06BFEAb7b9bE12"
```
mhluongo, line 27: :tada:

mhluongo (approved):
Confirmed that this resolves the issues I've had swapping. Merging and moving to further QA for release.

PR #991, opened 2022-02-14
Add Webpack configs with experimental feature set

cairomassimo:
Builds two variants per supported browser:
the first one has only some features enabled (ready for the public),
the second ("experimental") has all the features enabled (for QA).

The two feature sets are configured respectively with the files
`.env.features` and `.env.features.experimental`.

The build target location is unchanged for the standard feature set,
whereas experimental builds include the suffix  `-experimental`.
For example, for the Chrome browser, we now build:
`chrome/`, `chrome.zip`,
`chrome-experimental/` and `chrome-experimental.zip`.

## What to test

- Check that the build generates all the variants in the expected location.
- Install a standard build and an experimental one, and check that the enabled features change between the two as expected (for one feature at least).
- Make sure that CI is not broken, as well as anything that depends on it.
- Make sure that local environment configuration continues to work, so that we do not break any dev workflow.
- Evaluate the cost/benefit of doubling-up the duration of all builds. (Perhaps we could build experimental variants for `firefox` and `chrome` only?)

## Discussion

Now that we depend on three parameters (mode, browser, features) instead of two, `webpack.config.ts` may benefit from a further reorganization. (Not done in this PR to reduce its code footprint.) For example, all the configuration could be merged into a single function, which first builds a collection of partial configurations (depending on mode, browser and features) and then merges them together at the very last step.


cairomassimo:
Currently getting errors
```
Killed
error Command failed with exit code 137.
```
both on CI (systematically) and locally (sometimes).
It's definitely caused by OOMs locally, and likely on CI as well.
By building variants in parallel, the memory usage grows linearly with the number of variants. We need a fix for that before this can be merged.

Shadowfiend:
I see what this solves, but I'm not certain why we need it. Is there a specific problem we're trying to solve here?

Flagging this because I'm allergic to increasing the complexity of our webpack build, especially when we're exploding the number of potential output artifacts. I would much rather tailor a build solution to the narrow problem we're trying to solve, and make sure we don't have other tools that can help us. I'll list a few problems I see below and what I consider to be the right solutions:

- Particular devs may need to enable certain or all feature flags: `.env` is already changeable by devs to enable the feature flags they are trying to test, and gitignored
- Devs may need alternate builds locally for e.g. PR review: we can use our tools here, leveraging `git worktree` copies or separate clones to maintain a place specifically for non-feature-dev testing.
- We may want a QA build on GitHub for a particular feature: I would rather handle that at the Actions level rather than crank complexity on the main build if we can. We can use branch patterns and an extra Actions step for that, as a quick example. e.g., if a branch name matches `(^|/)<feature>/`, the env variable `ENABLE_<feature>` is set to `1` (or `HIDE_<feature>` is set to `0`, though I'd like us to move towards using positive verbiage for feature flags and assuming flags are off). Another option here is to have the branch name prompt a particular env package (similar to how this PR does `.env.features`, where `<package>/<rest of branch name>` would trigger copying `.env.<package>` to `.env` before running the build in Actions.

cairomassimo:
@Shadowfiend your comment makes a lot of sense. I'm closing this PR in favor of a better approach!

The goal here was to avoid temporary PRs such as #965. We discussed this last week, and noticed that having two build configurations, a standard one and one with all the features on, would cover 95% of the use cases. As @Shadowfiend suggested, we can achieve this better with separate GitHub actions/jobs that directly manipulate env variables.

BTW, following this idea, we could speed up local builds by having only one browser built at a time. If this makes sense to you, I can create an issue for it and do it when there is time.

PR #992, opened 2022-02-14
Fix issues with computing and displaying the current extension version

cairomassimo:
This PR fixes some issues left over by #828, namely:

1. Change the method used to compute the current version to `git describe --tags`. This fixes some cases where the current version was not computed correctly due to *unannotated tags* in the repo. In the future, we may consider updating the repo so that all the version tags are annotated (so they can be dated and signed), and revert this particular change.
2. The computed version is matched against the version in `manifest.json`. If a mismatch is found, the build is aborted.
3. The UI displaying the version was updated, loosely based on [this (Figma)](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=5250%3A126197). (With one difference with respect to the design in https://github.com/tallycash/extension/pull/828#issuecomment-1016612434: version is displayed inside the block with light background, because it's closely related to release name and the feedback button -- version number is mostly useful when reporting feedback. @VladUXUI please have a look at it.)

Note: configured CI to fetch the entire Git history (needed to compute the version).

## What to test

- Check that CI and local builds work.
- Check that the version is displayed properly in the menu screen.
- Evaluate costs/benifits of fetching the Git history in CI, for the purpose of computing the current version. (In the future, we can limit the history to, say, 1000 commits, and we should still be able to run `git describe --tags` as long as the most recent tag is not too far in the history.)

## Screenshot

Screenshot from 2022-02-14 16-12-39

instead of

Screenshot from 2022-02-14 16-21-08


Shadowfiend:
Hmmm… Could we use `NPM_PACKAGE_VERSION` and suffix whatever `git rev-parse --short HEAD` spits out? We can then check that `NPM_PACKAGE_VERSION` matches with the `manifest.json` version instead. This will include the commit hash even on production builds, but that's probably a good thing anyway since tags are mutable in git. Dates, times, etc, are less important.

The reason I mention this is that it should let us drop the `GitRevisionPlugin` build-time dependency (which is an excellent vector for a supply-chain attack), since the rev-parse call should basically just be `require('child_process').execSync('git rev-parse --short HEAD').toString('UTF-8').trim()` (though we probably should write it more cleanly hehe).

If we want to keep the date as well, we can do `git show -s --format='%h (%cs)' HEAD` as the command instead.

cairomassimo:
@Shadowfiend yeah, I totally forgot about the NPM version. So now we have three sources of truth for the version:

- NPM package version
- Manifest version
- Git tags

What scares me is that these three can easily start to differ from one another.
For the future:

- We could remove the manifest version as a source of truth, and just generate it on build.
- About NPM package version and Git tags: they are both mutable. Also, storing the current commit hash does not provide a guarantee of integrity (one could build with local changes, or just tamper with the build system). Currently, the guarantee comes from the fact that a build was done by our CI, from our official repository. So I don't see any reason to prefer the NPM version or Git tags in this sense (we can just make sure not to change the tags).
- Is the NPM package version actually used? Is the package published on NPM? If not, then we could even remove the version, and rely on Git tags only. (That's my personal favourite option.)
- Using the NPM version as a source of truth also works. In that case, checking that it matches the most recent Git tag can be useful. However, I prefer the previous alternative because it handles non-official versions automatically and nicely, and has a single source of truth.

Besides that, @Shadowfiend your comment doesn't seem to be blocking for this PR (as it applies to stuff already in main). Is that right? Thanks!

Shadowfiend (approved):
Let's roll this on in and deal with the underlying generation strategy separately as @cairomassimo suggested.

VladUXUI:
Sorry Massimo, i got lost in the different PR's.
Let's use the original design. you are correct that showing version + date is mostly for Reporting bugs. But that is from our perspective, our needs. The needs of the information for the user might not be the same as ours.
This information also is a way for users to make sure they are up to date, and shows that we are continuously working on it and upgrading it.

PR #1000, opened 2022-02-15
🌒 Update slide up menu to always have clickable backdrop

### LINKED ISSUE #999: Add background to slide up modal
color green-120 70%
modal

henryboldi:
Per Vlad's instructions during our UI sync

slide_menu

Closes #999

⚠️ Haven't written a proper description here yet

### DIFF
--- a/ui/components/Shared/SharedSlideUpMenu.tsx
+++ b/ui/components/Shared/SharedSlideUpMenu.tsx
@@ -1,6 +1,6 @@
+import React, { ReactElement, CSSProperties, useRef } from "react"
 import classNames from "classnames"
-import React, { ReactElement, useState, useEffect, CSSProperties } from "react"
-import { useDelayContentChange } from "../../hooks"
+import { useDelayContentChange, useOnClickOutside } from "../../hooks"
 
 export type SharedSlideUpMenuSize =
   | "auto"
@@ -17,7 +17,6 @@ interface Props {
   children: React.ReactNode
   customSize?: string
   size: SharedSlideUpMenuSize
-  showOverlay?: boolean
   alwaysRenderChildren?: boolean
 }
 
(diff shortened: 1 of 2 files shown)


ui/components/Shared/SharedSlideUpMenu.tsx
```diff
@@ -30,15 +29,12 @@ const menuHeights: Record<SharedSlideUpMenuSize, string | null> = {
...
-    size,
-    children,
-    customSize,
-    showOverlay,
-    alwaysRenderChildren,
-  } = props
+  const { isOpen, close, size, children, customSize, alwaysRenderChildren } =
+    props
+
+  const ref = useRef(null)
```
Shadowfiend, line 35: Let's start naming these refs for what they're refing. I know we have a couple of other uses where it's just named `ref`, but we don't have to change those in this PR—let's just avoid introducing new unmapped names like this for now.

Shadowfiend (dismissed):
Light naming note. Need to test drive this, hopefully early Thursday before merging.

ui/components/Shared/SharedSlideUpMenu.tsx
henryboldi: Good point 😌

ui/components/Shared/SharedSlideUpMenu.tsx
```diff
@@ -30,15 +29,12 @@ const menuHeights: Record<SharedSlideUpMenuSize, string | null> = {
...
-    customSize,
-    showOverlay,
-    alwaysRenderChildren,
-  } = props
+  const { isOpen, close, size, children, customSize, alwaysRenderChildren } =
+    props
+
+  const slideUpMenuRef = useRef(null)
+
+  useOnClickOutside(slideUpMenuRef, close)
```
Shadowfiend, line 37: One more idea that occurred to me in the middle of the night for no apparent reason and would be it's own PR lol...

Most of our uses of this declare a null ref, then pass it to this hook. It feels like we could refactor the hook to be a little more hook-like by considering it an extension to `useRef` and making it return a ref instead of taking it as a param, e.g.:

```
  const slideUpMenuRef = useOnClickOutsideRef(close)
```

It could take an optional second parameter that sets a non-null initial value for the ref. I can think of one or two scenarios where we might still need a ref passed in, but I don't know that they'll be in our code base any time soon. 

Interested in thoughts, though again this isn't urgent, blocking, or otherwise important haha.

Shadowfiend:
> Need to test drive this, hopefully early Thursday before merging.

Reviewer testing should be considered required before merging unless something crazy is going on 🙏

henryboldi:
> Reviewer testing should be considered required before merging unless something crazy is going on 

Does turning on auto-merge hinder this? @Shadowfiend, I don't follow

Shadowfiend:
Yeah, auto-merge basically means as soon as builds pass the PR will land. Only the reviewer should enable it.

Shadowfiend:
A'iiight just tested this, looks glorious. Let's do it :guitar:

henryboldi:
I thought auto-merge still requires approval

Issue #940, opened 2022-02-03
Different keyrings mixed together


VladUXUI:
### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

I added two keyrings to Tally

### What did not work?

![image](https://user-images.githubusercontent.com/80846886/152361407-70319bd2-c61f-4164-865b-ba162a5fc13b.png)

- The above two addresses are derived from 2 different keyrings, but they are bundled together.
- Clicking on Add address, it adds an address to the first Keyring.

This is tricky if users check out tally using a New recovery phrase and than decide to import

They should all have their own category, if we have more from same category, we should number them as Import 1, Import 2...:
New seed = Tally
Import phrase = Import
Ledger = Ledger 
Read-only = Read-only

SJQuality:
Looks like this was resolved in https://github.com/tallycash/extension/pull/1001 closing

PR #1001, opened 2022-02-15
Differentiate between Tally Wallets & Imported Wallets in Accounts Panel

0xDaedalus:
We want to have a separate section for "Tally" wallets in the accounts list as mentioned in #940.  This PR introduces a new piece of state to keyrings which identify them as either being imported or created form a new seed. 

 ~It depends on https://github.com/tallycash/hd-keyring/pull/6 being merged and published to npm.~

image


background/package.json
```diff
@@ -37,7 +37,7 @@
     "@ledgerhq/hw-app-eth": "^6.23.0",
     "@ledgerhq/hw-transport": "^6.20.0",
     "@ledgerhq/hw-transport-webusb": "^6.20.0",
-    "@tallyho/hd-keyring": "^0.3.3",
+    "@tallyho/hd-keyring": "file:../../../tally/hd-keyring",
```
0xDaedalus, line 40: This must be changed to the correct version for tests to pass.

Shadowfiend:
Hmmm… Why do we want to push the source all the way into the keyring package? It feels like we could just track it in the extension, no?

0xDaedalus:
We certainly can - I figured since we are conceptually adding a new descriptive property to a keyring that it made sense to make a change to the hd-keyring package and augmenting the `HDKeyring` class but I can definitely see the argument for keeping the logic confined to the extension.

beemeeupnow:
What do you think about 'origin' instead of 'source'? (for some reason it feels like it more immediately conveys the intended meaning)

After having played around in the code changed by this it feels to me that updating [the hd-keyring PR](https://github.com/tallycash/hd-keyring/pull/6) to support arbitrary metadata as mentioned [here](https://github.com/tallycash/hd-keyring/pull/6#discussion_r809411317) and using that makes the most sense if the association belongs to the key (as opposed to address/account).

0xDaedalus:
Alright - significant changes since PR creation deserve significantly updated descriptions.  The flow now looks like this

1.  A user adds an account to tally
2.  When the account is added - the `keyringService`'s  private #keyringMetadata field is updated to map the source of the address to the keyring id.  This metadata is also emitted to redux.
3.  When the keyring service is locked - the in-memory #keyringMetadata is encrypted into the vault and cleared from memory.  It is also cleared from redux.
4. When the keyring service is unlocked - #keyringMetadata is decrypted from the vault, updated in memory (in keyringService) and sent to redux.

Shadowfiend:
> When the keyring service is locked - the in-memory #keyringMetadata is encrypted into the vault and cleared from memory. It is also cleared from redux.

This, we can't do, unfortunately, as it will mean that if the user walks away from their computer for X minutes and the service autolocks, their account selection will be completely fudged up. You'll notice the reducer for `updateKeyrings` explicitly ignores keyring updates while the keyrings are locked—we should do the same for metadata.

Without having gone deep here, I would also suggest we emit keyrings and their metadata together in the keyrings event—possible I'm missing something, but I think that would make things easier in the whole stack and generally not add complexity anywhere.

0xDaedalus:
> This, we can't do, unfortunately, as it will mean that if the user walks away from their computer for X minutes and the service autolocks, their account selection will be completely fudged up. You'll notice the reducer for `updateKeyrings` explicitly ignores keyring updates while the keyrings are locked—we should do the same for metadata.

I see this now - thank you!

> Without having gone deep here, I would also suggest we emit keyrings and their metadata together in the keyrings event

Sounds good to me - i debated wether to split or not when developing and am happy to batch them together

background/main.ts
```diff
@@ -829,9 +829,12 @@ export default class Main extends BaseService<never> {
       this.store.dispatch(setKeyringToVerify(generated))
     })
 
-    keyringSliceEmitter.on("importKeyring", async ({ mnemonic, path }) => {
-      await this.keyringService.importKeyring(mnemonic, path)
-    })
+    keyringSliceEmitter.on(
+      "importKeyring",
+      async ({ mnemonic, path, source }) => {
+        await this.keyringService.importKeyring(mnemonic, source, path)
```
wojciech-turek, line 835: We have agreed to name it seedPhrase, mnemonic is a banned word. Let's change it before Matt sees it.

wojciech-turek (approved):
I have given it a test run, it displays correct information and while adding multiple wallets is buggy in a sense that its not possible to create new internal wallet and then add an imported one (or other combinations) but the accounts are differentiated and separated into their respective types as designed.

My comments on the code are just soft suggestions. Feels like this is ready to go!

background/main.ts
mhluongo: Heh `mnemonic` for technical, "recovery phrase" for copy. Let's avoid "seed phrase" everywhere!

0xDaedalus:
> while adding multiple wallets is buggy in a sense that its not possible to create new internal wallet and then add an imported one

I wasn't able to recreate this issue but I'll go ahead and merge this into `main` and open up another PR for the above suggestions as well as (hopefully) fixes to the buginess.

Shadowfiend:
> while adding multiple wallets is buggy in a sense that its not possible to create new internal wallet and then add an imported one

> I wasn't able to recreate this issue but I'll go ahead and merge this into main and open up another PR for the above suggestions as well as (hopefully) fixes to the buginess.

Just a process note: please don't approve or merge a PR that seems to have introduced a regression. Keeping main releasable requires diligence from all sides.

Also, ideally don't merge your own PRs. The merge by the reviewer both symbolically and git-wise cements the shared ownership of the outcomes. If you intend to merge but need to wait for a main build, just flip on auto merge. 

Excited to see this land 😄

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -187,34 +192,61 @@ const signingMethodTypeToAccountType: Record<
...
+  if (signingMethod == null) {
+    return AccountType.ReadOnly
+  }
+  if (signingMethodTypeToAccountType[signingMethod.type] === "ledger") {
+    return AccountType.Ledger
+  }
+  if (addressSources[address] === "import") {
+    return AccountType.Imported
+  }
+  return AccountType.NewSeed
```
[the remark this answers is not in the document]
Shadowfiend: Let's make sure to do this. Consistency is important and once this kind of stuff hits a release it requires migrations to update.

PR #1006, opened 2022-02-16
Remove .only's and fix broken integration tests

0xDaedalus:
This PR remotes two `it.only`'s from keyring-integration.test.ts and configured our linter to throw an error if it detects a `.only` test block.  I went into this PR with the assumption that the inactive tests were written correctly but underlying behavior had changed while they were inactive - so went with the route of changing the code to match the tests although it is certainly possible that the tests themselves need to change as well.

### DIFF
--- a/background/services/keyring/index.ts
+++ b/background/services/keyring/index.ts
@@ -299,6 +299,8 @@ export default class KeyringService extends BaseService<Events> {
    * used outside the extension.
    */
   getKeyrings(): Keyring[] {
+    this.requireUnlocked()
+
     return this.#keyrings.map((kr) => ({
       // TODO this type is meanlingless from the library's perspective.
       // Reconsider, or explicitly track which keyrings have been generated vs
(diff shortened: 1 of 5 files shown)


background/services/keyring/index.ts
```diff
@@ -88,6 +88,7 @@ export default class KeyringService extends BaseService<Events> {
     // Emit locked status on startup. Should always be locked, but the main
     // goal is to have external viewers synced to internal state no matter what
     // it is. Don't emit if there are no keyrings to unlock.
+    await super.internalStartService()
```
0xDaedalus, line 91: This was added to fix [these tests](https://github.com/tallycash/extension/blob/main/background/tests/keyring-integration.test.ts#L379) by allowing assertions based on [this mock](https://github.com/tallycash/extension/blob/main/background/tests/keyring-integration.test.ts#L389) to pass.

background/services/keyring/index.ts
```diff
@@ -450,8 +451,12 @@ export default class KeyringService extends BaseService<Events> {
   // //////////////////
 
   private emitKeyrings() {
-    const keyrings = this.getKeyrings()
-    this.emitter.emit("keyrings", keyrings)
+    if (this.locked()) {
+      this.emitter.emit("keyrings", [])
+    } else {
+      const keyrings = this.getKeyrings()
```
0xDaedalus, line 457: `getKeyrings` throws an error if called when the keyringService is locked so we avoid calling it in the case that the keyring is locked.

mhluongo (approved):
Looks great 😁

PR #994, opened 2022-02-14
Multi-network Prep: Prep the codebase for multi-network refactoring

Shadowfiend:
The refactors here are light and shouldn't change any behavior;
instead, they push most of the current-network management into
`ChainService`, where it can be refactored from a central location
rather than having a global `getEthereumNetwork` function that has
to be targeted piecewise. Based extensively on work in #732.

### Testing

Testing here should just be that everything works roughly as expected
across the board. Again, behavioral changes are expected to be minimal.

PR #1009, opened 2022-02-18
The Dude Does Not Provide: Lose a bespoke provider in erc20 lib

Shadowfiend:
Specifically for the purpose of handling asset approval amount updates
at signing time, the erc20 library file instantiated an Alchemy provider
separate from the system's existing chain communication infrastructure.
Since @0xDaedalus hooked up annotation enrichment for transaction requests,
this step was no longer necessary.

It turns out removing this provider to remove a network-specific code path
(done in the last commit) was a pretty serious journey, but the outcome has
cleaned up a good chunk of technical debt.

Here are the changes:

- The price decimal heuristic is now extracted from the currency enrichment
  functions.
- `sameEVMAddress` can handle null values. TS-wise, it never receives them,
  but due to loose null checking in Ethers we can have nulls masquerading as
  undefineds.
- `enrichAssetAmountWithMainCurrencyValues` handles undefined price points
  transparently instead of requiring the caller to check for them explicitly.
- `CompleteAssetAmount` now takes just one type parameter through some internal
  trickery. The result is more ergonomic.
- Transaction requests are enriched every time their data is updated, leading
  so that updates to e.g. approval amounts is handled.

These are the simple changes. The big change to support all of this, however,
is that transaction signing renderers all work with a `TransactionAnnotation`
now instead of working with ERC20 parsing from the erc20 library. This opens
up the space for actually starting to use the renderers for swaps and such,
since the enrichment service provides a central place to resolve these kinds of
calls.

To do this properly required updating all of the signing renderers to
use asset amounts and our internal functions for handling and enriching
those amounts with localized main currency data, since that's what is
carried in the annotations. This is great, but it means there were a lot
of changes.

### Testing

This very heavily hits the transaction signing flow. Please make sure that
signing continues to work, and in particular check ETH sends, dApp signing,
swaps, and approval transactions to make sure these are all still working well.

For approval transactions, please make sure editing the approval amount
still works. It should actually also support entering “infinite” or “infinity”
as the amount, as well as supporting arbitrary decimal amounts of the
asset.

### DIFF
--- a/background/accounts.ts
+++ b/background/accounts.ts
@@ -38,7 +38,7 @@ export type AccountBalance = {
 /**
  * An address on a particular network. That's it. That's the comment.
  */
-export type AddressNetwork = {
+export type AddressOnNetwork = {
   address: HexString
   network: EVMNetwork
 }
@@ -46,7 +46,7 @@ export type AddressNetwork = {
 /**
  * A domain name, with a particular network.
  */
-export type NameNetwork = {
+export type NameOnNetwork = {
   name: string
   network: EVMNetwork
 }
--- a/background/accounts.ts
+++ b/background/accounts.ts
@@ -1,5 +1,5 @@
 import { AnyAssetAmount } from "./assets"
-import { Network } from "./networks"
+import { EVMNetwork } from "./networks"
 import { HexString } from "./types"
 
 /**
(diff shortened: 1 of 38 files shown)


Shadowfiend:
Got a little excited here lol. Draft until #994 lands.

Shadowfiend:
… Oh I did something very wrong in my weekend branch fixing heh. #994… Ended up including this PR.

Closing the PR :scream:

mhluongo:
Realized, merged it anyway, and meant to follow up 😅

PR #1017, opened 2022-02-22
WIP UI: Normalize HexString Values

dills122:
# Issue
Not all hex strings were being normalized.

# Solution
Leverage the existing functionality from `@tallyho/hd-keyring` and `background/lib/utils/index.ts` to normalize hex strings including block hashes, account strings, and transaction hashes.

### DIFF
--- a/background/lib/utils/index.ts
+++ b/background/lib/utils/index.ts
@@ -112,7 +112,8 @@ export function getEthereumNetwork(): EVMNetwork {
 }
 
 export function truncateAddress(address: string): string {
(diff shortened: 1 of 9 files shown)


Shadowfiend:
Closing this, with apologies, as we've drifted significantly from the code that was in place when the work started

Ultimately if and when we pursue this I recommend we:
- Use a nominal type to capture normalized hex strings (see https://github.com/tahowallet/extension/blob/main/background/types.ts#L37-L42 for an example with `NormalizedEVMAddress`).
- Break up the PRs into components so that we can merge more easily.

PR #1026, opened 2022-02-23
Re-enable showing errors when signing with Ledger

cairomassimo:
This commit reverts 9b8e592c72989f4150c394613bda9ada36c0c298 (which in turn reverted a change causing a bug in non-Ledger signing) and fixes the original bug.

## What to test

- Signing works regularly for non-Ledger accounts.
- Correct error slide-ups are shown when signing iwth Ledger and, either:
  - Ledger is disconnected
  - a different Ledger is connected (i.e., not containing the signer account)


fernya (approved):
The Ledger flow works as expected, the non-Ledger flow shows the appropriate "Sign" button.

PR #1029, opened 2022-02-23
Death and Taxes: Charge fees regardless of the gated 0x API

mhluongo:
This change means the community multisig will go back to receiving fees for every swap.

### DIFF
--- a/background/redux-slices/0x-swap.ts
+++ b/background/redux-slices/0x-swap.ts
@@ -105,15 +105,11 @@ const zeroXApiBase =
   process.env.ZEROX_API_KEY.trim() !== ""
     ? "gated.api.0x.org"
     : "api.0x.org"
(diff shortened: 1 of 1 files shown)


Shadowfiend:
To be clear, this change means the multisig will for the first time get fees from dev builds :)

mhluongo:
Yep, here for it!

Shadowfiend:
100% same, just wanted to clarify this isn't going back to anything--only `main` builds have ever had the gated params passed along. We did briefly pass the affiliate address on all builds. 

Also forgot we upload artifacts for PRs now--those have likewise never had fees enabled.

Shadowfiend (approved):
Let's do it.

Shadowfiend (approved):
Again post-merge 😅

PR #1040, opened 2022-02-24
Asset similarity refactor

Shadowfiend:
The refactor here gives asset similarity its own file in lib/, introduces the
concept of similarity keys that allow merging larger numbers of assets more
efficiently, and using a common set of similarity metrics across the
system. “Similarity keys” fundamentally simulate how we were doing asset
merging based on a symbol + normalized contract address combo for Ethereum
assets, but open the door for other similarity correlators.

Note also that, unlike before, there is a final similarity check for two assets
that are considered the same by similarity key. This allows all asset merging
to be mediated by the score given by `scoreAssetSimilarity`, even when
similarity keys are used to speed that process up.

### Testing

Check that assets still look reasonable in the swaps UI. Additionally, make sure
that no assets have disappeared from the wallet overview or per-account asset
list. Finally, check that making a transaction that involves a new asset that
isn't in your balances properly reflects the new asset in your balances
(this could take 1-2 minutes in outlier cases).


background/lib/asset-similarity.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+  if ("decimals" in a && "decimals" in b) {
+    if (a.decimals === b.decimals) {
+      score += 1
+    } else {
+      score -= 1
+    }
+  } else if ("decimals" in a || "decimals" in b) {
+    score -= 1
+  }
```
cairomassimo, line 29:
```suggestion
  if ("decimals" in a || "decimals" in b) {
    if (a.decimals === b.decimals) {
      score += 1
    } else {
      score -= 1
    }
  }
```

background/lib/asset-similarity.ts
```diff
@@ -0,0 +1,115 @@
...
+      a.homeNetwork.name === b.homeNetwork.name &&
+      a.homeNetwork.chainID === b.homeNetwork.chainID
+    if (sameNetwork) {
+      score += 1
+    } else {
+      score -= 1
+    }
+  } else if ("homeNetwork" in a || "homeNetwork" in b) {
+    score -= 1
+  }
```
cairomassimo, line 41:
```suggestion
  if ("homeNetwork" in a || "homeNetwork" in b) {
    const sameNetwork =
      a.homeNetwork?.name === b.homeNetwork?.name &&
      a.homeNetwork?.chainID === b.homeNetwork?.chainID
    if (sameNetwork) {
      score += 1
    } else {
      score -= 1
    }
  }
```

cairomassimo (commented):
Minor, non-blocking code suggestion (on moved code, actually)

background/lib/asset-similarity.ts
Shadowfiend: Unfortunately, this won't pass type checking, as `a.decimals` is illegal if `"decimals" in a` is false. We could do this, however:

```suggestion
  if ("decimals" in a && "decimals" in b && a.decimals === b.decimals) {
      score += 1
  } else if ("decimals" in a || "decimals" in b) {
    score -= 1
  }
```

(Modulo prettier reformatting.) I do think that would be clearer overall.

background/lib/asset-similarity.ts
Shadowfiend: Same as above, we're proving the *existence* of the property, not its definedness, and without existence TypeScript won't let us dereference it. That said, I think we can do the same massage as above.

I need to see if there's enough usage of that network check in the codebase to justify extracting a `sameNetwork(a: Network, b: Network)`-type utility as well.

Shadowfiend:
Going to hit this in my next PR rather than here.

mhluongo (approved):
Refactor looks good, QA went well. LFG

PR #1042, opened 2022-02-25
Simplify implementation of Swap price quote fetching

cairomassimo:
EDIT: this PR now contains some code simplification for Swaps. The original goal has changed (see comments)

## Original description

It seems that the promise returned by `dispatch` sometimes resolves to `undefined` for no particular reason (on my machine?).
Instead of relying on it for Swap `fetchPrice`, we now save the response in the Redux state, and use a selector to read it from there.
This was the only portion of code where the return value of `dispatch` was used.

This change was necessary in order to test #1041, but it's kept in a distinct PR to simplify review.


Shadowfiend:
> for no particular reason

I would much rather understand the reason before approving this change. We should be able to rely on async thunk `dispatch`es to return across the app, so if that's not working it's worth burning some time on figuring out why.

Separately, wanted to flag that not saving this in redux was a deliberate decision, and is reflective of the fact that most of the state we're saving in redux for swaps right now is due to how transaction signing works (which transaction queueing will have to change/fix). I much prefer an outcome where we're saving *less*, not more, in redux for internal dApps, whose state is generally much more tied to the UI being shown rather than to global application state.

That said, if we can't figure it out and need to patch this short-term we can always land this PR as-is and circle back later.

cairomassimo:
@Shadowfiend the problem was caused by our patch to `webext-redux` not being applied on my machine (perhaps after updating a package, switching branch, and/or mixing up `yarn` and `npm` by accident).

This PR contains some improvements of `Swap.tsx`, which may be worth merging. I'll work some more on it (perhaps removing latest request/response entirely from Redux state) and update this PR (or open a new one if we prefer to).

cairomassimo:
Now that the main goal of this PR changed, I'll give a better look to the code to see if there are more related improvements before marking this as ready-to-review.

greg-nagy:
@cairomassimo What problem does this PR solve in its current state? 

If I understand the changes here it moves state from one place to another. 

I personally don't have strong feelings about the place of the state, 
but if something is not broken then it should not be fixed. 

Unless there is an underlying reasoning/guiding thought, which I am not aware of (yet)

cairomassimo:
@greg-nagy this PR started to fix a bug, which turned out not to be a bug (local misconfiguration). However, the PR contained some code cleanups that we could save (e.g., 39 lines or code removed).

So there is no rush to merge it, we can wait after launch and if it rots too much we can just close it :man_shrugging:

greg-nagy:
@cairomassimo In this case, I would prefer to close this now. 

If the only reason we want to change something that's working well is that it saves us 39 lines then I would prefer not to do it.

cairomassimo:
> I would prefer to close this now.

Makes sense. Closing.

PR #1044, opened 2022-02-25
Remove /slices in UI readme

gasolin:
Tracing source and found inaccurate file structure in [ui/README.md](ui/README.md)
I guess the /slices is moved into background [background/redux-slices/ui.ts](background/redux-slices/ui.ts)


Shadowfiend (approved):
Yep, great point, thanks for the fix!

Also highlights the fact that we *don't* have this section for the background README, unfortunately.

mhluongo:
Thanks @gasolin! Mind signing those commits so we can merge?

PR #1045, opened 2022-02-25
Assistive assets: Abstract asset helpers for easier multi-network handling

Shadowfiend:
The primary addition here is `AssetDataHelper`, accessible as
`chainService.assetData.*`. This wraps token metadata, balance, and
transfer lookup. Currently it does all Alchemy-specific calls and
only tries to fall back for token metadata, but it provides the
foundation for non-Alchemy fallbacks for networks that may not have
Alchemy-specific calls or support.


background/services/chain/asset-data-helper.ts
```diff
@@ -0,0 +1,155 @@
...
+export default class AssetDataHelper {
+  constructor(private providerTracker: ProviderManager) {}
+
+  private withProviderForNetwork<T>(
+    network: EVMNetwork,
+    handler: (provider: SerialFallbackProvider) => T
+  ): T | undefined {
+    const provider = this.providerTracker.providerForNetwork(network)
+
+    if (typeof provider === "undefined") {
```
cairomassimo, line 43: Non-blocking: I've seen this `typeof x === "undefined"` all around the codebase. Why not `x === undefined` instead? AFAIK it's guaranteed equivalent in strict JS and TypeScript.

background/services/chain/asset-data-helper.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  async getTokenBalances(
+    addressOnNetwork: AddressOnNetwork,
+    smartContractAddresses?: HexString[]
+  ): Promise<SmartContractAmount[]> {
+    return (
+      (await this.withProviderForNetwork(
+        addressOnNetwork.network,
+        async (provider) => {
```
cairomassimo, line 61: I would replace this (and the two other instances below) with something like:

```typescript
const provider = await this.getProviderForNetwork(addressOnNetwork.network)
if (provider === undefined) return []
// ... body of the callback
```

It looks much simpler. I don't see enough value in extracting this pattern into a higher-order function.

background/services/chain/asset-data-helper.ts
```diff
@@ -0,0 +1,155 @@
...
+          try {
+            // FIXME Allow arbitrary providers?
+            if (
+              provider.currentProvider instanceof AlchemyWebSocketProvider ||
+              provider.currentProvider instanceof AlchemyProvider
+            ) {
```
cairomassimo, line 67: Is it possible to invert the logic here, and make the `if` into an early return (outside of `try`)?
We are 7 indentation levels deep.
If not, maybe extracting a method would make it look better.

background/services/chain/asset-data-helper.ts
```diff
@@ -0,0 +1,155 @@
...
+    tokenSmartContract: SmartContract
+  ): Promise<SmartContractFungibleAsset | undefined> {
+    return this.withProviderForNetwork(
+      tokenSmartContract.homeNetwork,
+      async (provider) => {
+        try {
+          if (
+            provider.currentProvider instanceof AlchemyWebSocketProvider ||
+            provider.currentProvider instanceof AlchemyProvider
+          ) {
```
cairomassimo, line 98: Same as above

background/services/chain/asset-data-helper.ts
```diff
@@ -0,0 +1,155 @@
...
+  ): Promise<AssetTransfer[]> {
+    return (
+      this.withProviderForNetwork(
+        addressOnNetwork.network,
+        async (provider) => {
+          try {
+            if (
+              provider.currentProvider instanceof AlchemyWebSocketProvider ||
+              provider.currentProvider instanceof AlchemyProvider
+            ) {
```
cairomassimo, line 134: Same as above

background/services/chain/asset-data-helper.ts
```diff
@@ -0,0 +1,155 @@
...
+ * optimizations.
+ */
+export default class AssetDataHelper {
+  constructor(private providerTracker: ProviderManager) {}
+
+  private withProviderForNetwork<T>(
+    network: EVMNetwork,
+    handler: (provider: SerialFallbackProvider) => T
+  ): T | undefined {
+    const provider = this.providerTracker.providerForNetwork(network)
```
cairomassimo, line 41: I would replace this with a simpler method `getProviderForNetwork`. See below.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -255,9 +248,78 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
-  // Overriding internal functionality here to support event listener
-  // restoration on reconnect.
-  // eslint-disable-next-line no-underscore-dangle
+  /**
+   * Subscribe to pending transactions that have been resolved to a full
+   * transaction object; uses optimized paths when the provider supports it,
+   * otherwise subscribes to pending transaction hashes and manually resolves
+   * them with a transaction lookup.
+   */
+  async subscribeFullPendingTransactions(
```
cairomassimo, line 257: This method is quite long and wide (indentation-wise).
Would benefit from extracting methods from its body and/or a rework.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -255,9 +248,78 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
+
+            handler(transaction)
+          } catch (error) {
+            logger.error(
+              `Error handling incoming pending transaction: ${result}`,
+              error
+            )
+          }
+        }
+      )
```
cairomassimo, line 291: E.g., these lines could be extracted into a `private` method `subscribeFullPendingTransactionsAlchemy(...)` (with same signature as `subscribeFullPendingTransactions`)

background/services/chain/serial-fallback-provider.ts
```diff
@@ -255,9 +248,78 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
+    } catch (error) {
+      const errorString = String(error)
+      if (errorString.match(/unsupported subscription/i)) {
```
cairomassimo, line 294: We are silently ignoring an error if the condition is false.

Perhaps
```typescript
      if (!errorString.match(/unsupported subscription/i)) throw error
```
or at the bare minimum we need a comment saying that we are ignoring errors purposfully.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -255,9 +248,78 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
+              handler(transaction)
+            }
+          } catch (innerError) {
+            logger.error(
+              `Error handling incoming pending transaction hash: ${transactionHash}`,
+              innerError
+            )
+          }
+        })
+      }
```
cairomassimo, line 314: This could also be another method, say, `subscribeFullPendingTransactionsStandard`.

cairomassimo (changes requested):
Some valueable code improvements that should be done before merging this

background/services/chain/asset-data-helper.ts
Shadowfiend: It's only guaranteed by eslint rules, unfortunately. `undefined` can also be overridden globally by setting `window.undefined` in any dependency, so even if the file itself is in strict mode it's not protected from this kind of scope poisoning.

I'm intending on turning on a raft of TypeScript-aware rules that will flag these kinds of conditionals and make sure we're being fully defensive with them sometime after launch.

background/services/chain/asset-data-helper.ts
Shadowfiend: Strong disagree as to the simplicity here (this seems like a familiarity thing, not a simplicity thing), but it's a pattern that's not really in use in our codebase and should go.

background/services/chain/serial-fallback-provider.ts
Shadowfiend: Yikes, yep. Thanks for flagging.

background/services/chain/asset-data-helper.ts
cairomassimo: I just investigated this a bit. It looks like we are safe.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#description):

> In modern browsers (JavaScript 1.8.5 / Firefox 4+), undefined is a non-configurable, non-writable property, per the ECMAScript 5 specification. (Even when this is not the case, avoid overriding it.)

I tested this:

```javascript
function a() { undefined = 1; return undefined }
function b() { 'use strict'; undefined = 1; return undefined }
```

In the above example, `a` still returns `undefined`, not `1`, while `b` throws `TypeError`. IIUC, in no case ECMAScript 5+ allows global `undefined` to be overwritten.

ES5 has 99.5% browser support (https://caniuse.com/es5), i.e., everything apart from IE8 (which we definitely don't target).

Shadowfiend: You're absolutely right! In fact, we had this conversation when we originally introduced the approach. Unfortunately the above fails in the case of a local variable (unless `use strict` is enabled); as a minimal test case:

```javascript
> (function() { let undefined = "blap"; let booyan = "blap"; return booyan === undefined })()
< true
```

On paper this shouldn't be a concern, but given that we're writing relatively sensitive code, we chose to operate with a little extra paranoia around this particular issue.

I haven't been mentioning it across the board because it's a bit excessive to enforce this without a linter rule, and my intent was to switch such a rule on at some point… However, the rule that does this ([`no-undefined`](https://eslint.org/docs/rules/no-undefined)) doesn't support autofix… So I'm thinking it's time to pull back from that stance.

greg-nagy: I had this very same question and argument for sooo long. It's so calming to finally have an answer.

background/services/chain/serial-fallback-provider.ts
Shadowfiend: I went a slightly different path here, since the “standard” path is really the final fallback. The Alchemy path is in its own function and returns `"unsupported"` as a non-error return value if the provider returns an unsupported error, at which point we fall back on this standard approach.

Other Alchemy errors are thrown up as proper errors, and propagated into callers.

Shadowfiend:
This guy should be ready for another look.

greg-nagy (approved):
I think these changes are good to 🚢

@Shadowfiend Would you mind shareing a couple of thoughts about what you have in mind for dealchemyfing the codebase?

Shadowfiend:
> Would you mind shareing a couple of thoughts about what you have in mind for dealchemyfing the codebase?

There are 1-3 RFBs in this… Keep an eye out :)

PR #1047, opened 2022-02-26
Remove /slices in UI readme

gasolin:
replace with signed commit in #1044

Tracing source and found inaccurate file structure in [ui/README.md](ui/README.md)
I guess the /slices is moved into background [background/redux-slices/ui.ts](background/redux-slices/ui.ts)


mhluongo (approved):
Thanks @gasolin! :muscle:

PR #1050, opened 2022-02-26
✨ support build edge extension

gasolin:
with a small config change its possible to load Tally on the Edge browser (windows/mac)


Shadowfiend (approved):
Let's do it!

PR #1053, opened 2022-02-28
Improve approve spend limit input UX

cairomassimo:
Implements the following UX changes:

- empty input mean no limit (no need to type "Infinite"),
- show "Infinite" as input placeholder,
- show snackbar instead of silently ignoring invalid input.

### DIFF
--- a/ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
+++ b/ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
@@ -12,10 +12,11 @@ import {
   truncateAddress,
 } from "@tallyho/tally-background/lib/utils"
 import { updateTransactionOptions } from "@tallyho/tally-background/redux-slices/transaction-construction"
+import { setSnackbarMessage } from "@tallyho/tally-background/redux-slices/ui"
 import { AssetApproval } from "@tallyho/tally-background/services/enrichment"
 import { ethers } from "ethers"
 import { hexlify } from "ethers/lib/utils"
-import React, { ReactElement, useEffect, useState } from "react"
+import React, { ReactElement, useState } from "react"
 import { useDispatch } from "react-redux"
 import FeeSettingsText from "../NetworkFees/FeeSettingsText"
 import SharedAssetIcon from "../Shared/SharedAssetIcon"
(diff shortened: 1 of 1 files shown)


cairomassimo:
> Any chance we can split out the default from the snackbar change?

@Shadowfiend, I didn't get which of the two you want to be merged. Anyway @VladUXUI prepared a new design, so we might as well implement the final design and merge it when ready. Let me know :)

cairomassimo:
Will implement the updated UI spec then set this PR back to ready-to-review.

henryboldi (changes requested):
Overall this is an awesome addition! 👏👑 So a couple things come to mind which are mostly UI/UX,

- The input validation `snackbar` only appears when clicking save — but not when hitting enter.

- If the user can only enter numbers, the input field should probably only allow numbers.

- When the input field is open the user can't scroll far enough to edit gas. Not a big deal, but still something to note.
can't scroll

- We don't use snackbars to display errors on any other input. How about sticking with the pattern of outlining the input area in red and adding the error below? cc @VladUXUI

-  If it doesn't already exist, it'd probably make sense to get a design from @VladUXUI for this.

greg-nagy:
@cairomassimo Can you please refresh this PR?

cairomassimo:
I fixed:

> The input validation snackbar only appears when clicking save — but not when hitting enter.

and

> We don't use snackbars to display errors on any other input. How about sticking with the pattern of outlining the input area in red and adding the error below?

Screenshot from 2022-03-14 11-05-04

The others require more work and are a bit out of scope for this PR (which has already grown in scope a lot)

> - If the user can only enter numbers, the input field should probably only allow numbers.
> - When the input field is open the user can't scroll far enough to edit gas. Not a big deal, but still something to note.

I will open issues for them.

cairomassimo:
This was already fixed in main:

> When the input field is open the user can't scroll far enough to edit gas. Not a big deal, but still something to note.

About this:

> If the user can only enter numbers, the input field should probably only allow numbers.

I leave to @VladUXUI to decide if we should fix it (and, in that case, we can open the issue)

henryboldi (approved):
Looks great - thanks for these updates! 👑👏

There appears to be no way in the UI to return the approve limit back to infinity. Not strictly blocking. I'll open up an issue to track this.

PR #1054, opened 2022-02-28
Fix page numbering in Ledger import screen

### LINKED ISSUE #1048: Ledger path pages numbering
### Discord Discussion Link

n/a

Chrome

Try out Ledger support

### What did not work?

It's working, just a minor UI update. 

It shows on the bottom
0-6
6-12

I think it would be better to either have 
0-5
6-11

Or
1-6
7-12

Maybe 1-6 would be better as it's then matching MM numbering.

Screen Shot 2022-02-26 at 13 20 04

Screen Shot 2022-02-26 at 13 20 10

### Version

Community Edition v0.11.2

### Relevant log output

_No response_

cairomassimo:
Show first and last address indexes, both inclusive.
Fixes #1048.


ui/pages/Ledger/LedgerImportAccounts.tsx
```diff
@@ -33,8 +33,8 @@ function usePageData({
   const paths: string[] = []
 
   const firstIndex = pageIndex * addressesPerPage
-  const lastIndex = (pageIndex + 1) * addressesPerPage
-  for (let i = firstIndex; i < lastIndex; i += 1) {
+  const lastIndex = (pageIndex + 1) * addressesPerPage - 1
+  for (let i = firstIndex; i <= lastIndex; i += 1) {
```
Shadowfiend, line 37: Not sure the best place to do this (probably in the `current_page`?), but let's also heed the suggestion in the issue to start the UI count at 1 for the first address, rather than 0.

Can merge this PR + do it separately or hold until that happens–up to you!

Shadowfiend (approved):

ui/pages/Ledger/LedgerImportAccounts.tsx
cairomassimo: The index is zero-based because it's the one we use in the address derivation path.
(About the code, I usually prefer code in high-level language such as JS to be symmetric with bounds, so both inclusive in this case. I didn't do it originally, which I didn't like. No problem in adopting a different convention, just let me know.)

PR #1059, opened 2022-03-01
Do not spam Blocknative

### LINKED ISSUE #949: Only check gas prices frequently when the extension visible
... otherwise, we only need gas prices ...

* after installation
* at extension startup
* every 10 minutes to every hour in the background

This will prevent a bunch of wasted BlockNative and RPC calls.

fernya:
Fixes #949

Waiting for #1041 to be merged to have a proper event trigger

Now we can configure integer multiples for selected alarms and 

Current implementation polls Blocknative on:
  - startup,
  - every 2 minutes when the Wallet is shown,
  - every 60 minutes when the Wallet is hidden,
  - and every time when the Wallet appears and more than 2 minutes elapsed since the last poll

# QA
If the event trigger is finalized, then one can check
- if the prices are updated correctly,
- no traffic goes to Blocknative if the Wallet is hidden

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -85,6 +85,7 @@ import {
 import { ETHEREUM } from "./constants"
 import { HIDE_IMPORT_LEDGER } from "./features/features"
 import { SignatureResponse, TXSignatureResponse } from "./services/signing"
+import { PollingPeriod } from "./services/chain"
 
 // This sanitizer runs on store and action data before serializing for remote
 // redux devtools. The goal is to end up with an object that is directly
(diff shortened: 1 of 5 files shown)


Shadowfiend (commented):
I'm not sure this approach will work. There can be multiple versions of the Tally popup open at any given moment (e.g. one or more dApp confirmation popups + the popover from the Tally icon), so to do this successfully we'd need to have a counter based on the source. I think @cairomassimo actually had the core of a solution in #1041 based on opening/closing ports, which gives you uniqueness on the source popup.

fernya:
I think I'll prototype the polling setting using this trigger method. If its done, then it could be melted together with @cairomassimo's approach.

mhluongo:
Excited to ramp balance checks and other stuff up and down based on this as well.

background/main.ts
```diff
@@ -1146,6 +1147,12 @@ export default class Main extends BaseService<never> {
         )
       }
     )
+
+    uiSliceEmitter.on("setUiVisibility", async (visibility: boolean) => {
```
0xDaedalus, line 1151: you may want to rename `visibility` => `visible` to match what the variable name in `redux-slices/ui.ts`

mhluongo:
Closing, though we should revisit this idea down the road.

PR #1060, opened 2022-03-01
Books Cannot Be Killed By Fire - Removing Addresses from the Accounts Panel

0xDaedalus:
This PR introduces the ability to remove non read-only addresses from the accounts list. For keyring-addresses, this is done through a new #hiddenAccounts property inside of keyringService that acts as a filter for what addresses to emit with keyrings when emitting via the emitKeyrings method. Note that users cannot delete their currently selected address.

Removed addresses that are associated with a keyring are easily able to be re-added with the `+ Address` button - and addresses are not removed from accountsToTrack when they are removed so that users have the UX of being able to immediately see the current state of addresses they have previously hidden when unhiding them.

### Testing
- [ ]  User should be able to add multiple read-only, imported, and tally addresses.
- [ ]  User should be able to remove a read-only address from the accounts list and add it back
- [ ]  Removing the last address of a tally & imported keychain should remove the Account from the accounts list
- [ ]  Users should be able to remove and re-add the same address from an Account with multiple addresses
- [ ]  Users should be able to remove an account (delete all addresses associated with it) and re-add it using its seed phrase.

https://user-images.githubusercontent.com/94649004/157472901-659649e5-4ee3-4ff5-9c8f-f1dbc45f9809.mov


mhluongo:
Love this! Quick thought just from the GIF — let's tighten up that copy with something like

> Removing this address doesn't delete your recovery phrase or any private keys. Instead...

Thought being that we want to always use "recovery phase" rather than "seed", and that nothing is ever really deleted "from" your phrase anyway

0xDaedalus:
Great point - here is an updated screenshot of the confirm menu.

image

0xDaedalus:
> Also, are we doing this only for read-only?

For this PR yes - otherwise the diff would have been quite large.  Removal for keychain addresses is up next.

ui/components/AccountItem/AccountItemSummary.tsx
```diff
@@ -0,0 +1,124 @@
...
+          background-size: cover;
+          width: 48px;
+          height: 48px;
+          border-radius: 12px;
+        }
+        .avatar_selected_outline {
+          width: 52px;
+          height: 52px;
+          border-radius: 16px;
+          border: 2px solid #22c480;
```
VladUXUI, line 71: I think it would be good if we would use colour var everywhere. This becomes important when we are doing the light version or other colour combinations for accessibility

VladUXUI, line 71: I think it would be good if we would use colour var everywhere. This becomes important when we are doing the light version or other colour combinations for accessibility

VladUXUI (approved):

0xDaedalus:
Vlad that's a great point - I went ahead and fixed the issue in a number of other places in a separate PR: https://github.com/tallycash/extension/pull/1077

ui/components/Shared/SharedPanelAccountItem.tsx
```diff
@@ -1,8 +1,10 @@
 import React, { ReactElement } from "react"
 
 import { AccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
+import { HexString } from "@tallyho/tally-background/types"
 
-import SharedLoadingSpinner from "./SharedLoadingSpinner"
+import AccountItemOptionsMenu from "../AccountItem/AccountItemOptionsMenu"
+import AccountItemSummary from "../AccountItem/AccountItemSummary"
```
Shadowfiend, line 7: If the container is under `Shared`, should this also be under `Shared` so that it's clear it's a shared component? This feels right to me, but open to disagreement for sure.

0xDaedalus: You know I debated this with myself for a while - I do agree it feels weird that we have a `Shared` component that is basically just an `li` that immediately imports other components not in the `Shared` directory.  The point I was struggling with was at-what-point does something go from `shared` to feature-specific (e.g. the menu to meclearly is `AccountItem` specific).

Maybe the answer here is to change `AccountItemSummary` => `SharedAccountItemSummary` and to render `AccountItemOptionsMenu` via the `children` prop and pass it into `SharedPanelAccountItem`?

Shadowfiend: The summary rename sounds like a good plan, as we'll be using that a couple of places. I think we might could eliminate this component entirely as it's not clear it's providing much value at this point?

Shadowfiend: (None of this is blocking btw.)

0xDaedalus:
Now that ##1079 has been merged in the scope of this PR has increased.  All addresses (Import, Tally, Read-only) should now be removeable.

Shadowfiend (commented):
Left a few notes on the keyring hiding piece—which is not what this PR is about, so I'm not sure how that code ended up here :grimacing: Since we're not focused on that, let's leave any related changes for the PR that implements it, though we can discuss ofc.

Posting it in its own review so we can keep it separate.

background/services/keyring/index.ts
```diff
@@ -325,9 +327,13 @@ export default class KeyringService extends BaseService<Events> {
       [newKeyring.id]: { source },
     }
     newKeyring.addAddressesSync(1)
+    const [address] = newKeyring.getAddressesSync()
+    // Make sure that the first address of a keyring is not "hidden".
+    if (this.#hiddenAccounts[address]) {
+      this.#hiddenAccounts[address] = false
+    }
```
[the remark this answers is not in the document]
0xDaedalus: > I think we also need to persist the hidden accounts info, maybe as part of the metadata you introduced?

How do you feel about doing this in a subsequent PR?  I think I'm definitely going to need to do this when I add the ability to rename / change image on addresses.  Maybe as a separate #addressMetadata field but I haven't put any real thought into it yet.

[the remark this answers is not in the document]
mhluongo: > Should we delete the keyring at that point? I feel like the answer is yes.

Are we talking about deleting key material? If so, I don't think we should until we offer a way for people to do password-backed exports.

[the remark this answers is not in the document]
0xDaedalus: Hmmmnnnn yeah I see your point - I do think it makes sense to minimize the chances of users losing their accounts - but at the same time "your keyring exists inside of the extension until you uninstall" is not really a good state to be in either.  Maybe its a question of priority?  Do we feel like we can punt on _true_ keyring deletion for a few weeks?

0xDaedalus:
Updated warning when users are attempting to remove the last address on a keyring.

image

ui/components/AccountItem/AccountItemOptionsMenu.tsx
```diff
@@ -0,0 +1,142 @@
+import { AccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
+import { HexString } from "@tallyho/tally-background/types"
+import React, { useRef, useState } from "react"
+import { useOnClickOutside } from "../../hooks"
+import SharedSlideUpMenu from "../Shared/SharedSlideUpMenu"
+import AccountItemRemovalConfirm from "./AccountItemRemovalConfirm"
+import RemoveAddressLabel from "./AccountItemRemoveAddressLabel"
+
+interface AccountItemOptionsMenuProps {
```
Shadowfiend, line 9: Non-blocking: we should really switch to using `type =` for these, we've made the switch consistently everywhere except for components.

ui/components/AccountItem/AccountItemOptionsMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+  address: HexString
+  isSelected: boolean
+}
+
+const AccountItemOptionsMenu: React.FC<AccountItemOptionsMenuProps> = ({
+  hideMenu,
+  accountTotal,
+  address,
+  isSelected,
+}) => {
```
Shadowfiend, line 21: No real need to block on this, but we've chosen to use `function` instead of `const = ()` for these guys, and avoid `React.FC` as there are very mixed opinions about its usefulness (see https://github.com/facebook/create-react-app/pull/8177).

We also generally inline `export` with the declaration, rather than doing it as a separate statement.

ui/components/AccountItem/AccountItemOptionsMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+interface AccountItemOptionsMenuProps {
+  hideMenu: boolean
+  accountTotal: AccountTotal
+  address: HexString
+  isSelected: boolean
+}
+
+const AccountItemOptionsMenu: React.FC<AccountItemOptionsMenuProps> = ({
+  hideMenu,
```
Shadowfiend, line 17: This feels like it leaks parent intent into the component. If the parent doesn't want this rendered, the parent just shouldn't render it—IMO, having the component not render itself is something that is either so useful that we should add it to all components—or not useful enough to add it to any. I fall on the latter train :grin:

ui/components/AccountItem/AccountItemOptionsMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+  if (hideMenu) {
+    return <></>
+  }
+  return (
+    <>
+      <SharedSlideUpMenu
+        size="custom"
+        customSize="336px"
+        isOpen={showAddressRemoveConfirm}
+        close={() => {
```
Shadowfiend, line 38: Oof, definitely not blocking but we should really name this `onCloseClick` to follow our event handler naming. Not sure how I've missed it before now 😅

ui/components/AccountItem/AccountItemRemovalConfirm.tsx
```diff
@@ -0,0 +1,125 @@
...
+)
+
+const LoudWarning: React.FC = () => (
+  <span>
+    <h3>
+      Removing this address will remove its associated account from the UI.
+    </h3>{" "}
+    Are you sure you want to proceed?
+  </span>
+)
```
Shadowfiend, line 36: Why have either of these be components? They can just be constants, right?

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
```diff
@@ -228,6 +229,9 @@ export default function AccountsNotificationPanelAccounts({
                     {accountTotalsByKeyringId.map((accountTotal) => {
                       const lowerCaseAddress =
                         accountTotal.address.toLocaleLowerCase()
+                      const isSelected =
+                        lowerCaseAddress === selectedAccountAddress
```
Shadowfiend, line 233: We should use `sameEVMAddress` here.

I'm also partial to inlining it, tbh, but not super opinionated on that front.

ui/components/AccountItem/AccountItemOptionsMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { useOnClickOutside } from "../../hooks"
+import SharedSlideUpMenu from "../Shared/SharedSlideUpMenu"
+import AccountItemRemovalConfirm from "./AccountItemRemovalConfirm"
+import RemoveAddressLabel from "./AccountItemRemoveAddressLabel"
+
+interface AccountItemOptionsMenuProps {
+  hideMenu: boolean
+  accountTotal: AccountTotal
+  address: HexString
+  isSelected: boolean
```
Shadowfiend, line 13: Hmmm… It's not clear to me why this is needing to be passed. It looks like we're passing it through so we can render it in the account item summary in the removal slide-up, but we're also not actually rendering this menu at all unless the item isn't selected, so it feels like this isn't a useful prop to pass through? We can just use `false` for it when we render it in the slide-up, right?

ui/components/AccountItem/AccountItemRemoveAddressLabel.tsx
```diff
@@ -0,0 +1,46 @@
+import classNames from "classnames"
+import React from "react"
+
+interface RemoveAddressProps {
+  hoverable?: boolean
+}
+
+const RemoveAddressLabel: React.FC<RemoveAddressProps> = ({ hoverable }) => {
```
Shadowfiend, line 8: I'm not really sure how much bang this component brings for its new-file-and-unclear-prop buck. The two uses of this label are fairly different in their intent, with one being a button and the other being an actual header. We may not even want to use the same containing element (some sort of header tag in the slide-up menu wouldn't be the wrong thing, for example).

We do gain some light sharing of styling, but because the semantic relationship is tenuous between the two uses, we already see some of the abstraction leaking by needing this `hoverable` prop that only applies changes to the base styles. And looking more closely at the design even some of those styles that aren't overriden aren't actually shared—the button form, for example, uses 16px text, vs the header form which uses 24px (which makes sense, since one is a header!).

Mostly we're just sharing the flex setup, font color, and line-height. In practice, I expect we'll want to apply these consistently among all items in the account options menu, so those styles should live there, and the styling of the slide-up menu headers will mostly be independent.

The last thing that these consistently share is the icon, but here I'd like to suggest that we stick to @henryboldi approach of using `mask-image`. Even though `filter` works, the outcome makes it very non-obvious that we're trying to achieve an identical color between the icon and the text! `mask-image` lets us set the color declaratively in the CSS, which makes it much clearer what the code intent is for future updates.

Shadowfiend (changes requested):
Left a few comments, though didn't make it through all the code.

Few more notes from testing:

- The fact that the item under the options menu still receives hover events makes the menu rendering look buggy.
- The options menu seems taller than the design dictates, the button text larger (I mention that in one of the other comments), and the X also seems larger.
- The options menu should be a few pixels off the right side of the accounts menu, vs flush with it.
- I imagine we should have a hover state for the options menu's X button (cc @VladUXUI on this one, may be wrong).
- ~~I'm guessing the hover state for the remove account button isn't a text color change, but a background color change, in preparation for having more than one item in the menu.~~ -> missed the area of the design where this is defined, color was the right thing here.
- I expect the hit area of the menu is divided between the X and the remove account button; I don't think there should be any area that is “neither”. (Clicking this “neither” area triggers the underlying account to be selected, which is pretty unexpected, but I think the above is the real issue.)
- I believe the currently-selected account should also have an options menu.

background/services/keyring/index.ts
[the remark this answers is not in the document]
Shadowfiend: > How do you feel about doing this in a subsequent PR?

If we're exposing account removal for keyrings, not great. We need to have that right otherwise a user can remove addresses, restart their browser, and see them appear again. Not a good look.

Shadowfiend:
A note about the review above: the only blockers I think are the design notes. The rest is refactoring-related but doesn't need to land before merge IMO.

ui/components/AccountItem/AccountItemOptionsMenu.tsx
0xDaedalus: ✅  - I'll go ahead and do this in a separate PR since I imagine the change will touch quite a few different components.

0xDaedalus:
Feedback addressed for the most part -  tyvm for the deliberate and detailed feedback!  Some notes:

> I believe the currently-selected account should also have an options menu.

I don't think we should do this _yet_.  The only option in the menu right now is to remove an address - and I don't think we should allow them to remove their currently selected address.  Once we have more things in the menu (like rename address) I agree we certainly should have the selected account have an options menu (without the `Remove Address` option).

> I imagine we should have a hover state for the options menu's X button (cc @VladUXUI on this one, may be wrong).

Happy to do this with blessing from Vlad

I think I've addressed all of the design-notes - should be good-to-go for another round of testing.

Shadowfiend:
> I don't think we should allow them to remove their currently selected address

Can you talk about this some more? Unless it's the last remaining address in the wallet (and even then tbh), why not? I can eg delete or archive the current email I'm viewing in gmail or Mail, for example.

VladUXUI:
> Updated warning when users are attempting to remove the last address on a keyring.
> 
> image

Couple of things i'm seeing here:

- [ ] The main button seems to have the attention colour. Let's please use our primary BT component here (i'm aware it's currently white over gold)

- [ ] The balance seems to be very far away from the menu. for accounts that are not selected.

VladUXUI:
I also agree that need to allow users to delete the connected account. As users will most likely switch to the account they want to delete in order to check assets and make sure that is the one they are removing.
By not allowing them to do so, we are asking them to take too many extra steps and and it adds a bit of "Was this the account i want to remove" factor.

On another note, allowing the above will also fix the issue were a menu appears and disappears, and the balance moves left/right.

0xDaedalus:
1. Sounds good - you guys are right - allowing users to delete the connected account does make more sense ✅.  I've gone ahead and implemented that as well as @VladUXUI's comments above and have updated the original description of the PR with a new video of the whole flow.

Shadowfiend:
The sizing still doesn't look right unfortunately. Right now we're implementing sizing for the popup via a combo of:

- Padding on the options `ul`
- Fixed heights for the `button` elements.

I recommend dropping the fixed heights across the board and moving to a single padding on the `li`s at 14px, which aligns with the padding visible across the board in the design. The right margin for the close button can then move from 5px to 4px (for a total right spacing of 18px, again matching the design). That'll also set us up for future menu items, should they appear :grin:

At that point, the text will look tight to the X, because the text is still too big—it should be 16px rather than the current 18, and the weight should be 500 rather than the current `bold` which [corresponds to 700](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#values).

Lastly, we still need a hover state for that X and the triggering `⋮`:

> for hover state. We use the same state for all icons (that are green/white) 
>
> image

Shadowfiend:
This guy's been cooking for long enough that once we fix the design issues above I'm planning on giving it the green light, btw.

VladUXUI:
Any chance we could see an updated screenshot for this?

Shadowfiend:
@VladUXUI the video in the description has been updated to the latest version :)

Shadowfiend:
> I think I see the right-spacing as 12px - am I looking in the wrong place?

NOPE I was looking at the wrong spacing because I was too zoomed out lol. I see 10, with respect to the box around the X—but in practice that box is more likely to match the padding requirement (vs the edge of the X which will require pixel measurement).

> Do either of you have a recommendation for a tool that lets you see pixel spacing between elements in the DOM?

I cheat real hard and use the screenshot cross (Cmd-Shift-4) to measure and hit Esc before it takes the screenshot 😅 Click-and-drag gives you pixel counts on the screenshot area 🤷

0xDaedalus:
> I cheat real hard and use the screenshot cross (Cmd-Shift-4) to measure and hit Esc before it takes the screenshot 😅 Click-and-drag gives you pixel counts on the screenshot area 🤷

🤯 🤯 🤯

background/services/keyring/index.ts
```diff
@@ -360,7 +372,16 @@ export default class KeyringService extends BaseService<Events> {
       throw new Error("Keyring not found.")
     }
 
-    const [newAddress] = keyring.addAddressesSync(1)
+    const keyringAddresses = keyring.getAddressesSync()
+
+    // If There are any hidden addresses, show those first before adding new ones.
+    const newAddress =
+      keyringAddresses.find(
+        (address) => this.#hiddenAccounts[address] === true
```
Shadowfiend, line 380: May need to normalize here.

background/services/keyring/index.ts
```diff
@@ -369,6 +390,37 @@ export default class KeyringService extends BaseService<Events> {
     return newAddress
   }
 
+  async hideAccount(address: HexString): Promise<void> {
+    this.#hiddenAccounts[address] = true
```
Shadowfiend, line 394: May need to normalize here as well.

Shadowfiend (approved):
Okay, this seems to work well across the board. Slightly weird that if I remove address 1 and then add it back it goes to the end… But it's a corner case and this feature is an improvement on the current state of play for folks, so let's not block.

Left a comment on the component structure here that's worth addressing, but not worth blocking on. Let's roll it :rocket:

greg-nagy:
I was preparing myself to get through this when I saw it was merged. Thanks @Shadowfiend 

And awesome work @0xDaedalus.

This PR and feature is truly a good showcase of eng and design and product cross over. I was happy to go over this convo.

PR #1079, opened 2022-03-03
Enable hiding non `read-only` addresses.

0xDaedalus:
This PR introduces the ability to remove non read-only addresses from the accounts list.  This is done through a new `#hiddenAccounts` property inside of `keyringService` that acts as a filter for what addresses to emit with keyrings when emitting via the `emitKeyrings` method.

Removed addresses are easily able to be re-added with the `+ Address` button - and addresses are not removed from `accountsToTrack` when they are removed so that users have the UX of being able to immediately see the current state of addresses they have previously hidden when unhiding them.

This PR is blocked until #1060 is merged.

https://user-images.githubusercontent.com/94649004/156622214-21e56318-0bdc-441b-ab77-a09c6fa58aad.mov

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -171,7 +171,6 @@ function migrateReduxState(
 const reduxCache: Middleware = (store) => (next) => (action) => {
   const result = next(action)
   const state = store.getState()
-  ;(window as any).store = store
 
   if (process.env.WRITE_REDUX_CACHE === "true") {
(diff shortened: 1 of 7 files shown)


mhluongo (approved):
Concept ACK, I like this approach. Going to try to QA and go a little deeper later, so not merging yet.

PR #1080, opened 2022-03-03
Immutably Update this.#keyringMetadata to prevent exception.

0xDaedalus:
Without this fix - attempting to import a wallet after first creating a new Tally wallet resulted in an exception being thrown in the background thread and the importing `status` in the Keyring Redux Slice being stuck at `pending`.

Not sure why this did not show up until after `main` was merged into #1001 - but this should fix the issue of not being able to add multiple separate `Tally` and `Import` keyrings.

To Test:

- Add a `Tally` wallet from a new Seed.
- Import a wallet from a metamask seed.
- You should see both wallets in the Account List.
- Any combination of Tally / Imported wallets should be supported.

### DIFF
--- a/background/services/keyring/index.ts
+++ b/background/services/keyring/index.ts
@@ -320,13 +320,15 @@ export default class KeyringService extends BaseService<Events> {
       ? new HDKeyring({ mnemonic, path })
       : new HDKeyring({ mnemonic })
     this.#keyrings.push(newKeyring)
-    this.#keyringMetadata[newKeyring.id] = { source }
+    this.#keyringMetadata = {
+      ...this.#keyringMetadata,
+      [newKeyring.id]: { source },
+    }
     newKeyring.addAddressesSync(1)
     await this.persistKeyrings()
 
     this.emitter.emit("address", newKeyring.getAddressesSync()[0])
     this.emitKeyrings()
-
     return newKeyring.id
   }
 


henryboldi (approved):
Descriptive PR, clean code, and works nicely! 👑

Shadowfiend:
What was the exception you were seeing? It doesn't look like any path should leave the metadata field corrupted, so I'm puzzled as to what this particular change fixed.

0xDaedalus:
You know I am super puzzled as well.  I have definitely used this pattern before.
The error was:

```
Cannot add property {id} Object is not extensible
```

where {id} was the id of the keyring.

Shadowfiend:
I don't like this one bit heh.

As a small thing, can we make the emit (https://github.com/tallycash/extension/blob/bf05d7c0e2d47aef959140c5d492597592aaf64c/background/services/keyring/index.ts#L529) and the persist (https://github.com/tallycash/extension/blob/bf05d7c0e2d47aef959140c5d492597592aaf64c/background/services/keyring/index.ts#L544) usages of this use a copy? That is, `{ ...this.#keyringMetadata }` instead of `this.#keyringMetadata`. We probably shouldn't let this mutable private variable escape the access area of the KeyringService.

Not sure if that's the direct issue, basically wondering if something deep in the depths of redux or webext-redux or something is freezing the object or so.

PR #1081, opened 2022-03-03
Rename "NewSeed" accounts to "Internal"

0xDaedalus:
Per discussion in https://github.com/tallycash/extension/pull/1001.  We want to avoid using the `new seed` nomenclature.

### DIFF
--- a/background/redux-slices/accounts.ts
+++ b/background/redux-slices/accounts.ts
@@ -17,7 +17,7 @@ export const enum AccountType {
   ReadOnly = "read-only",
   Imported = "imported",
   Ledger = "ledger",
-  NewSeed = "newSeed",
+  Internal = "internal",
 }
 
 const availableDefaultNames = [
--- a/background/redux-slices/keyrings.ts
+++ b/background/redux-slices/keyrings.ts
@@ -7,7 +7,7 @@ import { Keyring, KeyringMetadata } from "../services/keyring/index"
 
 type KeyringsState = {
   keyrings: Keyring[]
-  keyringMetadata: { [keyringId: string]: { source: "import" | "newSeed" } }
+  keyringMetadata: { [keyringId: string]: { source: "import" | "internal" } }
   importing: false | "pending" | "done"
   status: "locked" | "unlocked" | "uninitialized"
   keyringToVerify: {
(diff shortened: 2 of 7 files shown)


background/redux-slices/accounts.ts
```diff
@@ -17,7 +17,7 @@ export const enum AccountType {
   ReadOnly = "read-only",
   Imported = "imported",
   Ledger = "ledger",
-  NewSeed = "newSeed",
+  Internal = "internal",
```
Shadowfiend, line 20: Just a random note with no urgency or high importance, but given how we're using this, I'm very comfortable if we decide we want to simplify to an `export type AccountType = "read-only" | "imported" ...` thing instead of an `enum`.

I've cooled a bit on enums when their requirements are easily fulfilled by a literal type union. cc @wojciech-turek because I know I once recommended them to you heartily lol.

That would also allow us to elide some of the conversion steps we're having to do in this and the previous PRs.

background/redux-slices/keyrings.ts
```diff
@@ -7,7 +7,7 @@ import { Keyring, KeyringMetadata } from "../services/keyring/index"
 
 type KeyringsState = {
   keyrings: Keyring[]
-  keyringMetadata: { [keyringId: string]: { source: "import" | "newSeed" } }
+  keyringMetadata: { [keyringId: string]: { source: "import" | "internal" } }
```
Shadowfiend, line 10: We should extract a type for this and use it consistently, though I don't think that needs to block this PR. It could also be that we have a type for `KeyringAccountType = "import" | "internal"` and then compose it into `type AccountType = KeyringAccountType | ...`.

Shadowfiend (approved):
Change looks good… Out of time to test drive right now, but will try to give it a whirl tonight if no one else beats me to it.

mhluongo:
Thinking about this... does it require a Redux migration @0xDaedalus?

henryboldi (approved):
@mhluongo, I checked the codebase for 0.12.0 and there's no use of "newSeed" in there. So it doesn't appear a Redux migration is needed in terms of this rename. That said, is a migration needed for `keyringMetadata` in general, @0xDaedalus? I'm assuming the keyring service initializes it, but not certain.

Anyway, test driven and working ✅👏

0xDaedalus:
@mhluongo I don't think that _this_ PR needs a new migration but @henryboldi brings up a great point about migrating `keyringMetadata`.  Looking at the code I don't _think_ a migration is needed (keyringService initializes `keyringMetadata`) but better safe than sorry.  Will test now.l

PR #1089, opened 2022-03-04
Fix flash of unsized popup

### LINKED ISSUE #924: Flash of extension popup missing height and width
I believe introduced in https://github.com/tallycash/extension/pull/893/, the extension popup flashes without a height and width. Demo GIF below,

flash

cairomassimo:
Fixes #924.

Flash was caused by dynamic loading of CSS.
This commit restores the popup sizing as part the static, global CSS,
but uses a popup-specific class applied to `<body>`, to avoid poisoning
the style of full-page UIs.

### DIFF
--- a/ui/public/popup.html
+++ b/ui/public/popup.html
@@ -5,7 +5,7 @@
     <link rel="stylesheet" href="./fonts.css" />
     <link rel="stylesheet" href="./index.css" />
   </head>
-  <body>
+  <body class="popup">
     <div id="tally-root" />
     <script src="../ui.js" type="text/javascript"></script>
   </body>
--- a/ui/pages/Popup.tsx
+++ b/ui/pages/Popup.tsx
@@ -187,12 +187,6 @@ export function Main(): ReactElement {
       <>
         <style jsx global>
           {`
-            body {
-              width: 384px;
-              height: 594px;
-              scrollbar-width: none;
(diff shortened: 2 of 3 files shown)


ui/public/popup.html
```diff
@@ -5,7 +5,7 @@
     <link rel="stylesheet" href="./fonts.css" />
     <link rel="stylesheet" href="./index.css" />
   </head>
-  <body>
+  <body class="popup">
```
Shadowfiend, line 8: Should this only be applied when a popup is rendered? Not sure how the styled-jsx block works when it's global but defined in a component.

henryboldi: It's my understanding that class `popup` would still be accessible as a class for  `styled-jsx global`. The use of `styled-jsx` causes a flash of unsized popup while the JS is still rendering. `Styled-jsx`'s production build config should hypothetically make this a non-issue, so it precompiles. That'd be the robust fix. But that's a whole can of worms. 

This issue occurred when the popup width and height was moved from `index.css` into `popup.tsx`. This fix is essentially a reversal to what we originally had, while also distinguishing the popup from the new tab page. So @cairomassimo's approach makes a lot of sense to me.

henryboldi (approved):
Works well. Nicely done KING 👏👑

PR #1090, opened 2022-03-04
Fix swap list searchbox to the top

### LINKED ISSUE #1038: Weird scrolling in swap token list
Title and search bar are scrolling along with the list, causing overlap of close button and search bar. I'm assuming this is not the intended behviour, and that only the list should scroll, @VladUXUI?
Screenshot from 2022-02-24 15-11-48

jagodarybacka:
Fixes #1038

Fixed problem with tokens list searchbox that was scrolling with the list.
Now the list is scrolling but the searchbox stays on the top.

### DIFF
--- a/ui/components/Shared/SharedAssetInput.tsx
+++ b/ui/components/Shared/SharedAssetInput.tsx
@@ -125,7 +125,7 @@ function SelectAssetMenuContent<T extends AnyAsset>(
         </div>
       </div>
       <div className="divider" />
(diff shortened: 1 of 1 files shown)


wojciech-turek:
I think there is a small line underneath the panel that isn't necessary and covers part of the view. We should probably remove it. 

image

wojciech-turek (approved):
Works and looks good, shipping!

PR #1094, opened 2022-03-05
🤘 Forks Out: Use a local mainnet fork

greg-nagy:
This PR has got itself a multi-tiered joke ... 
But it's no fun if I share.

# What 

This PR adds the possibility to run a local mainnet fork and updates the extension codebase to use this fork.
With as little derivation from the original code paths as possible.

# Why

## Problem

We as dApp devs want a dev environment that walks like mainnet quacks like the mainnet but doesn’t cost gas. So we can develop our dApps in a production-like environment but doesn’t cost us an arm and a leg.

The solution is 2 fold: 
1) create a fork for the mainnet locally 
2) use the most dev-friendly wallet EVER  so I can interact with all the live dApps and contracts, but my own transactions are sent to the forked mainnet.

## Blocker issue 
(Murphy)

The only settings that hardhat fork works — since ~Saturday — if you set the chainId to 1337 -> hosted dApps won’t work if they receive that.

# How / Workaround

Set fork to have chainId 1337 and implement a kind of “man in the middle” solution in the extension.
Let’s replace the chainId in outgoing txs to 1337
modify ethers.js to use the localhost URL when the mainnet is selected

# Setup 

For more detailed description see `./dev-utils/local-chain/README.md`

Quick Start:
```
cd dev-utils/local-chain
yarn install
yarn start
```

# Demo
[Recordings and screenshot of usage](https://drive.google.com/drive/u/2/folders/1c2jPRuqPI-wUVgeYJsWebQ_IdTvMtwnG)

# Setup from the ground up 

- get the code
```
git clone https://github.com/tallycash/extension tally
cd tally
git checkout fork-demo
```
- setup Tally - terminal 1
```
// nvm use
npm install -g yarn
yarn install
yarn start --config-name chrome
```
- Start the mainnet fork - terminal 2
```
cd dev-utils/local-chain
yarn install
yarn start
```
- install tally in chrome
  - create a new profile in chrome
  - go to `chrome://extensions`
  - turn on developer mode (right upper corner)
  - `Load unpacked` (left upperish corner)
  - load tally from `./tally/dist/chrome`
  - import a wallet using this seed phrase `test test test test test test test test test test test junk`
  - go to https://app.uniswap.org/
  - connect you new wallet as injected
  - try to swap (might not succeed for the first time ... probably won't :) )


dev-utils/local-chain/contracts/Greeter.sol
```diff
@@ -0,0 +1,22 @@
...
+
+    function greet() public view returns (string memory) {
+        return greeting;
+    }
+
+    function setGreeting(string memory _greeting) public {
+        console.log("Changing greeting from '%s' to '%s'", greeting, _greeting);
+        greeting = _greeting;
+    }
+}
```
wojciech-turek, line 22: Do we need that contract here at all?

dev-utils/local-chain/scripts/sample-script.js
```diff
@@ -0,0 +1,32 @@
+// We require the Hardhat Runtime Environment explicitly here. This is optional
```
wojciech-turek, line 1: Do we need this file?

dev-utils/local-chain/test/sample-test.js
```diff
@@ -0,0 +1,19 @@
+const { expect } = require("chai")
```
wojciech-turek, line 1: Also, seems like default hardhat files.

wojciech-turek (dismissed):
This looks good for me, some files generated from new hardhat project seem to be unnecessary but the whole mainnet fork functionality works as expected. 

No blockers.

Does it require more testing using various scenarios so that we can consider is a fully fledged testing environment? **Yes.**
Is it good enough for our current needs and should me merge this now? **Also yes.**

dev-utils/local-chain/hardhat.config.js
```diff
@@ -0,0 +1,40 @@
+require("@nomiclabs/hardhat-waffle")
```
mhluongo, line 1: It'd be nice if we used our other Hardhat conventions here, eg typesafe config in Typescript, etc. You can see an example in https://github.com/tallycash/contracts/

background/constants/networks.ts
```diff
@@ -44,6 +44,13 @@ export const BITCOIN: Network = {
   family: "BTC",
 }
 
+export const FORK: EVMNetwork = {
```
mhluongo, line 47: This deserves a comment :)

mhluongo (commented):
Cool idea. I'd like to hear @Shadowfiend sound off on the overall structure, but provisional ACK

dev-utils/local-chain/contracts/Greeter.sol
greg-nagy: I left the contract in place because tbh I expect some `hello-world` level contracts as an entry point to have something to interact with. 

Maybe mock contracts?

wojciech-turek (approved):
Shipping!

Shadowfiend:
You can't impersonate an account you don't have the keys for in the wallet, even if you're connected to a mainnet fork. The hardhat command line I believe may allow some aspects of this so that you can send yourself ETH or similar, but the account in the wallet will need to be imported as a mnemonic or generated as a fresh account.

greg-nagy:
> how do i impersonate and get 1000eth

You can import the `test test test test test test test test test test test junk` seed and send it through the wallet. 
https://hardhat.org/hardhat-network/docs/reference#initial-state

Or you can write simple nodejs scripts to do it automatically

PR #1098, opened 2022-03-06
Token Immutability: Include a custom token list via IPFS

mhluongo:
The list *should* make it easy to tweak token display. Eventually, we can make sure that it's given priority over other lists.

Unfortunately, though the list validates against the same schema in its generating repo, our validation setup tosses it out. @greg-nagy I'm hoping this is something you immediately know how to fix, as I believe this list will fix many of our issues with WETH.

### DIFF
--- a/background/services/preferences/db.ts
+++ b/background/services/preferences/db.ts
@@ -84,6 +84,27 @@ export class PreferenceDatabase extends Dexie {
           })
       })
 
+    // Add the new default token list
+    this.version(5)
+      .stores({
+        preferences: "++id",
+      })
+      .upgrade((tx) => {
+        return tx
+          .table("preferences")
+          .toCollection()
+          .modify((storedPreferences: Preferences) => {
+            // eslint-disable-next-line no-param-reassign
+            storedPreferences.tokenLists = {
+              ...storedPreferences.tokenLists,
+              urls: [
+                "https://ipfs.fleek.co/ipfs/bafybeid6yf4wa52cz3dkblqykktvfofvmqq2jiwfnyxtz4ypkafmthtqzi",
+                ...storedPreferences.tokenLists.urls,
+              ],
+            }
+          })
+      })
+
     // This is the old version for populate
     // https://dexie.org/docs/Dexie/Dexie.on.populate-(old-version)
     // The this does not behave according the new docs, but works
--- a/background/services/preferences/defaults.ts
+++ b/background/services/preferences/defaults.ts
@@ -5,6 +5,7 @@ const defaultPreferences: Preferences = {
   tokenLists: {
     autoUpdate: false,
     urls: [
+      "https://ipfs.fleek.co/ipfs/bafybeid6yf4wa52cz3dkblqykktvfofvmqq2jiwfnyxtz4ypkafmthtqzi", // the Tally community-curated list
       "https://gateway.ipfs.io/ipns/tokens.uniswap.org", // the Uniswap default list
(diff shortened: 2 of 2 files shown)


mhluongo:
To convince yourself the list validates using other setups, here's the [generating repo](https://github.com/tallycash/token-list/) as we all the [build script](https://github.com/tallycash/token-list/blob/main/utils/build.ts#L15). While the repo uses a *slightly* newer version of `@uniswap/token-lists`, I've tried the older version `extension` uses locally and also don't have issues with its validation.

Shadowfiend:
Which issues with WETH are you referring to? Not sure adding adding another potentially-conflicting list to the existing ones that aren't resolving conflicts deterministically help overall either 🤔

Shadowfiend:
Update:

https://github.com/tallycash/extension/blob/33ca4ae298441be6d9ddc07c950d7857c48d5cda/background/lib/token-lists.ts#L25

To include `isValidUniswapTokenListResponse.errors`, and you should see what failed validation.

mhluongo:
> Which issues with WETH are you referring to?

We're getting a lot of reports in Discord around WETH not showing up that are difficult to reproduce. Unsure if they're actually specific to WETH or if it's just the most common because everyone wants it to swap. Looking at the other lists, though, I don't think this will help — I thought it was a WETH9 vs 10 issue or that WETH wasn't in our lists at all. Alas.

Also trying to tackle this pain by [enabling ETH in swaps](https://github.com/tallycash/extension/pull/1103).

> Not sure adding adding another potentially-conflicting list to the existing ones that aren't resolving conflicts deterministically help overall either

Outside WETH, this list is just tokens we don't have elsewhere that users have requested. It shouldn't increase the confusion — if anything, this gives us a spot to start building a groomed canonical list.

greg-nagy:
The issue here is that `Polygon (Matic)` does not match the uniswap schema for a token name `^[ \w.'+\-%\/À-ÖØ-öø-ÿ:]+$`
I think I will add a shortcut for this and merge the PR

mhluongo:
Hm, I need to figure out why the token list didn't catch this validation error itself... this will be a constant foot gun otherwise.

https://github.com/tallycash/token-list/runs/5436139458

greg-nagy:
Wanted to include this in the release but the below solution feels really dirty. 

So 
- if you want to fix it in the token list repo, I am down to wait for it with this PR and start the release without this 
- I can fix the token list repo if you are swamped
- We can use the below changes as a temp patch but hm ... 


`export const isValidUniswapTokenListResponse: ValidateFunction` 

```
  if (!isValidUniswapTokenListResponse(json)) {
    const validationError =
      isValidUniswapTokenListResponse.errors &&
      isValidUniswapTokenListResponse.errors[0]
    if (validationError && validationError.instancePath !== "/tokens/5/name") {
      throw new Error(`Invalid token list at ${url}`)
    }
  }
```

greg-nagy:
Blargh ... newer ajv version in the token list repo ...

mhluongo:
Let's fix the token list itself. Hell, we can just snip Polygon for this release if we need

greg-nagy:
Don't have access to the repo :(

mhluongo:
> Don't have access to the repo :(

Odd. Alright, figuring that out — but in the meantime I patched the list and tested locally, and we're good.

The logos don't show up yet because our interface doesn't know how to deal with IPFS rn, but that's okay IMO.

greg-nagy (approved):
The issue is gone after the token list patch.  🐾

PR #1102, opened 2022-03-06
feat: Support Legacy (MEW/MyCrypto) Ledger HW wallet derivation path

### LINKED ISSUE #1086: Ledger derivation path dropdown menu is missing the Legacy m/44'/60'/0' path
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/948817668511780885

Chrome

On the "select Ledger accounts" screen, I was trying to access a Ledger account derived using the Legacy derivation path m/44'/60'/0'

Screen Shot 2022-03-03 at 6 51 15 PM

### What did not work?

The Legacy derivation path m/44'/60'/0' was not available from the derivation path dropdown menu. However, I was able to add it as a custom path and successfully access the desired wallet address in that manner.

Screen Shot 2022-03-03 at 6 32 56 PM

Screen Shot 2022-03-03 at 6 33 17 PM

### Version

Community Edition v0.12.0

### Relevant log output

```shell
N/A
```

kimpers:
This closes #1086, implementing support for the legacy Ledger derivation path `m/44'/60'/0'` , which was used early on in MyEtherWallet and MyCrypto wallets. Allowing you to connect old Ledger accounts to Tally

https://user-images.githubusercontent.com/5450545/156929614-f251656b-9639-4586-9d0f-bb560cd3540d.mov


ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
```diff
@@ -24,6 +24,10 @@ const initialDerivationPaths: { value: string; label: string }[] = [
     value: "m/44'/61'/0'/0",
     label: "Trezor",
   },
+  {
+    value: "m/44'/60'/0'",
+    label: "Legacy (MEW/MyCrypto)",
```
Shadowfiend, line 29: Was this used across the board or was it only used for ledger in these older versions? Asking because this dropdown is also used when importing a mnemonic directly, so we should to make sure if it's ledger-related we give that context here.

Shadowfiend:
@VladUXUI thoughts on the wrapping behavior here? Believe we can easily make it so that once it's selected it's on one line with an ellipsis and an OS tooltip on hover.

Shadowfiend (commented):
@VladUXUI thoughts on the wrapping behavior here? Believe we can easily make it so that once it's selected it's on one line with an ellipsis and an OS tooltip on hover.

ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
kimpers: I'm fairly certain that this is only for Ledger, and only really for accounts created several years ago :)

I would expect anyone needing this to know exactly what they need as you need to set it up in any other wallet you use

VladUXUI: I think we need to rename with something that gives a bit more context? " Ledger - Legacy", or "Ledger - Old" for a bit more international inclusiveness.
Also after the item is selected, the derivation path should not be visible, only the name.

kimpers: I agree that something like `Ledger - Legacy` might be more readable! I have addressed the requested changes, and here's an updated interaction vid 🍻 

https://user-images.githubusercontent.com/5450545/157107798-959e18e8-5446-4d77-ac81-c3bd5afc66f2.mov

FWIW I have been using Tally as my main wallet with the local build 🙌

VladUXUI:
Looks very nice on my end. Will let Antonio do the actual review

greg-nagy:
@fernya I would like you to pick this up, if you are down to it.

fernya:
> @fernya I would like you to pick this up, if you are down to it.

On board again, I'll do it ASAP.

ui/components/Shared/SharedSelect.tsx
```diff
@@ -98,7 +99,7 @@ export default function SharedSelect(props: Props): ReactElement {
           tabIndex={0}
         >
           <span>
-            {showValue && activeIndex
+            {showValue && activeIndex && !currentHideActiveValue
```
fernya, line 102: Nice forward-thinking!

fernya (approved):
Seems great, with the pre-approval of @VladUXUI I think its mergeable material.

PR #1103, opened 2022-03-06
:zany_face: Etherium: Support buying & selling ETH via 0x swaps

mhluongo:
No, that's not how you spell Ethereum :zany_face: 

This PR hacks in native ETH support for swaps on 0x. Thanks to @Shadowfiend's existing work, it was a pretty simple change. I think the tech debt is a reasonable tradeoff, and this will buy us time while contracts go live.

## Testing

- [ ] Make sure that you can buy an ERC-20 with ETH
- [ ] Make sure that you can sell an ERC-20 and receive ETH
- [ ] Double check that the rest of the swaps asset list is still correct


background/redux-slices/0x-swap.ts
```diff
@@ -132,10 +132,17 @@ function build0xUrlFromSwapRequest(
...
-  // assets are added (e.g., ETH), switch to `.symbol` for those.
-  const sellToken = assets.sellAsset.contractAddress
-  const buyToken = assets.buyAsset.contractAddress
+  // When available, use smart contract addresses. The only non-smart
+  // contract asset supported current is ETH. Once others are added, switch
+  // those to `.symbol`.
+  const sellToken =
+    "contractAddress" in assets.sellAsset
+      ? assets.sellAsset.contractAddress
+      : "ETH"
```
Shadowfiend, line 141: I would just make this `.symbol` and leave the asset filtering to the pros (ie the stuff feeding this 😁).

background/redux-slices/0x-swap.ts
```diff
@@ -233,20 +240,26 @@ export const fetchSwapPrice = createBackgroundAsyncThunk(
...
-
-    const existingAllowance: BigNumber =
-      await assetContract.callStatic.allowance(
-        await signer.getAddress(),
-        quote.allowanceTarget
+    let needsApproval = false
+    // If we aren't selling ETH, check whether we need an approval to swap
+    // TODO Handle other non-ETH base assets
+    if (
+      quote.sellTokenAddress !== "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
```
Shadowfiend, line 247: I immediately died. 

Can we extract this to a constant maybe? `ZEROX_ETH_SENTINEL_ADDRESS` or something.

ui/pages/Swap.tsx
```diff
@@ -220,9 +227,12 @@ export default function Swap(): ReactElement {
...
       return
     }
-    if (!isSmartContractFungibleAsset(sellAsset)) {
+    if (
+      !isSmartContractFungibleAsset(sellAsset) ||
+      sellAsset.symbol !== "ETH"
+    ) {
       logger.error(
-        "Attempting to approve transfer of a non-contract asset.",
+        "Attempting to approve transfer of a non-contract non-ETH asset.",
```
Shadowfiend, line 235: This should stay as is I think. Calling this callback for ETH should still be an error.

ui/pages/Swap.tsx
```diff
@@ -433,7 +443,7 @@ export default function Swap(): ReactElement {
           <SharedActivityHeader label="Swap Assets" activity="swap" />
           <div className="form">
             <div className="form_input">
-              <SharedAssetInput
+              <SharedAssetInput<SmartContractFungibleAsset | FungibleAsset>
```
Shadowfiend, line 446: I think because we use `isSmartContractFungibleAsset`, which is a type guard, we should be able to use bare `FungibleAsset`, since type guards coerce rather than narrow. Maybe we're relying on narrowing elsewhere? Otherwise I don't see a strong need for this. 

If it does happen, maybe we should introduce an `AnyFungibleAsset` type that is a central place to enable narrowing, similar to `AnyAsset` and some of the other `Any*` types.

Shadowfiend (commented):
Couple of thoughts. Off hand, none of them seem blocking.

background/redux-slices/0x-swap.ts
mhluongo: Lol sure. I've gotten so used to this constant I've forgotten we don't all have it memorized 😵

ui/pages/Swap.tsx
mhluongo: I tried without and it was a no-go. `AnyFungivleAsset` makes sense to me

background/redux-slices/0x-swap.ts
mhluongo: Fixed

background/redux-slices/0x-swap.ts
mhluongo: Fixed

ui/pages/Swap.tsx
mhluongo: Fixed

ui/pages/Swap.tsx
mhluongo: Saving this for the contract integration if you don't mind

Shadowfiend (approved):
Code lgtm, hoping to do a test pass this evening + merge if no one beats me to it.

background/redux-slices/0x-swap.ts
```diff
@@ -15,8 +15,8 @@ import { ERC20_ABI } from "../lib/erc20"
 import { COMMUNITY_MULTISIG_ADDRESS } from "../constants"
 
 interface SwapAssets {
-  sellAsset: SmartContractFungibleAsset
-  buyAsset: SmartContractFungibleAsset
+  sellAsset: SmartContractFungibleAsset | FungibleAsset
+  buyAsset: SmartContractFungibleAsset | FungibleAsset
```
cairomassimo, line 19: Non blocking (and I'm not 100% sure about it): I would rather use a discriminated union type here instead of a simple union, to replace the `"contractAddress" in x` checks with more legible `x.type === "smart-contract-asset"`.

ui/pages/Swap.tsx
```diff
@@ -75,13 +76,17 @@ export default function Swap(): ReactElement {
...
     accountBalances?.assetAmounts.filter(
-      (assetAmount): assetAmount is CompleteSmartContractFungibleAssetAmount =>
-        isSmartContractFungibleAsset(assetAmount.asset)
+      (
+        assetAmount
+      ): assetAmount is CompleteAssetAmount<
+        SmartContractFungibleAsset | FungibleAsset
+      > =>
+        isSmartContractFungibleAsset(assetAmount.asset) ||
+        assetAmount.asset.symbol === "ETH"
```
cairomassimo, line 89: Nitpick: `"ETH"` is a magic string (untyped, I suppose).

cairomassimo (approved):
Couple of minor observations. Can merge as it is!

greg-nagy (approved):
The code looks and ETH > ERC20 works, which is a big win. 

ERC20 > ETH did not work for me because ETH was not selectable as a buy token. 

Nonetheless I think this is an improvement as is, so let's 🚀

PR #1108, opened 2022-03-07
Allow signing messages (incl. sign-in-with-ethereum) with Ledger and other related improvements

cairomassimo:
This PR makes signing messages work with Ledger, by reusing shared code.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1046,23 +1046,54 @@ export default class Main extends BaseService<never> {
       }) => {
         this.store.dispatch(signDataRequest(payload))
 
-        const resolveAndClear = (signature: string) => {
-          this.keyringService.emitter.off("signedData", resolveAndClear)
+        const clear = () => {
+          if (HIDE_IMPORT_LEDGER) {
+            // eslint-disable-next-line @typescript-eslint/no-use-before-define
+            this.keyringService.emitter.off("signedData", resolveAndClear)
+          } else {
+            this.signingService.emitter.off(
+              "personalSigningResponse",
+              // eslint-disable-next-line @typescript-eslint/no-use-before-define
+              handleAndClear
+            )
+          }
           signingSliceEmitter.off(
             "signatureRejected",
             // eslint-disable-next-line @typescript-eslint/no-use-before-define
             rejectAndClear
           )
-          resolver(signature)
+        }
+
+        const handleAndClear = (response: SignatureResponse) => {
+          clear()
+          switch (response.type) {
+            case "success-data":
+              resolver(response.signedData)
+              break
+            default:
+              rejecter()
+              break
+          }
+        }
+
+        const resolveAndClear = (signedData: string) => {
+          clear()
+          resolver(signedData)
         }
 
         const rejectAndClear = () => {
-          this.keyringService.emitter.off("signedData", resolveAndClear)
-          signingSliceEmitter.off("signatureRejected", rejectAndClear)
+          clear()
           rejecter()
         }
 
-        this.keyringService.emitter.on("signedData", resolveAndClear)
+        if (HIDE_IMPORT_LEDGER) {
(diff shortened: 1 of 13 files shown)


background/main.ts
```diff
@@ -1051,23 +1046,54 @@ export default class Main extends BaseService<never> {
       }) => {
         this.store.dispatch(signDataRequest(payload))
 
-        const resolveAndClear = (signature: string) => {
-          this.keyringService.emitter.off("signedData", resolveAndClear)
+        const clear = () => {
+          if (HIDE_IMPORT_LEDGER) {
+            // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
0xDaedalus, line 1051: Idle thought - should we relax this rule?  I've always been of the opinion that its a bit too strict for modern js

background/services/ledger/index.ts
```diff
@@ -455,10 +455,32 @@ export default class LedgerService extends BaseService<Events> {
...
-    this.emitter.emit("signedData", joinSignature(signature))
-    return joinSignature(signature)
+    const signature = await eth.signPersonalMessage(
+      accountData.path,
+      Buffer.from(message).toString("hex")
+    )
+    this.emitter.emit(
+      "signedData",
+      joinSignature({
+        r: `0x${signature.r}`,
```
0xDaedalus, line 475: this object can probably be extracted into a variable since its reused on line 480

background/services/provider-bridge/authorization.ts
```diff
@@ -0,0 +1,43 @@
...
+  PermissionRequest,
+} from "@tallyho/provider-bridge-shared"
+import { sameEVMAddress } from "../../lib/utils"
+import { HexString } from "../../types"
+
+export function checkPermissionSignTypedData(
+  params: unknown[],
+  enablingPermission: PermissionRequest
+): void {
+  const walletAddress = params[0] as HexString
```
0xDaedalus, line 14: the `as`'s in this file make me a bit nervous even though I know they should be totally safe.  What do you think about making these functions into type guards instead of returning void and using `as`?  If we do that I think we can get some extra typesafety downstream if we ever need it.

ui/pages/PersonalSign.tsx
```diff
@@ -1,68 +1,62 @@
...
     typeof signerAccountTotal === "undefined"
   ) {
     return <></>
   }
 
   const handleConfirm = () => {
-    if (signingDataRequest !== undefined) {
-      dispatch(signData(signingDataRequest))
-    }
+    if (signingMethod === null) return
```
0xDaedalus, line 55: Nitpick but I feel obliged to mention that one of the most disastrous apple bugs ever happened because of [lack of curly braces after an if statement](https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/). Obviously Javascript is not C but the same class of error is possible (although not in this case - since you're returning).

0xDaedalus (commented):
A few ancillary comments - none are blocking.

ui/pages/PersonalSign.tsx
mhluongo: Not a nitpick at all IMO, we should avoid this syntax.

cairomassimo: My personal take is: no curly-braces iff on the same line. I find it especially useful to handle exceptional conditions early, and either `throw` or `return` on the same line. Not sure if this can be enforced by the linter though.

Let's configure `eslint` any way you prefer, and apply all the fixes in one shot?

background/main.ts
cairomassimo: Totally agree.

cairomassimo:
Thank @0xDaedalus for the comments! I've applied a couple of changes and re-tested it. Merging :)

Issue #1078, opened 2022-03-03
Ui/Ux Fixes


VladUXUI:
- [ ]  Swap Page: - Margin top seems to be 17px, let's change this to 8px. This way you don't have to scroll to see the full button.

- [ ]  Community banner in settings has a bit of empty space on the left side. 3px or so. Nitpicky but it looks off

_________

- [ ] Asset Card: - The hover effect doesn't change the Send icon colour, thus making it hard to see

- [ ] Send button should be clickable

<img width="366" alt="image" src="https://user-images.githubusercontent.com/80846886/156596380-07818eab-14c9-4f40-b9bf-65a9c9de0b18.png">

<img width="391" alt="image" src="https://user-images.githubusercontent.com/80846886/156596695-1c93e9d8-d4ce-4c43-b099-1b092d582161.png">

https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=7301%3A178747

beemeeupnow:
I'll create a new issue if you want, but I figured this goes with the theme here.

Now that the Ledger item is shown, the alignment needs work so that the 'Create new wallet' item isn't cut off at the bottom:
![add_screen_needs_adjustment](https://user-images.githubusercontent.com/95515141/156701940-dfcdb3b7-d817-42a3-afbd-8c1b206a9536.png)

You can scroll, but there's no visible scroll bar and unless more options will be added I think the current ones can fit.

VladUXUI:
Thanks for bringing it up @beemeeupnow . I didn't add this in because Henry is working on the new version for this page.

![image](https://user-images.githubusercontent.com/80846886/156712832-acaab992-01c2-48d2-bdcf-f9041e75a059.png)

beemeeupnow:
looks slick, thank you for the info
 
missing a 't' in that pic :smile:

PR #1110, opened 2022-03-07
💅✨  UI polish

henryboldi:
Covers a few issues mentioned by @VladUXUI in #1078:

Asset Card

- [x] "Send button should be clickable"
- [x] "The hover effect doesn't change the Send icon color, thus making it hard to see"

Swap Page

- [x] "Margin top seems to be 17px, let's change this to 8px. This way you don't have to scroll to see the full button."

And adds a notice that ETH swaps are not yet available.

warning1


ui/components/Wallet/WalletAssetListItem.tsx
```diff
@@ -58,7 +60,20 @@ export default function WalletAssetListItem(props: Props): ReactElement {
             </div>
           </div>
           <div className="right">
-            <span className="icon_send_asset" />
+            {!contractAddress && (
+              <Link
```
greg-nagy, line 64: This change gives the send icon 0px height and renders it invisible.

ui/components/Wallet/WalletAssetListItem.tsx
```diff
@@ -58,7 +60,20 @@ export default function WalletAssetListItem(props: Props): ReactElement {
...
+                to={{
+                  pathname: "/send",
+                  state: {
+                    symbol: assetAmount.asset.symbol,
+                    contractAddress,
+                  },
+                }}
+                className="icon_send_asset"
```
greg-nagy, line 72: I think this icon and the span are duplicates.

greg-nagy (changes requested):
This PR is a beauty :)

But the send button dissapers on my machine (osx, chrome) becuase of the 0px height of the icon. :(

ui/components/Wallet/WalletAssetListItem.tsx
henryboldi: Crap I think I missed some changes in my commit on this. Ty for catching it

ui/components/Wallet/WalletAssetListItem.tsx
henryboldi: Should be fixed now :)

greg-nagy (approved):
I think this PR is good improvmenet so let's 🚀

Our current implementation diverges from the design in the usage of send and swap icons but that's good for a follow up pr

design: 
CleanShot 2022-03-09 at 09 18 33

current implementation:
CleanShot 2022-03-09 at 09 14 00

PR #1112, opened 2022-03-07
Make non-end incremented Ledger derivation path imports possible

### LINKED ISSUE #1105: Update default Ledger derivation path
--> -m/44'/60'/0'/0,

fernya:
Fixes #1105 

Other wallets may increment different parts of the original path.
I chose character 'x' to signal the _only_ place where the root
path should be incremented. If 'x' is not present, we do the
original stuff - add another incrementing part at the end.

Screenshot 2022-03-08 at 10 03 31


ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
```diff
@@ -11,10 +11,17 @@ import SharedModal from "../Shared/SharedModal"
...
 
 // TODO make this network specific
-const initialDerivationPaths: { value: string; label: string }[] = [
+const initialDerivationPaths: {
+  value: string
+  label: string
+}[] = [
+  {
+    value: "m/44'/60'/x'/0/0",
+    label: "MM Live",
```
mhluongo, line 20: What is "MM Live"?

ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
```diff
@@ -11,10 +11,17 @@ import SharedModal from "../Shared/SharedModal"
...
+  },
   {
     value: "m/44'/60'/0'/0",
-    label: "Ethereum",
+    label: "BIP44",
```
mhluongo, line 24: No one will know that this means I'm afraid.

mhluongo (commented):
Quick thoughts!

ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
fernya: When I wrote that, I was unsure if that was just a foxy thing, or Ledger Live does that too.
RN i know that Ledger Live also does that!

ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
fernya: Reverting back, thanks!

cairomassimo (approved):
Looks ok!

PR #1113, opened 2022-03-07
Allow '_' as a variable name

0xDaedalus:
Using an underscore as a signifier for an unused variable is a common convention and our linter should allow it.  This is an extension of the current airbnb style-guide rule we are using found [here](https://github.com/iamturns/eslint-config-airbnb-typescript/blob/master/lib/shared.js#L39)

[See discussion here](https://github.com/tallycash/extension/pull/1060#discussion_r820857886).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,6 +41,10 @@ module.exports = {
       {
         selector: "variable",
         format: null,
+        filter: {
+          regex: "^_+$",
+          match: true,
+        },
         custom: {
           regex: "^_+$",
           match: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,6 +28,25 @@ module.exports = {
     // The rule is also mostly irrelevant to this codebase due to TypeScript
     // usage (where .tsx is required).
     "react/jsx-filename-extension": [0],
+    "@typescript-eslint/naming-convention": [
+      "error",
+      // Allow camelCase variables (23.2), PascalCase variables (23.8), and UPPER_CASE variables (23.10)
+      {
+        selector: "variable",
+        leadingUnderscore: "allow",
+        format: ["camelCase", "PascalCase", "UPPER_CASE"],
+      },
+      // Allow camelCase functions (23.2), and PascalCase functions (23.8)
+      {
+        selector: "function",
+        format: ["camelCase", "PascalCase"],
+      },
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -28,6 +28,25 @@ module.exports = {
...
+      {
+        selector: "function",
+        format: ["camelCase", "PascalCase"],
+      },
+      // Airbnb recommends PascalCase for classes (23.3), and although Airbnb does not make TypeScript recommendations, we are assuming this rule would similarly apply to anything "type like", including interfaces, type aliases, and enums
+      {
+        selector: "typeLike",
+        format: ["PascalCase"],
+      },
+    ],
```
Shadowfiend, line 49: So, I actually explicitly *don't* want to accidentally allow all leading underscores here. Here's what I suggest we do instead:

```suggestion
    "@typescript-eslint/naming-convention": [
      ...airbnbTypeScriptNamingConventionRules,
      // Allow underscore-only identifiers to indicate ignored positional variables.
      {
        selector: "variable",
        format: [],
        custom: {
          regex: /^_+$/,
          match: true
        }
      }
    ],
```

This requires an additional require up top:

```
const {
  rules: { "@typescript-eslint/naming-convention": airbnbTypeScriptNamingConventionRules },
} = require("eslint-config-airbnb-typescript/lib/shared/rules/best-practices")
```

I believe.

Shadowfiend (dismissed):
Code luks gud.

.eslintrc.js
```diff
@@ -28,6 +35,18 @@ module.exports = {
...
+      ...airbnbTypeScriptNamingConventionRules,
+      // Allow underscore-only identifiers to indicate ignored positional variables.
+      {
+        selector: "variable",
+        format: null,
+        custom: {
+          regex: "^_+$",
+          match: true,
+        },
+      },
```
Shadowfiend, line 48: Looks like this has to be:

```suggestion
      {
        selector: "variable",
        format: null,
        custom: {
          regex: "^_+$",
          match: true,
        },
      },      {
        selector: "variable",
        format: [],
        filter: {
          regex: "^_+$",
          match: true,
        },
        custom: {
          regex: "^(_+)$",
          match: true,
        },
      },
```

Shadowfiend (approved):
Ok, code looks good and tested locally to confirm it works 💪

PR #1114, opened 2022-03-07
Blue Grouse Migration 🐦 - migrate redux state to include keyringMetadata.

0xDaedalus:
Now that we use keyringMetadata to store extra information about keyrings, we need a migration to add it to the redux store of existing tally users.

### How to test:

-  uninstall tally wallet from chrome (or open incognito)
- `git fetch --all`
- `git checkout e400653e13ef9cb42279a9fb3f2cafbaa5b5f495` // version 0.12
- `yarn dev`
-  Add a new tally wallet & import a wallet (or two)
- `ctrl+c` the yarn dev process
- `git checkout migrate-keyringmetadata`
- `yarn dev`
-  click the update button in chrome://extensions

After these steps are completed - the app should be useable and the accounts list should properly display your previously created/imported wallets. 

Note: The imported wallets will display as `Tally` wallets since we did not keep track of the `source` of a wallet in version 0.12


background/main.ts
```diff
@@ -194,6 +194,13 @@ const REDUX_MIGRATIONS: { [version: number]: Migration } = {
       networks,
     }
   },
+  5: (prevState: Record<string, unknown>) => {
+    const { ...newState } = prevState
+
+    ;(newState.keyrings as Record<string, unknown>).keyringMetadata = {}
```
cairomassimo, line 200: This line is confusing. Perhaps we should type `prevState` as `any`, and silence the linter about it?

Ideally, we could keep the typing information for the previous versions of the state (as long as migrations are there).
Since we don't, at the moment, I would either

- use `any`, or
- rebuild the state from scratch, checking all the types in the process (this is likely unfeasible),

and avoid backflips with types.

cairomassimo (commented):
Minor code-style suggestion

background/main.ts
0xDaedalus: Good suggestion - I try to avoid using `any` whenever possible but I think in this case you're right that rebuilding state from scratch is pretty unfeasable.

greg-nagy: Always rebel against any :)

greg-nagy (approved):
0 typo a day keeps the doctor away 

Let's 🚀

Issue #593, opened 2021-12-12
Copy Address from top menu Name: 


VladUXUI:
Add the ability to copy the current account address:

![Design for account address copying from a tooltip on the account avatar/name.](https://user-images.githubusercontent.com/80846886/145706686-ff1d3d8a-b230-45e9-9969-5b937f13a091.png)

The way this should work:
- If you hover over the name or the avatar, a tooltip appears after 1-2 seconds with the address and a copy icon
- The user can move their cursor down into the tooltip to copy the address by clicking. Clicking anywhere into the tooltip copies the address.
- If the user clicks on the name or avatar, even when the tooltip is open, the the account switcher menu still opens

mr-michael:
Discussion @ https://www.flowdock.com/app/cardforcoin/tally/threads/6M3k8PXzvIYU4XgJIx6bPp87W3o

Shadowfiend:
From that discussion, the way this is expected to work:

- If you hover over the name or the avatar, a tooltip appears after 1-2 seconds with the address and a copy icon
- The user can move their cursor down into the tooltip to copy the address by clicking. Clicking anywhere into the tooltip copies the address.
- If the user clicks on the name or avatar, even when the tooltip is open, the the account switcher menu still opens

Naxsun:
Dropping this in here. Not sure if this is already part of the scope of this. The user suggested to also have a copy address option here.
https://discord.com/channels/808358975287722045/875438141610270740/951120858842275851

![image](https://user-images.githubusercontent.com/71384922/157640179-f6001024-c276-4362-a69a-499dcea32b6c.png)

jagodarybacka:
Main part implemented and merged, linked issues are improvements.

PR #1118, opened 2022-03-08
Create copy address tooltip

jagodarybacka:
Implements #593

### What

Allow users to copy their address in one click. Add tooltip to the user profile button on the top menu bar. User can hover this item and after delay tooltip will become visible. User can click on the tooltip and copy their wallet address. 

### Test
Hover avatar or shortened address in the top menu. After tooltip appears click anywhere on the tooltip. Success feedback message should appear. You should have your address copied and ready to be pasted anywhere else.


ui/components/TopMenu/TopMenuProfileButton.tsx
```diff
@@ -1,31 +1,118 @@
...
-import React, { ReactElement } from "react"
+import { setSnackbarMessage } from "@tallyho/tally-background/redux-slices/ui"
+import React, { ReactElement, useState, useRef } from "react"
+import { useDispatch } from "react-redux"
 import { useBackgroundSelector } from "../../hooks"
 import SharedCurrentAccountInformation from "../Shared/SharedCurrentAccountInformation"
 
+const TOOLTIP_DELAY = 1000
+
+function TopMenuProfileTooltip(props: {
```
jagodarybacka, line 10: Do you think Tooltip should be in a separate file? For now it only applies to the Profile Button so I'm not sure what is your preference

greg-nagy: Let's refactor when we need it. 

Probably  when doing the same for raw tx data

ui/components/TopMenu/TopMenuProfileButton.tsx
```diff
@@ -1,31 +1,118 @@
 import { selectCurrentAccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
-import React, { ReactElement } from "react"
+import { setSnackbarMessage } from "@tallyho/tally-background/redux-slices/ui"
+import React, { ReactElement, useState, useRef } from "react"
+import { useDispatch } from "react-redux"
 import { useBackgroundSelector } from "../../hooks"
 import SharedCurrentAccountInformation from "../Shared/SharedCurrentAccountInformation"
 
+const TOOLTIP_DELAY = 1000
```
greg-nagy, line 8: @VladUXUI To me the tooltip delay feels slow but I would like to have you make the call on this

CleanShot 2022-03-09 at 10 50 36

greg-nagy (commented):
The code looks good, let's wait for the design.

ui/components/TopMenu/TopMenuProfileButton.tsx
jagodarybacka: I've moved tooltip to its own file as stylesheet got bigger. If tx data will need the same tooltip then we can move it to shared components folder

ui/components/TopMenu/TopMenuProfileButton.tsx
VladUXUI: Agreed the delay seems a bit slow. If it's 1000 now, let's maybe go with 500

ui/components/TopMenu/TopMenuProfileButton.tsx
```diff
@@ -1,31 +1,78 @@
...
   const { onClick } = props
 
+  const showTooltip = () => {
+    timerRef.current = window.setTimeout(
+      () => setShouldDisplayTooltip(true),
+      TOOLTIP_DELAY
+    )
+  }
+  const hideTooltip = () => {
+    clearTimeout(timerRef.current ?? 0)
```
cairomassimo, line 30: Better be more explicit:
```suggestion
    if (typeof timerRef.current !== undefined) clearTimeout(timerRef.current)
    timerRef.current = undefined
```

Also, using `null` instead of `undefined` as default avoids the need of `typeof ...` (which I think is not necessary anyway - unless we want to replace all the occurrences of `undefined` with `void 0`)

ui/components/TopMenu/TopMenuProfileButton.tsx
```diff
@@ -1,31 +1,78 @@
...
+  }
+
+  const copyAddress = () => {
+    if (address) {
+      navigator.clipboard.writeText(address)
+      hideTooltip()
+      dispatch(setSnackbarMessage("Address copied to clipboard"))
+    }
+  }
+
```
cairomassimo, line 45: There is a tiny "memory leak" here (not really relevant, but good to know).
```suggestion

  useEffect(() => () => {
    hideTooltip() // cleanup
  })
```
to make sure the timeout is never triggered when the component is not mounted.

ui/components/TopMenu/TopMenuProfileTooltip.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <button type="button" className="tooltip" onClick={copyAddress}>
+      <span className="tooltip_label">Copy to clipboard</span>
+      <span className="tooltip_icon" />
+      <style jsx>{`
+        .tooltip {
+          display: flex;
+          align-items: center;
+          position: absolute;
+          z-index: 999999999;
```
cairomassimo, line 17: So aggresive :grin:

cairomassimo (commented):
Code is good overall!
Just a couple of minor comments

ui/components/TopMenu/TopMenuProfileTooltip.tsx
greg-nagy: just very very explicit :)

ui/components/Shared/SharedCurrentAccountInformation.tsx
```diff
@@ -16,13 +16,16 @@ export default function SharedCurrentAccountInformation({
 }: Props): ReactElement {
   return (
     <div className={classNames("account_info_wrap", { hover: showHoverStyle })}>
-      {name?.includes(".") ? name : shortenedAddress}
+      <span className="account_info_label">
+        {name?.includes(".") ? name : shortenedAddress}
```
greg-nagy, line 20: That's one way to make this decision ... :) 

Not a blocker for sure, but this behavior is a bit weird. 
I click on a copy for an eth address and I get the address on the clipboard. 
But makes sense. 

Could we make this a clearer @VladUXUI? 
CleanShot 2022-03-10 at 12 43 28

greg-nagy (approved):
Nothing blocker on the code and UX side, the feature works so let's 🚀

ui/components/Shared/SharedCurrentAccountInformation.tsx
Shadowfiend: I think this was why we explicitly said "copy address", fwiw.

PR #1119, opened 2022-03-08
Notify users when blind signing is disabled

### LINKED ISSUE #1106: Notify user to enable blind signing on devices before signing
Blind signing. This doesn't need to be activated from the beginning as it's only necessary when signing a transaction.
So let's have this in app. Screenshot bellow.
The i need help link should take users to https://support.ledger.com/hc/en-us/articles/4405481324433-Enable-blind-signing-in-the-Ethereum-ETH-app?docs=true

[Figma File](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=7374%3A178075)

image

_Originally posted by @VladUXUI in https://github.com/tallycash/extension/issues/1088#issuecomment-1060326302_

fernya:
Resolves  #1106

Several use-cases requires the unfortunate usage of blind signing
on the Ledger device. Notify the users before the can click on
Sign button on transaction signing pages.

Users need to click on "Refresh" button after they changed the setting
on their Ledgers.


cairomassimo:
@fernya now this PRs requires to activate blind signing even for sending ETH, which AFAIK should not need it.
If possible, let's fix this before merging!

fernya:
> @fernya now this PRs requires to activate blind signing even for sending ETH, which AFAIK should not need it. If possible, let's fix this before merging!

Good observation @cairomassimo , now I think its fixed along the idea of @Shadowfiend

ui/components/SignTransaction/SignTransactionContainer.tsx
```diff
@@ -114,6 +115,9 @@ export default function SignTransactionContainer({
         {signingLedgerState === "multiple-ledgers-connected" && (
           <SignTransactionMultipleLedgersConnected />
         )}
+        {signingLedgerState === "activate-blind-signing" && (
+          <SignTransactionLedgerActivateBlindSigning />
+        )}
```
cairomassimo, line 120: Regarding my other comment about `useLedgerSigningState`, here we could add a `isArbitraryDataSigningRequired: boolean` prop to this component. It should be set in `SignTransaction`, `SignData`, and `PersonalSign` according to the specific thing to be signed. (Be careful of conflicts with #1108.)

ui/components/SignTransaction/SignTransactionLedgerActivateBlindSigning.tsx
```diff
@@ -0,0 +1,27 @@
...
+      title="Activate blind signing"
+      helpMessage="Take the following steps:"
+      steps={[
+        <>Open Ethereum app on Ledger</>,
+        <>Navigate to Settings</>,
+        <>Enable blind signing</>,
+      ]}
+      onRefreshClick={() => {
+        dispatch(connectLedger())
+      }}
```
cairomassimo, line 19: Can we do this for every error? Now the refresh button is not bound to anything at all. If this refreshes the connection that's what we should do always.

ui/components/SignTransaction/SignTransactionSlideUpContentLayout.tsx
```diff
@@ -5,10 +5,14 @@ export default function SignTransactionSlideUpContentLayout({
   title,
   helpMessage,
   steps,
+  onRefreshClick,
+  onHelpClick,
 }: {
   title: ReactNode
   helpMessage: ReactNode
   steps: ReactNode[]
+  onRefreshClick?: () => void
```
cairomassimo, line 14: As per my other comment, we could actually remove this argument and always run `dispatch(connectLedger())` on refresh (if it is safe to do so, let's test it)

background/redux-slices/ledger.ts
```diff
@@ -91,19 +93,27 @@ const ledgerSlice = createSlice({
...
       ) {
         immerState.currentDeviceID = null
       }
-
       const device = immerState.devices[deviceID]
       if (!device) return
       device.status = status
+
+      if (typeof isBlindSigner === "undefined") return
+      device.isBlindSigner = isBlindSigner
```
cairomassimo, line 116: Not relevant anymore per previous comment, but

```suggestion
      device.isBlindSigner = isBlindSigner ?? false
```

background/redux-slices/ledger.ts
```diff
@@ -91,19 +93,27 @@ const ledgerSlice = createSlice({
...
       immerState,
       {
-        payload: { deviceID, status },
-      }: { payload: { deviceID: string; status: LedgerConnectionStatus } }
+        payload: { deviceID, status, isBlindSigner },
+      }: {
+        payload: {
+          deviceID: string
+          status: LedgerConnectionStatus
+          isBlindSigner?: boolean
```
cairomassimo, line 101: This shouldn't be optional. (The one in the state could be optional to avoid a migration, but here it's not needed AFAIK)

cairomassimo (commented):
Some suggestions for improvements nice to have before merging

background/main.ts
```diff
@@ -843,9 +843,13 @@ export default class Main extends BaseService<never> {
...
     this.store.dispatch(resetLedgerState())
 
-    this.ledgerService.emitter.on("connected", ({ id }) => {
+    this.ledgerService.emitter.on("connected", ({ id, metadata }) => {
       this.store.dispatch(
-        setDeviceConnectionStatus({ deviceID: id, status: "available" })
+        setDeviceConnectionStatus({
+          deviceID: id,
+          status: "available",
+          isBlindSigner: metadata.ethereumBlindSigner,
```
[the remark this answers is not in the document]
fernya: I think the `isBlindSigner` clicks the fastest for readers, I'll match the nomenclature.

ui/components/SignTransaction/useSigningLedgerState.tsx
```diff
@@ -22,9 +23,16 @@ export function useSigningLedgerState(
     if (connectedDevices.length === 0) return "no-ledger-connected"
     if (state.ledger.usbDeviceCount > 1) return "multiple-ledgers-connected"
 
+    const txHasData =
+      state.transactionConstruction.transactionRequest?.input !== null &&
+      state.transactionConstruction.transactionRequest?.input !== undefined &&
+      state.transactionConstruction.transactionRequest?.input.length > 0
```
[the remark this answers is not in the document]
fernya: One new TS syntactic sugar for each day keeps the doctor away!

ui/components/SignTransaction/SignTransactionLedgerActivateBlindSigning.tsx
fernya: Sure, I'll do this.

background/main.ts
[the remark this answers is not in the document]
cairomassimo: I'd prefer `isBlindSigningEnabled` over `isBlindSigner`. Longer, but clearer.

background/services/ledger/index.ts
```diff
@@ -183,7 +197,10 @@ export default class LedgerService extends BaseService<Events> {
           id: this.#currentLedgerId,
           type,
           accountIDs: [idDerviationPath],
-          metadata: { ethereumVersion: ethVersion },
+          metadata: {
+            ethereumVersion: appData.version,
+            ethereumBlindSigner: appData.arbitraryDataEnabled !== 0,
+          },
```
cairomassimo, line 203: DRY :)
Extract `metadata` into a variable and re-use it here

background/services/ledger/index.ts
```diff
@@ -166,13 +173,20 @@ export default class LedgerService extends BaseService<Events> {
         throw new Error("Can't derive meaningful identification address!")
       }
 
-      const ethVersion = (await eth.getAppConfiguration()).version
+      const appData = await eth.getAppConfiguration()
```
cairomassimo, line 176: Perhaps using destructuring here makes it clearer what we need? (Not necessary, just an impression)

cairomassimo (commented):
A couple of more suggestions

greg-nagy (approved):
Feature works, nothing blocker on the code side. 

Now only the tx signing is handled, a follow up PR will come with the rest of the signing types.

PR #1123, opened 2022-03-09
🔁 Add balance reload button - temporary solution

henryboldi:
reload_button


cairomassimo (commented):
I love this (besides the usual abuse of event-based design).
I think we need a second opinion before mergin though, given the impact on UX.

wojciech-turek:
I don't love this, but we need it.
I made a swap and refreshed and everything was updated nicely.

My idea for re-fetching balances is that we should trigger balances update after every wallet transaction + periodical check to check if any tx were made outside Tally wallet.

wojciech-turek:
Is @VladUXUI on board with this approach for now?

0xDaedalus:
I think the compromises here are worth the tech debt.  As @henryboldi said this is definitely a throwaway solution but I think the value of having a messy workaround for one of the biggest complaints in time for launch is high enough to warrant the debt.

background/main.ts
```diff
@@ -1165,6 +1165,10 @@ export default class Main extends BaseService<never> {
         )
       }
     )
+
+    uiSliceEmitter.on("refreshBackgroundPage", async () => {
```
mhluongo, line 1169: Doing this might kill other stuff, versus eg emitting an event to do an ETH and ERC-20 balance check :thinking:

mhluongo (commented):
Left a couple quick thoughts

ui/components/Wallet/WalletAccountBalanceControl.tsx
```diff
@@ -37,6 +39,75 @@ function ReadOnlyNotice(): ReactElement {
   )
 }
 
+function BalanceReloader(): ReactElement {
+  const dispatch = useDispatch()
+
+  const [isSpinning, setIsSpinning] = useState(false)
+
+  // 0 = never
+  const [timeWhenLastReloaded, setTimeWhenLastReloaded] = useLocalStorage(
```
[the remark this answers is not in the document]
henryboldi: Using component state won't persist the guard after reopening the popup. Sure, we can add it to Redux but that's more to rip out later

background/main.ts
henryboldi: I'm not sure what the root cause of our balance troubles are. Maybe there's something about the balance check that's broken? There's more to dig into here. But in the interest of time for a temp solution, I just wanted to emulate the solution that has worked for members in Discord (reloading the extension)

greg-nagy: I think this event is probably unnecessary because the async thunk runs in the background thread so the reload could live in the asyncThunk directly

ui/components/Wallet/WalletAccountBalanceControl.tsx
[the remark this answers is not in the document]
greg-nagy: I am torn tbh. This being a temporary solution convinced me but usually, these types of temp solutions stick with us the longest.

greg-nagy (approved):
Code looks good, it's supposed to be a short lived feature™️  it works and users wants it

@henryboldi can you please put this behind a feature flag in a follow up PR?

PR #1127, opened 2022-03-09
Fix styles of assets list

jagodarybacka:
Make divider smaller to match design.
Fix height of the list as "coming soon" notice bar is added.

Screenshot 2022-03-09 at 14 24 03


henryboldi (approved):
😍

PR #1128, opened 2022-03-09
create the same assets on release-* and main

greg-nagy:
Why?

At the moment the assets from main are used as release asset. 
But internal QA happens on the release branch and during the QA PRs can be merged into main.
This way the released asset is different from the QAd version which beats the purpose of the internal QA. 
(or we stop merging PRs to main during internal QA)

How? 

This way release branch generates the release build which can be used
for store uploads. This way the QAd code is the same as released.

Simplify asset upload into a single step on every branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,27 +22,20 @@ jobs:
         node-version: "${{ steps.nvm.outputs.NVMRC }}"
         cache: "yarn"
     - run: yarn install --frozen-lockfile
-    - name: Build artifacts for verification
-      if: github.ref != 'refs/heads/main'
+    - name: Dev build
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -22,27 +22,20 @@ jobs:
         node-version: "${{ steps.nvm.outputs.NVMRC }}"
         cache: "yarn"
     - run: yarn install --frozen-lockfile
-    - name: Build artifacts for verification
-      if: github.ref != 'refs/heads/main'
+    - name: Dev build
+      if: ${{ github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/heads/release-') }}
```
mhluongo, line 26: A little surprised this is necessary

mhluongo (approved):
Let's give it a shot!

PR #1131, opened 2022-03-09
AAAVEE AAVEE AAVE - Add prioritized asset sorting to SharedAssetInput

0xDaedalus:
This PR adds an easily customizable priority list so that when users use swaps they are not immediately presented with a bunch of AAVE tokens with no icons.  The list of assets I chose is pretty much random - I went through our lists and picked some assets that actually had icons.  Completely open to changing up the list if desired.

### To Test:
Open the swaps page - try selecting a buy asset - if you see UST and KEEP at the top of the list - great success!

https://user-images.githubusercontent.com/94649004/157496668-d7d5bfb8-c5be-44ca-9996-b5f80634d411.mov

### DIFF
--- a/ui/components/Shared/SharedAssetInput.tsx
+++ b/ui/components/Shared/SharedAssetInput.tsx
@@ -20,6 +20,32 @@ import SharedAssetItem, {
 } from "./SharedAssetItem"
 import SharedAssetIcon from "./SharedAssetIcon"
 
+// List of symbols we want to display first.  Lower array index === higher priority.
+// For now we just prioritize somewhat popular assets that we are able to load an icon for.
+const SYMBOL_PRIORITY_LIST = [
+  "UST",
+  "KEEP",
+  "ENS",
+  "CRV",
+  "FTM",
+  "GRT",
+  "BAL",
+  "MATIC",
+  "NU",
+  "AMP",
+  "BNT",
+  "COMP",
+  "UMA",
+  "WLTC",
+  "CVC",
+]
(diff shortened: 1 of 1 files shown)


ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -20,6 +20,32 @@ import SharedAssetItem, {
 } from "./SharedAssetItem"
 import SharedAssetIcon from "./SharedAssetIcon"
 
+// List of symbols we want to display first.  Lower array index === higher priority.
+// For now we just prioritize somewhat popular assets that we are able to load an icon for.
+const SYMBOL_PRIORITY_LIST = [
+  "UST",
```
henryboldi, line 26: Based on the top tokens on Uniswap,
```suggestion
  "DAI ",
  "USDC",
  "USDT",
  "WBTC",
  "UST",
```
How about we include these?

0xDaedalus: I am happy to and I looked at the top tokens on uniswap as well - but the only issue is that none of those tokens except `UST` have icons in tally (yet) so it looks pretty bad when opened with them at the top.

henryboldi (approved):
Tested and LGTM! 👏👑

ui/components/Shared/SharedAssetInput.tsx
henryboldi: Ah ah good point

PR #1133, opened 2022-03-10
Properly display infinite token approval

### LINKED ISSUE #726: Properly display infinite token approvals

jagodarybacka:
FIxes #726
### What is new
Mark with red color infinite spend limit. User can change spend limit and it will be white. 
Add tooltip to explain spend limit role.

### DIFF
--- a/ui/components/Shared/SharedTooltip.tsx
+++ b/ui/components/Shared/SharedTooltip.tsx
@@ -1,26 +1,15 @@
 import React, { ReactElement, useState } from "react"
 
 type VeriticalPosition = "top" | "bottom"
-type HorizontalPosition = "left" | "center" | "right"
 
 interface Props {
   verticalPosition?: VeriticalPosition
-  horizontalPosition?: HorizontalPosition
-  width?: number
+  width: number
   children: React.ReactNode
 }
 
-function getHorizontalPosition(horizontal: HorizontalPosition, width?: number) {
-  switch (horizontal) {
-    case "center":
-      return width ? `right: -${width / 2 + 8}px;` : ""
-    case "left":
-      return "right: 0;"
-    case "right":
-      return "left: 0;"
-    default:
-      return ""
-  }
+function getHorizontalPosition(width: number) {
+  return `right: -${width / 2 + 4}px;`
 }
 
 function getVerticalPosition(vertical: VeriticalPosition) {
@@ -35,12 +24,7 @@ function getVerticalPosition(vertical: VeriticalPosition) {
 }
 
 export default function SharedTooltip(props: Props): ReactElement {
-  const {
-    children,
-    verticalPosition = "bottom",
-    horizontalPosition = "right",
-    width,
-  } = props
+  const { children, verticalPosition = "bottom", width } = props
   const [isShowingTooltip, setIsShowingTooltip] = useState(false)
 
   return (
@@ -72,7 +56,7 @@ export default function SharedTooltip(props: Props): ReactElement {
             display: block;
           }
           .tooltip {
-            width: ${width ? `${width}px` : "auto"};
+            width: ${width}px;
             position: absolute;
             box-shadow: 0 2px 4px rgba(0, 20, 19, 0.24),
               0 6px 8px rgba(0, 20, 19, 0.14), 0 16px 16px rgba(0, 20, 19, 0.04);
@@ -84,7 +68,7 @@ export default function SharedTooltip(props: Props): ReactElement {
             border-radius: 3px;
             padding: 12px;
             ${getVerticalPosition(verticalPosition)}
-            ${getHorizontalPosition(horizontalPosition, width)}
+            ${getHorizontalPosition(width)}
           }
         `}
       </style>
--- a/ui/components/Shared/SharedTooltip.tsx
+++ b/ui/components/Shared/SharedTooltip.tsx
@@ -1,11 +1,46 @@
 import React, { ReactElement, useState } from "react"
 
+type VeriticalPosition = "top" | "bottom"
+type HorizontalPosition = "left" | "center" | "right"
+
 interface Props {
+  verticalPosition?: VeriticalPosition
+  horizontalPosition?: HorizontalPosition
+  width?: number
   children: React.ReactNode
 }
 
+function getHorizontalPosition(horizontal: HorizontalPosition, width?: number) {
+  switch (horizontal) {
+    case "center":
+      return width ? `right: -${width / 2 + 8}px;` : ""
+    case "left":
+      return "right: 0;"
+    case "right":
+      return "left: 0;"
+    default:
+      return ""
+  }
+}
+
+function getVerticalPosition(vertical: VeriticalPosition) {
+  switch (vertical) {
+    case "bottom":
+      return "top: 16px; margin-top: 5px;"
+    case "top":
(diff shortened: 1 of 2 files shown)


ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -162,7 +176,12 @@ export default function SignTransactionSpendAssetInfoProvider({
               </div>
             ) : (
               <>
-                <span className="spend_amount">
+                <span
+                  className={classNames(
+                    { spend_amount: true },
```
cairomassimo, line 181:
```suggestion
                    "spend_amount",
```

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -227,6 +256,9 @@ export default function SignTransactionSpendAssetInfoProvider({
                 font-size: 16px;
                 line-height: 24px;
               }
+              .spend_amount.has_error {
+                color: #d64045;
```
cairomassimo, line 260: We should have a shared color for this, I guess. If not, we should probably add one.

cairomassimo (commented):
Minor code style suggestions

cairomassimo (commented):
Position of tooltip does not match position/layout in the specs (https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=7667%3A186256) on my machine.
Screenshot from 2022-03-10 14-09-33

specs

Screenshot from 2022-03-10 14-11-38

@VladUXUI it's not clear from the specs how it should be positioned

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -162,7 +176,11 @@ export default function SignTransactionSpendAssetInfoProvider({
               </div>
             ) : (
               <>
-                <span className="spend_amount">
+                <span
+                  className={classNames("spend_amount", {
+                    has_error: approvalLimitString === null,
```
Shadowfiend, line 181: Not blocking, but I would mark this as "dangerous" rather than "has error", since an infinite approval isn't strictly an error, we just want to make sure the user is aware it's dangerous.

jagodarybacka:
Regarding tooltip position - right now we don't have any control over the tooltip position and either I can allow to pass the desired position to the tooltip component or we can think about something smarter. What do you think @cairomassimo ?

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -127,7 +129,19 @@ export default function SignTransactionSpendAssetInfoProvider({
             )}
           </span>
           <form onSubmit={(event) => event.preventDefault()}>
-            <span className="speed_limit_label">Spend limit</span>
+            <div className="spend_limit_header">
+              <span className="spend_limit_label">Spend limit</span>
+              <SharedTooltip>
```
greg-nagy, line 134: Please make the tooltip configurable through props and center the position of this tooltip

greg-nagy (changes requested):
Good PR, code looks good, agreed on some improvements, leaving this review mostly as a record of that convo.

ui/components/Shared/SharedTooltip.tsx
```diff
@@ -1,11 +1,46 @@
...
+  verticalPosition?: VeriticalPosition
+  horizontalPosition?: HorizontalPosition
+  width?: number
   children: React.ReactNode
 }
 
+function getHorizontalPosition(horizontal: HorizontalPosition, width?: number) {
+  switch (horizontal) {
+    case "center":
+      return width ? `right: -${width / 2 + 8}px;` : ""
```
greg-nagy, line 16: The positioning on the center is not perfect but kinda works 
CleanShot 2022-03-11 at 14 50 00



Left and right are not good

CleanShot 2022-03-11 at 14 51 01


CleanShot 2022-03-11 at 14 49 20

greg-nagy (changes requested):
The horizontal positioning is not yet perfect, but I would the center solution is already shipable :)

ui/components/Shared/SharedTooltip.tsx
jagodarybacka: How would you like right and left to work then?

greg-nagy: I don't think we need them now, so I am ok with simply just removing those and keeping the center as default. 

If you want to fix them properly then I believe the normal behavior would be to close the right side of the pop-up to the right side of the window while showing the whole text so the user can read it.

jagodarybacka: But you can get properly positioned right and left - this tooltip has defined width as this is the simplest way to get center position. If you'll set smaller width in props then it will be fine. It's up to developer and design of given tooltip to decide about width and position I guess.

greg-nagy: I am ok keeping only the center position as a default. 

But if I am presented with an option to position left and right I would not expect the component to move my content out of the visible area. I understand that it can mean that in the end the component is not positioned left or right.

This is "position correctly, not a problem if the content is not visible" vs "make the content visible, and position according to the instruction if there is enough space".

jagodarybacka: ok, so lets keep center position as an only option for now

Shadowfiend: Quick thoughts:
- some sort of clamping is reasonable. 
- any version of this will require judicious application by the dev, I expect, because we can’t predict what constraints a tooltip will exist in. 
- with that in mind and as a future reference, I think it’s fine to land a PR that brings tools to the table and then improve it in follow up work, given the current usage doesn’t hit it.

Shadowfiend: (The above comment shouldn’t change the current path here, I’m really focused on future approached here.)

jagodarybacka: @Shadowfiend I agree, as for now this is the only tooltip we have implemented. As I'm scrolling through figma there will be more tooltips in near future. IMO ideal solution would be probably to left the decision to the tooltip component as where it should be rendered but I think it is strongly out of scope of this task.

greg-nagy (approved):

ui/components/Shared/SharedTooltip.tsx
greg-nagy: > with that in mind and as a future reference, I think it’s fine to land a PR that brings tools to the table and then improve it in follow up work, given the current usage doesn’t hit it.

I will think about this. I usually have a different approach but this made me think and wonder if I should reevaluate my approach.

PR #1136, opened 2022-03-10
Rename copy address tooltip

### LINKED ISSUE #1135: Change copy from Copy to clipboard to Copy address
Greg and Antonio both have a good point that copy to clipboard is not clear enough what you are copying if it's an ENS name and not an Address.

> That's one way to make this decision ... :) 
> 
> Not a blocker for sure, but this behavior is a bit weird. 
> I click on a copy for an eth address and I get the address on the clipboard. 
> But makes sense. 
> 
> Could we make this a clearer @VladUXUI? 
> CleanShot 2022-03-10 at 12 43 28

_Originally posted by @greg-nagy in https://github.com/tallycash/extension/pull/1118#discussion_r823628299_

jagodarybacka:
Resolves #1135

Rename from _Copy to clipboard_ to _Copy address_ to improve UX as user should be informed what will be copied.


mhluongo (approved):
I just ran into this myself :)

PR #1139, opened 2022-03-10
Good BN: Use ethers for a few cases instead of other libs

Shadowfiend:
Other libs tend to pull in many copies of bn.js through broken
dependency trees. Ethers... Does not.


greg-nagy (approved):
tested, ledger import and signing works fine

PR #1142, opened 2022-03-10
Decoding EIP2612 Permit Messages - 🦄  Uniswap and beyond...

### LINKED ISSUE #1100: Decode sign-permit TX
Currently the page looks like this. 
We have some information here that won't make sense for most users. Like:
- Value (Format should be 20 USDC)
- Deadline (12 Mar 2022)
- Addresses should be shortened with a link to enterscan, same pattern as we use everywere [0x3843...32j8d (new tab Icon)]
- We also need to show the DAPP that is requesting the signature

image

0xDaedalus:
This PR Adds a formatted screen for signing uniswap permit transactions.

Closes #1100

### To test:
- Navigate to https://app.uniswap.org/#/swap?chain=mainnet
- [ ] Attempt to permit an asset you have not previously granted permission for.
- [ ] Attempt to permit an asset you have previously granted permission for.
- [ ] Attempt to permit an asset you have not previously granted permission for and is not in our current token list.
- [ ] Attempt to permit an asset and sign message with Ledger (I have not been able to test this yet)

You should see a beautifully formatted message pop up!

Signing typed data for Dapps other than Uniswap should continue to work as usual - with the added benefit that addresses will attempt to display their ens names and have links to Etherscan.

Screen Shot 2022-03-10 at 3 16 27 PM


background/services/signing/types.ts
```diff
@@ -0,0 +1,11 @@
+import { EIP712TypedData } from "../../types"
+
+export type SignTypedDataRequest = {
```
0xDaedalus, line 3: I moved these two types in here because the signing-service, internal-ethereum-provider, and now enrichment-service depend on them and it felt weird to me that we had so many backend services depending on types in a redux file (not to mention the circular dependencies that were caused with enrichment-service).  Happy to change though if we think there is a better place for them.

background/services/signing/types.ts
```diff
@@ -0,0 +1,10 @@
...
+  account: string
+  typedData: EIP712TypedData
+}
+
+export type SigningMethod =
+  | { type: "keyring" }
+  | { type: "ledger"; deviceID: string; path: string }
```
Shadowfiend, line 10: This should 100% be in here, since dependencies should flow redux -> service, not vice versa. To be enforced once we split redux slices to their own package <_<

background/redux-slices/signing/types.ts
```diff
@@ -0,0 +1,65 @@
+import { EnrichedSignTypedDataRequest } from "../../services/enrichment"
```
Shadowfiend, line 1: Small note: we generally should be avoiding `types` files. Types are part and parcel of the code they're used in, not a separate addendum. There are a few in the codebase still from places where we couldn't quite do that or where we were lazy (e.g. https://github.com/tallycash/extension/commit/5913c99d548a6728c22ca4204a7bf191af5c2b66).

Splitting code apart may still be a good idea if a file is getting unwieldy or if there are circular interdependency issues, but along semantic lines rather than types-vs-not lines.

My suggestion here is that we pull EIP* types and out alongside the `SignTypedDataRequest` type that were moved to `services/signing/types.ts` to a `background/utils/signing.ts` file, and bring along various non-redux utilities like `parseSigningData` and `checkEIP4361`. We may even want that file to be more specific, as `typed-data-signing.ts` or so.

0xDaedalus:
@mhluongo shared[ a neat link](https://eips.ethereum.org/EIPS/eip-2612) that I was not aware of.  I didn't know that uniswap's permission request actually adheres to an EIP.  I think this means that all we can potentially support a lot more dapps than uniswap in this PR if we can get a list (crowdsource?) of `dappName: contractAddress` that use EIP-2612 to request permission.

mhluongo:
> In light of this I think it makes sense to rename this PR to a more general `Decode EIP-2612 Transactions`

Strong agree. Note that I still think it's very, very important to...
* Special case known dApps to bootstrap trust. Eg showing the Uniswap logo here if we've confirmed this comes from their specific dApp is :ok_hand: 
* Make sure people can tell who the `spender` address is. It's by far the most security-critical piece of this screen (followed by token contract, amount, and perhaps deadline)

background/services/signing/types.ts
greg-nagy: > To be enforced once we split redux slices to their own package <_<
🙏

greg-nagy (approved):
let's 🚀

I think a follow-up PR is warranted but this is an awesome feature and it works!

- Navigate to https://app.uniswap.org/#/swap?chain=mainnet
- [x] Attempt to permit an asset you have not previously granted permission for.
- [x] Attempt to permit an asset you have previously granted permission for.
- [ ] Attempt to permit an asset you have not previously granted permission for and is not in our current token list.
  - > how do I do this? If it's not in my token list than its value is 0 and it will give insufficient balance or something
- [x] Attempt to permit an asset and sign message with Ledger (I have not been able to test this yet)

0xDaedalus:
> Attempt to permit an asset you have not previously granted permission for and is not in our current token list.
>> how do I do this? If it's not in my token list than its value is 0 and it will give insufficient balance or something.

Ah - by `not in our token list` I mean not in our current list of ~300 tokens.  For example this is what I see when I preview verygoodverynice.eth (a random address I found with APE tokens).

image

Maybe this test isn't needed - I just like to be extra-dilligent.

background/services/enrichment/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+    // Uniswap v2 Router
+    "0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45": "🦄 Uniswap",
+  }
+
+export function isEIP2612TypedData(
+  typedData: EIP712TypedData
+): typedData is EIP2612TypedData {
+  if (typeof typedData.message.spender === "string") {
+    if (
+      // Must be on main chain
```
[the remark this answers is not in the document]
0xDaedalus: You're right it does not have to be

background/services/enrichment/types.ts
```diff
@@ -76,3 +77,34 @@ export type EnrichedEVMTransactionSignatureRequest =
...
+
+export type TypedDataField = {
+  value: string
+  type: "address" | "string"
+}
+
+export type EIP2612SignTypedDataAnnotation = {
+  type: "EIP-2612"
+  source: string
+  displayFields: {
```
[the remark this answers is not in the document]
0xDaedalus: I do think it makes sense to align on the actual EIP spec (deadline & value).  I think the design in figma was perhaps based on the `DAI` implementation of EIP-2612 (See the `Backwards Compatibility` section in your above link).

Issue #1084, opened 2022-03-03
Sending ETH w/o unlock clunkily clears send page


henryboldi:
To reproduce:
1. Enter the send amount
2. Enter the destination
3. Click send
4. Directs the user to the unlock wallet page (enter password)
5. Redirects you back to the send page with the form blank

The expected behavior should be, after entering their password, direct the user to the sign transaction page for sending the ETH — aka picking up where they left off.

Using `v0.12.0-55-gbf05d7c0` on Chrome.

cairomassimo:
I managed to reproduce it, but only if the sequence of steps is the very first thing you do once you start the extension (i.e., after a reload, no need to re-install).

I think it's related to a flash of error that I've been seeing for a while now (in `main`). I'm investigating further.

cairomassimo:
The problem is here:

https://github.com/tallycash/extension/blob/ca919f6f241d7034289611959f29642618d6d1e2/ui/pages/SignTransaction.tsx#L49

When we update the location via `history.push`, the Redux state is not yet updated and the `SignTransaction` page redirects back to "safety". Should be fixed with proper transaction queueing.

PR #1154, opened 2022-03-14
Fix glitches with page transition animations

cairomassimo:
WIP.

It was an attempt for #1084 but so far it does not solve it. Not sure if we can ever merge this.


mhluongo:
Closing as stale!

PR #1159, opened 2022-03-15
Create shared icon component

jagodarybacka:


ui/components/Shared/SharedIcon.tsx
```diff
@@ -0,0 +1,63 @@
+import classNames from "classnames"
+import React, { ReactElement } from "react"
+
+type Props = {
+  type: "mask" | "background"
+  icon: string
```
0xDaedalus, line 6: Did we come to a consensus on `.png` vs `.svg`?  I think it would be such a big win here for community contributors if could type this as a union type of all of our valid icons.  Probably out of the scope of this PR but something to keep in mind.

0xDaedalus (dismissed):
This looks awesome!  The only blocking issue is that I do think we should decide on `mask-image` vs `background-image` and stick to it.

ui/components/Shared/SharedIcon.tsx
```diff
@@ -0,0 +1,63 @@
+import classNames from "classnames"
+import React, { ReactElement } from "react"
+
+type Props = {
+  type: "mask" | "background"
```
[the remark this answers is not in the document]
jagodarybacka: What about multi color icons? I assume they are right now using background-image. Maybe they deserve another component like `SharedImage` or something similar?

ui/components/Shared/SharedIcon.tsx
jagodarybacka: As I remember we wanted to move entirely to svgs but as this will require some work I wanted to cover both svg and png for now.

ui/components/Shared/SharedIcon.tsx
```diff
@@ -0,0 +1,63 @@
...
+  icon: string
+  width: number
+  height?: number
+  color?: string
+  hoverColor?: string
```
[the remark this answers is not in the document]
VladUXUI: That is correct. Some icons have mouse over. I have them as buttons in my design

ui/components/Shared/SharedIcon.tsx
VladUXUI: I think we decided to go with SVG. But the decision that wasn't yet made is if we are doing a sprite or no

ui/components/Shared/SharedIcon.tsx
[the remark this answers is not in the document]
VladUXUI: If it help our icon design system does not include multiple coloured icons. Nor do i foresee that in the near feature

ui/components/Shared/SharedIcon.tsx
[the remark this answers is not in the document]
jagodarybacka: @VladUXUI I believe mouse over color is covered here as hover color. Do we have icons that are changing color for good after clicking on them? Or changing color during mouse press on them?

ui/components/Shared/SharedIcon.tsx
jagodarybacka: Ok so do you guys think that we should leave only "mask" type of icons + allow only svg icons in the SharedIcon component?

0xDaedalus: I think yes to restricting this component to only `mask` type icons and maybe create a separate issue to:
a) convert all icons in the public folder from `.svg` => `.png`
b) restrict this component to `.svg` components.

Shadowfiend: I think it’s easier than this: make a list of all non-SVG icons inside the component and transparently (to the caller) adjust how we render them, maybe by using a separate internal component. We already basically have the code in this PR, we just ask the caller to make the decision right now. 

For anything else, assume SVG.

jagodarybacka: @Shadowfiend But it's not like `.png` needs `background` type and `.svg` needs `mask` type, we have a lot of `.png` that are used with `mask`. The only scenario I can think of where we need `background` is where icon needs multiple colors. But Vlad wrote a comment above that no icon will have multiple colors.

So I'm leaning into this solution: leave this component with `mask` type only, for both png and svg as we don't have all icons exported as svgs yet. What do you think?

greg-nagy (approved):
I think this is an improvement as is. 

There was some really good discussion in the conversation though, so @jagodarybacka please capture those ideas in github issues. 

And let's 🚢  this bad boy :)

Shadowfiend:
Yeah it should not render as a button unless it has a click handler defined tbh. Even then I'm not convinced it makes sense for the icon component to be in charge of click handling/button-ing; we could straightforwardly make an icon button component that uses the icon component. Cc @jagodarybacka

jagodarybacka:
@Shadowfiend I think we should then figure out what we want to do with SharedButton component - as SharedButton is pretty hard to customize it would be nice to have some button component that would take care of less things than `SharedButton`

Issue #1158, opened 2022-03-15
Wallet Balances Not Updating


0xDaedalus:
ERC-20 Token and ETH balances are not correctly updating in the application.  It is likely that #1082, #1067, and #1109 are all related to this issue.  This issue is here to coordinate bug-hunting efforts to help us get to the root cause quicker.

# Related issues
- [x] #1067 
- [ ] #1082

0xDaedalus:
This morning @Shadowfiend proposed a solution where we update the [`periodInMinutes`](https://github.com/tallycash/extension/blob/main/background/services/indexing/index.ts#L85) of account-wide token refresh from `30` to `1` minute - but unfortunately this does not fully address the issue - however I did learn some things while recreating / debugging.

1.  This issue seems to affect both ERC-20 transactions and regular ol' ETH.  The wallet did not pick up ETH sent to my wallet for over 4 hours. (The ETH appeared in the wallet after hitting the Refresh button).

2.  When performing a swap from ETH => USDC, I saw my ETH balance decrease , but my USDC balance stayed the same.  (The USDC did not appear in the wallet even after hitting the Refresh button)

3.  Re-installing the wallet resulted in my USDC balance disappearing completely (although I clearly see it on etherscan). The last calls I see to alchemy over the websocket connection are (in order): `eth_subscribe` `eth_getBalance` `eth_call` and `eth_blockNumber`.  The extension finishes "Diggin Deeper" and all I'm left in my wallet balances is ETH.

This suggests to me that:
a) decreasing the interval at which we call `handleTokenRefresh` will not do much (it seems we're having difficulty even creating a correct list of assets to track in the first place).
b) The "Refresh Button" does not seem to be as effective a fix as we had hoped.

VladUXUI:
Is there any chance that you have "hide balance under $2" activated and your balance was bellow that?

kelvindevv:
Talk to a Polygon Dev on the [Live support page](https://forummaticnetwork.org/) so that your issue can be looked into.

0xDaedalus:
No, these were all balances above $15

0xDaedalus:
I think I’ve found a reliable way to recreate the issue of swap transactions not going through:

- install fresh tally wallet
- import a wallet capable of swapping
- open background page console.
- navigate to the network tab
- turn off your laptop’s wifi
[...]

PR #1162, opened 2022-03-15
So Refreshing 🍋 - Update token balances every minute.

0xDaedalus:
This PR fixes a regression introduced post 0.12.1  that prevented the wallet from getting _any_ token balances - as well as reduces the interval at which we check for token balances down to 1 minute (from 30).  It partially addresses the symptoms caused by #1158 but probably not the root cause.

### To Test
• complete a swap for (or send yourself) a token you do not already have in your wallet
• wait a maximum of 1 minute
• the token and balance should show up in your asset list

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -266,7 +266,7 @@ export default class ChainService extends BaseService<Events> {
    * provider exists.
    */
   providerForNetwork(network: EVMNetwork): SerialFallbackProvider | undefined {
-    return this.providers[network.name]
+    return this.providers[network.name.toLowerCase()]
   }
 
   /**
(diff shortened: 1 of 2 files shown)


background/services/chain/index.ts
```diff
@@ -266,7 +266,7 @@ export default class ChainService extends BaseService<Events> {
    * provider exists.
    */
   providerForNetwork(network: EVMNetwork): SerialFallbackProvider | undefined {
-    return this.providers[network.name]
+    return this.providers[network.name.toLowerCase()]
```
0xDaedalus, line 269: This is somewhat of a dirty fix but I am not sure what effects normalizing all EVMNetwork name's to lowercase might have so chose this implementation for now.

henryboldi (approved):
Tested and it definitely fixes this critical issue,
> PR fixes a regression introduced post 0.12.1 that prevented the wallet from getting any token balances

That said, I haven't had the chance to see if this fixes the balance updates on a stale account. Decent chance it does?

That said, let's get this out ASAP 👏. Nicely done sir

PR #1163, opened 2022-03-15
Fix sign transaction navigating away the first time it is opened

### LINKED ISSUE #1084: Sending ETH w/o unlock clunkily clears send page

cairomassimo:
Fixes #1084.

`webext-redux` is (necessarily) async, which means that the UI can observe inconsistent state if it is updated (syncrounously) before a Redux dispatch is fulfilled. This was the case when requesting a transaction signature via Redux and at the same time navigating to `/sign-transaction`.

This PR defines a new hook `useIsBackgroundSettled` which returns `true` once any pending actions are dispatched to Redux and the state is up-to-date.

This is definitely a "hacky" fix. A better solution would be to reduce the use of non-background state (e.g. in `window.location`), and replace it with Redux-backed state (which is maintained consistent thanks to the Redux state-action pattern).


ui/hooks.ts
```diff
@@ -175,3 +176,18 @@ export function useLocalStorage(
...
+
+/**
+ * Returns true once all pending redux updates scheduled before the first render
+ * (if any) have been applied, and false otherwise.
+ */
+export function useIsBackgroundSettled(): boolean {
+  const [settled, setSettled] = useState(false)
+  const dispatch = useBackgroundDispatch()
+  useEffect(() => {
+    Promise.resolve(dispatch(noopAction())).then(() => {
```
Shadowfiend, line 188: `Promise.resolve` here is just a workaround for the typing, right?

Shadowfiend (approved):
I hate it, let's do it 😅 Did some quick testing to make sure there weren't any surprising regressions—didn't spot anything.

PR #1166, opened 2022-03-15
❌ Remove all image assets with "Tally" brand

henryboldi:
Per @karlmoll's QA,
- [x] Removes all image assets with Tally.
- [x] Updates set password screen width to show full updated logo.


0xDaedalus (changes requested):
I still see `Tally` on the set password screen 

image

0xDaedalus (approved):

PR #1173, opened 2022-03-16
💔 It's not you: remove git dependency from the build process

greg-nagy:
https://www.youtube.com/watch?v=QVYgoejO3ac

Depending on the git to check version number during
build time proved to be a headache, so we have returned
to using the version from `package.json` file.

The idea is good though, probably should be placed into
the CI pipe as a step so the build itself does not depend
on the git tag.

(It caused issues with the firefox release and also internally)

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import path from "path"
 import webpack, {
   Configuration,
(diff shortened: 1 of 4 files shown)


webpack.config.ts
```diff
@@ -112,16 +92,8 @@ const baseConfig: Configuration = {
...
       // FIXME webpack version.
     }) as unknown as WebpackPluginInstance,
-    gitRevisionPlugin,
     new DefinePlugin({
-      "process.env.GIT_COMMIT_HASH": JSON.stringify(
-        gitRevisionPlugin.commithash()
-      ),
-      "process.env.GIT_COMMIT_DATE": JSON.stringify(
-        gitRevisionPlugin.lastcommitdatetime()
-      ),
```
Shadowfiend, line 122: Non-blocking: I would love it if we could still try to pull the commit hash + date at the end per the last suggestion @ https://github.com/tallycash/extension/pull/992#issuecomment-1039643629 . These will be very useful for QA builds, and we can have the combined string stay blank if the `git` command is unavailable or fails for whatever reason.

cairomassimo:
I'm merging this and creating an issue to address https://github.com/tallycash/extension/pull/992#issuecomment-1039643629 in the future.

PR #1174, opened 2022-03-16
Make proper trigger for blind signing notification + eyecandy

fernya:
This PR intends to correctly implement the blind signing notification for all types of handled signing flows (message/typed data/transaction).

### Test instructions

1. **Turn off blind signing in Ethereum app**
1. **Try one use-case below**
1. **Follow instructions if shown**
1. **Turn off blind signing again in Ethereum app**

#### Test cases

- No slide-up expected when sending ETH in a transaction to another address
- No slide-up expected when permitting the contract to swap USDC coins on app.uniswap.org
- Slide-up expected when a USDC coin swap is finalized on app.uniswap.org
- Slide-up expected when a login attempt is signed on login.xyz


ui/components/SignTransaction/useSigningLedgerState.tsx
```diff
@@ -5,9 +5,8 @@ export type SigningLedgerState =
   | "no-ledger-connected"
   | "wrong-ledger-connected"
   | "busy"
-  | "available"
   | "multiple-ledgers-connected"
-  | "activate-blind-signing"
+  | { arbitraryDataEnabled: boolean }
```
Shadowfiend, line 9: It seems like the intent here is to let the caller know the ledger is available but has arbitrary data enabled or disabled, and then let the caller decide whether it needs to further prompt them to enable arbitrary data.

I would suggest we keep the core state structure functioning as a string enum, but add an additional optional field; something like:

```typescript
export type SigningLedgerState =
  { state: "no-ledger-connected" | "wrong-ledger-connected" | ... } |
  { state: "available", arbitraryDataEnabled: boolean }
```

In the calling code, you can leverage the type system to handle this much more transparently than the current JSON stringification:

```typescript
    const isLedgerAvailable = signingLedgerState?.status === "available"

    const mustEnableArbitraryDataSigning =
      isLedgerAvailable &&
      isArbitraryDataSigningRequired &&
      ! signingLedgerState.arbitraryDataEnabled

    const canLedgerSign =
      isLedgerAvailable && !mustEnableArbitraryDataSigning
```

Separately, I would encourage consistency in calling this “arbitrary data enabled” or “arbitrary data signing enabled” throughout the codebase, rather than switching between them—this reduces the cognitive load (and grep complexity haha) of tracing what is happening where, and where data is flowing through.

fernya: Thanks @Shadowfiend, when I typed in the stringification I felt dirty. 😅

cairomassimo (commented):
I made essentially the same comment as @Shadowfiend before finding it down below :man_shrugging:

greg-nagy (approved):
whoopsie daisy let's merge this baby 👶

PR #1175, opened 2022-03-16
Balance Update Problems

0xDaedalus:
This draft pr is here to so we can easily share code related to exploring the problem of balances not updating consistently.  Relevant to #1158

### DIFF
--- a/background/services/chain/serial-fallback-provider.ts
+++ b/background/services/chain/serial-fallback-provider.ts
@@ -43,8 +43,8 @@ function waitAnd<T, E extends Promise<T>>(
  * ms to back off before making the next attempt.
  */
 function backedOffMs(backoffCount: number): number {
-  const backoffSlotStart = BASE_BACKOFF_MS ** backoffCount
-  const backoffSlotEnd = BASE_BACKOFF_MS ** (backoffCount + 1)
+  const backoffSlotStart = BASE_BACKOFF_MS ** Math.sqrt(backoffCount)
+  const backoffSlotEnd = BASE_BACKOFF_MS ** Math.sqrt(backoffCount + 1)
 
   return backoffSlotStart + Math.random() * (backoffSlotEnd - backoffSlotStart)
 }
@@ -54,9 +54,7 @@ function backedOffMs(backoffCount: number): number {
  * either closing or already closed. Ethers does not provide direct access to
  * this information, nor does it attempt to reconnect in these cases.
  */
-function isClosedOrClosingWebSocketProvider(
-  provider: JsonRpcProvider
-): boolean {
+function isNotConnectedWebSocketProvider(provider: JsonRpcProvider): boolean {
   if (provider instanceof WebSocketProvider) {
     // Digging into the innards of Ethers here because there's no
     // other way to get access to the WebSocket connection situation.
@@ -65,7 +63,8 @@ function isClosedOrClosingWebSocketProvider(
 
     return (
       webSocket.readyState === WebSocket.CLOSING ||
-      webSocket.readyState === WebSocket.CLOSED
+      webSocket.readyState === WebSocket.CLOSED ||
+      webSocket.readyState === WebSocket.CONNECTING
     )
   }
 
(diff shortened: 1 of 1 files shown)


background/services/chain/serial-fallback-provider.ts
```diff
@@ -43,8 +44,8 @@ function waitAnd<T, E extends Promise<T>>(
  * ms to back off before making the next attempt.
  */
 function backedOffMs(backoffCount: number): number {
-  const backoffSlotStart = BASE_BACKOFF_MS ** backoffCount
-  const backoffSlotEnd = BASE_BACKOFF_MS ** (backoffCount + 1)
+  const backoffSlotStart = BASE_BACKOFF_MS ** Math.sqrt(backoffCount)
+  const backoffSlotEnd = BASE_BACKOFF_MS ** Math.sqrt(backoffCount + 1)
```
cairomassimo, line 48: This formula does not make dimensional sense (we are raising seconds to a fractional power - I hope we don't have any physicists in the team :)

I think we are looking for this instead (standard exponential backoff):

```suggestion
  const backoffSlotStart = BASE_BACKOFF_MS * 2 ** backoffCount
  const backoffSlotEnd = BASE_BACKOFF_MS * 2 ** (backoffCount + 1)
```

If not, please add a comment to the code explaining why.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -54,9 +55,7 @@ function backedOffMs(backoffCount: number): number {
  * either closing or already closed. Ethers does not provide direct access to
  * this information, nor does it attempt to reconnect in these cases.
  */
-function isClosedOrClosingWebSocketProvider(
-  provider: JsonRpcProvider
-): boolean {
+function isNotConnectedWebSocketProvider(provider: JsonRpcProvider): boolean {
```
cairomassimo, line 58: Avoiding negated logic here would make it clearer.

This method should be `isWebSocketProviderConnected` instead (probably returning `webSocket.readyState === WebSocket.CONNECTED` or similar), and `if (!isWebSocketProviderConnected(...))` where appropriate.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -199,8 +198,8 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
             params
           )
 
-          return await waitAnd(backoff, async () => {
-            if (isClosedOrClosingWebSocketProvider(this.currentProvider)) {
+          return waitAnd(backoff, async () => {
```
cairomassimo, line 201: Nice that you removed the redundant `await` here.

BTW there is an `eslint` plugin (maybe in `typescript-eslint`) that checks them. It disallows `return await x` when semantically equivalent to `return x`, and actually _forces_ you to use `await` when it makes a difference (i.e., inside `try` clauses).

cairomassimo (commented):
A couple of comments :)

background/services/chain/serial-fallback-provider.ts
0xDaedalus: The idea here was just to get our backoff time to be a little less aggressive (right now it averages something like 150ms backoff, 19s backof, 6 min backoff - and this made it more like 150ms, 1s, 4s, 10s, etc..).  From further testing it looks like an **even more** aggressive backoff timing than what we have right now may be desirable so this looks like it's definitely going to change.

background/services/chain/serial-fallback-provider.ts
0xDaedalus: You know - maybe we _want_ the `return await` here?  [The rule you're referencing](https://eslint.org/docs/rules/no-return-await) has some interesting context.

> Using return await inside an async function keeps the current function in the call stack until the Promise that is being  awaited has resolved, at the cost of an extra microtask before resolving the outer Promise.

The cost of the extra microtask might be worth the tradeoff of making debugging easier (especially with how hard it has been to debug the hanging swaps issue).  @greg-nagy what do you think?

greg-nagy: My first reaction is to get rid of the unnecessary awaits. 
But the thing is that in this specific case you are down in the trenches. I don't see how keeping it there would make debugging easier but that's on me. Please choose whatever makes your life easier.

0xDaedalus:
Closed in favor of #1226

PR #1189, opened 2022-03-17
Fix broken USD network fees labels

### LINKED ISSUE #1161: Transaction Gas Fee Prices not displaying
### Discord Discussion Link

_No response_

Chrome

Adjust the gas fee for a Tally Ho sawp.

### What did not work?

The Dollar values did not populate for the various gas levels:

image

### Version

_No response_

### Relevant log output

_No response_

jagodarybacka:
Resolves #1161

### What has been fixed

Network fees have estimated values in gwei and they should show estimated values in USD too. There was a problem that USD values were always `$0.00`.

### How it was fixed

As user has possibility to set gas limit in the input and override suggested gas limit there was attempt to set 
```gasLimit = BigInt(overrideValue)``` but as `overrideValue = ""` and `BigInt("") === 0n` we had a problem where `gasLimit` was `0`. 

Now we will check if `overrideValue` is an empty string before conversion to BigInt.

### DIFF
--- a/ui/components/NetworkFees/NetworkSettingsSelect.tsx
+++ b/ui/components/NetworkFees/NetworkSettingsSelect.tsx
@@ -61,6 +61,7 @@ const gasOptionFromEstimate = (
           2
         )
       : undefined
+  const dollarValue = feeAssetAmount?.localizedMainCurrencyAmount
 
   return {
     confidence: `${confidence}`,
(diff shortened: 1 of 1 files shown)


jagodarybacka:
@0xDaedalus isn't it how it should be working? Empty gas limit means that we can't calculate dollar value, entering it manually or using suggested value (if you click "approve asset" you will have final estimation with pre filled gas limit) means we can display dollars. At least that is what I understand from code :)

0xDaedalus:
You're absolutely right - I should have had my coffee this morning ☕ .   My gas limit comment is off as well I was confusing `Gas Limit` for `Gwei`

0xDaedalus (approved):

PR #1193, opened 2022-03-18
Fix infinite route history dispatch problem

jagodarybacka:
Fixes problem that caused #1122

### What

We had infinite loop of dispatches and rerenders of the Popup page because of updating route history with arrays with the same content.

### Fix

We check if the route history has changed before dispatching it.


greg-nagy (approved):
thank you ❤️

I really wanted this to be released 🚢

Issue #438, opened 2021-11-19
[Feature]: UI string localization


Umiiii:
### Discord Discussion Link

https://discord.com/channels/808358975287722045/875438141610270740/911210317302870067

### What would you like to be able to do?

To allow people who doesn't speak English can also join the crypto world, we should localize all the string inside tally extension.

For now, the alpha version of tally hardcoded all its string.

So I suggested there should be these steps (in general, draft):

- [ ] Extract all hardcoded strings 
- [ ] Define the key of localized strings (i.e. `tally.wallet.transaction.send` could be `Send` in `en-us` and `发送交易` in `zh-CN`)
- [ ] Find a suitable collaboration translate website (such as app.lokalise.com) so we can ask community to join the localization
- [ ] Write some build scripts to update the localization strings

Shadowfiend:
This sounds great! As noted in Discord, would love to see a working group develop alongside this. We've def taken the quick path at the moment, but have had i18n in the back of our minds. A lot of the localizable pieces are localized where possible in the current code (e.g., currency stringification, decimalization).

gasolin:
#1195 introduced the basic structure of i18n on the extension side.

may need to think more about how to support string localization to the extension

there are 3 goals to meet

1. to find a way to sync each messages.json file (ex, when en/messages.json changed, how to propagate to others)
2. easiness to leverage the community's help, new translations requests will happen in each release cycle
3. monitor the translation progress of each language

for example, to evolve from this PR, the translation flow could start from picking an online translate tooling such as https://translatewiki.net/ , https://crowdin.com/ or others

- when en/messages.json is changed, upload to the translate site
- notice translators to start. translate
- download the translations from the translate site

gasolin:
https://discord.com/channels/808358975287722045/808360140008456203/975749765901983824

just setup one https://hosted.weblate.org/projects/tallycash/
if all settled, weblate supports webhook to get updated from tally repo en/message.json and support push PR to tally 🤩
- can login via github, once you login/registered, pass your username and I can add you to the project user to explore together

current lack parts:
- Configure repository hooks for automated flow of updates to Weblate.
- Configure push URL for automated flow of translations from Weblate.
- Define translation instructions to give translators a guideline.
- Add screenshots to show where strings are being used.
- Use flags to indicate special strings in your translation.
- Enable add-on: Cleanup translation files
- Update all translation files to match the monolingual base file. For most file formats, this means removing stale translation keys no longer present in the base file.

[...]

PR #1195, opened 2022-03-18
init extension i18n support

gasolin:
related to #438

refer official i18n doc https://developer.chrome.com/docs/webstore/i18n/
to localize some strings

截圖 2022-03-18 下午11 13 43

截圖 2022-03-18 下午11 13 52 

to add new locale, just add _locale/es/messages.json and copy from en/messages.json

Learned while porting:
- the message format is verbose, and json is not a good format to maintain multiple locale files. Need think about a better way to
  - merge new strings for locale files 
  - collaborate the translations work
- localize ui/pages/Tab.tsx broken the UI, need more tweak
- some strings (ex: `$`) need to be present as the encoded version in messages.json 
refer https://www.w3schools.com/tags/ref_urlencode.asp

### DIFF
--- a/ui/_locales/en/messages.json
+++ b/ui/_locales/en/messages.json
@@ -8,7 +8,7 @@
     "description": "in Wallet > WalletAccountBalanceControl"
   },
   "menuHideSmallAssetBalance": {
-    "message": "Hide asset balance under $$2",
+    "message": "Hide asset balance under $$$1",
     "description": "in pages > Menu"
   },
   "menuSetAsDefault": {
--- a/ui/_locales/en/messages.json
+++ b/ui/_locales/en/messages.json
@@ -8,7 +8,7 @@
     "description": "in Wallet > WalletAccountBalanceControl"
   },
   "menuHideSmallAssetBalance": {
-    "message": "Hide asset balance under %242",
+    "message": "Hide asset balance under $$2",
     "description": "in pages > Menu"
   },
   "menuSetAsDefault": {
--- a/ui/_locales/zh_TW/messages.json
+++ b/ui/_locales/zh_TW/messages.json
@@ -8,7 +8,7 @@
     "description": "in Wallet > WalletAccountBalanceControl"
   },
   "menuHideSmallAssetBalance": {
-    "message": "隱藏小於 $$2 的資產",
+    "message": "隱藏小於 $$$1 的資產",
     "description": "in pages > Menu"
   },
   "menuSetAsDefault": {
--- a/ui/_locales/zh_TW/messages.json
+++ b/ui/_locales/zh_TW/messages.json
@@ -1,18 +1,18 @@
 {
   "totalAccountBalance": {
-    "message": "Total account balance",
+    "message": "帳戶總額",
     "description": "in Wallet > WalletAccountBalanceControl"
   },
   "readOnlyNotice": {
-    "message": "Read-only mode",
+    "message": "唯讀模式",
     "description": "in Wallet > WalletAccountBalanceControl"
   },
   "menuHideSmallAssetBalance": {
-    "message": "Hide asset balance under %242",
+    "message": "隱藏小於 %242 的資產",
     "description": "in pages > Menu"
   },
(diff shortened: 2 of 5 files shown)


Shadowfiend:
Stunning :heart_eyes:

> some strings (ex: $) need to be present as the encoded version in messages.json

This is infuriating lol. However, it feels like a lot of those strings will probably be UI-generated anyway. e.g., right now we have a setting for hiding <$2 balances, but we ultimately want the user to be able to configure their main currency of choice, and the balance amount will need to adapt at that point.

Anyway, that's for the future.

Question: do you want to do more than this, or were you interested in landing this PR as-is and then continuing the work separately?

Shadowfiend:
Also of note, I ran across https://github.com/Lusito/web-ext-translator at some point. I haven't even run it yet, but if it's good we could definitely consider making it a primary part of the translation flow.

Shadowfiend:
@gasolin wanted to bubble this back up:

> Question: do you want to do more than this, or were you interested in landing this PR as-is and then continuing the work separately?

If you're wanting to land this as-is we can probably do it in the next few work days.

gasolin:
@Shadowfiend I think the PR is ready to land if you like this approach.

https://github.com/Lusito/web-ext-translator looks good but may need to think more to 
1. find a way to sync each messages.json file (ex, when en/messages.json changed, how to propagate to others)
2. easiness to leverage the community's help, new translations requests will happen in each release cycle
3. monitor the translation progress of each language

for example, to evolve from this PR, the translation flow could start from picking an online translate tooling such as https://translatewiki.net/
- use github action to upload the en/messages.json to the translate site when en.json is changed
- download the other language translations from the translate site

Is there a place to involve in the discussion? 
can't find the dev related channel in discord

gasolin:
@Shadowfiend will move the main translation related discussion to #438

ui/_locales/en/messages.json
```diff
@@ -1 +1,18 @@
...
+  "totalAccountBalance": {
+    "message": "Total account balance",
+    "description": "in Wallet > WalletAccountBalanceControl"
+  },
+  "readOnlyNotice": {
+    "message": "Read-only mode",
+    "description": "in Wallet > WalletAccountBalanceControl"
+  },
+  "menuHideSmallAssetBalance": {
+    "message": "Hide asset balance under %242",
```
Shadowfiend, line 11: Can we parametrize this? It's hardcoded right now but it will almost certainly become a preference.

Shadowfiend (approved):
Solid start 💪

ui/_locales/en/messages.json
Shadowfiend: (Incidentally this also addresses some of the encoding issue—anything currency-related should be provided by the extension, which can localize separately from the string around it.)

Shadowfiend:
Darn it, we really need a bot to check this stuff automatically… @gasolin we need you to sign your commits; see https://github.com/tallycash/extension#commit-signing .

gasolin:
> Darn it, we really need a bot to check this stuff automatically… @gasolin we need you to sign your commits; see https://github.com/tallycash/extension#commit-signing .

the sign is removed when using github rebase UI on web....

ui/_locales/en/messages.json
gasolin: seems possible, https://developer.chrome.com/docs/extensions/mv3/i18n-messages/

gasolin: turn the string to accept a param

Shadowfiend (approved):
Forward!

henryboldi:
I believe merging this introduced a little bug w/ the double dollar signs here? Or am I missing something?
double_dollar

Shadowfiend:
Hmmm... this should be getting interpolated by the browser automatically. This in chrome?

henryboldi:
Yes. Haven't checked it out in Firefox 😬

gasolin:
oops, here's the fix https://github.com/tallycash/extension/pull/1311

PR #1196, opened 2022-03-18
Remove "ETH coming soon" notice

0xDaedalus:
Now that we support swapping from ETH into other ERC-20 Tokens.  The notice is no longer needed.

### To Test
- Open up the Swaps page, select ETH as your `From` Asset, the notice should not be there.


henryboldi (approved):
LGTM 👏

PR #1198, opened 2022-03-18
Disable Editing of Slippage Tolerance in Swaps Network Settings.

### LINKED ISSUE #863: Swaps - "Slippage tolerance" button does nothing when clicks
### Discord Discussion Link

_No response_

Chrome

I clicked slippage tolerance, expecting to see a dropdown to change %

### What did not work?

Button changed color on hover... when I clicked it, it had no effect:

Screen Shot 2022-01-20 at 8 16 12 PM

Screen Shot 2022-01-20 at 8 16 58 PM

### Version

_No response_

### Relevant log output

_No response_

0xDaedalus:
Closes #863

Since this is a setting we do not ~~have control over~~ currently allow the user to change we should just display is as a value and not a dropdown.

Before:
image

After: 
image

### DIFF
--- a/ui/components/Swap/SwapTransactionSettingsChooser.tsx
+++ b/ui/components/Swap/SwapTransactionSettingsChooser.tsx
@@ -72,9 +72,7 @@ export default function SwapTransactionSettingsChooser({
             <div className="settings_wrap">
               <div className="row row_slippage">
                 <span className="settings_label">Slippage tolerance</span>
-                <SharedButton type="secondary" size="medium" icon="chevron">
-                  1%
-                </SharedButton>
+                <span>1%</span>
               </div>
               <div className="row row_fee">
                 <NetworkSettingsSelect


Shadowfiend:
Fwiw, we do have control, we just haven't discussed what the interaction for editing this should be. I believe the current implementation deviates already from the original design.

Right now this just gets hardcoded a few lines above your edit:

https://github.com/tallycash/extension/blob/5e066be74e623b95f6b691e34cb47b3bcc185faa/ui/components/Swap/SwapTransactionSettingsChooser.tsx#L50-L54

Shadowfiend:
Related/unrelated, we could use the swap quote's response @ https://github.com/tallycash/extension/blob/85e2c1f0ab68d5deecc8a693390757e3263f703c/ui/pages/Swap.tsx#L333 to update the swap settings with a suggested gas limit based on the `estimatedGas` field (https://github.com/tallycash/extension/blob/ca11f3bf95df60bbe1359243180801db36f0a108/background/lib/validate/0x-swap.ts#L15 ) and get better network cost estimates sooner now that @jagodarybacka's fixes have landed for that calculation.

0xDaedalus:
Is this something we want to get in before release?  As it stands now having the 1% be a dropdown menu with 1 option makes it looks like a bug.  Happy to work on this though if we want to prioritize it.

Shadowfiend:
> As it stands now having the 1% be a dropdown menu with 1 option makes it looks like a bug.

Dis true.

Shadowfiend (approved):
Let us do.

PR #1200, opened 2022-03-18
Do not merge: Webauthn test

fernya:
Currently is has serious issues with life, @greg-nagy might see the culprit in this

### DIFF
--- a/background/services/ledger/index.ts
+++ b/background/services/ledger/index.ts
@@ -84,44 +84,57 @@ type Events = ServiceLifecycleEvents & {
 export const idDerviationPath = "44'/60'/0'/0/0"
 
 async function deriveAddressOnLedger(path: string, eth: Eth) {
-  const derivedIdentifiers = await eth.getAddress(path)
-  const address = ethersGetAddress(derivedIdentifiers.address)
-  return address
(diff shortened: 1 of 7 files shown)


background/services/ledger/index.ts
```diff
@@ -84,39 +84,63 @@ type Events = ServiceLifecycleEvents & {
...
-  }
+  return [address, LedgerType.LEDGER_NANO_S]
+}
 
-  if (extensionDeviceType === LedgerType.UNKNOWN) {
-    return [undefined, extensionDeviceType]
-  }
+function getTransport(): Promise<Transport> {
+  const isChrome =
+    /Chrome/.test(navigator.userAgent) && /Google Inc/.test(navigator.vendor)
```
Shadowfiend, line 135: Small note realizing that this is not final, but we should plan on feature-detecting rather than checking the user agent here. `typeof navigator.usb === "object"` seems like a reasonable check to pull from the previous iteration.

greg-nagy:
@fernya sorry, but haven't got to it :(

Shadowfiend:
Closing this as webauthn isn't currently on the roadmap.

PR #1201, opened 2022-03-18
Display snackbar message on transaction broadcast

### LINKED ISSUE #1069: Swaps: Indicate that the transaction was submitted before returning to any other screen
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/948429175377784842

After clicking the button to sign swap transaction, the wallet goes back to swap screen without any indication to the user as to whether or not the transaction was actually sent.

Because there is no transaction queue yet, it doesn't even show up under 'Activity', which may lead one to believe that it was not successful until it actually goes through and shows up.

UX improvement: Indicate to the user that the transaction was actually sent and will show up when completed.

0xDaedalus:
Resolves #1069 

Added a snackbar message so users have feedback that their transaction has successfully been broadcast.

### To Test:
- Perform a swap
- Ensure that `Transaction signed, broadcasting...` is visible in the extension immediately after.

image


ui/pages/SignTransaction.tsx
```diff
@@ -81,7 +81,7 @@ export default function SignTransaction({
           location.state.redirectTo.state
         )
       } else {
-        history.goBack()
+        history.push("/")
```
0xDaedalus, line 84: would the change from `goBack` to always redirecting to `/` cause an issue in any flows outside of Swap?

Shadowfiend: Almost certainly yes.

greg-nagy (approved):
oh the joy this brings ... 🥰

Shadowfiend (commented):
Couple of non blocking notes.

background/services/chain/index.ts
```diff
@@ -621,21 +622,26 @@ export default class ChainService extends BaseService<Events> {
...
-            "alchemy"
-          )
-          this.releaseEVMTransactionNonce(transaction)
-
-          return Promise.reject(error)
-        }),
+        this.providers.ethereum
+          .sendTransaction(serialized)
+          .then((transactionResponse) => {
+            this.emitter.emit("transactionSent", transactionResponse.hash)
```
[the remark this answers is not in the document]
0xDaedalus: Ooooh yes I love this 📈

PR #1204, opened 2022-03-19
Fix swap gas limit allowing negative numbers

### LINKED ISSUE #1191: Gas limit can be negative on Swap Network Settings
We should not allow a negative value for `Gas limit`.

image

alleycat0001:
Closes #1191 

Also took care of a FIXME to validate gasLimit at input time and update NetworkFeeSettings.gasLimit type to an optional bigint from string.


Shadowfiend:
Hey there, thanks for the contribution! Quick thing before we move this forward, we'll need you to sign your commits; see https://github.com/tallycash/extension#commit-signing .

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -212,12 +202,17 @@ export default function NetworkSettingsSelect({
...
-      gasLimit: newGasLimit,
-    })
+    try {
+      const parsedGasLimit = BigInt(newGasLimit)
+      if (parsedGasLimit >= 0n) {
+        onNetworkSettingsChange({
+          ...networkSettings,
+          gasLimit: parsedGasLimit,
+        })
+      }
```
Shadowfiend, line 212: Could we add an error note to the input in this case? `SharedInput` has an `errorMessage` prop that will display an error.

This could be as simple as “Invalid gas limit” or so.

Shadowfiend (commented):
Glorious. Left a couple of quick notes, handing off to @jagodarybacka to finish reviewing here and get this merged.

background/redux-slices/transaction-construction.ts
```diff
@@ -278,9 +278,21 @@ export const selectDefaultNetworkFeeSettings = createSelector(
...
+    suggestedGasLimit,
+  }): {
+    feeType: NetworkFeeTypeChosen
+    gasLimit: bigint | undefined
+    suggestedGasLimit: bigint | undefined
+    values: {
+      maxFeePerGas: bigint
+      maxPriorityFeePerGas: bigint
+    }
+  } => ({
```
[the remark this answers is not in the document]
alleycat0001: Much better, done ✔️

ui/components/NetworkFees/NetworkSettingsSelect.tsx
alleycat0001: Added. Do note that the only way this occurs is when the user clicks the built-in browser down arrow when the input is at 0. Looks like the shared input when set to `number` type doesn't allow any other invalid input.

jagodarybacka:
Hi @alleycat0001 I've reviewed changes and they are looking good.

Input is not working perfectly yet in my opinion, as there is a possibility to get en error message while input is still showing positive number but I think it should be discussed with @VladUXUI (if we want to show negative numbers with error message or just not allow to go below 0). 

Steps to reproduce:
image
- enter positive number into input (for example `123`)
- place cursor on the beginning
- type minus sign `-`

VladUXUI:
Awesome work here!
I would go with showing an error for negative numbers. This way at least user get's some feedback

jagodarybacka:
@alleycat0001 I've added issue https://github.com/tallycash/extension/issues/1227 You can take it if you want 😄

alleycat0001:
@jagodarybacka feel free to assign to me 👍 

As discussed I will:
1. Do the `parseAndValidateFunction` work in a separate PR and get that merged
2. Come back to this PR, rebase, and update to use `parseAndValidateFunction`.

jagodarybacka:
@alleycat0001 Hmm, I can't assign you to the issue but I will assign myself and I will review your PR.

Shadowfiend:
> Come back to this PR, rebase, and update to use parseAndValidateFunction.

@alleycat0001 Would love to actually merge this PR with the dropped error if that doesn't mess with your flow too much! We already have other fields behaving this way (not accepting input vs displaying an error), and if we're following up with a fix for it I think the typing win (of having proper `bigint`s) here lets us clean up certain other in-progress PRs like #1215 as well.

alleycat0001:
@Shadowfiend  my bad, I think I misinterpreted your comment. Pushed a commit that removes the `gasLimit` validation message ✔️

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -212,12 +202,17 @@ export default function NetworkSettingsSelect({
   }, [updateGasOptions])
 
   const setGasLimit = (newGasLimit: string) => {
-    // FIXME Make gasLimit a bigint and parse/validate here, as close to the user
-    // FIXME entry as possible.
-    onNetworkSettingsChange({
-      ...networkSettings,
-      gasLimit: newGasLimit,
-    })
+    try {
```
Shadowfiend, line 205: One more bit after testing—if `newGasLimit.trim()` is empty, we should `onNetworkSettingsChange` with `gasLimit: undefined` or without `gasLimit`, either way. Otherwise we don't have a way to signal “use the estimated gas” after entering anything.

alleycat0001: Done, I should have caught that myself.

Shadowfiend: That's what review is for! 🙌

Shadowfiend (commented):
Green light from me, but I didn't get a chance to do a final test drive. @jagodarybacka if you get a chance to test and then want to merge and layer on the dollar gas price stuff on top, I'm 100% behind that 🚀

jagodarybacka (approved):
Checked, looks good ✨

Issue #1101, opened 2022-03-06
Swap - Sign-TX 


VladUXUI:
Current issues that make it difficult to verify what you are signing:

- [x] #1224
- [x] USD value is missing the $ sign in front
- [ ] Advanced tab missing: Advanced tab should consist of as many details as we can offer (Decoded). We should at a minimum try and show Function Type and the contract you are interacting with. 
- [x] Gas should have USD value as well (covered in this ticket https://github.com/tallycash/extension/issues/957)
- [x] Raw Data tab 

<img width="860" alt="image" src="https://user-images.githubusercontent.com/80846886/156914728-1e91b9ed-a209-48ff-9570-035695319487.png">

Shadowfiend:
Things that should be doable this week:

- Raw data tab.
- $ in front of USD amounts.
- USD gas conversion (probably).

jagodarybacka:
Raw data was added in #1134

0xDaedalus:
Closed b/c stale

Issue #374, opened 2021-10-28
Asset Transfer


Shadowfiend:
- [x] #131 
- [x] End-to-end send token flow

wojciech-turek:
@henryboldi It is, imo we close this issue when #369 is done.

mhluongo:
Closed by #1263

PR #1210, opened 2022-03-21
Show currency on sign tx page

jagodarybacka:
Mentioned by #1101
### What
Show ETH and $ signs on the sign tx page to let user know how much they are going to spend

### Demo
Screenshot 2022-03-21 at 12 25 57


ui/components/SignTransaction/SignTransactionSignInfoProvider.tsx
```diff
@@ -69,12 +71,9 @@ export default function SignTransactionSignInfoProvider({
           <div className="container">
             <div className="spend_amount_label">Spend Amount</div>
             <div className="spend_amount">
-              <div className="eth_value">
-                {completeTransactionAssetAmount.localizedDecimalAmount}
-              </div>
+              <div className="eth_value">{ethValue} ETH</div>
```
0xDaedalus, line 74: I think that `enrichAssetAmountWithMainCurrencyValues` should return an `asset` which should have a `symbol` property you can use here instead of hard-coding `ETH`.

0xDaedalus (changes requested):

greg-nagy (changes requested):
This works fine for ETH but nothing else. 

I am not sure where you should get the information for the asset. Please try to figure it out, but if it proves to be a pain just reach out and we can figure it out.

CleanShot 2022-03-22 at 14 32 13

Shadowfiend:
Let's 100% not block this merge on that.

Recognizing simple asset transfers and presenting them correctly is something we can put under #374 (we can use existing enrichments for it), and recognizing asset transfers as a general class and swaps as a general class is something we're going to require enrichment for (likely via Blocknative), and that we'll have to make some tough calls on how we interpret and display. Recognizing our own swaps is doable with less work but will probably require some architectural fiddling.

greg-nagy (approved):
I lean on the opinion from @Shadowfiend 

let's 🚀

PR #1212, opened 2022-03-21
Fix height of token list

jagodarybacka:
### What 

Fix height of list as ETH bar was removed

#### Before
Screenshot 2022-03-21 at 14 07 44

#### After
Screenshot 2022-03-21 at 14 06 37


0xDaedalus (approved):
looks great 🎉 🎉

PR #1215, opened 2022-03-21
Show currency network fee

### LINKED ISSUE #957: Network fee updates/bugs
Current UI specs ([Figma](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=6062%3A153814)) show the network fees as USD amounts when signing on Ledger (along with Gwei), but the currently-implemented keyring signing flow does _not_ show an USD amount unless the user opens the fee settings selector.

Somewhere in [Figma](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=1875%3A63567) the main currency amount is shown in the keyring flow as well.

Showing the main currency (USD) amount when signing with a Ledger requires some extra code, which could be useful for non-Ledger flow as well.

We should:

- confirm which is the intended UI (@VladUXUI),
- implement this once, taking into account all the use cases (keyring signing, Ledger signing, future use cases)

jagodarybacka:
Mentioned by #1101
Resolves #957

### What

* Show dollar value of network fee in the network fee settings button - value is corresponding to the value that user will see after opening settings panel. Value changes after user changes selected option or gas limit. 
Screenshot 2022-03-21 at 15 58 14

* Fix border styles issue - elements of network fee settings panel


0xDaedalus:
lgtm!

Shadowfiend:
Sorry, forced this back to draft as I want to give #1204 an opportunity to land. It cleans up types in such a way that should reduce the code here.

@jagodarybacka feel free to make any adjustments locally and then force-push to replace the merges.

ui/components/NetworkFees/FeeSettingsButton.tsx
```diff
@@ -10,7 +10,7 @@ export default function FeeSettingsButton({
 }: FeeSettingsButtonProps): ReactElement {
   return (
     <button className="settings" type="button" onClick={onClick}>
-      <FeeSettingsText />
+      <FeeSettingsText showDollarValue />
```
0xDaedalus, line 13: I wonder if we can assume that if a dollar value is available then we should _always_ show it.  As a user of the wallet I think I'd like this a lot but maybe its too broad of a generalization.

This is just me musing and not a blocking comment.

ui/components/NetworkFees/FeeSettingsText.tsx
```diff
@@ -4,15 +4,60 @@ import {
...
+import { selectMainCurrencyPricePoint } from "@tallyho/tally-background/redux-slices/selectors"
+import { enrichAssetAmountWithMainCurrencyValues } from "@tallyho/tally-background/redux-slices/utils/asset-utils"
+import { PricePoint } from "@tallyho/tally-background/assets"
 import React, { ReactElement } from "react"
+import { getGasLimit } from "@tallyho/tally-background/redux-slices/utils/transacation-utils"
 import { useBackgroundSelector } from "../../hooks"
 
-export default function FeeSettingsText(): ReactElement {
+const getFeeDollarValue = (
+  currencyPrize: PricePoint | undefined,
```
0xDaedalus, line 20: I think we want `currencyPrice` here

ui/components/NetworkFees/FeeSettingsText.tsx
```diff
@@ -4,15 +4,60 @@ import {
...
+  networkSettings: {
+    gasLimit: string
+    suggestedGasLimit: bigint | undefined
+    values: { maxFeePerGas: bigint; maxPriorityFeePerGas: bigint }
+  }
+) => {
```
0xDaedalus, line 26: lets explicitly type what this function returns

0xDaedalus (changes requested):
Visually looks great and handlers small and large dollar values well.  Just a couple of small code comments and we're good to go!

ui/components/NetworkFees/FeeSettingsButton.tsx
Shadowfiend: I tend to agree we shouldn't hide this behind a prop unless we know for sure we want to hide it in certain circumstances.

Definitely need to QA that layouts are ready for that behavior though.

ui/components/NetworkFees/FeeSettingsText.tsx
```diff
@@ -4,15 +4,53 @@ import {
...
+    values: { maxFeePerGas, maxPriorityFeePerGas },
+  } = networkSettings
+  const gasLimit = networkSettings.gasLimit ?? networkSettings.suggestedGasLimit
+
+  if (!gasLimit) return undefined
+
+  const { localizedMainCurrencyAmount } =
+    enrichAssetAmountWithMainCurrencyValues(
+      {
+        asset: ETH,
```
Shadowfiend, line 33: Rather than hardcode this, let's use the current price point's first asset. Price points are in a stable order when they reach us here, and this will be more future-proof for other EVM chains that need similar fee logic.

ui/components/NetworkFees/FeeSettingsButton.tsx
Shadowfiend: Believe this still outstanding. Are we avoiding dollar values in other places for a reason?

Shadowfiend (approved):
Couple of quick morning notes/questions before I actually get started for the day 😅

Shadowfiend (approved):
Gave it a test drive, looks amazing. Let's :guitar:

Issue #1213, opened 2022-03-21
UI tries to unlock KeyringService twice


Naxsun:
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/954648702784520232

### What browsers are you seeing the problem on?

Brave

### What were you trying to do?

Unlock the wallet

### What did not work?

The wallet thinks it's already unlocked and gives the user an 'Incorrect Password' message.

Logs of a different user than above linked message below, removed intermediate area of logs because below field is limited. Can share the full logs separately.

### Version

Community Edition v0.12.1

### Relevant log output

```shell
logger.tss:65 Async thunk failed Error: KeyringService is already unlocked!
    at KeyringService.unlock (index.tss:161:13)
    at main.tss:914:33
    at index.jss:306:13
    at Array.map (<anonymous>)
    at Emittery.emit (index.jss:304:23)
    at async keyrings.tss:146:3
    at async utils.tss:129:9
genericLogger @ logger.tss:65
error @ logger.tss:106
(anonymous) @ utils.tss:131
await in (anonymous) (async)
(anonymous) @ redux-toolkit.esm.jss:1112
step @ redux-toolkit.esm.jss:38
(anonymous) @ redux-toolkit.esm.jss:19
(anonymous) @ redux-toolkit.esm.jss:87
__async @ redux-toolkit.esm.jss:69
(anonymous) @ redux-toolkit.esm.jss:1096
(anonymous) @ redux-toolkit.esm.jss:1095
(anonymous) @ index.jss:8
(anonymous) @ alias.jss:20
dispatchResponse @ wrapStore.jss:94
(anonymous) @ serialization.jss:38
logger.tss:60  🐛 Tally debug @ 2022-03-21T11:07:21.027Z
geturl.jss:30          GET https://api.blocknative.com/gasprices/blockprices 429
(anonymous) @ geturl.jss:30
(anonymous) @ geturl.jss:8
geturl_awaiter @ geturl.jss:4
getUrl @ geturl.jss:13
(anonymous) @ index.jss:169
(anonymous) @ index.jss:8
web_lib_esm_awaiter @ index.jss:4
(anonymous) @ index.jss:165
_fetchData @ index.jss:164
lib_esm_fetchJson @ index.jss:311
getBlockPrices @ index.tss:101
getBlockPrices @ gas.tss:28
(anonymous) @ index.tss:659
pollBlockPrices @ index.tss:658
handler @ index.tss:188
[...]

PR #1217, opened 2022-03-21
Don't fail to unlock already unlocked keyrings.

0xDaedalus:
Partially addresses #1213 until we can figure out the problem on the UI side.

### To Test
I am not sure how to recreate this naturally in the application so these are the steps I took to reliably create inconsistent state:

#### Setup
- add `;(window as any).keyringService = this.keyringService` on the last line of `internalStartService` in `main.ts`.
- add an optional `broadcast` parameter to the `#unlock` method that defaults to true.
- wrap line `143` in `services/keyrings/index.ts` in an `if (broadcast) { ... }` clause.
- pass `false` to `this.#unlock` on like `222` in `services/keyrings/index.ts`

#### Execution
After making the above changes to the codebase
- ensure your wallet is locked by inspecting `keyringService.locked()` in the `background page` console.
- If it is not locked - `keyringService.lock()`.
- Navigate to a page that requires you to unlock your wallet (for example - adding a new address to an account).
- Once the `Enter Password` screen is open - do not enter your password.
- Open the background page console and type `keyringService.unlock(your_password)`
- Verify the keyring is unlocked (`keyringService.locked() === false`)
- The extension should still be asking you for a password (this is the state inconsistency we created above)
- Type in your password to the extension
- You should see a `warning` in the `background page` console about keyrings already being unlocked - and the extension should successfully accept your password and unlock.

### DIFF
--- a/background/services/keyring/index.ts
+++ b/background/services/keyring/index.ts
@@ -24,6 +24,7 @@ import { ETH, FORK, MINUTE } from "../../constants"
 import { ethersTransactionRequestFromEIP1559TransactionRequest } from "../chain/utils"
 import { HIDE_IMPORT_LEDGER, USE_MAINNET_FORK } from "../../features/features"
 import { AddressOnNetwork } from "../../accounts"
+import logger from "../../lib/logger"
 
 export const MAX_KEYRING_IDLE_TIME = 60 * MINUTE
 export const MAX_OUTSIDE_IDLE_TIME = 60 * MINUTE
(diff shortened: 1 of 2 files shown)


background/services/keyring/index.ts
```diff
@@ -158,7 +168,9 @@ export default class KeyringService extends BaseService<Events> {
     ignoreExistingVaults = false
   ): Promise<boolean> {
     if (!this.locked()) {
-      throw new Error("KeyringService is already unlocked!")
+      logger.warn("KeyringService is already unlocked!")
+      this.#unlock()
```
Shadowfiend, line 172: Given that we consider the double unlock a bug that needs fixing separately, I think:

- It's fine (maybe even safer!) not to mark activity when we get it.
- It's fine not to emit `locked`.
- If we do want to do those things, we should repeat the code here. Extracting a function named `#unlock` that doesn't actually unlock anything is not an ideal outcome IMO.

Another way to think about it is that if we add this behavior to the double unlock path, we should also test for it. I personally don't think it's worth the time :grin:

greg-nagy (approved):
let's 🚢

I am starting to fell in love with your PRs dude ... ❤️

PR #1219, opened 2022-03-21
Make "Add address +" link available for Ledger addresses

### LINKED ISSUE #1187: Account selection - Missing add address for Ledger
### Discord Discussion Link

_No response_

Chrome

Connect my second address from my Ledger

### What did not work?

Button is not there 

image

### Version

_No response_

### Relevant log output

```shell
Version: v0.12.2-1-g53105dba (16 Mar 2022)
```

fernya:
Resolves #1187 

# Things that could be broken (but not)

- Try to add addresses to imported wallets (should add one new address at once)
- Try to add addresses to internal wallets (should add one new address at once)
- Try to add addresses to a current Ledger account (should open the onboarding flow)

https://user-images.githubusercontent.com/499307/159330164-0be40daf-63de-42e6-88c3-e00b54570fec.mov

### DIFF
--- a/background/redux-slices/accounts.ts
+++ b/background/redux-slices/accounts.ts
@@ -306,6 +306,6 @@ export const removeAccount = createBackgroundAsyncThunk(
   "account/removeAccount",
   async (address: HexString, { dispatch, extra: { main } }) => {
     dispatch(accountSlice.actions.deleteAccount(address))
-    main.removeAccount(address, { type: "keyring" })
+    main.removeAccount(address, { type: "keyring", keyringID: null })
   }
 )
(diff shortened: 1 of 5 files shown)


Shadowfiend (changes requested):
This behavior is already split up strangely, and I'd rather avoid doubling down on it.

I suggest we:
- Move `keyringId` to be part of the keyring `SigningMethod`, the same way that the Ledger account type carries ledger id information.
- Add a handler function that takes the `SigningMethod` and does the right thing for adding an address.

I'm blocking here because we've already started having folks try to contribute additional external signer integrations, and I'd like to not continue down the path of complicating that contribution process. Let's use `SigningMethod` for its intended purpose and start building around it.

cc @0xDaedalus if you feel like you want to pick up the refactor here (since you originally introduced `keyringId`), we can also merge this PR as-is and do the refactor separately.

fernya:
Thanks @Shadowfiend, I'll do the `SigningMethod` refactor as you suggested.

Shadowfiend:
We closed the associated issue as not a priority, so going ahead and closing this PR as well for now. We can always reference it later if we pick the work up!

PR #1221, opened 2022-03-22
add ENS contractName on SignTransaction modal

jack-the-pug:
### Demo

ens-contract-name

### Motivation

By having a human readable contract address (if available) when signing tx, frontend hijacking attack will be much harder.

### Implementation

Currently only support displaying the ENS name (if exists) for the `tx.to` address when `annotation.type == 'contract-interaction'`.


background/services/enrichment/index.ts
```diff
@@ -144,10 +144,14 @@ export default class EnrichmentService extends BaseService<Events> {
...
         }
       } else {
+        const toName = await this.nameService.lookUpName(transaction.to, ETHEREUM, false);
         // Fall back on a standard contract interaction.
         txAnnotation = {
           timestamp: resolvedTime,
           type: "contract-interaction",
+          displayFields: {
+            contractName: toName
+          }
```
Shadowfiend, line 154: No need for the extra indirection here imo:

```suggestion
          contractName: toName
```

Shadowfiend (commented):
Yaaassss 🙌 

Left a quick note on the data here but this is awesome 🔥

ui/components/SignTransaction/SignTransactionSignInfoProvider.tsx
```diff
@@ -59,6 +60,10 @@ export default function SignTransactionSignInfoProvider({
             ) : (
               <>
                 <div className="label">Send to</div>
+                {annotation?.type === "contract-interaction" &&
+                annotation.contractName ? (
+                  <div className="send_to_ens">{annotation.contractName}</div>
```
Shadowfiend, line 65: Let's make this `send_to_name`. `NameService` is very deliberately not named `EnsService`, and `contractName` is equally importantly not `contractEns` :grin:

Shadowfiend (dismissed):
Nothing blocking left here, but I may not get a chance to test drive tonight. Good to go when someone gives it a spin though 💪

mhluongo (approved):
:rocket:

PR #1226, opened 2022-03-22
Update Provider Reconnect & Subscribe Logic (Pending Swaps Fix)

0xDaedalus:
This PR fixes the issue where swaps would sometimes be stuck in
a `pending` state in the case that our websocket provider disconnected
or was never able to connect in the first place.

To fix this, a few changes were implemented to improve the resilience
of the `SerialFallbackProvider` in the face of disconnects:

- A bug where overriding `detectNetwork` was causing forever blockages
  deep in Ethers was fixed by dropping the override.
- When a WebSocket is connecting, requests are backed off until it is
  done connecting.
- Resubscriptions now wait briefly before being attempted to give the
  websocket a chance to connect first.
- The backoff curve was adjusted so that it maxes out closer to 75s
  before falling back, vs >>10 minutes before.
- Finally, when the SerialFallbackProvider falls back to backup providers,
  it continues to try to reconnect to the primary provider until it
  succeeds. This fixes cases where a fallback is necessary but we still
  want WebSocket subscriptions to be live, since it will eventually
  reconnect to the known-WebSocket primary provider. 

### To Test
- [ ]  Import a wallet with enough assets to execute a swap.
- [ ] Turn off wifi.
- [ ]  Fully Quit Chrome.
- [ ]  Reopen Chrome.
- [ ]  Attempt to execute a swap.
- [ ] Your swap should work, be reflected in the `Activity` tab, and not be stuck in a `pending` state.


background/services/chain/serial-fallback-provider.ts
```diff
@@ -72,6 +73,20 @@ function isClosedOrClosingWebSocketProvider(
   return false
 }
 
+/**
+ * Returns true if the given provider is using a WebSocket AND the WebSocket is
+ * connecting. Ethers does not provide direct access to this information.
+ */
+function websocketProviderIsConnecting(provider: JsonRpcProvider): boolean {
+  if (provider instanceof WebSocketProvider) {
+    // eslint-disable-next-line no-underscore-dangle
```
Shadowfiend, line 82: Let's keep the eslint explainer comment from above here. eslint disables should always include explanations :)

background/services/chain/serial-fallback-provider.ts
```diff
@@ -451,12 +468,21 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
-      const provider = this.currentProvider as WebSocketProvider
+    if (provider instanceof WebSocketProvider) {
+      const websocketProvider = provider as WebSocketProvider
+
+      if (
+        isClosedOrClosingWebSocketProvider(provider) ||
+        websocketProviderIsConnecting(provider)
+      ) {
+        return false
+      }
```
Shadowfiend, line 485: We can just do this as a top-level `if`, right?

0xDaedalus: We definitely can - but it kind of feels like websocket related logic belongs inside of the websocket part of `resubscribe` - although in general I do agree that the more nested the `if`'s the smellier the code.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -469,29 +495,54 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
+      return false
+    }
 
     logger.debug("Subscriptions resubscribed...")
+    return true
+  }
+
+  private async attemptToReconnectToPrimaryProvider(): Promise<unknown> {
+    // Attempt to reconnect to primary provider every 15 seconds
+    return waitAnd(15_000, async () => {
```
Shadowfiend, line 533: No Magic Numbers—let's make sure we're putting constants like this in named variables.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -154,6 +169,11 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
         throw new Error("WebSocket is already in CLOSING")
       }
 
+      if (websocketProviderIsConnecting(this.currentProvider)) {
+        return await waitAnd(100, async () => {
```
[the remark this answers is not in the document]
Shadowfiend: Also as below, let's put this in a named constant.

Shadowfiend (commented):
Few notes throughout, some I forgot to bundle in a review :grimacing:

background/services/chain/serial-fallback-provider.ts
Shadowfiend: Yeah I can see that cut both ways for sure. The way we've tended to write code across the extension, we do tend towards less nesting and earlier returns. It's not always my favorite strategy, but it's consistent with the rest of the codebase.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -469,29 +495,54 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
     logger.debug("Subscriptions resubscribed...")
+    return true
+  }
+
+  private async attemptToReconnectToPrimaryProvider(): Promise<unknown> {
+    // Attempt to reconnect to primary provider every 15 seconds
+    return waitAnd(15_000, async () => {
+      const primaryProvider = this.providerCreators[0]()
+      // Give the provider 2 seconds to establish websocket connection
+      return waitAnd(2_000, async (): Promise<unknown> => {
```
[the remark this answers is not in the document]
0xDaedalus: I'll go ahead and update the comment - its because if we synchronously try to resubscribe the `primaryProvider` will almost always still be `Connecting` rather than `Closed` or `Open`.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -16,8 +15,10 @@ import { transactionFromAlchemyWebsocketTransaction } from "../../lib/alchemy"
 const BASE_BACKOFF_MS = 150
 // Reset backoffs after 5 minutes.
 const COOLDOWN_PERIOD = 5 * MINUTE
-// Retry 3 times before falling back to the next provider.
-const MAX_RETRIES = 3
+// Retry 8 times before falling back to the next provider.
+// This generally results in a wait time of around 30 seconds before falling back since we
+// usually have multiple requests going out at once.
```
Shadowfiend, line 20: Might be useful to include the top-end range here (I think ~75s max if we hit the high end of each backoff slot and backoffs happen end to end instead of in parallel?).

background/services/chain/serial-fallback-provider.ts
```diff
@@ -431,6 +448,7 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
    * has been somehow set out of range, resets it to 0.
    */
   private async reconnectProvider() {
+    await this.disconnectCurrentProvider()
```
Shadowfiend, line 451: I'm nervous about this. We sometimes call `reconnect` when the provider is already disconnecting; while right now this may not throw exceptions, I'm not sure we can be sure that will be the case.

Shadowfiend (commented):
Couple more comments as I dig deeper here.

Also of note, this chunk:

https://github.com/tallycash/extension/blob/3547c72b103852ab2e344a44e6044eaa2a512764/background/services/chain/serial-fallback-provider.ts#L205-L212

May need to go away now that we're consistently trying to resubscribe to the first provider. We may need to shift to a different strategy if we loop around, OR `attemptToReconnectToPrimaryProvider` should detect that the current provider index has looped back to `0` and stop attempting in that case.

Thoughts?

background/services/chain/serial-fallback-provider.ts
0xDaedalus: Are you sure we need to worry about that?  It looks like [the `close` method the underlying websocket library uses](https://github.com/websockets/ws/blob/dd3d5a710b490d7b4211f92e882b34b5114c8bc2/lib/websocket.js#L287) gracefully handles calling `close` while the websocket is disconnecting - and is something I think we can expect any widely-used ws library to handle.

0xDaedalus:
> May need to go away now that we're consistently trying to resubscribe to the first provider. We may need to shift to a different strategy if we loop around, OR attemptToReconnectToPrimaryProvider should detect that the current provider index has looped back to 0 and stop attempting in that case.
>
> Thoughts?

I like the sound of `attemptToReconnectToPrimaryProvider` detecting the current provider index and stopping if it's 0.  I have a feeling we'll see more changes in here when we introduce transaction queueing so I think keeping the delta small for now makes sense.

background/services/chain/serial-fallback-provider.ts
Shadowfiend: In general I'm nervous about treating a mutating operation like `close` as idempotent unless it's part of the definition of the operation. 

It also feels like a stretch to give a reconnect function the responsibility for disconnecting. If this is how all reconnects handle disconnecting, it's fine, but right now it feels like we're disconnecting in a couple of different places and it's not clear exactly why we have to do it again here in the cases where we've already done it in the caller. 

All that said, I don't think it's a blocker for this merge... Just want to make sure we're thinking through the flow rigorously as we shape this provider up over time.

Shadowfiend:
Code here looks good overall; excited for it. Testing and merge tomorrow!

Shadowfiend (approved):
Tested it and it worked 💪 Overall resilience is something we'll have to observe in the wild, but underlying behavior is much improved and generally doesn't result in failure of epic proportions on the provider's part. Let's :ship:

PR #1228, opened 2022-03-22
💉 Hot-shot: Inject window-provider at build time into provider bridge

greg-nagy:
- https://www.urbandictionary.com/define.php?term=hot-shot
- https://www.imdb.com/title/tt0102059/

The async fetch call to load the `window-provider.js` runtime
caused some really bad race conditions, so we needed a way
to have the content of the `window-provider.js` readily available
when the provider-bridge starts running.

This functionality had to work in both dev and prod environments
so the solution was to create a webpack plugin for this.

The trickiest part was finding the proper hooks in webpack
and escaping the source code so it can be stored as a string.


build-utils/inject-window-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+              // eslint-disable-next-line no-underscore-dangle
+              assets[WINDOW_PROVIDER_FILENAME]._children[0]._value
+
+            // need to encode so it can be used as a string
+            // in non optimised builds the source is a multi line string > `` needs to be used
+            // but ${ needs to be escaped separatly otherwise it breaks the ``
+            windowProviderSource = encodeURI(windowProviderSource).replaceAll(
+              "${",
+              "\\${"
+            )
```
Shadowfiend, line 30: I think you can just `JSON.stringify` here and skip the URI encoding, no?

build-utils/inject-window-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+            windowProviderSource = `\`${windowProviderSource}\``
+            const providerBridgeSource =
+              // @ts-expect-error this exist
+              // eslint-disable-next-line no-underscore-dangle
+              assets[PROVIDER_BRIDGE_FILENAME]._children[0]._value
+
+            // @ts-expect-error this exist
+            // eslint-disable-next-line no-underscore-dangle,no-param-reassign
+            assets[PROVIDER_BRIDGE_FILENAME]._children[0]._value =
```
Shadowfiend, line 39: Rather than munging internals here again, I believe you can just:
```typescript
  assets[PROVIDER_BRIDGE_FILENAME] = new sources.RawSource(
    providerBridgeSource.replace...
  )
```

(`sources` is importable from `webpack` in the existing import above.)

Shadowfiend (changes requested):
I think these are the right bones, but the approach feels like it's extremely brittle. I think we can clean it up a little by:

- Using the public API for Webpack `Source`s.
- Using JSON stringification~~/parsing~~ instead of URI encoding.
- Lifting the window provider to a top-level variable.
- Dropping the placeholder.

For the last two, what I mean is adding, at the top level of provider-bridge/index.ts:

```typescript
const INJECTED_WINDOW_PROVIDER_SOURCE = ""
```

Then replacing that directly:

```typescript
    source.replace(
      'const INJECTED_WINDOW_PROVIDER_SOURCE = ""',
      `const INJECTED_WINDOW_PROVIDER_SOURCE = ${windowProviderSourceString}`,
    )
```

We should also pull these magic strings to the top level if possible.

build-utils/inject-window-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+      PLUGIN_NAME,
+      (compilation: Compilation) => {
+        compilation.hooks.processAssets.tap(
+          {
+            name: PLUGIN_NAME,
+            stage: Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_INLINE,
+          },
+          (assets) => {
+            let windowProviderSource =
+              // @ts-expect-error this exist
```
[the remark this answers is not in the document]
greg-nagy: \o/ thank you!

greg-nagy:
* Using the public API for Webpack Sources. - in progress, will update
* ~~Using JSON stringification/parsing instead of URI encoding.~~ - this has a problem with ```` and `${}`
* ✅Lifting the window provider to a top-level variable.
* ~~Dropping the placeholder.~~ - can't do this :( When the source is optimized/minified we lose the variable names, but not the strings.

build-utils/inject-window-provider.ts
Shadowfiend: I'm not seeing that locally—if you inject the string directly it seems to work? You don't `JSON.parse` the string when injecting though, you just inject the stringified version directly as the `textContent`.

build-utils/inject-window-provider.ts
greg-nagy: \o/ thanks

Shadowfiend:
> When the source is optimized/minified we lose the variable names, but not the strings.

That's annoying but makes sense. We have to run after optimization so we're inlining an optimized version, but we can't reference existing variable names because we're running after optimization. Blah lol.

greg-nagy:
> That's annoying but makes sense. We have to run after optimization so we're inlining an optimized version, but we can't reference existing variable names because we're running after optimization. Blah lol.

Tbh that was my original plan, had the hook for it, was almost ready with the source code render part too, but then realized, that I would have the unoptimized code in the variable which would have not been ideal.

build-utils/inject-window-provider.ts
```diff
@@ -30,18 +29,17 @@ export default class InjectWindowProvider {
             )
             windowProviderSource = `\`${windowProviderSource}\``
             const providerBridgeSource =
-              // @ts-expect-error this exist
               // eslint-disable-next-line no-underscore-dangle
-              assets[PROVIDER_BRIDGE_FILENAME]._children[0]._value
+              assets[PROVIDER_BRIDGE_FILENAME].source() as string
```
Shadowfiend, line 33: This `as` isn't actually true! Without knowing exactly when and where Webpack decides to provide a string or a buffer, it's impossible to make this assumption safely.

`toString()` is a safer way to ensure we have a string here.

I think you can also remove a bunch of `eslint-disables` as well now.

build-utils/inject-window-provider.ts
greg-nagy: My bad. I had leftover code and that caused the problem. Thank you.

build-utils/inject-window-provider.ts
greg-nagy: I think the as is actually true in this compiler and compilation hook. But the `toString()` is more robust.

Removed the eslint rules

Shadowfiend:
There are a bunch of outdated comments sprinkled throughout. Other than that I think we're about good to go here.

Hate that we had to take this approach, love that it's done and is working 🙌

build-utils/inject-window-provider.ts
Shadowfiend: Incidentally using `ReplaceSource` would save us the trouble of having to convert the potential buffer to a string (because `indexOf` exists on `string` and `buffer` both and is enough to do the replacement). 

Def not blocking tho, just musing.

Shadowfiend:
Will plan on testing and merging here tomorrow, pending the comment cleanup. 

Is there a site that has shown this fairly consistently? Sounds like login.xyz is one.

greg-nagy:
>> When the source is optimized/minified we lose the variable names, but not the strings.

> That's annoying but makes sense. We have to run after optimization so we're inlining an optimized version, but we can't reference existing variable names because we're running after optimization. Blah lol.

Actually, I realized that we could have the empty string in the source code, inject the replacement string in an early hook and then replace it the same way it's happening now. 

Don't think it worth it though.

greg-nagy:
> There are a bunch of outdated comments sprinkled throughout.

I cleaned these up in my last commit, not sure if I caught all of them though.

Shadowfiend (approved):
Looks good, tests good, let's do it.

Issue #649, opened 2021-12-15
Shared address display component + copy button


mhluongo:
We display addresses all over the place. We should have shared address display components with copy buttons.

I'd propose
- [ ] one that doesn't include a copy button but still copies on click. This can even be styled like text... whatever is takes to fit our look and feel while making truncated addresses copyable.
- [ ] another the includes an actual button, as currently used on the Receive page.

Thoughts @VladUXUI @henryboldi? Here's where the discussion got started.

>  It's terrible. We probably want a component that does this + a copy button + a `title` with the full address.

_Originally posted by @Shadowfiend in https://github.com/tallycash/extension/pull/546#discussion_r769231648_

VladUXUI:
We have a Ui component and a ux for this. Could potentially be rethought. 
Basically clicking on name/ens would copy the address + show a "address copied to clipboard" snackbar.

> We display addresses all over the place.
I'm uncertain if it's a good idea to let people copy from any/more places by clicking on an address. I think this could lead to people making mistakes and sending funds to wrong addresses. Having one or two source of truth from were you can copy an address might be safer. But definitely worth some exploration/testing.

![image](https://user-images.githubusercontent.com/80846886/146153082-d39968da-301b-4957-bf48-2402ee5a3d71.png)

I suspect we will soon need to show the address under the account name though(but first i want to test with just the name + show address on mouse over).

Shadowfiend:
> one that doesn't include a copy button but still copies on click.

This is not great for accessibility and discoverability; I would def prefer to show an explicit button for it... And I don't think we're so tight for space anywhere that it would be a problem? I realize it makes for some extra line noise though.

mhluongo:
Accessibility should be about an increase in functionality to serve more people IMO, not a decrease to some sort of lowest common denominator. I'm bringing up a button-less control because I'd rather make sure all our addresses are copy-able than make people go to EtherScan.

Maybe our designs solve this everywhere, but remember this will one day include tx call args, transfer recipients, etc.

Shadowfiend:
🤔 I'm not suggesting we make things not copyable because of the button, right? Just that we should use a button unless we really can't.

0xDaedalus:
Closed as part of #1231

PR #1231, opened 2022-03-23
Copy Waste: Add SharedAddress component with address copying

Shadowfiend:
Add `SharedAddress` component and use it in transaction signing
and activity details.

Current limitations:
- Tooltip is an OS tooltip using `title`, preferences may vary.
- We don't propagate name system information and don't yet have
  a design for how to tell the user which name system was used
  to determine the name they see.
- Names are not resolved in as many places as might be ideal.
- There is no icon indicator that a copy is possible.

That said, anywhere a name is resolved it is displayed, tooltip
has the address, and clicking copies, so the basis for the
functionality we want is there and we are in a strictly better
place with these changes.

Screenshots:

Screen Shot 2022-03-22 at 21 57 10

Screen Shot 2022-03-22 at 21 57 30

See #649.

### DIFF
--- a/ui/components/Shared/SharedAddress.tsx
+++ b/ui/components/Shared/SharedAddress.tsx
@@ -1,0 +1,68 @@
+import { truncateAddress } from "@tallyho/tally-background/lib/utils"
+import { setSnackbarMessage } from "@tallyho/tally-background/redux-slices/ui"
+import { NameResolverSystem } from "@tallyho/tally-background/services/name"
+import React, { ReactElement, useCallback } from "react"
+import { useBackgroundDispatch } from "../../hooks"
+import SharedTooltip from "./SharedTooltip"
+
+type SharedAddressProps = {
+  address: string
+  name?: string | undefined
+  nameResolverSystem?: NameResolverSystem
+}
+
+/**
+ * The SharedAddress component is used to render addresses that can optionally
+ * be represented as resolved names, and that are copiable by clicking.
+ *
+ * The component always expects an `address` prop. If an optional `name` prop
+ * is passed, it is shown; otherwise, a truncated version of the address is
+ * shown. The component tooltip always has a tooltip with the full address.
+ * Additionally, clicking the component will copy the address to the clipboard
+ * and present the user with a snackbar message indicating this has occurred.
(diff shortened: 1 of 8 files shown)


Shadowfiend:
Rebased on main, added hover color, and added extra info in tooltip.

ui/components/Shared/SharedAddress.tsx
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * If the optional `nameResolverSystem` property is provided, an info tooltip
+ * is included in the component to inform the user of which name resolver
+ * system was used to resolve the passed `name`. If no `name` is passed, the
+ * `nameResolverSystem` prop is ignored.
+ */
+export default function SharedAddress({
+  name,
+  address,
+  nameResolverSystem: nameSourceSystem,
```
0xDaedalus, line 32: Just curious why use `nameSourceSystem` inside of the component but `nameResolverSystem` outside.

0xDaedalus (dismissed):
Tested and works great!  I do want to mention though that the user experience of hovering over text - seeing it turn gold - cursor turning into a pointer - and the resulting interaction from clicking NOT being a link is a bit jarring (my mind wants it to be a link that opens in etherscan after seeing gold hover).  Not at all blocking but just wanted to share my subjective experience.

ui/components/Shared/SharedAddress.tsx
Shadowfiend: This was me missing a sloppy auto-rename.

Shadowfiend:
Fixed all those issues :point_up_2:

0xDaedalus (approved):

Shadowfiend:
> I do want to mention though that the user experience of hovering over text - seeing it turn gold - cursor turning into a pointer - and the resulting interaction from clicking NOT being a link is a bit jarring (my mind wants it to be a link that opens in etherscan after seeing gold hover).

Talk to the @mhluongo s if the world lol. I absolutely hate it but it's how most implementations work due to the pervasiveness of the action.

mhluongo:
Just speaking up for our users :stuck_out_tongue:

PR #1232, opened 2022-03-23
🚨🚓🚨 - Alert on failed transaction broadcasts.

0xDaedalus:
This PR gets rid of the `transactionSent` event in `chainService` in favor of a try / catch inside of the `broadcastSignedTransaction` event handler that reports a success or failure based on the result of `broadcastSignedTransaction` method in `chainService`.

### Test Criteria
- [ ] A successful transaction should result in a snackbar message of `Transaction signed, broadcasting...`
- [ ] An unsuccessful transaction should result in a snackbar message of `Transaction failed to broadcast.`

The easiest way to simulate an unsuccessful transaction is to edit the code and throw an error inside of `broadcastSignedTransaction`

This PR addresses [an excellent suggestion](https://github.com/tallycash/extension/pull/1201#discussion_r831486778) made in #1201

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -642,12 +642,6 @@ export default class Main extends BaseService<never> {
       this.store.dispatch(blockSeen(block))
     })
 
-    this.chainService.emitter.on("transactionSent", () => {
-      this.store.dispatch(
-        setSnackbarMessage("Transaction signed, broadcasting...")
-      )
-    })
-
     transactionConstructionSliceEmitter.on("updateOptions", async (options) => {
       const {
         values: { maxFeePerGas, maxPriorityFeePerGas },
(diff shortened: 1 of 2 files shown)


Shadowfiend (dismissed):
Yus and yussss let's do it :rocket:

Shadowfiend:
Okay, so unfortunately I just gave swaps a test drive, and a couple of things stood out:
- When a swap transaction completes, it goes to the wallet page, but not to the activity tab :(
- The snackbar message doesn't seem to show up on swap transaction failure.

I ran out of ETH in my test account before I could test swap transaction success snackbar 😅

Also of note, I did see success + failure snackbar messages when doing simple ETH transfers.

0xDaedalus:
> When a swap transaction completes, it goes to the wallet page, but not to the activity tab :(

I agree that this makes sense but I didn't add it in #1201 because it does not match what is in figma.  I can go ahead and make another issue for that.

> The snackbar message doesn't seem to show up on swap transaction failure

:sadface: looking into this now

Shadowfiend:
> I can go ahead and make another issue for that and address immediately.

Maybe let's hold there and check in with @VladUXUI in chat in the morning, I might have invented the behavior 😅

0xDaedalus:
Ok so, the issue is that we also call `broadcastSignedTransaction` [here](https://github.com/tallycash/extension/blob/main/background/services/internal-ethereum-provider/index.ts#L170) which is outside of the try/catch block on the `transactionConstructionSliceEmitter` message handler.  I wasn't aware of this other use but its not a problem on `main` since we set the snackbar based on an emit from `broadcastSignedTransaction` regardless of what context it is called in.

In light of this do you think it makes sense to go back to the event based pattern and simply add alerting for failed transactions as part of this enhancement?

Shadowfiend:
Yep, 100% agree and feel a little silly for having forgotten it 😅

0xDaedalus:
Closed in favor of #1250

PR #1238, opened 2022-03-23
📜 Fix janky moving "+ add wallet" switcher footer

### LINKED ISSUE #1185: Account selection - I can scroll passed Add wallet
### Discord Discussion Link

_No response_

Chrome

nothing much

### What did not work?

Everything worked, i just managed to scroll past the add wallet button to an empty field of green

image

### Version

_No response_

### Relevant log output

```shell
Version: v0.12.2-1-g53105dba (16 Mar 2022)
```

henryboldi:
~~Just a tiny one line change does the trick!~~  In all cases so far, either we've got something (typically search) fixed to the top of the menu content, so the content handles the scroll, or there's no scroll at all. Given that, and in the interest of solving this issue, we're explicitly stating that the responsibility of scroll is the slide up menu content's.

Screenshot,

accounts2

Other slide up menus remain working,
other

Closes #1185


0xDaedalus:
Found some undesirable behavior:

https://user-images.githubusercontent.com/94649004/159760160-671c6691-c08d-47f0-9b58-6764ceef17eb.mov

henryboldi:
Interesting 🤔

Shadowfiend:
Guessing the positioned parent used for the remove address dropdown is probably no longer the scrolling element.

It feels like the right fix here isn't for shared slide-ups not to allow scrolling, but for the layout of the particular slide-up to be adjusted, maybe? More code, but I think the outcome will probably be overall more stable and lead to fewer surprises later. IMO it makes sense for a general-purpose fixed-height container like the shared slide-up to handle child element overflow by default.

ui/components/AccountItem/AccountItemOptionsMenu.tsx
```diff
@@ -118,7 +121,7 @@ export default function AccountItemOptionsMenu({
           }
           .options {
             position: absolute;
-            right: 8;
+            right: 6;
```
Shadowfiend, line 124: Non-blocking: let's include a unit here.

ui/components/AccountItem/AccountItemOptionsMenu.tsx
```diff
@@ -128,6 +131,8 @@ export default function AccountItemOptionsMenu({
             justify-content: space-between;
             width: 212px;
             border-radius: 4px;
+            height: 52px;
+            margin-top: -46px;
```
Shadowfiend, line 135: Could this just be a `top: 0` or similar? Negative top margin on an absolutely-positioned element seems weird.

ui/components/Shared/SharedSlideUpMenu.tsx
```diff
@@ -84,7 +84,6 @@ export default function SharedSlideUpMenu(props: Props): ReactElement {
           .slide_up_menu {
             width: 100%;
             height: var(--menu-height);
-            overflow-y: auto;
```
Shadowfiend, line 87: Do we still want to keep the overflow-y: auto vs removing it entirely? Looks like it was added [to enable scrolling in the first place](https://github.com/tallycash/extension/commit/d2f0196241b4a1b11b849a48905288c5ea111cf0), but maybe something has changed.

Shadowfiend (commented):
Couple of notes, solution looks good :muscle: Want to confirm that we're still scroll-capable in slide-up menus in Chrome + FF tho.

ui/components/Shared/SharedSlideUpMenu.tsx
henryboldi: Great call out. I can't find a slide up menu that removing this impacts tbh — well, in terms of functionality given existing implementations of the slide up menu contents. I don't think that solution linked ended up solving the problem. In all cases so far, we've got something fixed to the top, so the content handles the scroll, or there's no scroll at all. I think `overflow-y` is safe to remove. _Maybe_ it will return at some point, but wouldn't bet on it considering the designs I see in Figma. That said, editing a shared component does heighten concern here. And I don't want to lean too heavy on QA 😬

ui/components/AccountItem/AccountItemOptionsMenu.tsx
henryboldi: Updated ✅

ui/components/AccountItem/AccountItemOptionsMenu.tsx
henryboldi: Yes that's wayyyy better 🙇

ui/components/Shared/SharedSlideUpMenu.tsx
Shadowfiend: Testing in FF just now, I think you need to explicitly set this to `overflow: hidden;` otherwise the parent scrolls before this does, leaving the original issue of an unpinned bottom bar in place.

Shadowfiend:
Did not get to test this before wrapping up yesterday. @0xDaedalus if you get a chance would love to land it, otherwise I'll have a look Monday.

0xDaedalus (approved):
Its beautiful 😍

PR #1239, opened 2022-03-23
💖 Improve wallet asset cards

### LINKED ISSUE #1074: Asset Card
Asset Card currently looks like this
image

- [ ] Swap icon doesn't get brighter on mouse over. Loosing visibility

- [ ] The send is not a button, although it should be

image

https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=7301%3A178747

### LINKED ISSUE #1075: Asset Card icons
We seem to be missing swap on the asset card icons:

image

Creating a different ticket for this, as it might not be that easy to implement. Given that not all assets can be traded.

henryboldi:
Per Vlad's review this PR:
- [x] Adds a swap button on the assets which directs users to the swap page with the asset preselected
swap1
swap2

- [x] Changes the color of the icons to be brighter while hovering the asset list item
cards_hover
 
Closes #1075, closes #1074


henryboldi:
>  then going back to the wallet page and clicking on another asset does not pre-select the new asset in the swap (the old asset sticks).

Not seeing this issue on my end 😬

0xDaedalus:
> Not seeing this issue on my end 😬
Roger that - this is how I've been consistently able to reproduce:

- Clean Install
- Import wallet
- Click on swap asset from wallet page
- Choose a buy asset and enter a quantity but don't swap.
- Go back to wallet page, choose a different asset to swap (the previous asset will stick).

https://user-images.githubusercontent.com/94649004/159918210-6c78fe11-0c3f-45c4-a75f-20d07b80d8a7.mov

Confirming that you can now edit the pre-selected asset but the above issue still exists in this latest update.

henryboldi:
@0xDaedalus, thank you for the reproduction steps! super helpful ☺️

I believe it's fixed now.

ui/pages/Swap.tsx
```diff
@@ -119,7 +119,8 @@ export default function Swap(): ReactElement {
   const {
     assets: { sellAsset: savedSellAsset, buyAsset: savedBuyAsset },
     amount: savedSwapAmount,
-  } = savedQuoteRequest ?? {
+  } = (!locationAsset && savedQuoteRequest) || {
+    // ^ If coming from an asset item swap button, let the UI start fresh
```
Shadowfiend, line 123: Feels like this would be more explicit if we put it in the `savedQuoteRequest` assignment:

```typescript
    const savedQuoteRequest =
      typeof locationAsset !== "undefined" ?
        // If an asset is being forced by location information,
        // invalidate any saved quote and start fresh.
        undefined :
        useBackgroundSelector(selectLatestQuoteRequest)
```

henryboldi: React hooks shouldn't be called conditionally 😬

`React Hook "useBackgroundSelector" is called conditionally. React Hooks must be called in the exact same order in every component render.eslint[react-hooks/rules-of-hooks](https://reactjs.org/docs/hooks-rules.html)`

Shadowfiend: Yepppp gp gp.

Shadowfiend:
@0xDaedalus just a heads up that this is blocked on your requested changes :)

PR #1244, opened 2022-03-24
Change "Execute Swap" to "Confirm Swap"

0xDaedalus:
Having "Execute Swap" be the verbiage for approving was causing confusion when users were presented with an additional signing page after clicking "Execute Swap".


henryboldi (approved):
👨‍🍳😋

PR #1245, opened 2022-03-24
ADSR(elease): Add automated release artifacts for tags

Shadowfiend:
> [The most common kind of envelope generator has four stages: attack, decay, sustain, and release (ADSR).](https://en.wikipedia.org/wiki/Envelope_(music)#ADSR)
> 
> [- **Attack** is the time taken for initial run-up of level from nil to peak, beginning when the key is pressed.](https://en.wikipedia.org/wiki/Envelope_(music)#ADSR)
> [- **Decay** is the time taken for the subsequent run down from the attack level to the designated sustain level.](https://en.wikipedia.org/wiki/Envelope_(music)#ADSR)
> [- **Sustain** is the level during the main sequence of the sound's duration, until the key is released.](https://en.wikipedia.org/wiki/Envelope_(music)#ADSR)
> [- **Release** is the time taken for the level to decay from the sustain level to zero after the key is released.](https://en.wikipedia.org/wiki/Envelope_(music)#ADSR)

This commit adds builds for tag pushes. Instead of uploading a single
build ZIP to the actions result, it:
- Creates a new draft GitHub release with GitHub's own automatically-
  generated release notes.
- Marks the release as a prerelease if it includes `-pre` or starts with
  `v0`.
- Attaches each release ZIP to the release as an artifact.


mhluongo:
I like this approach a lot! It does seem like it [deserved a README update[(https://github.com/tallycash/extension/blob/780d6d4eb817d05300a8458e534a10d1fa12e4ca/README.md#releasing-a-version) though.

Shadowfiend:
🤔 this doesn't change anything about what's enumerated in the README, just makes automatic some stuff that was still manual. What would you like to see listed?

PR #1247, opened 2022-03-24
Return hexadecimal strings in response to `eth_chainId`.

### LINKED ISSUE #1242: OpenSea thinks the extension on mainnet is on the wrong network

0xDaedalus:
Closes #1242

According to [EIP-695](https://eips.ethereum.org/EIPS/eip-695), `eth_chainId` should return a hexadecimal string representing the chain id number (we were previously returning a `number`).  `net_version` should return an integer but in general `eth_chainId` is generally preferred over `net_version`.


background/services/internal-ethereum-provider/index.ts
```diff
@@ -117,7 +118,9 @@ export default class InternalEthereumProviderService extends BaseService<Events>
           typedData: JSON.parse(params[1] as string),
         })
       case "eth_chainId":
-        return this.chainService.ethereumNetwork.chainID
+        return ethers.utils.hexlify(
+          Number(this.chainService.ethereumNetwork.chainID)
+        )
```
Shadowfiend, line 123: Best to skip the external library and the possible precision loss IMO. `BigInt` + `toString(16)` does the same with minimal external oddity potential.

Shadowfiend (approved):
Wondering if the EVMNetwork chain if should be hex actually, but not blocking. Can't test rn but this looks good 👌

mhluongo (approved):
Tested and working great! :ship:

PR #1248, opened 2022-03-25
📏🤏 Adjust height of select dropdown to imply more options

### LINKED ISSUE #1183: Derivation Path - fix size and position of dropdown
image

Was thinking something like this, what i'm trying to fix here is that some users said it's not clear enough that there are more options/can add custom derivation path.

So instead of a scroll-bar maybe hiding half of an item is good enough to let users know there is more to be seen.

I've set this to 140px and it seems to work nicely.

We can revisit moving custom path button to the top later on

henryboldi:
Adjusts the height of select dropdown to imply more options,
dropdown

The solution is by @VladUXUI, so I've added him as the co-author on the commit :)

Closes #1183


Shadowfiend:
> The solution is by @VladUXUI, so I've added him as the co-author on the commit :)

I love this and am now considering how we can implement broadly 💪

mhluongo (approved):
:rocket:

PR #1250, opened 2022-03-25
Show "Transaction failed to broadcast." on failed send

0xDaedalus:
This PR adds a `Transaction failed to broadcast.` snackbar notification in the case that there is an error after signing but before sending a transaction.

### Test Criteria
- [ ] A successful transaction should result in a snackbar message of `Transaction signed, broadcasting...`
- [ ] An unsuccessful transaction should result in a snackbar message of `Transaction failed to broadcast.`

The easiest way to simulate an unsuccessful transaction is to edit the code and throw an error inside of `broadcastSignedTransaction`

This PR addresses [an excellent suggestion](https://github.com/tallycash/extension/pull/1201#discussion_r831486778) made in #1201


background/services/chain/index.ts
```diff
@@ -83,7 +83,8 @@ const TRANSACTION_CHECK_LIFETIME_MS = 10 * HOUR
 interface Events extends ServiceLifecycleEvents {
   newAccountToTrack: AddressOnNetwork
   accountBalance: AccountBalance
-  transactionSent: HexString
+  transactionSend: HexString
+  transactionSendFailure: undefined
```
Shadowfiend, line 87: Let's include the transaction and error here, even if we don't currently consume them.

Shadowfiend (approved):
Teensy note, but not blocking. Will test drive.

PR #1252, opened 2022-03-25
Replace `webext-redux` using `window.postMessage` and Channel Messaging API

cairomassimo:


mhluongo:
We've bailed on this spike, closing it out!

PR #1257, opened 2022-03-27
#1227, shared input parser prop function

### LINKED ISSUE #1227: Add parseAndValidate function prop to SharedInput
Issue from discussion #1204

### What
Add `parseAndValidateFunction` prop to SharedInput, of type `(value: string) => T | { error: string }`. 
If the function returns `{ error: string }` the shared input would mark an error. If the function returns a `T`, the shared input would emit `onChange` the way it currently does, but with a `T` instead of a string.

Use new prop with gas limit input

### Why
This lets us handled parsed values, and would let us remove some extra state handling and weird behaviors we're doing in other places like SharedAssetInput as well.

alleycat0001:
Closes #1227

Additional discussion: https://github.com/tallycash/extension/pull/1204#issuecomment-1075235712

Started with the suggested `parseAndValidationFunction`, but made some modifications after getting into the weeds.

**Made `SharedInput` generic**

`SharedInput<T = string>`. Defaulting to `string` so that the component is fully backwards compatible with existing uses as the output of `<input />` is a `string`. This was done because the `parser` and `onChange` props need a type to "attach" on to, or otherwise infer from.

**Deviated from suggested parsing function signature**

Suggested: `parseAndValidationFunction: (value: string) => { error: string } | T`
Implemented: `parser: (value: string) => { state: 'error', message: string } | { state: 'parsed', parsed: T }`.

-  `parser` is just a bit more succinct than `parseAndValidationFunction` :)
- The return type `{ error: string } | T` has issues where if `T` includes `null` or `undefined` you'd get `if (null.error)` which would give you an error. You can do some work arounds but ultimately a proper tagged union just makes it cleaner to work with.
- Additionally It also allows for other parse states, like `warning`, e.g. "That gas limit is really high Dog!". If this happens then the `if/else` turns into a nice clean `switch`.
- I explored `parseOrThrow` where the user returns `T` or throws. Some other form component libs do this pattern, but meh, the above works really nicely.

If we ever truly want OS style inputs as [mentioned](https://github.com/tallycash/extension/pull/1204#issuecomment-1075235712), we will need to build a `<Form />` component. Reason being is you'll need to centralize the logic of parsing/validating as well as controlling the `onSubmit` so the user can enter any value but not submit the form.


Shadowfiend:
I'll have a deeper look at the PR tomorrow, but wanted to articulate a little more of the thinking behind the original suggestions without necessarily implying any changes need to happen in this particular change set:

> parser is just a bit more succinct than parseAndValidationFunction

More succinct, but less clear! The intent was to have this function validate even if there's no parsing to be done, and to do post-parsing validation when relevant (eg min-max after a bigint parse)---something non-obvious if the name only references parsing. 

The reason to have one for both is that it allows central handling of validation, whether or not it's parsing-related, vs having to add control flow to split the two. Splitting would also potentially require three phases: pre parse validation, parsing and errors from parsing, and post parsing validation. Instead, one function can handle any custom control flow the caller wants, but the name requires a bit more clarity. 

> If we ever truly want OS style inputs as mentioned, we will need to build a <Form /> component. Reason being is you'll need to centralize the logic of parsing/validating as well as controlling the onSubmit so the user can enter any value but not submit the form.

Not necessarily; validation errors can be set via the [constraint validation API](https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation) and automatically fail form submissions without additional handling.

Shadowfiend:
... Also thanks for tackling this! ❤️

alleycat0001:
> More succinct, but less clear! The intent was to have this function validate even if there's no parsing to be done, and to do post-parsing validation when relevant (eg min-max after a bigint parse)---something non-obvious if the name only references parsing.

That's fair. I'm not strongly opinionated on the naming and can certainly update.

> Not necessarily; validation errors can be set via the [constraint validation API](https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation) and automatically fail form submissions without additional handling.

Oh nice. Suppose I'll find some time to go and review the available Web APIs to see what else I've missed :)

Shadowfiend:
Didn't make it here today, but it's on my list for tomorrow!

alleycat0001:
No problem. I totally understand the pre-launch hustle :)

ui/components/Shared/SharedInput.tsx
```diff
@@ -49,6 +59,22 @@ export default function SharedInput(props: Props): ReactElement {
...
+      if (result.state === "error") {
+        setParserError(result.message)
+      } else {
+        setParserError(null)
+        onChange?.(result.parsed)
+      }
+    } else {
+      onChange?.(value as unknown as T)
+    }
+  }
```
Shadowfiend, line 76: What do you think about making the split `{ error: string } | { parsed: T }`? This feels less noisy in the parser function, but still gives us a good way to discriminate at the top level. It's also clear in the handling code what path is doing what:

```suggestion
  const onInputChange: ChangeEventHandler<HTMLInputElement> = ({
    target: { value: inputValue },
  }) => {
    if (parser) {
      const result = parser(inputValue)
      if ("error" in result) {
        setParserError(result.error)
      } else {
        setParserError(null)
        onChange?.(result.parsed)
      }
    } else {
      onChange?.(value as unknown as T)
    }
  }
```

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -249,11 +231,29 @@ export default function NetworkSettingsSelect({
...
       <div className="info">
         <div className="limit">
-          <SharedInput
+          <SharedInput<bigint | undefined>
             id="gasLimit"
             value={networkSettings.gasLimit?.toString() ?? ""}
             placeholder={networkSettings.suggestedGasLimit?.toString() ?? ""}
             onChange={setGasLimit}
+            parser={(value) => {
+              try {
```
Shadowfiend, line 240: I would limit the `try` to the handling that we expect will produce the error we're looking for. That is, I would exclude the handling of empty strings from the `try`, since if that throws it's not going to be because the gas limit isn't a number.

ui/components/Shared/SharedInput.tsx
```diff
@@ -31,9 +40,10 @@ export default function SharedInput(props: Props): ReactElement {
     errorMessage,
     autoFocus = false,
     autoSelect = false,
+    parser,
   } = props
-
   const inputRef = useRef<HTMLInputElement | null>(null)
+  const [parserError, setParserError] = useState<string | null>(null)
```
Shadowfiend, line 46: Not really a huge deal, but we generally use `undefined` instead of `null` for most empty states.

Shadowfiend (dismissed):
Left a few notes. I don't think any of them is particularly blocking, but will hold on merging for a little bit in case you want to address here.

ui/components/Shared/SharedInput.tsx
alleycat0001: No problem, don't want to go against the grain. Updated ✔️

ui/components/NetworkFees/NetworkSettingsSelect.tsx
alleycat0001: Fair enough 👍 . Updated ✔️

ui/components/Shared/SharedInput.tsx
```diff
@@ -1,23 +1,32 @@
...
   value?: string | number | undefined
-  onChange?: (value: string) => void
+  onChange?: (value: T) => void
   onFocus?: () => void
   errorMessage?: string
   autoFocus?: boolean
   autoSelect?: boolean
+  parser?: (
+    value: string
+  ) => { state: "error"; message: string } | { state: "parsed"; parsed: T }
```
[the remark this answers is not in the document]
alleycat0001: Renamed to `parseAndValidate` 👍 

> I would flip this to require and then set a default prop that sets it to a simple string passthrough. This will clean up the handler function below since we'll always be able to rely on the parser existing.

❤️ . Done ✔️

ui/components/Shared/SharedInput.tsx
alleycat0001: Oh nice, the `"error" in result` gives TypeScript enough info to know `result.parsed` exists as T in the `else`. I did not consider that. I dig it. We can still add a `{ warning: string }` to the union if we want to expand out 👍 

✔️

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -249,11 +231,29 @@ export default function NetworkSettingsSelect({
       })}
       <div className="info">
         <div className="limit">
-          <SharedInput
+          <SharedInput<bigint | undefined>
```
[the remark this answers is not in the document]
Shadowfiend: This was me not being clear, sorry. I meant I think you should be able to leave off the explicit type parameter at the call site and have it inferred; it's definitely necessary in the `SharedInput` and `Props` definition.

ui/components/Shared/SharedInput.tsx
```diff
@@ -40,10 +38,10 @@ export default function SharedInput<T = string>(props: Props<T>): ReactElement {
     errorMessage,
     autoFocus = false,
     autoSelect = false,
-    parser,
+    parseAndValidate = (v) => ({ parsed: v as unknown as T }),
```
Shadowfiend, line 41: There's a problem here… We're allowing a `T` that is not a `string` to receive a default parser that doesn't actually produce a `T`, which breaks all of the type expectations of the rest of the component's props.

Unfortunately, I haven't been able to come up with a way to handle this without overloading `SharedInput` itself to have a simplified `string` path with optional parser/validator and a more complex non-`string` path with a required parser/validator.

As it is, we can do this:

```typescript
            <SharedInput
              label="ETH address or ENS name"
              onChange={(bi: bigint) => handleInputChange((bi + 2n).toString())}
              errorMessage={errorMessage}
            />
```

And no error will happen at compile-time, even though `onChange` will receive a `string`. We wouldn't even get an error at runtime, we'd just get bad behavior (bigint concatenation) :grimacing:

I feel like we may need to make `SharedInput` a shim over `SharedTypedInput<string>`, and define `SharedTypedInput` separately in this file as a non-default export with a required `parseAndValidate`.

Thoughts?

ui/components/NetworkFees/NetworkSettingsSelect.tsx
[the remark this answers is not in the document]
alleycat0001: No problem, I see what you mean. Looks like TS is smart enough to infer based on the `onChange` passed in and thus we can remove the explicit type parameters 👍 . Done ✔️

ui/components/Shared/SharedInput.tsx
```diff
@@ -70,6 +90,7 @@ export default function SharedInput(props: Props): ReactElement {
       />
       <label htmlFor={id}>{label}</label>
       {errorMessage && <div className="error_message">{errorMessage}</div>}
+      {parserError && <div className="error_message">{parserError}</div>}
```
Shadowfiend, line 93: We should probably migrate any external error providers to using the parser-validator approach in forthcoming work, or we should have a priority pipeline so that we only ever show one error. Right now our layouts aren't really set up to show more than one at a time.

Shadowfiend (approved):
At risk of regretting leaving typing inconsistencies in the codebase, I'm going to go ahead and merge this as-is and leave the solution for https://github.com/tallycash/extension/pull/1257#discussion_r837801917 as a follow-up. I think the win here is already significant, and the current usage of the updated parsing/validation strategy can be evolved alongside the strategy itself.

ui/components/Shared/SharedInput.tsx
alleycat0001: Hmmm. Yeah I really didn't like the necessity for `v as unknown as T` done before in `onInputChange`. It's definitely a code smell. It appeared like less of an issue there because it would only be used if no `parseAndValidate` prop was passed in and it took `value` directly from the browser event.

The default prop value for `parseAndValidate` is really nice but unfortunately it just moved the smell from one place to another.

I think your suggestion could certainly work and happy to explore that over the next couple days.

alleycat0001:
> At risk of regretting leaving typing inconsistencies in the codebase, I'm going to go ahead and merge this as-is and leave the solution for [#1257 (comment)](https://github.com/tallycash/extension/pull/1257#discussion_r837801917) as a follow-up. I think the win here is already significant, and the current usage of the updated parsing/validation strategy can be evolved alongside the strategy itself.

Sounds good. I'll work on that follow up at some point over the next couple days.

Issue #1209, opened 2022-03-21
Extension blank screens after auto update to 0.12.2


Naxsun:
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/954850856690405377

### What browsers are you seeing the problem on?

Brave

### What were you trying to do?

User: version 12.2, auto-updated in Brave. Tally window is completely black.

### What did not work?

Wallet gives blank screen:
![image](https://user-images.githubusercontent.com/71384922/159245743-6a503190-89f0-4882-9a62-07e3ffd86eb3.png)

"I tried restarting Brave, disabling/re-enabling Tally extension. Nothing seems to work."

Logs screenshot
![image](https://user-images.githubusercontent.com/71384922/159245722-bda5c973-174e-4aba-8627-04e213c215c2.png)

### Version

Community Edition v0.12.2

VladUXUI:
Happed to me as well, but maybe not exactly the same, but on Chrome.
I clicked refresh after a while and the screen was just stuck in blank screen until i closed.
I don't have logs unfortunately :(

<img width="394" alt="image" src="https://user-images.githubusercontent.com/80846886/159650793-8ed26bdc-1e0c-4794-baf8-5894a2d00d80.png">

henryboldi:
I haven't seen any more reports of this issue since https://github.com/tallycash/extension/pull/1258. Closing for now.

mhluongo:
Here's a PR for this. It's been going on since the Ledger integration

https://github.com/tallycash/extension/pull/1327

henryboldi:
My bad. Thank you!

henryboldi:
Logs were posted in #bug-reports 😁

<img width="603" alt="log" src="https://user-images.githubusercontent.com/1918798/164056533-4740240e-7612-4abf-90ad-3daf49593e76.png">

PR #1258, opened 2022-03-28
🤷‍♂️😬 Possible blank page on upgrade fixes

henryboldi:
This PR primarily adds a migration for `arbitraryDataEnabled`. `arbitraryDataEnabled` is the only new value I noticed in the Redux store since the appearance of the rare blank screen on upgrade bug. That said, I haven't been able to reproduce it. 

> I often found myself back in the onboarding page during web3-react PR dev - Greg

I've also added a guard to the onboarding to doubly prevent displaying it if an account has already been created. Granted, perhaps it'd be cleaner to place these guards/page permissions in `routes.tsx`.

#1209


greg-nagy (approved):
makes sense, code works, let's 🚢

PR #1165, opened 2022-03-15
RFB 2: UI Structure for Heterogeneous Signer Support

Shadowfiend:
Quoting from the proposal section:

> This RFB focuses on a scalable approach to architecting the UI flow for
signing, irrespective of whether it is transaction or message data being
signed. It does not speak to the interaction between transaction and message
data, nor to the handling of multiple pending signing requests (something that
has been informally referred to as "transaction queueing"), which is treated as
an orthogonal concern[.]

PR #1262, opened 2022-03-28
Modalicious: Make transaction signing modal

Shadowfiend:
Currently this is achieved not by overlaying signing, but by
replacing the location when a signature is in progress. Overlaying
signing proved too much for an initial pass due to how complex
the CSS transition logic is and how entangled it is with the page
structure.

Note also that the keyring handling here will likely be dropped in
favor of a signing-flow-specific approach via #1165 .

A few linter fixes and a small logger improvement are also in here.

## Testing

Try sending ETH, swapping, and using dApp transaction signing. All
should behave correctly.

### DIFF
--- a/background/redux-slices/transaction-construction.ts
+++ b/background/redux-slices/transaction-construction.ts
@@ -313,6 +313,12 @@ export const selectTransactionData = createSelector(
   (transactionRequestData) => transactionRequestData
 )
 
+export const selectIsTransactionPendingSignature = createSelector(
+  (state: { transactionConstruction: TransactionConstruction }) =>
+    state.transactionConstruction.status,
+  (status) => status === "loaded" || status === "pending"
+)
+
 export const selectIsTransactionLoaded = createSelector(
   (state: { transactionConstruction: TransactionConstruction }) =>
     state.transactionConstruction.status,
--- a/background/redux-slices/transaction-construction.ts
+++ b/background/redux-slices/transaction-construction.ts
@@ -8,6 +8,7 @@ import {
   REGULAR,
 } from "../constants/network-fees"
 import { USE_MAINNET_FORK } from "../features/features"
+import logger from "../lib/logger"
 
 import {
   BlockEstimate,
@@ -47,10 +48,7 @@ export enum NetworkFeeTypeChosen {
 }
 export type TransactionConstruction = {
   status: TransactionConstructionStatus
-  // @TODO Check if this can still be both types
-  transactionRequest?:
-    | EIP1559TransactionRequest
-    | EnrichedEIP1559TransactionRequest
+  transactionRequest?: EnrichedEIP1559TransactionRequest
   signedTransaction?: SignedEVMTransaction
   broadcastOnSign?: boolean
   transactionLikelyFails?: boolean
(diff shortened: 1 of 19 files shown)


background/redux-slices/transaction-construction.ts
```diff
@@ -8,6 +8,7 @@ import {
   REGULAR,
 } from "../constants/network-fees"
 import { USE_MAINNET_FORK } from "../features/features"
+import logger from "../lib/logger"
```
0xDaedalus, line 11: lets rm

background/redux-slices/transaction-construction.ts
```diff
@@ -47,10 +48,7 @@ export enum NetworkFeeTypeChosen {
 }
 export type TransactionConstruction = {
   status: TransactionConstructionStatus
-  // @TODO Check if this can still be both types
```
0xDaedalus, line 50: 📈 👍

background/redux-slices/transaction-construction.ts
```diff
@@ -252,6 +243,28 @@ export const {
...
+export const transactionSigned = createBackgroundAsyncThunk(
+  "transaction-construction/transaction-signed",
+  async (transaction: SignedEVMTransaction, { dispatch, getState }) => {
+    await dispatch(signed(transaction))
+
+    const { transactionConstruction } = getState() as {
+      transactionConstruction: TransactionConstruction
+    }
+
+    if (transactionConstruction.broadcastOnSign ?? false) {
```
0xDaedalus, line 262: Why do we need `?? false` here?

ui/components/Swap/SwapQuote.tsx
```diff
@@ -51,11 +51,7 @@ export default function SwapQuote({
   const handleConfirmClick = useCallback(async () => {
     const { gasPrice, ...quoteWithoutGasPrice } = finalQuote
 
-    // FIXME Set state to pending so SignTransaction doesn't redirect back; drop after
-    // FIXME proper transaction queueing is in effect.
-    await dispatch(clearTransactionState(TransactionConstructionStatus.Pending))
-
-    dispatch(
+    await dispatch(
```
0xDaedalus, line 54: we can rm `TransactionConstructionStatus` and `clearTransactionState` in this file

ui/pages/Swap.tsx
```diff
@@ -243,18 +243,12 @@ export default function Swap(): ReactElement {
       return
     }
 
-    // FIXME Set state to pending so SignTransaction doesn't redirect back; drop after
-    // FIXME proper transaction queueing is in effect.
-    await dispatch(clearTransactionState(TransactionConstructionStatus.Pending))
-
-    dispatch(
+    await dispatch(
```
0xDaedalus, line 246: unused imports in this file

0xDaedalus (approved):
Tested and LGTM outside of some non-blocking stuff.  Will merge later this evening

background/redux-slices/transaction-construction.ts
Shadowfiend: Explicit over implicit---undefined we should migrate away but in the meantime this explicitly makes it clear if the value is unset it's assumed false.

Shadowfiend:
I'll slap the unused imports in the follow up later tonight. For shame < hangs head >.

PR #1266, opened 2022-03-29
↕️ Sort overview tab's assets by total currency amount

henryboldi:
This PR sorts the overview tab's assets by total currency amount. The idea is to show the assets that users care most about first. The code adds a lil sort on `assets` before `.map` returns the rows. Test this by checking out the overview tab!

sorted_overview


ui/components/Overview/OverviewAssetsTable.tsx
```diff
@@ -22,47 +22,60 @@ export default function OverviewAssetsTable(props: Props): ReactElement {
...
-                  {asset.localizedUnitPrice}
+        {assets
+          .sort((a: CompleteAssetAmount, b: CompleteAssetAmount) => {
+            if (
+              a.localizedMainCurrencyAmount &&
+              b.localizedMainCurrencyAmount
+            ) {
+              return (
+                Number(b.localizedMainCurrencyAmount) -
+                Number(a.localizedMainCurrencyAmount)
```
Shadowfiend, line 33: Unfortunately there is no guarantee this will parse correctly; however, enriched assets include the number amount as well:

https://github.com/tallycash/extension/blob/70243b05764a7b6114fb0f7c1284be5e5c6ef8c9/background/redux-slices/utils/asset-utils.ts#L21

henryboldi: Updated 🙏

ui/components/Overview/OverviewAssetsTable.tsx
```diff
@@ -22,47 +22,54 @@ export default function OverviewAssetsTable(props: Props): ReactElement {
...
-              {asset.localizedUnitPrice ? (
-                <div>
-                  <span className="lighter_color">$</span>
-                  {asset.localizedUnitPrice}
+        {assets
+          .sort((a: CompleteAssetAmount, b: CompleteAssetAmount) => {
+            if (a.mainCurrencyAmount && b.mainCurrencyAmount) {
+              return b.mainCurrencyAmount - a.mainCurrencyAmount
+            }
+            return 0
```
Shadowfiend, line 30: Couple of questions:
- if either asset doesn't have a main currency amount (say, because we haven't been able to resolve its price yet), should we be enforcing some ordering? Similarly, if two assets have the same main currency amount, should we be enforcing ordering? It seems like we should maybe fall back on alphabetical (e.g. `a.symbol.localeCompare(b.symbol)` or so).
- Do we need the type annotations on a and b? `assets` should already be typed correctly, right?

henryboldi: > if either asset doesn't have a main currency amount (say, because we haven't been able to resolve its price yet), should we be enforcing some ordering? 

Meaning a fallback ordering? Like, alphabetical?

<br/>


> Do we need the type annotations

Removed 🙏

henryboldi:
Assuming that not having a fallback ordering is blocking?

Shadowfiend:
Not critical, just didn't get a chance to circle back. Anyone else is welcome to finalize review + merge here if they beat me to it!

If you get a chance to add the fallback beforehand, bonus points.

ui/components/Overview/OverviewAssetsTable.tsx
```diff
@@ -12,6 +12,27 @@ export default function OverviewAssetsTable(props: Props): ReactElement {
...
+      return 1
+    }
+
+    // Alphabetical fallback
+    if (a.asset.symbol > b.asset.symbol) {
+      return -1
+    }
+    if (a.asset.symbol < b.asset.symbol) {
+      return 1
+    }
```
Shadowfiend, line 31: `localeCompare` will do the right thing here, otherwise the alphabetical sort won't make sense in some places.

Shadowfiend (approved):
Giving this a test drive now. I feel bad that I took so long but happy that we got the alphabetical bit in :grimacing:

PR #1269, opened 2022-03-29
#1227, shared typed input shim

alleycat0001:
Following up on PR #1257, specifically the discussion [here](https://github.com/tallycash/extension/pull/1257#discussion_r837801917).

This updates `SharedInput.tsx` such that we now have a default export of `<SharedInput />` that is essentially a `<SharedTypedInput<string> />` and an additional export of `<SharedTypedInput<T> />`.

The purpose of this was to make `parseAndValidate` a required prop, avoiding a case where type expectations could be broken.

### DIFF
--- a/ui/components/Shared/SharedInput.tsx
+++ b/ui/components/Shared/SharedInput.tsx
@@ -21,10 +21,10 @@ interface Props<T> {
   errorMessage?: string
   autoFocus?: boolean
   autoSelect?: boolean
-  parseAndValidate?: (value: string) => { parsed: T } | { error: string }
+  parseAndValidate: (value: string) => { parsed: T } | { error: string }
 }
 
-export default function SharedInput<T = string>(props: Props<T>): ReactElement {
+export function SharedTypedInput<T = string>(props: Props<T>): ReactElement {
   const {
     id,
     label,
@@ -38,7 +38,7 @@ export default function SharedInput<T = string>(props: Props<T>): ReactElement {
     errorMessage,
     autoFocus = false,
     autoSelect = false,
-    parseAndValidate = (v) => ({ parsed: v as unknown as T }),
+    parseAndValidate,
   } = props
   const inputRef = useRef<HTMLInputElement | null>(null)
   const [parserError, setParserError] = useState<string | undefined>(undefined)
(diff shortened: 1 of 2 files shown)


ui/components/Shared/SharedInput.tsx
```diff
@@ -153,7 +153,20 @@ export default function SharedInput<T = string>(props: Props<T>): ReactElement {
   )
 }
 
-SharedInput.defaultProps = {
+const SHARED_INPUT_DEFAULT_PROPS = {
```
alleycat0001, line 156: Ensures this catches the case where `SharedTypedInput` is used directly is covered and the default props of it and its shim are in sync.

ui/components/Shared/SharedInput.tsx
```diff
@@ -153,7 +153,20 @@ export default function SharedInput<T = string>(props: Props<T>): ReactElement {
...
   type: "text",
   focusedLabelBackgroundColor: "var(--hunter-green)",
 }
+
+SharedTypedInput.defaultProps = {
+  ...SHARED_INPUT_DEFAULT_PROPS,
+}
+
+export default function SharedInput(props: Props<string>): ReactElement {
```
alleycat0001, line 165: Adding the return type here might be unnecessary but looking at other areas of code it looks like we tend to be pretty explicit.

ui/components/Shared/SharedInput.tsx
```diff
@@ -153,7 +153,20 @@ export default function SharedInput<T = string>(props: Props<T>): ReactElement {
...
+  return SharedTypedInput(props)
+}
+
+SharedInput.defaultProps = {
+  ...SHARED_INPUT_DEFAULT_PROPS,
```
Shadowfiend, line 170: This could just as easily be `...SharedTypedInput.defaultProps` and avoid the extra definition right? Not really a huge deal, just curious.

Shadowfiend (commented):
Amaze. Should review and merge later tonight hopefully.

alleycat0001:
Yeah something was annoying me about this at the dinner table after submitting this. I’ll update!

ui/components/Shared/SharedInput.tsx
Shadowfiend: Our lint config (typescript-airbnb) requires types on exported functions, so this was the move!

ui/components/Shared/SharedInput.tsx
alleycat0001: Updated ✔️

greg-nagy (approved):
generic typing ... 🤤

Ty

PR #1270, opened 2022-03-29
❤️‍🩹 Remedy route persistence complications

henryboldi:
`initialEntries` needs to be a reversed version of route history entities.
Without avoiding the initial load, entries will keep reversing. Given that
restoring our route history is a "POP" `history.action`, by specifying "PUSH"
we know that the most recent navigation change is by the user or explicitly
added.

That said, we can still certainly "POP" via history.goBack(). This case
is not yet accounted for. Because alternatively, comparing history lengths runs
into edge cases. For whatever reason, the first navigation change after
restoring the location always resets the navigation entities length to two. The
resetting entities length is something I'm looking into.

On the bright side, `history.goBack()` is not used on any page we persist.

### DIFF
--- a/ui/pages/Popup.tsx
+++ b/ui/pages/Popup.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useState, useRef, useEffect } from "react"
+import React, { ReactElement, useState, useEffect } from "react"
 import { MemoryRouter as Router, Switch, Route } from "react-router-dom"
 import { ErrorBoundary } from "react-error-boundary"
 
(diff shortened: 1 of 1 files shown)


Shadowfiend:
We're starting to dig pretty deep into the innards of react-router, and they're innards that change in the next major version, which makes me a bit uncomfortable.

I'm curious, can we just persist the latest entry instead of trying to persist the full history? It feels like this should be sufficient, but it's very possible I'm missing something. We could clear the entry if the user is on a page that isn't marked for persistence, and store it if they are on a page that is. This would eliminate the JSON serialization comparison, since we'd be dealing in strings, and feels like it would also be closer to the intended functionality.

Again, very possible I'm missing a sub-use case here, but if this is sufficient then we could maybe base this almost entirely on `transformedLocation`?

henryboldi:
> I'm curious, can we just persist the latest entry instead of trying to persist the full history?

Persisting pages with a back button wouldn't work. There'd be no history to go back to

Shadowfiend:
Ah, don't know that I knew we did that. Makes sense!

PR #1273, opened 2022-03-30
👻 False alarm: Remove brief error state from dapp connect

greg-nagy:
waiting for the dispatch to go through resulted in a
temporary state where the actual permission has already
been cleared up, but the window was not yet closed
resulting in a protective clause to briefly show the error state
which was confusing

https://discord.com/channels/808358975287722045/888500174685614090/958726241924554823

CleanShot 2022-03-30 at 16 37 04


greg-nagy:
Wanted to make sure that the whole back-and-forth process goes through without a problem. 
Since the action closes the window I was afraid that it could terminate and action propagation midflight resulting in a failed or inconsistent state.

But tbh I was probably too paranoid here, that's why it felt safe to remove now, and couldn't produce any issue when testing

PR #1275, opened 2022-03-30
Release v0.12.4 QA/Bump - SmokeTest

karlmoll:
Karl's first DevBuild!

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG.yml
+++ b/.github/ISSUE_TEMPLATE/BUG.yml
@@ -51,6 +51,7 @@ body:
       label: Version
       description: What version of the extension are you running?
       options:
+        - Community Edition v0.12.4
         - Community Edition v0.12.3
         - Community Edition v0.12.2
         - Community Edition v0.12.1
--- a/manifest/manifest.json
+++ b/manifest/manifest.json
@@ -1,6 +1,6 @@
 {
   "name": "Tally Ho",
-  "version": "0.12.3",
+  "version": "0.12.4",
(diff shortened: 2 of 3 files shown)


Shadowfiend:
I *believe* you missed a --tags on the git push here.

karlmoll:
> I _believe_ you missed a --tags on the git push here.

@Shadowfiend i definitely didn't explicitly add it. does the `yarn version --patch` handle the tag'ing?

karlmoll:
and should i delete this branch and re-push with `--tags`?

Shadowfiend:
I believe the version script does everything except push. No need to delete, just `git push --tags` should sync up the tag (and set up the draft release as a result).

karlmoll:
@Shadowfiend for some reason `yarn version --patch` is not updating my tags. 

@henryboldi and I saw some other weird behavior with patch, but I'll need to dig into this in the morning, I think

karlmoll:
ok, so when i cloned the repo again and ran the version patch, it set the tags correctly but it deleted the manifest.json

https://github.com/tallycash/extension/commit/e1a796969789e9fedeb5269cef65c97c74286112

karlmoll:
[actually when i look at the tags on my machine, i'm seeing `v0.12.3-195-gb222a27d` - @henryboldi and i saw some issues with the version bump script, but were trying to bypass it for now. maybe i need to dig deeper here. ](https://github.com/tallycash/extension/issues/1278)

greg-nagy:
Not sure if that helps but to me `yarn version` was the thing I needed. It asked me for the new version number and worked well.

Issue #1276, opened 2022-03-30
Swaps / High CPU usage when closing "Swap To" <SharedSlideUpMenu />


alleycat0001:
### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

Swapping tokens

### What did not work?

Swaps work. However, after selecting the `To` asset the extension freezes momentarily. Each of the spikes below are immediately after closing the asset chooser, specifically `<SharedSlideUpMenu />`.

I did not spend too much time digging but `<SharedSlideUpMenu />` does not virtually render its items. So on close that's a lot of DOM elements that need to be dumped.

![image](https://user-images.githubusercontent.com/99354130/160937942-7617fcd8-df3b-47eb-a034-401e2d729465.png)

Issue #1260, opened 2022-03-28
The `To` and `From` assets in the swap menu should be mutually exclusive.


0xDaedalus:
Users should not be able to select swaps that swap an asset for itself.

<img width="376" alt="image" src="https://user-images.githubusercontent.com/94649004/160425639-43275cb8-8a9b-4011-a2e2-6f6c51ad7c59.png">

mhluongo:
It's yours! :trophy:

mhluongo:
Closed by #1277

PR #1277, opened 2022-03-30
#1260, fix issue where to and from swap assets can be the same

alleycat0001:
Removes the ability for a user to select the same asset as the To and From.

The diff is a little bit big because I needed to move `buyAssets` down so it could reference `sellAsset`. I've noted below via comments what the real changes are.

As an aside, during testing I've found another issue and created #1276

### DIFF
--- a/ui/pages/Swap.tsx
+++ b/ui/pages/Swap.tsx
@@ -85,19 +85,6 @@ export default function Swap(): ReactElement {
         assetAmount.asset.symbol === "ETH"
     ) ?? []
 
-  const buyAssets = useBackgroundSelector((state) => {
-    // Some type massaging needed to remind TypeScript how these types fit
-    // together.
-    const knownAssets: AnyAsset[] = state.assets
-    return knownAssets.filter(
-      (asset): asset is SmartContractFungibleAsset | FungibleAsset =>
-        isSmartContractFungibleAsset(asset) ||
-        // Explicity add ETH even though it is not an ERC-20 token
-        // @TODO change as part of multi-network refactor.
-        (isFungibleAsset(asset) && asset.symbol === "ETH")
-    )
-  })
-
   const {
     symbol: locationAssetSymbol,
     contractAddress: locationAssetContractAddress,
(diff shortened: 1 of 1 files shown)


ui/pages/Swap.tsx
```diff
@@ -143,23 +130,41 @@ export default function Swap(): ReactElement {
...
+    // Some type massaging needed to remind TypeScript how these types fit
+    // together.
+    const knownAssets: AnyAsset[] = state.assets
+    return knownAssets.filter(
+      (asset): asset is SmartContractFungibleAsset | FungibleAsset =>
+        (isSmartContractFungibleAsset(asset) ||
+          // Explicity add ETH even though it is not an ERC-20 token
+          // @TODO change as part of multi-network refactor.
+          (isFungibleAsset(asset) && asset.symbol === "ETH")) &&
+        asset.symbol !== sellAsset?.symbol
```
alleycat0001, line 143: Ensure buy assets doesn't include the selected sell asset.

ui/pages/Swap.tsx
```diff
@@ -143,23 +130,41 @@ export default function Swap(): ReactElement {
...
+                {
+                  asset: sellAsset,
+                  amount: 0n,
+                  decimalAmount: 0,
+                  localizedDecimalAmount: "0",
+                },
+              ]
+        )
+  ).filter(
+    (sellAssetAmount) => sellAssetAmount.asset.symbol !== buyAsset?.symbol
```
alleycat0001, line 166: Ensure the sell assets don't include the selected buy asset.

henryboldi: Not blocking but here's some thoughts in terms of UX,

On the sell side, it might be surprising not to see all of the assets you own for selling, even if you've selected that asset to buy. So ideally, for selling, we may want to gray out the already chosen buy asset, and then if the user clicks it, it flips from the buy side to the sell side. For an example, Uniswap does this.

cc @VladUXUI

henryboldi (approved):
Makes sense! Great work ✅ 👑

I left a comment about the UX, but it's not vital by any means.

ui/pages/Swap.tsx
alleycat0001: I had that in mind as well. Totally agree that would be a better UX. Do we have an example of disabled `<SharedSlideUpMenu />` items somewhere else I can borrow from? Happy to update.

henryboldi: I don't believe these a design for disabled `<SharedSlideUpMenu />` items yet. Setting opacity to 0.6 would probably do the trick

PR #1280, opened 2022-03-31
📦  The Bad Batch: Account balance event batching

greg-nagy:
https://www.imdb.com/title/tt12708542/

> The 'Bad Batch' of elite and experimental clones make their way through an ever-changing galaxy in the immediate aftermath of the Clone Wars.

# What? 

This PR uses the balance updates in batches instead of asset by asset.
This includes the indexedDB layer and also the UI/redux parts.

# Why?

Because it turns out bursting events cause serious slow down in almost 
every layer of our application and also the most probable cause for huge CPU usage

# How? 

Best to follow commit descriptions. 

# Test

## Basic functionality 

- remove and add the extension 
- import a seed phrase
- see the assets show up
- add a couple more addresses

## Performance

This one is trickier as I couldn't come up with exact measurements. 
- checkout extension on main 
- download patch file into the extension folder ([performance-debug-patch.txt](https://github.com/tallycash/extension/files/8391697/performance-debug-patch.txt))
- `git apply performance-debug-patch.txt`
  - this changes the balance refresh period to `0.1` and dispalays all the redux actions on the console 
  > `0.1` is not supported by the standard but seems to be working in chrome
  - you should see many `account/updateAccountBalance` events in the console 
  roughly `imported addresses * tracked asset no`
  - open the chrome task manager and watch the CPU usage of tally extension for a while
- checkout this branch
  - on the console only a few `account/updateAccountBalance` should be visible 
    around the number of imported addresses
  - the CPU load should be smaller
  
  
 # Fun fact
 
-  Writing to indexedDB in a batch instead of one-by-one lowered the time it took to it's ~1/2-2/3
- console.profile('label') is a really neat tool


background/services/indexing/index.ts
```diff
@@ -31,7 +31,7 @@ import { EnrichedEVMTransaction } from "../enrichment/types"
 const FAST_TOKEN_REFRESH_BLOCK_RANGE = 10
 
 interface Events extends ServiceLifecycleEvents {
-  accountBalance: AccountBalance
+  accountBalance: AccountBalance | AccountBalance[]
```
0xDaedalus, line 34: This seems to lead to some awkward handling in the event listener.  I think you might make your life a lot easier if you changed this to `accountBalances: AccountBalance[]` and just passed in an array with 1 balance in the case that you have a single account balance

0xDaedalus (commented):
Gave it a quick look but have no tested - In general I think that you might gain a lot by moving away from `AccountBalance | AccountBalance[]` to just `AccountBalance[]`

background/services/indexing/index.ts
```diff
@@ -356,6 +357,20 @@ export default class IndexingService extends BaseService<Events> {
             contractAddress
           ))
 
+        if (amount > 0) {
+          if (knownAsset) {
+            // TODO: we might don't need this here? Or we just need it for the first time?
```
Shadowfiend, line 362: We do need this here, as asset transfers initiated by other people can trigger balance updates for tokens we aren't specifically tracking yet.

background/services/indexing/index.ts
```diff
@@ -356,6 +357,20 @@ export default class IndexingService extends BaseService<Events> {
...
+            await this.addAssetToTrack(knownAsset)
+          } else {
+            await this.addTokenToTrackByContract(
+              addressNetwork,
+              contractAddress
+            )
+            // TODO we're losing balance information here, consider an
+            // addTokenAndBalanceToTrackByContract method
```
Shadowfiend, line 370: I think this restructure means this is no longer true.

background/services/indexing/index.ts
```diff
@@ -366,19 +381,25 @@ export default class IndexingService extends BaseService<Events> {
...
-          await this.addTokenToTrackByContract(addressNetwork, contractAddress)
-          // TODO we're losing balance information here, consider an
-          // addTokenAndBalanceToTrackByContract method
+
+          return accountBalance
         }
+
+        return undefined
       })
+    ).then((returnArray) =>
```
Shadowfiend, line 390: We're in an `async` function, so I would avoid using `then` here, and rely on `await` instead—either with an additional intermediate variable, or by parenthesizing the `await` and then chaining to it.

background/services/indexing/index.ts
```diff
@@ -366,19 +381,25 @@ export default class IndexingService extends BaseService<Events> {
...
+      returnArray
+        .filter((promiseReturn) => promiseReturn.status === "fulfilled")
+        .map(
+          (fulfilledReturn) =>
+            (fulfilledReturn as PromiseFulfilledResult<AccountBalance>).value
```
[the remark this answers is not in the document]
Shadowfiend: We should probably introduce a helper here, because this is both verbose af and makes it hard not to use `as` in anger lol.

background/services/indexing/index.ts
```diff
@@ -366,19 +381,25 @@ export default class IndexingService extends BaseService<Events> {
...
+        )
+        .filter((value) => !!value)
```
Shadowfiend, line 397:
```suggestion
        .filter((value) => value !== undefined)
```

Shadowfiend (dismissed):
A variety of small nits around naming and style, but nothing really blocking IMO. Will let this sit for a bit in case you want to land any fixes before merging @greg-nagy, but otherwise plan on merging later today.

background/main.ts
```diff
@@ -643,10 +643,13 @@ export default class Main extends BaseService<never> {
 
   async connectChainService(): Promise<void> {
     // Wire up chain service to account slice.
-    this.chainService.emitter.on("accountBalance", (accountWithBalance) => {
-      // The first account balance update will transition the account to loading.
-      this.store.dispatch(updateAccountBalance(accountWithBalance))
-    })
+    this.chainService.emitter.on(
+      "accountsWithBalances",
+      (accountWithBalance) => {
```
Shadowfiend, line 648: The parameter here is still singular, but I want this too much to block on it :grin: Maybe we can sneak into another PR.

background/redux-slices/accounts.ts
```diff
@@ -163,9 +163,9 @@ const accountSlice = createSlice({
     },
     updateAccountBalance: (
```
Shadowfiend, line 164: Still need the action name update, see previous though on not blocking.

Shadowfiend (approved):
Let's.

Chocolatin'.

Gooooo!

PR #1281, opened 2022-04-01
Return Address: Introduce validation hooks, address/name validation, and SharedAddressInput

### LINKED ISSUE #1267: Can't send assets to ENS addresses
Right now the `Send` page only accepts full addresses in the `Send To` field and keeps the `Send` button grey if an ens address is entered.  It would be great if we could also send directly to ens names.

We will probably want to validate that a given ENS name is valid before allowing the user to click `Send`.

image

Shadowfiend:
This PR does more than I like to do in one PR, but some refactoring
needed to happen and we want `SharedAddressInput` ASAP, so here we are.

Big things:
- Hooks are refactored to `ui/hooks/*` with `ui/hooks/index.ts` acting as
  the entry point. Imports are unchanged.
- Validation hooks have been added, sync and async, one for parsing (as
  used via `parseAndValidate` in `SharedTypedInput`) and one for addresses/names
  (as used in `Send` and `SharedAddressInput`).
- `SharedTypedInput` was moved to `useParsedValidation`, which is just
  an abstraction of the underlying parse/validate logic. These might
  be further separated in the future---or not! It's useful to have a
  shared input that handles labeling, error rendering, etc all in one,
  as long as that's exactly what you need. Usage is unchanged.
- `SharedAddressInput` is a new `SharedInput` that handles address
  resolution. Sample use is in `OnboardingViewOnlyWallet`.
- `Send` uses the `useAddressOrNameValidation` hook directly since it
  has different styling than standard inputs.

## Testing

- Test that you can send tokens and ETH to ENS addresses.
- Test that you can send tokens and ETH to regular Ethereum addresses.
- Test that you can still add ENS and regular addresses as view-only wallets.
- Test that you can still properly set and clear the gas limit on network settings (this hits the `parseAndValidate` code path and thus the new `useParsedValidation` hook).

Closes #1267.


ui/hooks/validation-hooks.ts
```diff
@@ -0,0 +1,187 @@
...
+    if (trimmed === "") {
+      onValidChange(undefined)
+    } else if (isProbablyEVMAddress(trimmed)) {
+      onValidChange(trimmed)
+    } else {
+      setIsValidating(true)
+      validatingValue.current = trimmed
+
+      const resolved = (await dispatch(
+        resolveNameOnNetwork({ name: trimmed, network: ETHEREUM })
```
greg-nagy, line 162: Should we worry about canceling lookups in case of a new requests? 

In angular land this was usually where we strated piling RxJS operators.

greg-nagy (approved):
couple of nonblocking comments

let's ship this beauty ❤️

ui/hooks/validation-hooks.ts
Shadowfiend: This is what the `validatingValue` ref is currently used for, much like in swaps--it tracks the thing we're looking for a response to and ignores the response if it's not for the latest value. We *probably* want to abstract this to its own hook at some point, but I think the more useful first helper here is going to be a denounce hook so we're not triggering a million lookups as a user types.

greg-nagy: >  more useful first helper here is going to be a denounce hook so we're not triggering a million lookups as a user types.

This is one part of what I had in mind and the other is to cancel a sent request when it becomes irrelevant. 
We might be able to solve this — and we might need to solve this on the long run — but I expect this to be 
a bit of headache.

ui/hooks/validation-hooks.ts
```diff
@@ -0,0 +1,187 @@
...
+  const dispatch = useBackgroundDispatch()
+
+  const handleInputChange = async (newValue: string) => {
+    setRawValue(newValue)
+
+    const trimmed = newValue.trim()
+
+    setErrorMessage(undefined)
+    if (trimmed === "") {
+      onValidChange(undefined)
```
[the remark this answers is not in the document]
greg-nagy: Makes sense, thanks.

Sometimes I miss angular.

ui/hooks/validation-hooks.ts
Shadowfiend: Don't know that there's any real cancellation possible in this particular case, since once the request is issued it's out. We can stop paying attention to the response, but that's about it. For more complex stuff though thunks do have a cancellation signal we can leverage I believe.

PR #1282, opened 2022-04-01
Turbo Check: Implement accelerated balance checks for known transfers

Shadowfiend:
Direct commit message lift:

Enriched transaction annotations can give us high-certainty token
transfers for most ERC20 token transfers. In these scenarios, rather
than doing a within-1-minute balance check, the indexing service now
does an accelerated within-300-ms check for just those addresses,
which should considerably improve responsiveness to transaction
confirmation in those cases.

Additionally, fixed an issue in the accounts slice where normalized
addresses weren't being used to key into the accounts map consistently.
Balances should start behaving more consistently across the board now!

## Testing

- Do some swaps and other transactions that move around ERC20s. You
  should see fast balance updates fairly consistently, rather than
  having to wait up to a minute.

### DIFF
--- a/background/redux-slices/accounts.ts
+++ b/background/redux-slices/accounts.ts
@@ -8,6 +8,7 @@ import {
   AssetDecimalAmount,
 } from "./utils/asset-utils"
 import { DomainName, HexString, URI } from "../types"
+import { normalizeEVMAddress } from "../lib/utils"
 
 /**
  * The set of available UI account types. These may or may not map 1-to-1 to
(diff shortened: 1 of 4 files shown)


background/redux-slices/accounts.ts
```diff
@@ -171,18 +174,21 @@ const accountSlice = createSlice({
           asset: { symbol: updatedAssetSymbol },
         },
       } = updatedAccountBalance
-      const existingAccountData = immerState.accountsData[updatedAccount]
+
+      const updatedAccountKey = normalizeEVMAddress(updatedAccount)
+
+      const existingAccountData = immerState.accountsData[updatedAccountKey]
```
greg-nagy, line 180: \o/

background/services/indexing/index.ts
```diff
@@ -81,7 +85,7 @@ export default class IndexingService extends BaseService<Events> {
     super({
       tokens: {
         schedule: {
-          periodInMinutes: 1,
+          periodInMinutes: 10,
```
greg-nagy, line 88: Tbh, I am tempted to keep this as 1, especially it's not so CPU intensive anymore.

greg-nagy (approved):
I feel that although there is valuable insight in the review, this should be merged ... :D 

Works really well 🏎

CleanShot 2022-04-04 at 16 31 41

greg-nagy (approved):
let's 🚢

PR #1283, opened 2022-04-01
🎨 Update colors throughout with improved saturation

henryboldi:
Given a mismatch between how Figma's unmanaged color profile in rendered vs.
the colors rendered on the web in RGB, our implementation felt a bit dull in
comparison to the designs. After setting Figma's color profile explicitly to
sRGB we (Vlad) could pick colors that appear more vibrant, while constrained to
the color space that Chrome and Firefox can currently display. Safari has both
Chrome and Firefox beat in implementing `CSS Color Module Level 4`! So here we
are. Now we've got lovely vibrant colors selected by @VladUXUI. In fact, Vlad
sent me these colors already formatted with the code.

Kudos to Vlad —  these updated colors look like a nice boost! And to @mr-michael 
for bringing the discrepancy in saturation to our attention. It's like a layer
of dust has been wiped off 🧼

<br/>

<table>
  <tbody>
    <tr>
      <td>Before 🧦</td>
      <td><b>After ✨</b></td>
    </tr>
    <tr>
      <td>before_colors</td>
      <td>after_colors</td>
    </tr>
  </tbody>
</table>

### DIFF
--- a/ui/public/index.css
+++ b/ui/public/index.css
@@ -1,35 +1,40 @@
 :root {
-  --hunter-green: #002522;
+  --hunter-green: #002825;
   --castle-black: #0a0806;
-  --green-120: #001413;
-  --green-95: #193330;
-  --green-80: #33514e;
-  --green-60: #667c7a;
-  --green-40: #99a8a7;
-  --green-20: #ccd3d3;
-  --green-5: #f1f6f6;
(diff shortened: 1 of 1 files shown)


Shadowfiend (approved):
Less do eet.

PR #1284, opened 2022-04-01
🍫 Life is short: Psychological solution to high gas fees

greg-nagy:
CleanShot 2022-04-01 at 09 47 56

# What?

Let's make gas fees hurt a bit less, by utilizing the latest studies about the mood benefits of chocolate.

# Why?

Because life is too short!

# How? 

Let's not bother our brains with arbitrary numbers in a made-up currency. 
Let's use REAL things.

# Test

Grab a bar of chocolate when reviewing this PR.
Feels better, ain't it? 
Welcome :) 

# Demo

CleanShot 2022-04-01 at 09 51 29
CleanShot 2022-04-01 at 09 51 48


gasolin:
for tally context maybe a dog bone fits better :p

Shadowfiend:
Haha now that April 1 has passed I think we can close the PR :)

Incidentally we may have enough infrastructure that we could have set this up as the primary currency throughout the app by introducing a price point into the redux store :laughing:

PR #1285, opened 2022-04-01
Remove feature flag for Ledger support

fernya:
The feature is currently enabled in current distributed builds without any
observable regressions. Therefore removal of the flag is justifiable.


Shadowfiend (commented):
Going to test drive later but deeply excited to merge this. Feature flag cleanup here is overdue! :sunglasses:

background/main.ts
```diff
@@ -985,14 +939,9 @@ export default class Main extends BaseService<never> {
...
         const clear = () => {
-          if (HIDE_IMPORT_LEDGER) {
-            // Ye olde mutual dependency.
-            // eslint-disable-next-line @typescript-eslint/no-use-before-define
-            this.keyringService.emitter.off("signedTx", resolveAndClear)
-          } else {
-            // eslint-disable-next-line @typescript-eslint/no-use-before-define
-            this.signingService.emitter.off("signingTxResponse", handleAndClear)
-          }
+          // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
Shadowfiend, line 942: We need to leave this marked with the explanation as to why we're disabling the linter, but I'll handle this somewhere in a branch of mine.

background/main.ts
```diff
@@ -1091,11 +1032,8 @@ export default class Main extends BaseService<never> {
           rejecter()
         }
 
-        if (HIDE_IMPORT_LEDGER) {
-          this.keyringService.emitter.on("signedData", resolveAndClear)
-        } else {
-          this.signingService.emitter.on("signingDataResponse", handleAndClear)
-        }
+        this.signingService.emitter.on("signingDataResponse", handleAndClear)
```
Shadowfiend, line 1035: Looks like we can clean up some of the above code that is no longer used; I'll handle that separately as well.

Shadowfiend (approved):
Let's do eet.

PR #1286, opened 2022-04-01
Spot Check: Clean up some unused code, imports, and add names to activities

Shadowfiend:
Quick and easy.

Activity list improvements are pretty short:
- Contract interactions show the To address.
- The activity list shows the name when available for the displayed
  address. An OS tooltip shows the full address.
- Activity details show name + address when the name is available.
  The existing click-to-copy functionality remains.

## Testing

Look at some existing account history and check that:
- Any addresses that can be resolved display as names in
  the list and name + address in the details.
- Make sure OS tooltips exist with the full address.
- Make sure the details address/name can be clicked to copy.
- Make sure contract interactions show the To instead of From
  address.
- Make sure token and ETH sends to other addresses still show
  the To instead of From.
- Make sure token/ETH receives show the From instead of To.

Receipts:

Screen Shot 2022-04-01 at 17 17 06
Screen Shot 2022-04-01 at 17 17 45
Screen Shot 2022-04-01 at 17 18 00


ui/components/Wallet/WalletActivityListItem.tsx
```diff
@@ -23,10 +25,9 @@ function isReceiveActivity(activity: ActivityItem, account: string): boolean {
 }
 
 function isSendActivity(activity: ActivityItem, account: string): boolean {
-  return (
-    activity.annotation?.type === "asset-transfer" &&
-    sameEVMAddress(activity.annotation?.senderAddress, account)
-  )
+  return activity.annotation?.type === "asset-transfer"
```
greg-nagy, line 28: This mostly works well, but when sending to myself it's labeld as `Received` but `To:` is used.

CleanShot 2022-04-06 at 11 17 18

greg-nagy (approved):
the code is good, minor notes, nothing blocking 

the user satisfaction is 🔥

ui/components/Wallet/WalletActivityListItem.tsx
Shadowfiend: Yeah this one bugged me too but I didn't get a chance to tweak it. It's a mismatch between two places trying to decide what to display, and the "right" fix may involve some light refactoring I wasn't able to dig in on.

PR #1137, opened 2022-03-10
UNS domain support

mul1sh:
Replaces #1010 

Will fix https://github.com/unstoppabledomains/gitcoin-bounties/issues/3

PR #1288, opened 2022-04-03
TIP 2 – Integration of IDriss into Tally Ho! Wallet - Feature 1

idrisssystem:
Add IDriss-resolver in sending field.

As presented in the community calls on March 10 and April 1, 2022, users are able to input IDriss names (emails, phone numbers, and Twitter usernames) in the recipient field within the “Send Asset” tab and see wallet addresses that have been linked to them, as presented below:
tallyIdriss

Styling of elements and functionality has been adopted from PR #1137 

This PR is supporting **Feature 1** of **TIP 2 – Integration of IDriss into Tally Ho! Wallet** as found in the [Tally Ho! Governance Forum](https://gov.tally.cash/t/tip-2-integration-of-idriss-into-tally-ho-wallet-draft/596).

### DIFF
--- a/background/accounts.ts
+++ b/background/accounts.ts
@@ -41,6 +41,7 @@ export type AccountBalance = {
 export type AddressOnNetwork = {
   address: HexString
   network: EVMNetwork
+  alternative?: { name: string; address: string }[]
 }
 
 /**
--- a/background/lib/utils/index.ts
+++ b/background/lib/utils/index.ts
@@ -12,10 +12,12 @@ export function normalizeEVMAddress(address: string | Buffer): HexString {
 export function normalizeAddressOnNetwork({
   address,
   network,
+  alternative,
 }: AddressOnNetwork): AddressOnNetwork {
   return {
     address: normalizeEVMAddress(address),
     network,
+    alternative,
   }
 }
 
--- a/background/services/name/index.ts
+++ b/background/services/name/index.ts
@@ -17,7 +17,7 @@ import {
   addressBookResolverFor,
   ensResolverFor,
   unsResolver,
-  rnsResolver,
+  rnsResolver, idrissResolver,
 } from "./resolvers"
 import PreferenceService from "../preferences"
 import { isFulfilledPromise } from "../../lib/utils/type-guards"
@@ -126,6 +126,7 @@ export default class NameService extends BaseService<Events> {
       // for the given resource.
       ensResolverFor(chainService),
       unsResolver(),
+      idrissResolver(),
       ...(RESOLVE_RNS_NAMES ? [rnsResolver()] : []),
     ]
 
--- a/background/services/name/resolvers/idriss.ts
+++ b/background/services/name/resolvers/idriss.ts
@@ -24,7 +24,8 @@ export default function idrissResolver(): NameResolver<"idriss"> {
       const address = result["Public ETH"] ?? Object.values(result)[0]
       if (address) {
         const normalizedAddress = normalizeEVMAddress(address)
-        return { address: normalizedAddress, network }
+        const alternative = Object.entries(result).map(x=>({name:x[0], address:normalizeEVMAddress(x[1])}));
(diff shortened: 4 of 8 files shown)


geoist1:
> Wow, this looks awesome! 🎉 Thanks for presenting on the community call as well, I'm excited about what y'all are doing and it sounds like the community is too.
> 
> A small amount of level-setting on our end in terms of timelines and such:
> 
> * Right now the core team is pretty heads down on a forthcoming release, and will probably be in that mode for another 3-5 weeks, so it may take a bit to do a deep review here.
> * As a warning, it's unlikely that feature-by-feature voting will be happening via the DAO. I believe we're still shaping up how exactly that flow will go overall. That said, this is exactly the kind of feature we want to be able to integrate easily in the wallet core.
> * Generally when adding a different style of functionality (in this case, the ability to resolve a name to multiple addresses), we want to do it in a well-abstracted way, and that usually involves an RFB process that isn't all the way written down yet, otherwise we'll likely descend into a debt spiral. As you flagged, there's also a component that is having design look at it and consider the best way to integrate into the visual whole.
> 
> What I expect we'll have happen next purely from an integration perspective:
> 
> * I'm landing a pretty significant refactoring of the `NameService` this week. This will be the right home for IDriss's integration as well. This refactor still assumes a 1-to-1 mapping from a name to an address, however.
> * I'd like to draft an RFB for names that may resolve to more than one address. I expect that won't happen until the week of 18 April at the earliest (~2 weeks from now).
> * I would also like to draft a “this is how RFBs work” bit. Hopefully the timeline will be similar.
> * @VladUXUI will have to have a look here and see how this fits into the overall visual and UX goals. I can't commit to a timeline here, but it will be subject to the same “the team is focused on other things for the next few weeks” constraint. This feedback will probably factor into the RFB as well.
> 
> Happy to answer any questions here in the meantime, and hope to this integrated sooner rather than later!

Thank you for feedback! It's understandable that lots of processes are not fully worked out yet🙂 Looking forward to seeing how it all evolves.

geoist1:
@Shadowfiend Any updates here? 🙂

geoist1:
Hi @Shadowfiend, we recently added reverse resolving to this PR as well. Do you think we could start the deeper review here anytime soon?

Also, tagging @mr-michael here to let know we'd love to heavily co-market this integration in our channels (2.5k community members on Discord and 4.5k on Twitter) 🤝

Shadowfiend:
@geoist1 @mr-michael is going to be able to make the priority call here--we're stretched quite thin at the moment and my recollection is we wanted a more integrated design for multi name resolution and reverse resolution here as well.

geoist1:
@Shadowfiend Checking back here to see if it's a better time for you guys to review this? Optionally, we could limit the resolving to single results (most of the users have a single linked wallet anyway) to make the review much easier. I guess then we wouldn't need a special design for multi-name resolution, it would look exactly like ENS, UD, and RNS integration.

@mr-michael We've grown our user & community numbers a lot since August, which might be interesting for you from a co-marketing perspective (8.3k total address book registrations, 3.5k community members on Discord and 8.3k on Twitter). Here's our public Dune Analytics dashboard: https://twitter.com/IDriss_xyz/status/1585617423310327808

Shadowfiend:
Hey folks, apologies for the long silence. The short version here is that we're still not set up on the community side to clearly prioritize new integrations like this one with respect to other aspects of the roadmap, particularly when they have deeper architectural and design implications. We've been focused elsewhere since this PR was opened, and we don't have a clear roadmap for returning to that aspect of prioritization before Q4 at the earliest.

As such, for clarity in the PR list, I'm going to close this PR. Note that this does *not* mean zero interest in the integration, but rather that we should keep things at the community level (chat.taho.xyz, gov.taho.xyz) until we land on a good way to coordinate the team's time and work with the community's preferences. Again, we don't expect a lot of movement on that front until Q4 at the earliest.

Thanks again for bearing with us here, and I hope we can turn our attention to getting the processes ironed out here before too too long.

PR #1295, opened 2022-04-05
🏰 Pass IPFS asset icons through gateway

henryboldi:
This PR adds support for IPFS hosted asset icons. Many asset icons in the swap
lists are actually on IPFS, so we weren't resolving them. With this added
support, much more assets in the swap menus appear! This will be a noticeable
improvement for users.

Based on @Shadowfiend's comment, I moved the IPFS gateway functions already
being used in the name service to their own lib in `storage-gateway.ts`. The
asset icon component now simply checks if the URL is using the IPFS protocol,
and if so, it wraps it in an HTTP gateway.

To test browse around, particularly on the swap asset list menu, and see if
you're noticing more asset icons load in.


Shadowfiend (commented):
Oof, this is now the third place we're doing gateway stuff. Time to pull it up to a shared lib—let's move https://github.com/tallycash/extension/blob/main/background/services/name/index.ts#L56-L91 up to `background/lib/storage-gateway.ts` or so, and then reuse from here, `NameService`, and… The third place 😏

ui/components/Shared/SharedAssetIcon.tsx
```diff
@@ -12,15 +13,15 @@ export default function SharedAssetIcon(props: Props): ReactElement {
...
   const hasHardcodedIcon = hardcodedIcons.includes(symbol)
 
-  // Checks to see if it's an http(s) address because I've seen
-  // strings get here like ipfs://QmYNz8J1h5yefkaAw6tZwUYoJyBTWmBXgAY28ZWZ5rPsLR
-  // which won't load. Of if we have a hardcoded backup image
-  const hasValidImage =
-    (logoURL && logoURL.includes("http")) || hasHardcodedIcon
+  // If IPFS url, pass through gateway
+  let httpURL = logoURL
+  if (logoURL && new URL(logoURL) && logoURL.includes("ipfs")) {
```
mhluongo, line 18: I believe you want

```
if (logoURL && new URL(logoURL) && (new URL(logoURL)).protocol === "ipfs") {
```

... otherwise any URL that happens to include "ipfs", eg `https://example.com/ipfs/tokenImage.jpg`, will get munged here.

mhluongo (changes requested):
Let's parse URLs properly!

ui/components/Shared/SharedAssetIcon.tsx
henryboldi: Good point, thank you 🙏
<br/>

I think there should be a colon here though,
`.protocol === "ipfs:"`
> The protocol getter steps are to return this’s URL’s scheme, followed by U+003A (:).

Which was also missing in `function storageGatewayURL`

background/lib/storage-gateway.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  return newURL
+}
+
+// TODO eventually we want proper IPFS and Arweave support
+export function storageGatewayURL(url: URL): URL {
+  switch (url.protocol) {
+    case "ipfs:":
+      return changeURLProtocolAndBase(url, ipfsGateway)
+    case "ar":
```
mhluongo, line 34: Missed a `:` here!

henryboldi: Updated :)

ui/components/Shared/SharedAssetIcon.tsx
```diff
@@ -12,15 +13,15 @@ export default function SharedAssetIcon(props: Props): ReactElement {
...
   const hasHardcodedIcon = hardcodedIcons.includes(symbol)
 
-  // Checks to see if it's an http(s) address because I've seen
-  // strings get here like ipfs://QmYNz8J1h5yefkaAw6tZwUYoJyBTWmBXgAY28ZWZ5rPsLR
-  // which won't load. Of if we have a hardcoded backup image
-  const hasValidImage =
-    (logoURL && logoURL.includes("http")) || hasHardcodedIcon
+  // If IPFS url, pass through gateway
+  let httpURL = logoURL
+  if (logoURL && new URL(logoURL) && new URL(logoURL).protocol === "ipfs:") {
```
mhluongo, line 18: Okay one more: this should support `"ar:"` as well :sweat_smile:

mhluongo (dismissed):
Looking good — merging on conflict resolution + green build.

mhluongo (approved):
🌄

PR #1301, opened 2022-04-06
🔅 Use SVG icons pervasively

henryboldi:
This PR adds and uses many more SVG icons throughout the extension. This is
primarily achieved through edits to `SharedButton`. I've added two new props -
`iconSmall` and `iconMedium`, which are now type constrained to only the
available icons for their corresponding size. To test this, click around the
extension and see if any icons inside of buttons are looking funky.

Granted, I've left existing icon files and haven't tackled `SharedIcon`. My
sense is that `SharedIcon` should not be a button as it currently stands. I
think we should split it into `SharedIcon` and `SharedIconButton`, and then
used `SharedIcon` inside of `SharedButton`, but I haven't gotten there yet.


0xDaedalus:
Clicked around a bit and visually everything looks like its working!  Wanted to get your thoughts on an alternative to the `iconSmall` and `iconMedium` props though - what if we were to use a discriminated union?

For Example

```
type SmallIconProps = {
  iconSize: 'small',
  icon: a | b | c // all allowed small icons
}

type MediumIconProps = {
  iconSize: 'medium',
  icon: x | y | z // all allowed medium icons
}

type CommonProps = {
  // all the other props
}

type Props = CommonProps & (SmallIconProps | MediumIconProps)
```

The one thing that is a bit awkward about the current typing is that it is technically typesafe to declare both an iconSmall and an iconMedium (even though obviously wrong) - which a discriminated union would avoid.

Not blocking at all but thought I'd drop the thought here in case you agreed.

henryboldi:
That's a great idea! I'll update it :)

0xDaedalus (approved):
📈

PR #1303, opened 2022-04-08
Helpers and alignments: Handful of helper and alignment tweaks

Shadowfiend:
The three helpers here are used in later PRs and likely
useful across the code base, while the alignment tweaks
fix some margin jumps that happen during ENS resolution
on the asset Send page.

## Testing

Check that as you type a name/address on the Send page,
the network fee box doesn't move or jump, both while the
name/address is invalid and once it is valid or blank.


0xDaedalus (approved):
lgtm!

PR #1304, opened 2022-04-08
Resolver Revolver: Introduce NameResolver and refactor NameService to use it

Shadowfiend:
--> Draft until #1303 lands.

Two big differences here:

- The NameService is now able to use more than one resolver, though the
  only implemented resolver is still ENS.
- Lookups generally use *OnNetwork types, and results are similarly
  *OnNetwork types. Avatars are still returned as plain URLs, at least
  for now.

## Testing

Make sure that name resolution continues to work:

- Check that entering an ENS name resolves in the Send form.
- Check that entering an ENS name resolves in the Read-only form.
- Check that sending to an address with an ENS name correctly shows
  the ENS name in the Sign Transaction page.
- Check that sending to an address with an ENS name correctly shows
  the ENS name in the activity list.

### DIFF
--- a/background/services/name/index.ts
+++ b/background/services/name/index.ts
@@ -109,7 +109,10 @@ export default class NameService extends BaseService<Events> {
   /**
    * Cached resolution for avatar URIs that are not yet URLs, e.g. for EIP155.
    */
-  private cachedResolvedEIP155Avatars: Record<string, ResolvedAvatarRecord> = {}
+  private cachedResolvedEIP155Avatars: Record<
+    string,
+    ResolvedAvatarRecord | undefined
+  > = {}
 
   /**
    * Cached resolution for name records, by network family followed by whatever
@@ -118,7 +121,7 @@ export default class NameService extends BaseService<Events> {
   private cachedResolvedNames: {
     EVM: {
       [chainID: string]: {
-        [address: HexString]: ResolvedNameRecord
+        [address: HexString]: ResolvedNameRecord | undefined
       }
     }
   } = { EVM: { [ETHEREUM.chainID]: {} } }
@@ -208,18 +211,15 @@ export default class NameService extends BaseService<Events> {
     const { address: normalizedAddress, network } =
       normalizeAddressOnNetwork(addressOnNetwork)
 
-    if (
-      checkCache &&
+    const cachedResolvedNameRecord =
       this.cachedResolvedNames[network.family][network.chainID][
         normalizedAddress
       ]
-    ) {
+
+    if (checkCache && cachedResolvedNameRecord) {
       const {
         resolved: { nameOnNetwork, expiresAt },
-      } =
-        this.cachedResolvedNames[network.family][network.chainID][
-          normalizedAddress
-        ]
+      } = cachedResolvedNameRecord
 
       if (expiresAt >= Date.now()) {
         return nameOnNetwork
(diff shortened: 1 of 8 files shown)


background/services/name/index.ts
```diff
@@ -98,9 +104,24 @@ function storageGatewayURL(url: URL): URL {
...
+   * Cached resolution for avatar URIs that are not yet URLs, e.g. for EIP155.
+   */
+  private cachedResolvedEIP155Avatars: Record<string, ResolvedAvatarRecord> = {}
 
-  private cachedResolvedNames: Record<string, ResolvedNameRecord> = {}
+  /**
+   * Cached resolution for name records, by network family followed by whatever
+   * discrimant might be used within a network family between networks.
+   */
+  private cachedResolvedNames: {
```
0xDaedalus, line 118: It seems like this cache never invalidates - should we maybe have a max cache time of 30 days?

background/services/name/index.ts
```diff
@@ -120,177 +141,224 @@ export default class NameService extends BaseService<Events> {
...
+        )
+
+        if (metadata !== undefined && metadata.image !== undefined) {
+          const { image } = metadata
+          const resolvedGateway = {
+            ...baseResolvedAvatar,
+            resolved: { avatar: storageGatewayURL(new URL(image)) },
           }
+
+          this.cachedResolvedEIP155Avatars[avatarUri] = resolvedGateway
```
0xDaedalus, line 344: I think you might need `avatarUri.toLowerCase()` here for the check on line 314 to pass 100% of the time.

background/services/name/name-resolver.ts
```diff
@@ -0,0 +1,93 @@
...
+ * decide whether a resolver is _likely to be able to_ resolve a name, avatar,
+ * or address, and one in which the resolution is attempted. The static phase is
+ * used to avoid calls that are known to produce failures, such as attempting to
+ * resolve a Bitcoin name on a resolver that only supports Ethereum.
+ */
+export interface NameResolver<ResolverType extends string> {
+  type: ResolverType
+
+  /**
+   * Given a name and network, synchronously returns a boolean indicating
```
0xDaedalus, line 22: Given a *address and a network

0xDaedalus (commented):
Looks good but want to be explicitly sure that we're ok with a never-expiring cache.  Will test over the weekend.

background/services/name/index.ts
Shadowfiend: Yep, def true.

background/services/name/index.ts
Shadowfiend: I'm not super concerned because the entries do expire (the avatar ones don't though), and in fact do so quite quickly. We could have the cache grow uncontrollably but I'm not sure that's likely in the immediate term. 

Either way both caches need more work independent of this; here I just added network awareness.

background/services/name/name-resolver.ts
Shadowfiend: Yuh.

0xDaedalus (approved):
After some small tweaking lgtm 👍 

-  [x] Check that entering an ENS name resolves in the Send form.
-  [x] Check that entering an ENS name resolves in the Read-only form.
-  [x]  Check that sending to an address with an ENS name correctly shows
  the ENS name in the Sign Transaction page.
-  [x]  Check that sending to an address with an ENS name correctly shows
  the ENS name in the activity list.

PR #1305, opened 2022-04-08
Conflict Resolution: Introduce stub known contract and address book name resolution 

Shadowfiend:
---> Draft until #1304 lands.

Known contracts and the address book are stubbed in the
PreferenceService, and wired up with the expectation that they will both
eventually live in IndexedDB. Resolvers are implemented for both of
them, and used preferentially over ENS.

EIP2612 source name resolution is also moved to known contracts, making
Uniswap router resolution possible across the extension.

## Testing

- Make sure that EIP-2612 is still resolving correctly for Uniswap
  approval typed signatures.
- Check that the 0x Router is being shown instead of the full address
  on the Swap Sign Transaction page.
- Check that the 0x Router is being shown instead of the address in
  a swap transaction in the activity list.
- For a bonus, check that Uniswap is being shown in Sign Transaction
  screens for Uniswap.


background/services/name/index.ts
```diff
@@ -212,14 +234,23 @@ export default class NameService extends BaseService<Events> {
...
+        // Checked defined above.
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         this.cachedResolvedNames[network.family][network.chainID][
           normalizedAddress
-        ]
+        ]!
+
+      if (
+        addressOnNetwork.address ===
+        "0xDef1C0ded9bec7F1a1670819833240f027b25EfF"
```
0xDaedalus, line 250: I'm thinking it might have to be advantageous for this address to be stored as a named variable somewhere.  Otherwise people working with this code have to go to etherscan to look up what this is address is.

It's also unclear to me why we are `warn`ing in the various functions in here when we see this address.

background/services/name/index.ts
```diff
@@ -230,20 +261,46 @@ export default class NameService extends BaseService<Events> {
...
+    )
+      .filter(isFulfilledPromise)
+      .find(({ value: { resolved } }) => resolved !== undefined)?.value
+
+    if (
+      addressOnNetwork.address === "0xDef1C0ded9bec7F1a1670819833240f027b25EfF"
+    ) {
+      logger.warn(
+        "Got",
+        await Promise.allSettled(
```
0xDaedalus, line 286: I think we can avoid this second series of lookups here if we do something like the following above:

```
const nameLookups = await Promise.allSettled ... etc. (probably can be named better)
const firstMatchingResolution = nameLookups.filter ... etc.
```

and then pass `nameLookups` to this `warn`.

background/services/name/resolvers/known-contracts.ts
```diff
@@ -0,0 +1,27 @@
...
+      return false
+    },
+    canAttemptAddressResolution(): boolean {
+      return true
+    },
+
+    lookUpAddressForName:
+      preferenceService.lookUpAddressForContractName.bind(preferenceService),
+    async lookUpAvatar() {
+      throw new Error("Avatar resolution not supported in address book.")
```
0xDaedalus, line 22: probably want `in known contracts` here

background/services/name/index.ts
```diff
@@ -279,23 +346,28 @@ export default class NameService extends BaseService<Events> {
...
+      .filter(isFulfilledPromise)
+      .find(({ value: { resolved } }) => resolved !== undefined)?.value
 
-    if (resolved === undefined || resolved.resolved === undefined) {
+    if (
+      firstMatchingResolution === undefined ||
+      firstMatchingResolution.resolved === undefined
+    ) {
       return undefined
     }
```
0xDaedalus, line 365: Out of the scope of this PR but this is starting to not feel very DRY - will think about how we can condense these calls up if you think that would be beneficial

background/services/preferences/index.ts
```diff
@@ -1,10 +1,34 @@
...
+
+const BUILT_IN_CONTRACTS = [
+  {
+    network: ETHEREUM,
+    address: normalizeEVMAddress("0xDef1C0ded9bec7F1a1670819833240f027b25EfF"),
+    name: "0x Router",
+  },
+  {
+    network: ETHEREUM,
+    // Uniswap v2 Router
```
0xDaedalus, line 27: I named this wrong when I first made this change - looks like it is `Router 2` of `Uniswap v3`.  We can probably just change the comment to `Uniswap v3 Router`

0xDaedalus (commented):
Nothing blocking but have not taken for a test-drive yet.  Will do once its out of `Draft`.

background/services/name/index.ts
Shadowfiend: Eh, not really tbh. This log isn't necessary, I just forgot to clean up after my debugging 😭

background/services/name/index.ts
Shadowfiend: Same, this was more a debugging aid and needs to go IMO.

background/services/name/index.ts
Shadowfiend: I think it's not worth it unless we add another 1 or more things that route through preferences tbh. I thought about it for sure.

0xDaedalus (changes requested):
Everything except 0x-router resolving in the Activity List is working great!

image


- [x] Make sure that EIP-2612 is still resolving correctly for Uniswap
approval typed signatures.
- [x] Check that the 0x Router is being shown instead of the full address
on the Swap Sign Transaction page.
- [ ] Check that the 0x Router is being shown instead of the address in
a swap transaction in the activity list.
- [x] For a bonus, check that Uniswap is being shown in Sign Transaction
screens for Uniswap.

0xDaedalus (approved):
Upon further inspection pretty sure the 0x-router name not showing up was a cacheing issue on my end.  

Ahoy matey!
🚢 🚢 🚢

PR #1306, opened 2022-04-08
📋🐛 Export logs page from menu tab

henryboldi:


background/lib/logger.ts
```diff
@@ -58,6 +58,21 @@ const styles: LogStyles = {
   },
 }
 
+function purgeSensitiveFailSafe(log: string): string {
+  return log
+    .replace(/0x(.*?)[\s]/g, "[REDACTED]") // Replace hexadecimal segments
+    .replace(/\b[a-zA-Z0-9]{64}\b/g, "[REDACTED]") // Replace private key length segments
```
Shadowfiend, line 64: Three things here:
- let's make this one regex using `|`
- let's create the regex once at the top level
- let's add something for mnemonics, maybe as aggressive as any run of 12 or 24 letter-only words with lowercase letters.

background/lib/logger.ts
```diff
@@ -58,6 +58,21 @@ const styles: LogStyles = {
...
+}
+
+function saveLog(logLabel: string, stackTrace: string[] | undefined) {
+  localStorage.setItem(
+    "logs",
+    `${(localStorage.getItem("logs") ?? "").slice(
+      -5000
```
Shadowfiend, line 71: Let's maybe do 50k?

Shadowfiend (commented):
No time to do deep review before I'm out, but left a couple of quick notes. 

Regex approach is 👌

background/lib/logger.ts
henryboldi: Yes, that's more reasonable. Thank you! Updated 🙏

henryboldi:
Converting back to draft until this is updated with discussed design changes.

background/lib/logger.ts
```diff
@@ -133,6 +153,8 @@ function genericLogger(level: LogLevel, input: unknown[]) {
   }
 
   console.groupEnd()
+
+  saveLog(logLabel, stackTrace)
```
0xDaedalus, line 157: I think we want the input here as well - right?

henryboldi: Oh man 🙈. Absolutely! Thank you

0xDaedalus:
Re-tested and it looks like the logs are making it into the output ✅ 

Unfortunately it does not look like we're sanitizing them.  Adding explicit logging of sensitive data to `main.ts` does not sanitize.

For example I can export my seed phrase, address, and private key by adding the following to main.ts:  (real values substituted)
```
background/main.ts

logger.warn(
  `Seed Phrase:  ((a seed phrase)`
)

logger.warn(`Address: ((an address)`)

logger.warn(
  `Private Key:   ((a private key))`
)
```

henryboldi:
@0xDaedalus should be fixed now :)

PR #1311, opened 2022-04-11
update i18n param assignment, fix #1195

gasolin:
pass sign + amount in a single param, so no need for special variable

🐛 found a side bug that read-only ENS not work in the latest main branch

Fixes #1195.


Shadowfiend (approved):
Ahhh I see what went sideways here. Let's do it.

Shadowfiend:
> found a side bug that read-only ENS not work in the latest main branch

@gasolin is this something you captured in another issue?

gasolin:
> > found a side bug that read-only ENS not work in the latest main branch
> 
> @gasolin is this something you captured in another issue?

saw that happens when load the addon

PR #1312, opened 2022-04-11
🚦 Add countdown/delay on button for transaction signing

henryboldi:
This PR adds a delay to enabling the sign button when opening a sign
transaction screen. Additionally, it aims to enable/disable the sign
transaction button depending on if the window is blurred or focused, or if the
transaction details change. The idea is to make sure the user has a chance to
review the transaction and confirm there's no shenanigans before pulling the
trigger. As for implementation, this is achieved with hooks
in`SignTransactionContainer.tsx`. Granted, this enable/disable logic isn't
neatly tucked away as a single new hook of its own. That said, the
implementation feels somewhat convoluted to me. Input is welcome!

### DIFF
--- a/ui/components/SignTransaction/SignTransactionContainer.tsx
+++ b/ui/components/SignTransaction/SignTransactionContainer.tsx
@@ -72,23 +72,27 @@ export default function SignTransactionContainer({
     }
   }, [])
 
-  const onBlurFocusChange = useCallback(
-    (isFocus: boolean) => {
-      clearDelaySignButtonTimeout()
-      firstOpen.current = false
+  function onBlurFocusChange(isFocus: boolean) {
+    clearDelaySignButtonTimeout()
+    firstOpen.current = false
 
-      if (isFocus) {
-        delaySignButtonTimeout.current = window.setTimeout(() => {
-          setIsOnDelayToSign(false)
-          // Random delay between 0.5 and 2 seconds
-        }, Math.floor(Math.random() * (5 - 1) + 1) * 500)
-      } else {
-        setIsOnDelayToSign(true)
-        clearDelaySignButtonTimeout()
-      }
-    },
-    [clearDelaySignButtonTimeout]
-  )
+    if (isFocus) {
+      delaySignButtonTimeout.current = window.setTimeout(() => {
+        setIsOnDelayToSign(false)
+        // Random delay between 0.5 and 2 seconds
+      }, Math.floor(Math.random() * (5 - 1) + 1) * 500)
+    } else {
+      setIsOnDelayToSign(true)
+    }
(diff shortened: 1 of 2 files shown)


ui/components/SignTransaction/SignTransactionContainer.tsx
```diff
@@ -50,6 +58,57 @@ export default function SignTransactionContainer({
...
+      firstOpen.current = false
+
+      if (isFocus) {
+        delaySignButtonTimeout.current = window.setTimeout(() => {
+          setIsOnDelayToSign(false)
+          // Random delay between 0.5 and 2 seconds
+        }, Math.floor(Math.random() * (5 - 1) + 1) * 500)
+      } else {
+        setIsOnDelayToSign(true)
+        clearDelaySignButtonTimeout()
```
cairomassimo, line 87: One of the two calls to `clearDelaySignButtonTimeout` seems redundant.
Also, having only one invocation, we could inline the definition of `clearDelaySignButtonTimeout` and avoid `useCallback`.

ui/components/SignTransaction/SignTransactionContainer.tsx
```diff
@@ -50,6 +58,57 @@ export default function SignTransactionContainer({
...
+      // On first open of window mark as focused
+      onBlurFocusChange(true)
+    }
+
+    window.addEventListener("focus", () => onBlurFocusChange(true))
+    window.addEventListener("blur", () => onBlurFocusChange(false))
+
+    return () => {
+      window.removeEventListener("focus", () => onBlurFocusChange(true))
+      window.removeEventListener("blur", () => onBlurFocusChange(false))
```
cairomassimo, line 108: These are no-op's because `(() => onBlurFocusChange(true)) !== (() => onBlurFocusChange(true))`

cairomassimo (commented):
Tested, it works. Added a few comments on the code (some important, some minor)

Also, it should be possible to implement this by avoiding `useRef`s. Something like:

```typescript

const [blocked, setBlocked] = useState(true)
const [focusChangeNonce, setFocusChangeNonce] = useState(0)

useEffect(() => {
    const update = () => { setFocusChangeNonce(x => x+1) }
    window.addEventListener("focus", update)
    window.addEventListener("blur", update)

    return () => {
      window.removeEventListener("focus", update)
      window.removeEventListener("blur", update)
    }
}, [])

useEffect(() => {
  setBlocked(true);
  const timeout = setTimeout(() => { setBlocked(false) }, 1000)
  return () => {
    clearTimeout(timeout);
  }
}, [reviewPanel, focusChangeNonce])

```

cairomassimo (approved):

PR #1321, opened 2022-04-13
🚦 Wallet Router: Implement a wallet router and define a friendly api to wallet providers 

greg-nagy:
# What? 

Implement a wallet provider registry and some rules around switching between them, while being transparent through the `window.ethereum`

# Why?

Because wallet providers are overwriting each other based on confusing environmental conditions, and this is not how this should be handled.

# How? 

See the comments on the commits

# Test

- Install Tally from the dev build and metamask from the chrome store
- go to https://app.uniswap.org/ 
- click connect wallet and see `Injected`
- turn off the default wallet in tally > connect wallet > see `MetaMask`
- turn on the default wallet in tally > connect wallet > see `Injected`


cairomassimo:
+1 for the overall idea. However I think the design of the wallet router is too complex, and IMHO it's better to propose a good design from the very beginning to avoid losing a chance. (If it takes too long to design, the we can just do the override, without playing nice at first, and take our time to make a good proposal later on.)

- dApps should be able to choose among a collection of providers, so that
    - they can present a choice to the user,
    - the user could save a default provider per-dApp
- For this reason, I think that `window.ethereum` should be thought as legacy.
- Instead of `window.ethereum` as the main entry point, we should propose a new entry point, say, `window.ethereumProviders` which is simply an array of providers (e.g. `window.ethereumProviders = [window.tally, window.metamask]`). dApps should first look at `window.ethereumProviders` and if they don't find it only then use `window.ethereum`.
- We want the different provider to have to agree on as little as possible (because they all have to implement it):
  - on startup, they should add themselves to `window.ethereumProviders`,
  - if `!window.ethereumProviders.has(window.ethereum)`, then they should also add `window.ethereum` to `window.ethereumProviders` for backward compatibility,
  - they can set `window.ethereum` to whatever they want, we don't care (libraries should prefer `window.ethereumProviders` and ignore `window.ethereum` anyway), as long as `window.ethereumProviders.has(window.ethereum) === true` afterwards.
- The order of providers in `window.ethereumProviders` shouldn't be relevant, libraries could randomize it if they want.
- if possible, we can replace `window.ethereumProviders` with a custom implementation of `Array` which disallows deletion, and maybe even randomizes the order, but it has to be compatible with native array so that the implementation by other providers is as easy as possible
- I don't think we should propose a federated mechanism to choose a "default provider" that works for all the dApps. It risks being just another land of conquest, and an incentive for providers not to play nice. (dApps can always implement it themselves.)
- The providers should have a name, icon etc. to be recognizable, but I think we talked about this elsewhere.

greg-nagy:
@cairomassimo hm ... not the review I was hoping for. But let's see. 
We probably should focus on the minimum viable solution at this point.

Originally wanted to find the ideal solution because I thought I have it, but solving the problem is the most important thing now

So the requirements
- I as a user expect tally to work with the dApps I want to use.
- I as a user want tally to be available reliably in every dApp when it's installed
- I as a user want to be able to switch on and off the default provider injection feature so I can use tally and other wallets side by side
- I as a user expect tally to show up on dApp interaction when it's installed and is set to be a default wallet

Thought process:
  - If we wait for the backend to tell us if tally is set to default we end up in the race condition where the whole issue started. The communication can be slow, and it can happen frequently that the window provider is not available on `window.ethereum` when the dApp is looking for it resulting in tally not being detected.
    - conclusion: we need to inject window provider into `window.ethereum` as soon as possible
- If we inject tally as soon as possible and by default into `window.ethereum` then other wallets can overwrite it resulting in tally not being detected.
  - conclusion: we need to implement write protection after we have injected into `window.ethereum`
- if we inject into windoww.ethereum` as soon as possible and implement a write protection, it's not possible to turn off tally as a default wallet and restore the provider that was there before we arrived
  - conclusion: We need a new layer of abstraction that can be write protected but still able to switch providers transparently in `window.ethereum`

Given these requirements — if we agree on them — tthe minimal footprint for `window.walletRouter` could be `{currentProvider: window.tally, #previousProvider: window.ethereum, switchProviders() {...} , addProvider() {...}}`

Where is the error in my assessment, and what could be simplified?

greg-nagy:
Went ahead and made a couple of updates, and mostly they are aligned with your proposal, but I kept the `window.ethereum` entry point for pragmatic reasons.

src/window-provider.ts
```diff
@@ -7,21 +7,106 @@ import TallyWindowProvider from "@tallyho/window-provider"
...
+      previousProvider: window.ethereum,
+      providers: [
+        // deduplicate the providers array: https://medium.com/@jakubsynowiec/unique-array-values-in-javascript-7c932682766c
+        ...new Set([
+          // eslint-disable-next-line no-nested-ternary
+          ...(window.ethereum
+            ? // let's use the providers that has already been registered
+              // This format is used by coinbase wallet
+              Array.isArray(window.ethereum.providers)
+              ? [...window.ethereum.providers, window.ethereum]
```
Adamj1232, line 37: Nice!

jagodarybacka (approved):
Works as expected. Code is looking good and I like the idea you suggested here ✨

PR #1325, opened 2022-04-16
feat: make password warning more explicit

### LINKED ISSUE #1255: Select Password - Make it a warning
Ticket from @mr-michael, moved here

We need to make the password warning more explicit.

Right now it's too easy to miss.

Screen Shot 2022-03-26 at 2 14 52 PM

Users can't change this (currently) and we've already had a few password scares on Discord :(

I suggest something similar to this with a 'warning' label:

Screen Shot 2022-03-26 at 2 17 09 PM

rootial:
## Summary
close #1255 
make it more explicit to increase awareness of this note.
## Test
test in local build with firefox
image

## Reviewers
@henryboldi @VladUXUI


rootial:
@VladUXUI please let me know if the note looks good to you.

VladUXUI:
hey @rootial thanks for this!
Didn't get a chance to do a design for this and i think you are close.
Adding a link to our public figma here, so you can have a look.
https://www.figma.com/file/EiXpBpNz2HcoeFJftgC5oc/Tally-Public?node-id=701%3A3141

VladUXUI:
Also for colours let's always try to use our colour library, if new colours are needed i'm more than happy to look into that.
In this case you are right, our dark-120 (that we previously used for this) doesn't have enough contrast with the hunter green colour.
You are also spot on with the colour you chose #001413. We will change our green 120 to this.

VladUXUI:
Another note would be that we have SVG now! wohooo, so it's best if we use svg instead of png.
The icons are either 24x24 or 16x16. In this case we are using 24px

VladUXUI:
Also would be good to use the font library. If we need a new font size, i'm happy to look into that

rootial:
thanks for your suggestion. do we have svg icon for this case now? i dont't find it in code base. @VladUXUI

VladUXUI:
> thanks for your suggestion. do we have svg icon for this case now? i dont't find it in code base. @VladUXUI

We do yes! here is a link to it
https://github.com/tallycash/extension/blob/main/ui/public/images/icons/m/notif-attention.svg

rootial:
image
please take a look, update styles following figma

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -58,8 +59,18 @@ export default function KeyringSetPassword(): ReactElement {
         <div className="wordmark" />
       </div>
       <h1 className="serif_header">First, let&apos;s secure your wallet</h1>
-      <div className="subtitle">
-        You will NOT be able to change this password for now.
+
+      <div className="warning_wrap">
```
henryboldi, line 63: What are your thoughts on adding a generalized `SharedWarningBanner` and using it here? I think this component is something we'll want to reuse  🙏

VladUXUI:
Looking good on my end😍 ! Awesome work @rootial 👏

ui/components/Keyring/KeyringSetPassword.tsx
rootial: that's a good suggestion. Just add the component in latest commit. @henryboldi

jagodarybacka (approved):
Looks good, let's merge it! Thanks @rootial🔥

PR #1327, opened 2022-04-18
Uh Oh - Arbitrary extension version is enabled - Fix redux-state migration.

0xDaedalus:
This PR addresses two bugs:

The first is a bug where we were assigning `isArbitraryDatSigningEnabled` to the root LedgerState rather than assigning it to the state of each connected device.  This did not seem to have any adverse effect on the application but it did not conform to our typing.

The second is a much rarer bug where if a user was upgrading to version 0.13.0 from a pre 0.11.0 version of tally the application would become unresponsive due to pre 0.11.0 redux-states not having the "ledger" property.

### To Test:
-  uninstall tally extension
- `git checkout c88495eba2299771bb8be785d99832a0a89a6e0c` (version 0.10.0)
-  build it and add it to chrome
- open it up - add a wallet (preview is fine)
- `git checkout fix-migration-six`
-  update the wallet in chrome://extensions - version should change to 0.13.0.
-  you should _not_ get a black screen and the wallet should be useable.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -212,14 +212,23 @@ const REDUX_MIGRATIONS: { [version: number]: Migration } = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   6: (prevState: any) => {
     const { ...newState } = prevState
-    // If a user is upgrading from pre 0.11.0.  They will not have `ledger` in their redux store - so we need to check that it exists.
-    if (newState.ledger) {
-      Object.keys(newState.ledger.devices).forEach((deviceId) => {
-        ;(newState.ledger as LedgerState).devices[
-          deviceId
-        ].isArbitraryDataSigningEnabled = false
-      })
+
+    // A user might be upgrading from version without the `ledger` key in the redux store - so we
+    // initialize it here if that is the case.
+    if (!newState.ledger) {
+      newState.ledger = {
+        currentDeviceID: null,
+        devices: {},
+        usbDeviceCount: 0,
+      }
+      return newState
     }
+
+    Object.keys(newState.ledger.devices).forEach((deviceId) => {
+      ;(newState.ledger as LedgerState).devices[
+        deviceId
+      ].isArbitraryDataSigningEnabled = false
+    })
 
     return newState
   },
--- a/background/main.ts
+++ b/background/main.ts
@@ -85,6 +85,7 @@ import {
   SignDataRequest,
 } from "./utils/signing"
 import {
+  LedgerState,
   resetLedgerState,
   setDeviceConnectionStatus,
   setUsbDeviceCount,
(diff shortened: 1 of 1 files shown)


background/main.ts
```diff
@@ -211,7 +212,14 @@ const REDUX_MIGRATIONS: { [version: number]: Migration } = {
...
     const { ...newState } = prevState
-    newState.ledger.isArbitraryDataSigningEnabled = false
+    // If a user is upgrading from pre 0.11.0.  They will not have `ledger` in their redux store - so we need to check that it exists.
+    if (newState.ledger) {
+      Object.keys(newState.ledger.devices).forEach((deviceId) => {
+        ;(newState.ledger as LedgerState).devices[
+          deviceId
+        ].isArbitraryDataSigningEnabled = false
+      })
+    }
```
Shadowfiend, line 222: Do we just want to add the ledger key in this case so we're not having to deal with the potential for a missing key forever?

0xDaedalus: That is an interesting question and I think it comes down to how want to handle the general problem of a user potentially updating from an unknown point in the past.

If we go the path of adding keys that may be missing we might be introducing the following workflow for a contributor:
1) Make a change that needs a migration.
2) Check all previous migrations to make sure that that key of the key has been added in some migration. 
3) If no migrations add that key - add the key - otherwise don't.

Step 2 seems like it might be introducing implicit knowledge that's needed to successfully contribute here.

On the other hand if we don't go down the path of adding keys the flow might look like:
1) Make a change that needs a migration
2) Assume _all_ keys might not exist (I think we can codify this with stricter types.  We did this in previous migrations with `Record<string, unknown>` - I'm actually the lazy pup that introduced the pattern of `prevState: any`).
3) Add logic to skip the migration if the keys it depends on don't exist

Step 2 & 3 together seem to have less of an implicit knowledge requirement for contributing - but likely results in more code-per-migration.

I guess another approach might be to skip step 2) in the former example - and _always_ reconstruct state that might not exist.

I'm not sure of the best way forward here - curious to hear your (and others') thoughts.

Shadowfiend: I'm thinking step 2 is only a problem if we don't always migrate for new keys. If the rule is "new key = migration" then there isn't anything in the distant past to look at. 

How to enforce that there is a migration is an interesting question, but I think we can rely on code review in the short term?

Certainly stricter types and optional chaining is also possible, but could compound into gnarly handling of potential undefined everywhere when they don't carry any semantic value.

0xDaedalus: > but I think we can rely on code review in the short term?

I think this is a fine solution with the caveat that it makes me nervous since a miss in code-review will render the wallet unusable for some users.  Maybe as part of our QA process for releases we can add a step for a user to download v 0.10.0 and then upgrade to the new release version.  Anyway - pushing forward for now - will update and comment.

greg-nagy: - short term
  - Added migration check for the release QA plan, thanks for that 🤗
  - > If the rule is "new key = migration" then there isn't anything in the distant past to look at.
    - I think this mindset/rule of thumb could be good and be done until long term solution happens during review
- long term I would like to move the redux store into its own UI service and would like to change the persistence layer to be in indexedDB. That way handling the migration would be the same as with any other service

greg-nagy (approved):
let's 🚀

PR #1330, opened 2022-04-18
Back button improvements

### LINKED ISSUE #1323: Add back button in unlock screen
### Discord Discussion Link

na

Chrome

Go back from unlock screen

### What did not work?

There's no back button here. 

Screen Shot 2022-04-15 at 10 16 07

### Version

Community Edition v0.13.0

### Relevant log output

_No response_

jagodarybacka:
Resolves #1323
### What was done
* improved how Back button component handles going back in history
* going back to root path on onboarding page and setting password
* rejecting pending transaction and going back on password unlock page

### Testing
* Test back buttons on transaction approval flow
* Test back buttons on onboarding flow
<div style="display: flex">
image
image
</div>

### DIFF
--- a/ui/components/Keyring/KeyringSetPassword.tsx
+++ b/ui/components/Keyring/KeyringSetPassword.tsx
@@ -54,7 +54,7 @@ export default function KeyringSetPassword(): ReactElement {
   return (
     <section className="standard_width">
       <div className="top">
-        <SharedBackButton />
+        <SharedBackButton path="/" />
         <div className="wordmark" />
       </div>
       <h1 className="serif_header">First, let&apos;s secure your wallet</h1>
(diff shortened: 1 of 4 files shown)


jagodarybacka:
@henryboldi 
> how about we don't push the password page to the route history?

`keyring/unlock` is separate route, as I am aware react router can either override current path in the history or push new entry to the history (please let me know if this is not true). Neither of these solutions is good in our case, right? 

Making unlock page not related to routes is not an option I think and as we already had solution with back button redirecting to custom path I thought it is the easiest fix.

henryboldi:
Navigating to a React Router route doesn't have to add to the history. For example, you can use `<Redirect>` without the push property https://v5.reactrouter.com/web/api/Redirect/push-bool

henryboldi (approved):
Tested locally and it's smooth!! ✅👑 Awesome work

As an alternative solution, maybe in the future, we can explore the possibility of excluding the set password page from the route history. That way, the back button will work more closely to what's expected, rather than sending users to the home tab in all circumstances.

PR #1335, opened 2022-04-19
Adjust source map configs for performance and reproducibility

Shadowfiend:
Switch to eval source maps in development builds, which webpack claims
are more performant for rebuilds. At the same time, turn off source maps
for Firefox production builds, since these need to be fully reproducible
by the FF add-on verification team, and source maps currently inject
absolute path data into the packaged files, which makes the build
non-reproducible in a different environment.

Future work would ideally allow investigating how to have source maps
without injecting absolute paths.

## Testing

Run `yarn build --config-name firefox` in your clone. Then, clone to a
completely different place, maybe even on a different machine, and do the
same. Then check the generated ZIP files to ensure their unzipped contents
are textually identical (I used [`ksdiff`](https://kaleidoscope.app) for this step).

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -19,7 +19,7 @@ const supportedBrowsers = ["brave", "chrome", "edge", "firefox", "opera"]
 
 // Replicated and adjusted for each target browser and the current build mode.
 const baseConfig: Configuration = {
-  devtool: "source-map",
+  devtool: "eval",
   stats: "errors-only",
   entry: {
     ui: "./src/ui.ts",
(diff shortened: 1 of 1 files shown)


greg-nagy (changes requested):
Couldn't see the absolute paths in the dist output, but the CSP no-eval does not let the extension code run in dev mode. 

Could Matt's docker file idea work? 
Would we still need to solve this if we didn't minify the source? 

CleanShot 2022-04-20 at 12 21 50

Shadowfiend:
Derp, my bad. I switched to eval source maps because they are faster to rebuild, but forgot to double check it because that wasn't part of the main thing I was trying to achieve 🤦 

We can switch back to standard source maps without affecting any of the primary fix here.

It's possible we can use a docker container to get around it, but I'm not actually convinced it's definitely the case and I'm trying to eliminate yet another round of back and forth with them if possible.

greg-nagy:
I love the faster rebuild and would like to keep it if possible. 
Could we just use manifest merging and allow eval in dev mode?

Shadowfiend:
> Could we just use manifest merging and allow eval in dev mode?

Yes, but it would prevent us from finding actual things the CSP would block in production mode, so I don't think it's the path we want to take.

mhluongo (approved):
Dev builds aren't impacted, and the Firefox production build doesn't appear to include an absolute path!

PR #1340, opened 2022-04-20
Signing page improvements

jagodarybacka:
### What has been done
* improved `SharedSkeletonLoader` component 
  - made it more configurable via props
  - added animation
  - made it handle showing loaded data internally - based on `isLoaded` prop  
* refactored code around signing data and transaction
  - removed rendering empty React Fragments when transaction details are not yet loaded
  - handled `undefined` values that can happen because of lack of transaction data
  - all this refactoring was to make it possible to display meaningful UI while app is waiting for transaction data and to render `SkeletonLoaders` where it is impossible to display data.
* improved how data sign page is loading 
  - added skeleton loaders to the places that are loading a little bit longer
* improved how transaction sign page is loading - added loading page
  - as we have multiple transaction pages that are really similar to each other we can display loading page while app is loading transaction data. Loading page uses the same layout as transaction pages.

Screenshot 2022-04-22 at 14 18 43

### Testing
1. Sign message testing
Go to uniswap and try to swap USDC - click on `Allow uniswap protocol use your usdc`
2. Sign transaction testing
Try to swap assets or send eth to another address

**Expected:** loading screen should be visible for ~1s, then transaction should be handled as usual


henryboldi:
On initial checks, this appears to work properly. I know we discussed (offline) expediting the merge of this PR. However, admittedly, "refactored code around signing data" seems essential to QA rigorously. And I don't want to rush in a merge before the more detailed description is written

Shadowfiend:
> For posterity, would it be helpful to add the motivation and choices around "refactored code around signing data" here?

In general this should be in the commit messages as well! Answering the question “why” is at least as important as answering the question “what” :)

jagodarybacka:
@henryboldi Thank you! Description and commit messages fixed 🛠

ui/pages/SignDataDetailPanel.tsx
```diff
@@ -1,52 +1,90 @@
...
+  typedDataRequest,
+}: {
+  typedDataRequest: EnrichedSignTypedDataRequest
+}): ReactElement {
+  const {
+    typedData: { message },
+  } = typedDataRequest
   const keys = Object.keys(message)
 
+  if (keys.length > 2)
```
henryboldi, line 34: Why key length over 2?

henryboldi (changes requested):
Merging blocked by a regression to "Read-only accounts cannot sign"'s display.

Anyway, the UX improvement here is awesome! Amazing work!! 😁👑

<hr/>

### Non-blocking thoughts


I think these changes generally imply that if something is `undefined` then it's loading. I don't know if that's a correct assumption for us to make. How would we then handle a timeout or another kind of error? So if `undefined` doesn't mean loading, then we don't need to make all of these props optional, and we don't need to add a bunch of optional chaining. Well, the optional chaining might be unavoidable given that `SharedSkeletonLoader` accepts children.

What do you think of adding something to the effect of an `isLoading` prop with a [discriminating union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions)? So the data is required unless hypothetically `hasError` is true or `isLoading` is true.

ui/components/SignTransaction/SignTransactionContainer.tsx
```diff
@@ -145,7 +154,7 @@ export default function SignTransactionContainer({
                   {confirmButtonLabel}
                 </SharedButton>
               ))}
-            {!signerAccountTotal.signingMethod && (
+            {!signingMethod === null && (
```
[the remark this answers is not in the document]
greg-nagy: I think `bool === null` will always be false.

ui/components/Shared/SharedSkeletonLoader.tsx
```diff
@@ -2,29 +2,48 @@ import React, { ReactElement } from "react"
 import classNames from "classnames"
 
 export default function SharedSkeletonLoader(props: {
-  width: number
-  height: number
+  width?: number
+  height?: number
+  borderRadius?: number
+  margin?: string
```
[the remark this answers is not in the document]
greg-nagy: SharedSkeletonLoader.defaultProps might be a good idea here

ui/components/SignData/SignTypedDataInfo.tsx
```diff
@@ -7,18 +7,25 @@ import { EnrichedSignTypedDataRequest } from "@tallyho/tally-background/services
...
 type SignTypedDataInfoProps = {
-  typedDataRequest: EnrichedSignTypedDataRequest
+  typedDataRequest?: EnrichedSignTypedDataRequest
 }
 
 export default function SignTypedDataInfo({
   typedDataRequest,
 }: SignTypedDataInfoProps): ReactElement {
-  const { typedData, annotation } = typedDataRequest
+  const { typedData, annotation } = typedDataRequest ?? {
```
greg-nagy, line 16: I think this is also a good canidate for defaultProps

ui/pages/SignDataDetailPanel.tsx
greg-nagy: I think this comes from the deleted logic below from L38
The why is a good question though.

ui/components/SignTransaction/SignTransactionContainer.tsx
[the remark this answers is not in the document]
jagodarybacka: oh no, I missed that `!`... Thank you for noticing

ui/components/Shared/SharedSkeletonLoader.tsx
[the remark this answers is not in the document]
jagodarybacka: @henryboldi in this case I think `width`, `height` and `borderRadius` are important properties of this component, `customStyles` I would say should be just additional styles that can be whatever. 
As I'm thinking about it maybe margin could be included in `customStyles` but I would leave it as it is for now.

ui/components/SignData/SignTypedDataInfo.tsx
jagodarybacka: Turns out it is not even possible to get `typedDataRequest =  undefined` here 🙃

ui/pages/SignDataDetailPanel.tsx
jagodarybacka: So it was just moved/refactored - no regression should be visible. Why - as I can see from the commits history it was needed to implement new design of these fields. But I'm not sure if decision based on keys length is the best way to do it.

greg-nagy (approved):
Loving this feature :) 

Let's 🚀

PR #1343, opened 2022-04-21
🍁 420: Add commit hash to versions on menu page 

henryboldi:
By setting an env variable via the build GitHub action, this PR adds displaying
the first 7 digits of the triggering commit's SHA on the menu page. As we move
towards continuous deployment, some folks are doing QA on main via an artifact.
Since `main` is fluid between versions, using just the current version number
is insufficient to specify a build for a bug report.

version


greg-nagy:
@henryboldi this will make local build different from CI build so this will prolong the FF issue. 

I am not yet sure what would be a good solution for this problem, but an idea
- injecting the git hash into a file at version bump time
- in CI inject the current hash into the file overwriting the one there 
- if we are in a git repo and the hash is different from the one in the file > use the local git hash
- if we are not in a git repo then use the hash from the file

cc @Shadowfiend

Shadowfiend:
For now just don't use it on the production build, I would say. We can handle hash juggling with my work to auto submit artifacts, which can handle submitting sources with an appropriate env file as well.

PR #1345, opened 2022-04-21
Rename Addresses

0xDaedalus:
This PR adds the ability for users to rename their addresses!

https://user-images.githubusercontent.com/94649004/164518501-3dcbbded-3d12-418d-bbd7-56eb7e6284a8.mov

Some notes:
- Transactions in the Activity List are not retroactively updated - this will require going back through every transaction associated with a given address and re-annotating it.
- Once names are set they cannot be removed - only changed.  I didn't see much point of letting users go back to their arbitrarily assigned tally wallet names.

### To Test
- Change an addresses name as per the above video.
- Confirm that the name changes propagate throughout the application (with the above caveat in mind).


henryboldi:
Alternatively, we can put it behind a feature flag, so the PR can be reviewed + merged without the feature being entirely complete!

mhluongo:
https://giphy.com/gifs/why-not-both-dont-we-have-3ohfFhG5VDtDTzQv2o

0xDaedalus:
> Instead of suffixing [DRAFT] might I suggest you convert this to a draft?

Whoopsie forgot to change the title when converting it out of draft

mhluongo:
Giving this a look! I think we should try to merge #1347 first as @mul1sh has been working on it for a while, you okay with that @0xDaedalus?

ui/components/Shared/SharedInput.tsx
```diff
@@ -75,6 +77,13 @@ export function SharedTypedInput<T = string>(props: Props<T>): ReactElement {
           handleInputChange(event.target.value)
         }
         onFocus={onFocus}
+        onKeyDown={(e) => {
+          if (e.key === "Enter" && typeof onPressEnter !== "undefined") {
```
mhluongo, line 81: This sort of surprises me — are we using key events across the whole extension?

ui/components/Shared/SharedAccountItemSummary.tsx
```diff
@@ -19,6 +23,8 @@ export default function SharedAccountItemSummary(props: Props): ReactElement {
     localizedTotalMainCurrencyAmount,
   } = accountTotal
 
+  const dispatch = useDispatch()
```
mhluongo, line 26: Can this be dropped?

mhluongo:
Hmm, looking more closely, this might not even affect that PR.

ui/components/Shared/SharedInput.tsx
Shadowfiend: `onSubmit` and forms is def The Way™ here.

background/services/preferences/index.ts
```diff
@@ -76,8 +83,25 @@ export default class PreferenceService extends BaseService<Events> {
...
-  // TODO Implement the following 4 methods as something stored in the database and user-manageable.
+  // TODO Implement the following 6 methods as something stored in the database and user-manageable.
   // TODO Track account names in the UI in the address book.
+
+  addOrEditNameInAddressBook(newEntry: AddressBookEntry): void {
+    const correspondingEntryIndex = this.addressBook.findIndex((entry) =>
+      sameAddressBookEntry(newEntry, entry)
+    )
+    if (correspondingEntryIndex !== -1) {
+      this.addressBook[correspondingEntryIndex] = newEntry
```
Shadowfiend, line 94: Super excited for this, but… Pretty sure this won't survive browser restarts/extension reloads. For this address book to persist I think we need to move it to IndexedDB :grimacing: (I thought this was a nice-to-have, but fairly certain it's a req.)

On the bright side, once it's in IndexedDB you can just do direct lookups using the indices :grin:

0xDaedalus: The name seems to be persisting for me on extension upgrade, browser close + open, as well as full computer restart

henryboldi (changes requested):
- [x] Functions as expected and persists
        - Persists on close
        - Persists on restart extension
        - Persists on browser relaunch
- [x] The code has been updated based on the feedback

Excellent work sir! 👏👑

However, I'm seeing this visual bug. It looks like the bottom of the buttons are cut off. I think the spacing just needs to be adjusted.
edit name

0xDaedalus:
Good catch - tyty.  Updated visual: 
image

henryboldi (approved):
👑👑👑

PR #1347, opened 2022-04-21
UNS domain resolution support

mul1sh:
Replaces #1137

Will fix https://github.com/unstoppabledomains/gitcoin-bounties/issues/3

@mhluongo @Shadowfiend @0xJordi this is the final PR for UNS domain support.

Apologies for opening & closing alot of PR's before 🙁 , the reason for this is that I always get the error shown in the attached picture, when I rebased my changes from the upstream `origin/main` and this crashed tally for me locally. The solution for this has been to always close the PR, delete the old branch and then reopen a new branch and then do a new PR from there.

Since this is more or less complete and upto date with main, I don't expect to do this again. 🙂 

Screenshot_2022-04-20_at_9 34 41_PM


mul1sh:
Btw here is a video showing the functionality, really looking forward to the review  😃 

https://user-images.githubusercontent.com/2150634/164540802-9645f199-947c-4455-9ca9-66ea72ffb44c.mov

yarn.lock
```diff
@@ -2781,6 +2965,18 @@
   resolved "https://registry.yarnpkg.com/@uniswap/token-lists/-/token-lists-1.0.0-beta.25.tgz#e4279cf2581b297723d82eafd1bed35273070f07"
   integrity sha512-XemgUs8YiYP7Xci+Wb8vPSnyxV9aTO6J5XLPzhZmgDFO1RY5JMnB0As8PJFia3Pg9JSRZ7+AahzxowPYhnUVWw==
 
+"@unstoppabledomains/resolution@^7.1.4":
```
mhluongo, line 2968: I'm pretty sure you aren't using this in the PR — maybe revert `yarn.lock` outside what `node-fetch` needs? This is a pretty big dependency, and will likely require some tweaks to build and list on Firefox

background/services/name/resolvers/uns.ts
```diff
@@ -0,0 +1,169 @@
...
+import { AddressOnNetwork, NameOnNetwork } from "../../../accounts"
+import { ETHEREUM } from "../../../constants"
+import { sameNetwork } from "../../../networks"
+import { NameResolver } from "../name-resolver"
+
+/**
+ * Lookup a UNS domain name and fetch the owners address
+ */
+const lookupUNSDomain = async (domain: string) => {
+  const ALCHEMY_API_KEY = process.env.ALCHEMY_KEY
```
mhluongo, line 12: Nice hack, cool that Alchemy has support. We can do "native" resolution down the road.

mhluongo (commented):
Left a couple comments. Will leave greater review to @greg-nagy @Shadowfiend

background/services/name/resolvers/uns.ts
```diff
@@ -0,0 +1,169 @@
...
+      }
+    },
+    async lookUpAvatar(
+      addressOrNameOnNetwork: AddressOnNetwork | NameOnNetwork
+    ) {
+      const { network } = addressOrNameOnNetwork
+      const { address } =
+        "address" in addressOrNameOnNetwork
+          ? addressOrNameOnNetwork
+          : { address: undefined }
```
Shadowfiend, line 112: Thoughts on adjusting this branch to attempt to resolve? Eg:

```suggestion
          : await this.lookUpAddressForName(addressOrNameOnNetwork) ?? { address: undefined }
```

background/services/name/resolvers/uns.ts
```diff
@@ -0,0 +1,169 @@
...
+      if (unsRecords.length > 0) {
+        unsRecords.forEach(
+          (record: { attributes: { records: { [x: string]: never } } }) => {
+            const nftPicture = record.attributes.records["social.picture.value"]
+            if (nftPicture !== undefined || nftPicture !== null) {
+              avatarUrn = nftPicture
+            }
+          }
+        )
+      }
```
Shadowfiend, line 136: Not blocking, but this can be clarified with a `.map` + `.find`; roughly:

```typescript
      const data = (await reverseLookupAddress(address))?.data

      // check if any of the records returned has an NFT picture associated with it
      const avatarUrn = data
        .map((record: { attributes: { records: { [x: string]: string | undefined } } }) => {
          record.attributes.records["social.picture.value"]
        })
        .find(isDefined /* from lib/type-guards */)
```

background/services/name/resolvers/uns.ts
```diff
@@ -0,0 +1,169 @@
...
+
+      if (avatarUrn === undefined || avatarUrn === null) {
+        return undefined
+      }
+      // modify the nft picture value and make it a correct avatar urn
+      avatarUrn = avatarUrn.replace("1/erc721", "eip155:1/erc721")
+      avatarUrn = avatarUrn.replace("1/erc1155", "eip155:1/erc721")
```
Shadowfiend, line 143: Believe we should anchor these to the start so we don't replace a valid URN should one appear, plus I think we can use a capturing group to do this once:

```suggestion
      avatarUrn = avatarUrn.replace(/^1\/(erc721| erc1155)/, "eip155:1/$1")
```

Wondering also if we should allow NameService to accept erc1155 strings so we don't have to impersonate erc721 here...

Anyway, only the starting anchor is blocking here I think.

background/services/name/resolvers/uns.ts
```diff
@@ -0,0 +1,169 @@
...
+      }
+    },
+    async lookUpNameForAddress({
+      address,
+      network,
+    }: AddressOnNetwork): Promise<NameOnNetwork | undefined> {
+      // Get all the records associated with the particular ETH address
+      const data = await reverseLookupAddress(address)
+      // Since for a given address you can have mutliple UNS records, we just pick the first one
+      const name = data.data[0]?.attributes?.meta?.domain
```
Shadowfiend, line 157: Do we want to do a similar find as above where we pull the first defined one in case of weird records?

Shadowfiend (changes requested):
This is wonderful! And reverse resolving too, yay!

I left a few stylistic notes and a couple of implementation ones. Tried to call out the very few that should block merging the PR. 

Also wondering if we drop this behind a feature flag for testing/release timing purposes cc @greg-nagy @mr-michael .

background/package.json
```diff
@@ -57,6 +57,7 @@
   },
   "devDependencies": {
     "@reduxjs/toolkit": "^1.6.1",
+    "@types/node-fetch": "^2.6.1",
```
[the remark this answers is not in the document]
mul1sh: If I remove it, I get the following error

Screenshot 2022-04-22 at 10 14 11 AM

background/services/name/resolvers/uns.ts
```diff
@@ -0,0 +1,162 @@
...
+            attributes: { records: { [x: string]: string | undefined } }
+          }) => record.attributes.records["social.picture.value"]
+        )
+        .find(isDefined)
+
+      if (avatarUrn === undefined || avatarUrn === null) {
+        return undefined
+      }
+      // modify the nft picture value and make it a correct avatar urn
+      avatarUrn = avatarUrn.replace(/^1\/(erc1155 || erc721)/, "eip155:1/$1")
```
Shadowfiend, line 131:
```suggestion
      avatarUrn = avatarUrn.replace(/^1\/(erc1155|erc721)/, "eip155:1/$1")
```

I believe this is the only way the regex will work :point_up_2: but haven't triple-checked.

Shadowfiend (changes requested):
One small note on regex syntax, and I;ll to take this for a test drive today!

Any sample UNS domains that have known avatar info so I can test that piece?

mul1sh:
> One small note on regex syntax, and I;ll to take this for a test drive today!
> 
> Any sample UNS domains that have known avatar info so I can test that piece?

Yes 😃, please try with mine `mulili.crypto`

mul1sh:
> One small note on regex syntax, and I;ll to take this for a test drive today!
> 
> Any sample UNS domains that have known avatar info so I can test that piece?

Yes 😃, please try with mine `mulili.crypto`

Shadowfiend:
Avatar resolution seems to work, so does both forward and reverse resolution 💪 

One thing I noticed when adding a read-only address for `brad.crypto` is that we reverse-resolve to a different name in the extension:

Screen Shot 2022-04-22 at 11 57 36

There are some unrelated issues here (e.g., we need to handle longer names better in the UI), but it seems like sorting by id `ASC` over `DESC` when doing the `/owners` lookup will put `brad.crypto` first—I'm not sure what that sort means in practice, is it reasonable to switch to `ASC` or will that break other stuff?

mul1sh:
> Avatar resolution seems to work, so does both forward and reverse resolution 💪
💯 
> 
> One thing I noticed when adding a read-only address for `brad.crypto` is that we reverse-resolve to a different name in the extension:

Yes this is because the address associated with `brad.crypto` has many other UNS domains associated with it, but as noted below sorting ASC will fix this.
> 
> Screen Shot 2022-04-22 at 11 57 36
> 
> There are some unrelated issues here (e.g., we need to handle longer names better in the UI), but it seems like sorting by id `ASC` over `DESC` when doing the `/owners` lookup will put `brad.crypto` first—I'm not sure what that sort means in practice, is it reasonable to switch to `ASC` or will that break other stuff?

Nope it won't break anything, let me switch to this right now 🙂

Shadowfiend:
Sweet. I've captured the stacked issues that showed the brad.crypto problem as #1353 and #1354.

I'm going to drop this behind a feature flag and then merge so some of the non-dev folks get a chance to test it out before we go live. Thank you for all the work and the patience! I know this has been in the pipeline for a while (almost a month 😵) and appreciate your sticking with us as we fixed up the architecture a bit to make it work :pray:

mul1sh:
> Sweet. I've captured the stacked issues that showed the brad.crypto problem as #1353 and #1354.
> 
> I'm going to drop this behind a feature flag and then merge so some of the non-dev folks get a chance to test it out before we go live. Thank you for all the work and the patience! I know this has been in the pipeline for a while (almost a month 😵) and appreciate your sticking with us as we fixed up the architecture a bit to make it work 🙏

No worries, I didn't mind at all. And thank you too for the quick reviews, I much appreciate it. Looking forward to seeing it when it goes live 🙂 . Have a good one.

Shadowfiend (approved):
Let's go!

SJQuality:
Retested UNS to make sure it is still working properly on Main Build #2950

PR #1355, opened 2022-04-23
😊 The future is optimistic: Refactor ChainService to support Arbitrum and Optimism

mhluongo:
Add partial support for Arbitrum and Optimism in `ChainService`, guarded by a feature flag.

To test:

- [ ] Without the feature flag `MULTI_NETWORK` set, Make sure balances are still pulled when importing a new account from a Ledger
- [ ] Without the feature flag `MULTI_NETWORK` set, confirm account balances and asset transfers still function on mainnet
- [ ] Setting `MULTI_NETWORK` to `true` in `.env`, confirm account balances and transfers still function on mainnet
- [ ] After startup, you should see blocks from Optimism and Arbitrum in IndexedDB
- [ ] Make sure balances are still pulled when importing a new account from a Ledger
- [ ] Confirm gas prices populate work when sending ETH


background/services/chain/index.ts
```diff
@@ -6,6 +6,7 @@ import {
 import { getNetwork } from "@ethersproject/networks"
 import { utils } from "ethers"
 import { Logger } from "ethers/lib/utils"
+import { camelCase } from "lodash"
```
greg-nagy, line 9: It just cracked me up how not trivial this is and we need lodash for this 

https://stackoverflow.com/questions/2970525/converting-any-string-into-camel-case

background/services/chain/index.ts
```diff
@@ -192,28 +190,34 @@ export default class ChainService extends BaseService<Events> {
...
-          ),
-        () =>
-          new AlchemyProvider(
-            getNetwork(Number(this.ethereumNetwork.chainID)),
-            ALCHEMY_KEY
-          )
-      ),
-    }
+    this.supportedNetworks = USE_MULTI_NETWORK
+      ? [ETHEREUM, ARBITRUM_ONE, OPTIMISM]
```
greg-nagy, line 194: this is `EVM_MAIN_NETWORKS`

background/services/chain/index.ts
```diff
@@ -192,28 +190,34 @@ export default class ChainService extends BaseService<Events> {
...
+      : [ETHEREUM]
+
+    this.providers = Object.fromEntries(
+      this.supportedNetworks.map((network) => [
+        camelCase(network.name),
```
greg-nagy, line 199: This camelCase feels weird to me as a way to index the the providers. 
This makes it possible to write `this.providers.arbitrumOne.send()` and when we decide we remove arbitrum one the code will break. 

Put it another way. It feels that we should have a defined list of networks and use properties on them 

I was wondering what would be the simplest way to implement camelCase for this usecase:
`"Arbitrum One".split(" ").map((el,ind) => (ind === 0 ? el[0].toLowerCase() : el[0].toUpperCase()) +  el.slice(1)).join("")`

greg-nagy (changes requested):
It was fun to review this PR and also this is good addition. Ty 

When turning on the feature flag it breaks the gas estimation and everything fails that needs to be signed because of it

CleanShot 2022-04-27 at 14 33 00

And in `tally/chain` > blocks only `Arbitrum One` blocks seem to be stored. (Haven't dug deep enough to find out what is actually going on)

CleanShot 2022-04-27 at 14 44 22


This is behind feature flag, so I would be ok to merge this after recorded the state in an issue (happy to create the issue) but not merging because the test scenarios you have defined are failing. 

Would you like to fix it or merge as is?

background/services/chain/index.ts
mhluongo: Yeah, I've done this one by hand a few times... then saw that already had `lodash` in our dep tree... :see_no_evil:

background/services/chain/index.ts
mhluongo: It is for now! It likely won't be in the future though, as different services might have different support.

mhluongo:
Hmm, I definitely want to fix that issue before merge, regardless of the feature flag. I'll take a turn here.

background/services/chain/index.ts
mhluongo: Oof, my intent here was to add the same version we already had from the lockfile to `package.json` as good practice.. going to remedy this, we don't need new library code for `camelCase()` lol.

When lockfiles are sued correctly, the content hash can protect us from some supply chain issues. Longer term we'll want to explore Lavamoat + some other techniques to tame the dep tree, but  it's out of scope for where we are now IMO.

greg-nagy: I think we are good here. My understanding right now, is that yarn satisfies all the above version needs — coming from different libs — with the `4.17.21` version of the lib. And that's the version in `node_modules` also.

henryboldi: > saw that already had lodash in our dep tree... 🙈

Shouldn't we still explicitly list `lodash` as a package in package.json? If whatever package that we're using ends up dropping `lodash` then we'll break

mhluongo: Yep, and that was in face my intent. Fixing this :)

Shadowfiend: The real question I was going to ask is... Why camel case at all? This should be accessed as data via a constant rather than as direct property accesses, so why force camel casing?

henryboldi:
Just chiming in to say that I love how this PR is a behind a feature flag, rather than waiting for full implementation 😍

background/services/chain/index.ts
mhluongo: I was under the impression that you wanted to maintain this style of access, but if you don't care and no one likes it... :bomb:

Shadowfiend: Yeah nope. I'll reply in Greg's thread below, which I missed yesterday.

background/services/chain/index.ts
Shadowfiend: I think Matt's right that we will find ourselves at times wanting access to particular networks for particular things; however, I think we'll define that using a central set of constants and methods that interact with those rather than through direct named property access. In particular, our network constants let us look up providers by passing the relevant constant in, and that should be the correct path. Any direct property access by name should set off alarm bells, as Greg pointed out, because the property access is effectively unchecked *and mostly uncheckable*.

mhluongo: Should be resolved!

mhluongo:
This is ready for another look! Note I'm seeing blocks from Optimism, Arbitrum, and mainnet in IndexedDB locally.

I'm still looking into the block price issue though. Someone confirming that would be :ok_hand:

greg-nagy (approved):
Tests

- [x] Without the feature flag `MULTI_NETWORK` set, Make sure balances are still pulled when importing a new account from a Ledger
- [x] Without the feature flag `MULTI_NETWORK` set, confirm account balances and asset transfers still function on mainnet
- [x] Setting `MULTI_NETWORK` to `true` in `.env`, confirm account balances and transfers still function on mainnet
- [x] After startup, you should see blocks from Optimism and Arbitrum in IndexedDB
- [x] Make sure balances are still pulled when importing a new account from a Ledger
- [x] Confirm gas prices populate work when sending ETH

Looks good :)

SJQuality:
Behind a feature flag. This looks good with it off and through smoke testing.

PR #1358, opened 2022-04-25
Shared banner component refactor

jagodarybacka:
### What was done
* added possibility to close banner
* moved styling of banner content outside of component

Banner with close button can look like this ⬇️  (it is not included in the PR). Banners with close icon will be added later - as designed.
image


wojciech-turek (approved):
Works and looks good!

PR #1360, opened 2022-04-26
Improve styles of SharedBanner

jagodarybacka:
### What was done

Minor improvements to make Shared Banner easier to use.

- add possibility to apply custom styles
- fix alignment of close button and text when there is only one line of text

### DIFF
--- a/ui/components/Shared/SharedBanner.tsx
+++ b/ui/components/Shared/SharedBanner.tsx
@@ -7,6 +7,7 @@ type BannerProps = {
   icon?: string
   iconColor?: string
   iconAriaLabel?: string
+  customStyles?: string
 }
 
 type CanBeClosedProps =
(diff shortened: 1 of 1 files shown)


ui/components/Shared/SharedBanner.tsx
```diff
@@ -47,13 +52,18 @@ function Banner(props: BannerProps): ReactElement {
 }
 
 function BannerWithClose(props: BannerProps & { id: string }): ReactElement {
-  const { id, children, icon, iconColor, iconAriaLabel } = props
+  const { id, children, icon, iconColor, customStyles, iconAriaLabel } = props
```
henryboldi, line 55: Does the naming `customStyles` successfully imply that the supplied css will be appended rather than replaced?

henryboldi (approved):
Looks great!! 👏

PR #1363, opened 2022-04-27
🥗 Resolve CPU issues from `blocks` bloat in Redux store

### LINKED ISSUE #1351: Redux store constantly growing > the extension slowing down
(From @Shadowfiend:

"Strongly recommend looking at https://github.com/tallycash/extension/blob/680c0ec134664362b50cadb71ae183b96f0727af/background/redux-slices/assets.ts#L17 and implementing that TODO as a starting point.")

https://coda.io/d/Minimum-Viable-Launch_dznBnP8Wtkc/RFC-Performance-improvements_subOr#_lugBs

henryboldi:
When having Tally Ho installed for a significant period of time, the CPU usage
increases to pretty extreme levels. On the Chrome Store/production install I
have on Brave, I see 70,922 blocks and 114,214 balance entries saved in
IndexedDB. These blocks and balance entries accumulate over time. While
IndexedDB isn't the culprit of the CPU issues, I believe a significant portion
is making its way into the Redux store. The massive Redux store I believe is
the culprit for extreme CPU usage.

This PR removes the dependance on storing blocks in Redux. Their only
current use is for deriving an activity's timestamp. That approach is replaced
by [adding the transaction's associated block's timestamp to the transaction
annotation via the `EnrichmentService`](https://github.com/tallycash/extension/pull/1363/commits/161ec56f5a4f81572be3d401cd18f0eac410c946). Since individual blocks are no longer
used in the FE, this PR removes saving them in the networks slice at all (aside
from the latest `blockHeight`). By no longer including a huge number of blocks in
the Redux store, this change should significantly ameliorate the CPU issue.

#1351

### Testing instructions
1. Test to see that timestamps work on activities
  - On the home tab
  - On the activity detail panel
2. With Redux dev tools, ensure that no `blocks` are being added to the Redux
store (networks → evm["1"] → blocks)
3. Test the migration by checking out an older version of Tally Ho, let the block
data and activities load in, and then update to this branch 

Closes #1351


0xDaedalus (commented):
👑  Love how you link to commits in the PR description - going to steal that.

Know this is a draft - just dropping some initial comments.

background/redux-slices/activities.ts
```diff
@@ -87,6 +87,13 @@ const activitiesSlice = createSlice({
           )
         } else {
           activitiesAdapter.upsertOne(immerState[address], activityItem)
+          // Limit the number of activities in redux to 45
+          if (immerState[address].ids.length >= 45) {
```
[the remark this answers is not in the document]
henryboldi: Great question. `45` was pretty arbitrary. It was just to explore the impact it'd have. In reality, I don't think the number of activities is a top concern. Activity accumulation is fairly slow, and it's rare to get out of hand.

background/main.ts
```diff
@@ -78,7 +78,7 @@ import {
   typedDataRequest,
   signDataRequest,
 } from "./redux-slices/signing"
-
+import { blockFromEthersBlock } from "./services/chain/utils"
```
mhluongo, line 81: This is a smell

background/main.ts
```diff
@@ -778,9 +778,17 @@ export default class Main extends BaseService<never> {
...
     // Report on transactions for basic activity. Fancier stuff is handled via
-    // connectEnrichmentService
+    // connectEnrichmentService. Also dispatch block to be used for the timestamp
+    // and other supplementary info.
     this.chainService.emitter.on("transaction", async (transactionInfo) => {
       this.store.dispatch(activityEncountered(transactionInfo))
+      const block = blockFromEthersBlock(
+        transactionInfo.transaction.network,
+        await this.chainService.providers.ethereum.getBlock(
+          transactionInfo.transaction.blockHash || ""
```
mhluongo, line 788: A couple issues here

* `main.ts` shouldn't be interacting with providers like this. It's making multi-network more difficult, and it's compromising the layered architecture. This would quietly pull Ethereum blocks from non-Ethereum networks 
* The `ChainService` should make decisions on when to emit block info, not `main.ts`

background/redux-slices/networks.ts
```diff
@@ -38,11 +38,14 @@ const networksSlice = createSlice({
       ) {
         immerState.evm[block.network.chainID].blockHeight = block.blockHeight
       }
+    },
+
+    addBlock: (immerState, { payload: block }: { payload: AnyEVMBlock }) => {
```
mhluongo, line 43: I like the action being split between "here's the latest block seen" and "here's a block". But the actual pruning should be happening in the reducer rather than changing the behavior of `blockSeen` IMO.

mhluongo (changes requested):
Trying to impress the architectural lines here. Services decide what to do with data ingestion, Redux machinery is for deciding what's in the Redux store.

The price stuff looks great IMO, and is good to merge in a smaller PR. Concerns here are around the blocks.

background/redux-slices/activities.ts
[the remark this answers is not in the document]
mhluongo: So can we... not? and focus on fixing the problems we have now?

This makes my wallet personal less useful, and I think we can revisit here down the road by introducing an "infini-scroll" style lazy loading

background/redux-slices/networks.ts
henryboldi: How is it possible to know which block are needed if the activities are being added asynchronous?

background/main.ts
henryboldi: 100% knew this and did it anyway 😈 heh

background/redux-slices/activities.ts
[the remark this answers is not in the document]
henryboldi: Yup @mhluongo, it was already reverted when I posted my reply. And I had crossed out the change in the PR description.

background/redux-slices/networks.ts
mhluongo: If there's a block / activity relation we need to filter on in a reducer, are we can't... that implies we have the wrong slice structure I think.

mhluongo: I can see that being a longer-term  thing though

henryboldi:
> The price stuff looks great IMO, and is good to merge in a smaller PR.

Smart idea to separate the price stuff into a smaller PR. Will do!

background/main.ts
0xDaedalus: I see this got marked as resolved but as far as I understand it mhluongo's comment is still valid right?  We want to be exporting from `services/chain` rather than `services/chain/utils` or am I misunderstanding?

background/main.ts
```diff
@@ -230,6 +230,13 @@ const REDUX_MIGRATIONS: { [version: number]: Migration } = {
       ].isArbitraryDataSigningEnabled = false
     })
 
+    return newState
+  },
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  7: (prevState: any) => {
+    const { ...newState } = prevState
+    delete newState.networks.evm["1"].blocks // Only mainnet at this time
```
0xDaedalus, line 238: I think we need optional chaining here.  v0.10.0 which I believe is our first public release does not have `networks` in its redux slice on app load - so any users upgrading from that version or a later version that still doesn't have `networks` will have a bricked app unless they reinstall due to trying to access `evm` of `undefined`.

Lots of discussion on the topic [here](https://github.com/tallycash/extension/pull/1327#discussion_r852423241).

background/services/enrichment/types.ts
```diff
@@ -21,7 +21,11 @@ export type BaseTransactionAnnotation = {
    * When this transaction annotation was resolved. Including this means
    * consumers can more easily upsert annotations.
    */
-  timestamp: UNIXTime
+  annotatedAt: UNIXTime
+  /**
+   * The timestamp of the transaction's associated block if available.
+   */
+  timestamp: UNIXTime | undefined
```
0xDaedalus, line 28: What do you think about calling this "blockTimestamp"?  I think it'd improve clarity a bit when reading code inline w/o having to reference the above comment.

0xDaedalus (changes requested):
Henry Senpai you've outdone yourself again :bow:.  Only real blocking thing is the migration I think as well as maybe mhluongo's comment unless that has been resolved and I'm just misunderstanding.

Oh, whats the best way to test this PR?  I've clicked around and nothing seems obviously broken but curious to know if there is a better or more specific way.

background/services/enrichment/types.ts
henryboldi: I agree, good point!

henryboldi:
> Nothing blocking here except maybe mhluongo's comment unless that has been resolved

I believe they're resolved aside from longer-term ideas outside the scope of this PR 🙏

background/main.ts
henryboldi: Good catch! I meant to remove this in https://github.com/tallycash/extension/pull/1363/commits/9dba3ffcc72951053f23a12688b3380235b65963. It's not used in my changes at all anymore. Removed. ✔️

background/services/enrichment/index.ts
```diff
@@ -104,17 +105,29 @@ export default class EnrichmentService extends BaseService<Events> {
...
   ): Promise<TransactionAnnotation | undefined> {
     let txAnnotation: TransactionAnnotation | undefined
 
     const resolvedTime = Date.now()
+    let block: AnyEVMBlock | undefined
+
+    if (transaction?.blockHash) {
+      block = await this.chainService.getBlockData(
+        network,
+        transaction.blockHash ?? ""
```
mhluongo, line 122: If `transaction.blockhash` is `undefined` here, we don't want to use the block later, right?

background/services/enrichment/types.ts
mhluongo: Also nice to revert the `annotatedAt` rename and avoid the extra work.

mhluongo (commented):
Great approach, love removing all the blocks from the frontend and moving this to metadata summarized by the services. Waiting on a rename (and testing on my end) to approve, it's close!

background/services/enrichment/types.ts
henryboldi: > Also nice to revert the annotatedAt rename and avoid the extra work.

Can't say I agree with that one...

mhluongo: `timestamp` is a legacy name from the annotation, right? I don't see a migration for it...

And that said... there's also no migration to add `blockTimestamp` to existing transaction annotations, so I guess existing users won't have timestamps on their activities anymore?

mhluongo: Every rename or deletion in the Redux store is like changing a database schema in this architecture. Apologies if I'm missing something obvious, when to migrate / when we can avoid keeps surprising me.

henryboldi: You're right this needs a migration 😬

0xDaedalus (approved):
Looks good my comments have been addressed 👍

background/services/enrichment/index.ts
```diff
@@ -104,17 +105,29 @@ export default class EnrichmentService extends BaseService<Events> {
     network: EVMNetwork,
     transaction:
       | AnyEVMTransaction
-      | (Partial<EIP1559TransactionRequest> & { from: string }),
+      | (Partial<EIP1559TransactionRequest> & {
```
greg-nagy, line 108: This is a monster function. Respect for these changes 🔥

greg-nagy (approved):
Seeing what you have done here is some top level awesomeness. 

Couldn't find anything to add, so I am mostly here to express my happiness :) 

(Don't want to steal the thunder by merging this, but want to ... soo badly :) I definitely want this to be part of the next release)

PR #1372, opened 2022-04-28
Create SharedAccordion component

jagodarybacka:
### What has been done
* Created component with collapsible element. 
* It allows to add custom elements to the header - always visible, and to the collapsible content part.
* Added simple rotation animation to the arrow button. Added showing/hiding animation to the content.

It is not used anywhere yet. You can test it using this snippet.
```javascript
    <SharedAccordion
      headerElement={<div>Header</div>}
      contentElement={
        <>
          <div>element</div>
          <div>element</div>
          <div>element</div>
          <div>element</div>
          <div>element</div>
          <div>element</div>
          <div>element</div>
        </>
      }
    />
```


ui/components/Shared/SharedAccordion.tsx
```diff
@@ -0,0 +1,68 @@
...
+import React, { ReactElement, useEffect, useRef, useState } from "react"
+import SharedIcon from "./SharedIcon"
+
+export default function SharedAccordion({
+  headerElement,
+  contentElement,
+  contentHeight,
+}: {
+  headerElement: ReactElement
+  contentElement: ReactElement
```
henryboldi, line 11: I'm curious if an accordion component should accept multiple rows? This reads like a `SharedAccordionItem` to me

jagodarybacka: what do you mean by "multiple rows"? like multiple headers that each is opening different content?

henryboldi: > like multiple headers that each is opening different content?

Yes, typically accordion components allow multiple headers that each open different content.

jagodarybacka: Right, but I think we can add this later as we won't need it in the nearest future.

henryboldi (approved):
Excellent! 👏👑

PR #1373, opened 2022-04-28
The Man With No Name 🤠

0xDaedalus:
Fixes a bug where users could set the name of an account to an empty string.

### To Test
- [ ] Opening up the `Edit Name` menu and immediately clicking Save name should show an error message and not save the name.
- [ ] Opening up the `Edit Name` menu, placing cursor inside of the text input, and immediately pressing enter should show an error message and not save the name.
- [ ] Typing in a name and then deleting all characters should result in an error message being shown and submission should not be possible


ui/components/AccountItem/AccountItemEditName.tsx
```diff
@@ -32,6 +32,32 @@ export default function AccountItemEditName({
     }
   }, [newName, error, touched])
 
+  const onSubmit = React.useCallback(
```
henryboldi, line 35: Not blocking at all, but wanted to note that this project usually imports `useCallback` and `useEffect` etc directly via `import React, { useCallback, useEffect  } from "react"`

henryboldi (changes requested):
Nice! I love the consolidated `onSubmit` 🙇👑

Here's a change that might simplify this a bit. I believe 
```
  React.useEffect(() => {
    if (touched && newName.trim() === "") {
      setError("Name is required")
    } else {
      setError("")
    }
  }, [newName, error, touched])
  ```
can be removed because setting `error` is now handled in this `onSubmit` function. After this `useEffect` is removed, I don't believe there remains a use for the `touched` state, so that can be removed too.

ui/components/AccountItem/AccountItemEditName.tsx
0xDaedalus: good catch - will fix

0xDaedalus:
> After this useEffect is removed, I don't believe there remains a use for the touched state, so that can be removed too.

I don't think thats true - if we remove this the error message won't show up if someone types in a name and then deletes it without pressing enter - do you think thats something we don't need?

henryboldi:
I dig it :)

We might want to update our create password form to match, as it only checks on submit

PR #1375, opened 2022-04-28
Uns uns uns: Turn on UNS resolution feature

Shadowfiend:
This enables the UNS resolution feature flag for inclusion in the main
build of the extension.


SJQuality:
QA build #2918 Feature looks good in Chrome

greg-nagy (approved):
"left a couple of nonblocking comments, but I think this looks good" ... 😂

Let's do this! Maybe we can have a new release? :)

PR #1376, opened 2022-04-28
🥦 Prune asset prices older than a day from Redux store

henryboldi:
We have 2 charts designed, which are both Daily (Overview and Earn landing
page). More complex charts are not planned for the foreseeable future. So let's
prune prices older than a day. The price data is stored w/o this filter in
IndexedDB if we need it.

(cherry picked from commit 7f12585a84354e2a8e47cddc05d7457fe23fea0f)

#1351


mhluongo:
One thing I missed in the last PR: we should include a Redux store migration here that deletes all old `PricePoint`s, otherwise folks will keep suffering this bloat.

henryboldi:
> we should include a Redux store migration here that deletes all old PricePoints, otherwise folks will keep suffering this bloat.

But all existing users are getting new price data, and therefore all their old `PricePoint`s will be pruned, no?

mhluongo:
Great point! I guess the CPU usage will start to improve on their first price action rather than at startup, but maybe that's okay.

greg-nagy:
> Great point! I guess the CPU usage will start to improve on their first price action rather than at startup, but maybe that's okay.

I think that's totally ok. Came here to note the same, and when I realized how this works I was just like "hm, well done, dude :)  :manlynod:" :)

background/redux-slices/assets.ts
```diff
@@ -28,8 +28,13 @@ export type AssetsState = SingleAssetState[]
 export const initialState = [] as AssetsState
 
 function prunePrices(prices: PricePoint[]): PricePoint[] {
-  // TODO filter prices to daily in the past week, weekly in the past month, monthly in the past year
-  const pricesToSort = prices.map<[number, PricePoint]>((pp) => [pp.time, pp])
+  const pricesToday = prices.filter(
+    (pp) => pp.time > (Date.now() - 24 * 60 * 60 * 1000) / 1000 // 24 hours ago in UNIXTime
+  )
+  const pricesToSort = pricesToday.map<[number, PricePoint]>((pp) => [
```
greg-nagy, line 34: Was thinking a bit about skipping the intermediary variables and putting these into a chained call.

greg-nagy (approved):
This will be a good win or a good lesson to learn. 
We diverged here from the original intent: `filter prices to daily in the past week, weekly in the past month, monthly in the past year` and kept prices from the only the last day. 

I think in case of redux size it's good practise to err on the "smaller is better" side and solve the extra data fetching when we see we need it.

PR #1377, opened 2022-04-28
Hocus Focus: Implement focus and disabled styling for SharedInput and seed phrase input

Shadowfiend:
Add yellow highlight and label coloring to focused inputs:

Screen Shot 2022-04-28 at 17 58 25

And also the seed phrase textarea:

Screen Shot 2022-04-28 at 17 59 02

Also dropped some disabled styles, but these are not currently exposed.

Focus label styles are ignored in cases where the field is errored, though
the border still changes color when focused:

image

In the process, also fix a small issue with adding error styling (border/label
color) to shared inputs that fail validation or parsing. This is visible in the
gas limit field when entering a float number (e.g. `1.5`).


VladUXUI:
image
I like your version here for Error + Focus, if it's not too late, let's use that one

Shadowfiend:
Will try to sneak it in today 😈

Shadowfiend:
This fellow is ready for another round of review!

henryboldi (approved):
Works beautifully! 😍

SJQuality:
🐛 
https://github.com/tallycash/extension/issues/1400

PR #1378, opened 2022-04-29
🎚 Add "Set Tally Ho as default" toggle to set password page

henryboldi:
This PR adds the option of setting Tally Ho as the default wallet to the
"create password" screen. The idea is to increase the user's visibility of the
setting rather than a skipped-through info screen. To test, adjust the toggle
on the create password page, and then see if it matches with the menu tab. As
with using the menu tab, check whether Tally Ho openings when clicking
"injected" or MM.
 
Tally toggle


jagodarybacka (approved):
Looks good 💖

PR #1388, opened 2022-04-29
🫓 Fuckin' NaN: Fix long-tail token price bug

mhluongo:
If a token without a fiat pair was being price checked, the error prevented other tokens from getting a price.

Closes https://github.com/tallycash/extension/issues/1206

# To test

- [ ] Load a new wallet with many tokens
- [ ] Wait 10 minutes to see if prices are pulled. The bug would only allow the price of ETH through rather than many ERC-20s.

### DIFF
--- a/background/lib/prices.ts
+++ b/background/lib/prices.ts
@@ -127,12 +127,20 @@ export async function getEthereumTokenPrices(
     const price: number = Number.parseFloat(
       priceDetails[fiatSymbol.toLowerCase()]
     )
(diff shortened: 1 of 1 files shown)


henryboldi (approved):
Tested using `0x9eef87f4c08d8934cb2a3309df4dec5635338115`:
On `main` → Only ETH shows a price
On `fuckin-nan` → All prices (so far) appear ✅

An important fix! 🫓🍻👏 Excellent!

PR #1396, opened 2022-05-05
🔦 Switchable Switcher: Hook the protocol switcher up the to `MULTI_NETWORK` feature flag.

mhluongo:
In our slow, sustained march to multi-network support, this PR hooks up the protocol switcher skeleton to our multi-network feature flag.

To test

- [ ] Without setting the `MULTI_NETWORK` feature flag, make sure the "Ethereum" label in the top left corner of the wallet doesn't do anything on click or appear "clickable".
- [ ] Setting the `MULTI_NETWORK` feature flag to `true`, the label should have a drop-down chevron and open a menu when clicked.
- [ ] No other pages or components should be affected.


ui/components/TopMenu/TopMenuProtocolSwitcher.tsx
```diff
@@ -2,14 +2,17 @@ import React, { ReactElement } from "react"
...
+  enabled?: boolean
 }
 
 export default function TopMenuProtocolSwitcher({
   onClick,
+  enabled,
 }: Props): ReactElement {
   return (
-    <button type="button" onClick={onClick}>
+    <button type="button" onClick={() => enabled && onClick && onClick()}>
```
jagodarybacka, line 13:
```suggestion
    <button type="button" onClick={() => enabled && onClick?.()}>
```

ui/components/TopMenu/TopMenuProtocolSwitcher.tsx
```diff
@@ -2,14 +2,17 @@ import React, { ReactElement } from "react"
 
 type Props = {
   onClick?: () => void
+  enabled?: boolean
 }
 
 export default function TopMenuProtocolSwitcher({
   onClick,
+  enabled,
```
jagodarybacka, line 10: Can `TopMenuProtocolSwitcher` contain info if it is enabled based on `MULTI_NETWORK` instead of passing it in the prop? I think in all cases we want it to be based on that flag right?

jagodarybacka (dismissed):
Looks good 🥇 

Switching between protocols is not working for me but I guess it is not in the scope of this PR, am I right?

ui/components/TopMenu/TopMenuProtocolSwitcher.tsx
mhluongo: Where has this been all my life!

ui/components/TopMenu/TopMenuProtocolSwitcher.tsx
mhluongo: I'd rather not distinguish below this level, then revisit if we need it.

mhluongo:
> Switching between protocols is not working for me but I guess it is not in the scope of this PR, am I right?

Yep, that'll be dropping with some new Redux plumbing in a PR or two.

0xDaedalus (approved):
Embarcación 🚢

PR #1403, opened 2022-05-06
Porygon 🦆

0xDaedalus:
> [State-of-the-art technology was used to create Porygon. It was the first artificial Pokémon to be created via computer programming.](https://www.pokemon.com/us/pokedex/porygon)
image

This PR adds incredibly rudimentary support for connecting to the polygon network and seeing your assets in your wallet.  

The two main things to test in this PR are that it does not break existing functionality - and that we can import a wallet and see its MATIC balance correctly.

**Note that this PR does not implement switching between networks in the UI - this will come later**

#### Ethereum Tests
- [ ] Upgrading to this version does not break existing balance / import wallet functionality.
- [ ] Clean installing this version does not break existing balance / import wallet functionality.

#### Polygon Tests (set MULTI_NETWORK=true in your .env file)

On a clean install:

- [ ] You should be able to add a read-only address via ENS name lookup.
- [ ] Addresses with a MATIC balance on polygon should show that balance in the wallet (you can use `abaz.eth`)


background/lib/alchemy.ts
```diff
@@ -50,18 +50,29 @@ export async function getAssetTransfers(
     // excludeZeroValue: false,
   }
 
+  // Default Ethereum Mainnet categories
+  // https://docs.alchemy.com/alchemy/enhanced-apis/transfers-api#alchemy_getassettransfers-ethereum-mainnet
+  let category = ["external", "internal", "token", "erc20", "erc1155"]
```
mhluongo, line 55: Does mainnet not support ERC-721?

background/services/chain/index.ts
```diff
@@ -190,7 +194,8 @@ export default class ChainService extends BaseService<Events> {
     })
 
     this.supportedNetworks = USE_MULTI_NETWORK
-      ? [ETHEREUM, ARBITRUM_ONE, OPTIMISM]
+      ? // Just Polygon for now since we don't need fancy Alchemy ninjitsu to get it to work.
+        [ETHEREUM, POLYGON]
```
mhluongo, line 198: Bah, only forward!

>.>

background/services/chain/index.ts
```diff
@@ -495,7 +500,7 @@ export default class ChainService extends BaseService<Events> {
     const accountBalance: AccountBalance = {
       address: addressNetwork.address,
       assetAmount: {
-        asset: ETH,
+        asset: addressNetwork.network.baseAsset,
```
mhluongo, line 503: Feels like we should destructure the `addressNetwork` above this to avoid all the pre-pre-prefixing

background/services/name/index.ts
```diff
@@ -93,7 +93,12 @@ export default class NameService extends BaseService<Events> {
         [address: HexString]: ResolvedNameRecord | undefined
       }
     }
-  } = { EVM: { [ETHEREUM.chainID]: {} } }
+  } = {
+    EVM: Object.fromEntries(
+      EVM_MAIN_NETWORKS.map((network) => [network.chainID, {}])
```
mhluongo, line 98: :+1:

background/services/name/resolvers/ens.ts
```diff
@@ -19,15 +21,26 @@ export default function ensResolverFor(
...
       network,
     }: NameOnNetwork): Promise<AddressOnNetwork | undefined> {
+      // Lets assume that we can use ENS to look up names for any EVM compatible chain for now.
+      if (network.family !== "EVM") {
+        return undefined
+      }
+
       const address = await chainService
-        .providerForNetwork(network)
+        // Allow ENS lookups when on Arbitrum, Polygon, Optimism, etc..
```
mhluongo, line 42: Maybe worth a bit more explanation.

"Use ENS on Ethereum mainnet to resolve address on..."

mhluongo (commented):
Great start :)

background/lib/alchemy.ts
0xDaedalus: It does but why do we need it if we don't support NFTs yet?

background/lib/alchemy.ts
```diff
@@ -50,18 +50,29 @@ export async function getAssetTransfers(
...
+
+  if (addressOnNetwork.network.name !== "Ethereum") {
+    // Default L2 categories
+    // https://docs.alchemy.com/alchemy/enhanced-apis/transfers-api#alchemy_getassettransfers-testnets-and-layer-2s
+    category = ["token", "erc20", "erc721", "erc1155"]
+  }
```
Shadowfiend, line 61: Do we need these if they're the defaults anyway?

background/services/chain/index.ts
Shadowfiend: Tend to agree we shouldn't remove the other entries until we start making ship/don't ship calls.

background/services/name/index.ts
```diff
@@ -205,6 +210,10 @@ export default class NameService extends BaseService<Events> {
     const { address: normalizedAddress, network } =
       normalizeAddressOnNetwork(addressOnNetwork)
 
+    if (!this.cachedResolvedNames[network.family][network.chainID]) {
+      this.cachedResolvedNames[network.family][network.chainID] = {}
+    }
```
Shadowfiend, line 215:
```suggestion
    this.cachedResolvedNames[network.family][network.chainID] ??= {}
```

Personal idiomatic preference.

Shadowfiend: Though if we do this, why do the `fromEntries` dance above? Feels unnecessary, and I think we should keep this one rather than that one.

background/services/name/resolvers/ens.ts
```diff
@@ -19,15 +21,26 @@ export default function ensResolverFor(
       return true
     },
     canAttemptAddressResolution({ name, network }: NameOnNetwork): boolean {
-      return sameNetwork(network, ETHEREUM) && name.endsWith(".eth")
+      return (
+        name.endsWith(".eth") &&
+        !!supportedNetworks.find((supportedNetwork) =>
```
Shadowfiend, line 26: You're looking for [`.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) :)

Double coercion is ~~never~~ rarely the way :grin:

background/services/name/resolvers/ens.ts
```diff
@@ -19,15 +21,26 @@ export default function ensResolverFor(
...
+        )
+      )
     },
 
     async lookUpAddressForName({
       name,
       network,
     }: NameOnNetwork): Promise<AddressOnNetwork | undefined> {
+      // Lets assume that we can use ENS to look up names for any EVM compatible chain for now.
+      if (network.family !== "EVM") {
```
Shadowfiend, line 37: We won't reach this if `canAttemptAddressResolution` returns `false`, so we're already filtered to `supportedNetworks` here. I would drop this check.

background/services/name/resolvers/ens.ts
```diff
@@ -19,15 +21,26 @@ export default function ensResolverFor(
...
+        return undefined
+      }
+
       const address = await chainService
-        .providerForNetwork(network)
+        // Use ENS to Ethereum mainnet to resolve addresses on EVM networks (Polygon, Arbitrum, etc..)
+        .providerForNetwork(ETHEREUM)
```
Shadowfiend, line 43: Is this how ENS works? Or do they have a resolver per chain? Ethers supports resolvers per chain (see https://github.com/ethers-io/ethers.js/blob/cf47ec5fd5c7e6dda4da499b1c25ac1d1cc84b5a/packages/providers/src.ts/base-provider.ts#L1916-L1934), just not sure which is the right approach.

ui/hooks/validation-hooks.ts
```diff
@@ -145,6 +145,10 @@ export const useAddressOrNameValidation: AsyncValidationHook<
   const validatingValue = useRef<string | undefined>(undefined)
   const dispatch = useBackgroundDispatch()
 
+  const network = ETHEREUM
```
Shadowfiend, line 148: You could add to `uiSelectors.ts`:

```
export const selectCurrentNetwork = createSelector(
  (state: RootState) => state.ui.selectedAccount.network,
  (selectedNetwork) => selectedNetwork
)
```

And then pull the current network here or pass it as a parameter to the validation hook.

Truth be told, this is another place where we could really use the ability to resolve an address or network to multiple outcomes and show them for the user to select (e.g., show them that this name resolves on multiple networks and let them add one or more, etc). This would also be useful for #1288.

ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
```diff
@@ -16,14 +16,18 @@ export default function OnboardingViewOnlyWallet(): ReactElement {
   const [addressOnNetwork, setAddressOnNetwork] = useState<
     AddressOnNetwork | undefined
   >(undefined)
+  const network = ETHEREUM
```
Shadowfiend, line 19: Same as above re: using the current account's network as a starting point here (ultimately we'll probably want to separate the current network, but having a selector shim over it will let us make that refactor separate).

background/lib/alchemy.ts
mhluongo: I just thought it was weird it was in one list and not the other.

background/services/name/resolvers/ens.ts
mhluongo: I think this is right for now.

ui/pages/Onboarding/OnboardingViewOnlyWallet.tsx
0xDaedalus: I don't think this approach works here because its quite likely that a user does not have an account when viewing this page so there will be no `selectedAccount` to pull from.

ui/hooks/validation-hooks.ts
0xDaedalus: Same comment as below - I do think it is good to _try_ to pull in the selected account's network but a user is not guaranteed to have a set up an account when this hook runs if its running as part of `OnboardingViewOnlyWallet` page.

background/lib/alchemy.ts
0xDaedalus: Unfortunately yes - the `defaults` in the documentation seem to differ from the defaults that actually get used in alchemy's code when `category` is omitted.

Specifically - omitting the `category` property when querying L2 results in an error of `category "internal" not supported`.

background/lib/alchemy.ts
0xDaedalus: ah whoops - will remove from the other list as well.

mhluongo:
Two issues with testing, with `MULTI_NETWORK` set to true

1. Onboarding with an ENS read-only account ends up just showing the auto-genned account name

Screenshot from 2022-05-09 14-15-30

2. Alchemy asset transfer with Polygon don't appear to validate via JDT

Here's the response object and the logs
```
{
    "transfers": [
        {
            "blockNum": "0x1ab61eb",
            "hash": "0x61f577dda1c06be2616c0027d4a5840a8e74f2b968f453ca878eccb14dde51aa",
            "from": "0x0000000000000000000000000000000000000000",
            "to": "0x9eef87f4c08d8934cb2a3309df4dec5635338115",
            "value": null,
            "erc721TokenId": null,
            "erc1155Metadata": [
                {
                    "tokenId": "0x01",
                    "value": "0x01"
                }
            ],
            "tokenId": null,
            "asset": "BULLPLL",
            "category": "erc1155",
            "rawContract": {
                "value": null,
                "address": "0x32ea8cd902562564808e2d3a7127ae084380155d",
                "decimal": null
            }
        },
        {
            "blockNum": "0x1ab61eb",
            "hash": "0x61f577dda1c06be2616c0027d4a5840a8e74f2b968f453ca878eccb14dde51aa",
            "from": "0x0000000000000000000000000000000000000000",
            "to": "0x9eef87f4c08d8934cb2a3309df4dec5635338115",
            "value": null,
            "erc721TokenId": null,
            "erc1155Metadata": [
                {
                    "tokenId": "0x01",
                    "value": "0x01"
                }
            ],
            "tokenId": null,
            "asset": "BULLPLL",
            "category": "token",
            "rawContract": {
                "value": null,
                "address": "0x32ea8cd902562564808e2d3a7127ae084380155d",
                "decimal": null
            }
        },
        {
            "blockNum": "0x1ab9837",
            "hash": "0x402e162fc5ed918626f4995bcdd9478e2c98d58a88f0fcc14edab9a9d10bbd78",
            "from": "0x0000000000000000000000000000000000000000",
            "to": "0x9eef87f4c08d8934cb2a3309df4dec5635338115",
            "value": null,
            "erc721TokenId": null,
            "erc1155Metadata": [
                {
                    "tokenId": "0x01",
                    "value": "0x01"
                }
            ],
            "tokenId": null,
            "asset": null,
            "category": "erc1155",
            "rawContract": {
                "value": null,
                "address": "0x484e6eaec6e81d640c83a3f97c3b6a3d40825db3",
                "decimal": null
            }
        },
        {
            "blockNum": "0x1ab9837",
            "hash": "0x402e162fc5ed918626f4995bcdd9478e2c98d58a88f0fcc14edab9a9d10bbd78",
            "from": "0x0000000000000000000000000000000000000000",
            "to": "0x9eef87f4c08d8934cb2a3309df4dec5635338115",
            "value": null,
            "erc721TokenId": null,
            "erc1155Metadata": [
                {
                    "tokenId": "0x01",
                    "value": "0x01"
                }
            ],
            "tokenId": null,
            "asset": null,
            "category": "token",
            "rawContract": {
                "value": null,
                "address": "0x484e6eaec6e81d640c83a3f97c3b6a3d40825db3",
                "decimal": null
            }
        }
    ]
}
```

Screenshot from 2022-05-09 14-17-09

0xDaedalus:
> Onboarding with an ENS read-only account ends up just showing the auto-genned account name

In the interest of smaller and more-frequent PRs how would you feel about us tackling this in a future PR?

mhluongo:
I'm down! As long as we don't break with the feature flag off, leaving something unfinished is fair game.

mhluongo (approved):
I can't get this branch displaying MATIC with `MULTI_NETWORK` set to true... but I'm not seeing any issues without the feature flag, so I'm good to merge if you'd like @0xDaedalus

background/services/indexing/index.ts
```diff
@@ -523,9 +522,9 @@ export default class IndexingService extends BaseService<Events> {
     // TODO refactor for multiple price sources
     try {
       // TODO include user-preferred currencies
-      // get the prices of ETH and BTC vs major currencies
+      // get the prices of ETH, BTC, and MATIC vs major currencies
       const basicPrices = await getPrices(
-        [BTC, ETH] as CoinGeckoAsset[],
+        BASE_ASSETS as CoinGeckoAsset[],
```
henryboldi, line 527: YES

background/services/name/resolvers/ens.ts
```diff
@@ -19,15 +21,21 @@ export default function ensResolverFor(
       return true
     },
     canAttemptAddressResolution({ name, network }: NameOnNetwork): boolean {
-      return sameNetwork(network, ETHEREUM) && name.endsWith(".eth")
+      return (
+        name.endsWith(".eth") &&
```
henryboldi, line 25: Does this overlook [`Importing a DNS Domain Name to ENS`](https://medium.com/the-ethereum-name-service/step-by-step-guide-to-importing-a-dns-domain-name-to-ens-d2d15feb03e8)? Granted, I realize `name.endsWith(".eth")` doesn't come from your changes :)

henryboldi (approved):
LET'S GOOO!!!!!! 🔥🔥🔥

polygon

background/services/name/resolvers/ens.ts
Shadowfiend: Def does.

PR #1404, opened 2022-05-07
Error Vector Magnitude: Light refactoring for multi-network handling

Shadowfiend:
This implements queued transaction lookups for all networks with a
provider available, factors the provider tracking in `ChainService`
to follow the pattern from elsewhere of nesting all EVM chains under
an `evm` property for future work, and adds network tracking to
transaction requests.

## Testing

Things to be tested here:
- [ ] Transaction history population, as well as lookups for newly
      created transactions, continue to work properly on mainnet.
- [ ] Transactions can still be created, signed, etc.
- [ ] Mainnet fork still works as expected when USE_MAINNET_FORK is
      enabled; in particular, transactions can still be signed and
      mined on the mainnet fork.

### DIFF
--- a/background/networks.ts
+++ b/background/networks.ts
@@ -128,7 +128,7 @@ export type LegacyEVMTransaction = EVMTransaction & {
  */
 export type LegacyEVMTransactionRequest = Pick<
   LegacyEVMTransaction,
-  "gasPrice" | "type" | "nonce" | "from" | "to" | "input" | "value"
+  "gasPrice" | "type" | "nonce" | "from" | "to" | "input" | "value" | "network"
 > & {
   chainID: LegacyEVMTransaction["network"]["chainID"]
   gasLimit: bigint
@@ -164,6 +164,7 @@ export type EIP1559TransactionRequest = Pick<
   | "value"
   | "maxFeePerGas"
   | "maxPriorityFeePerGas"
+  | "network"
 > & {
   gasLimit: bigint
   chainID: EIP1559Transaction["network"]["chainID"]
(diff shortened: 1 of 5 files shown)


background/networks.ts
```diff
@@ -164,6 +164,7 @@ export type EIP1559TransactionRequest = Pick<
   | "value"
   | "maxFeePerGas"
   | "maxPriorityFeePerGas"
+  | "network"
 > & {
   gasLimit: bigint
   chainID: EIP1559Transaction["network"]["chainID"]
```
mhluongo, line 170: Should we remove `chainID`?

mhluongo:
Importing a read-only account without `MULTI_NETWORK` gives us this beaut

Screenshot from 2022-05-06 22-38-13

Screenshot from 2022-05-06 22-39-13

ENS names also aren't resolving, either when trying to input them in the read-only input or when trying to reverse-resolve an address. Not sure why that was hit.

mhluongo (commented):
Quick q on type shape

background/networks.ts
Shadowfiend: At some point we may want to, but we'll need to adapt the parent types a bit so they all conform. I don't think it's worth it right now tbh.

Shadowfiend:
Whoops yep, there's are mixed expectations on chain id vs network name for lookups here, and I didn't disentangle them enough. 1s.

mhluongo (approved):
Local testing look good!

Shadowfiend:
Including mainnet fork? That was probably the bit I had least confidence in since I've used it less.

PR #1410, opened 2022-05-07
Information Networks: Update accounts slice to track accounts by network

Shadowfiend:
> [**"The righteous need not cower before the drumbeat of human progress. Though the song of yesterday fades into the challenge of tomorrow, God still watches and judges us. Evil lurks in the datalinks as it lurked in the streets of yesteryear. But it was never the streets that were evil."**](https://civilization.fandom.com/wiki/Information_Networks_(SMAC))
> [   – Sister Miriam Godwinson, "A Blessed Struggle"](https://civilization.fandom.com/wiki/Information_Networks_(SMAC))

This is mostly a big ball of wax commit due to how entwined so many of
these pieces are. Here are the big changes:

- The accounts slice's accountsData key now has an `evm` subkey,
  followed by the chain ID, followed by the normalized EVM address.
- All accounts slice actions operate on AddressOnNetwork instead of bare
  addresses now.
- All account selectors operate on AddressOnNetwork instead of bare
  addresses.
- selectAccountTotalsByCategory is now
  selectCurrentNetworkAccountTotalsByCategory, and only selects account
  totals for the currently-selected network. Currently the selected
  network is derived as the network of the currently-selected account,
  and is exposed as a `selectCurrentNetwork` UI selector.
- Message and data signature requests use AddressOnNetworks in UI paths.
  Ledger `signMessage` calls reject signature requests for non-mainnet
  chains. signTypedData requests throw if a domain chainId is specified
  that does not match the account AddressOnNetwork (which used to be a
  bare address).
- A few components that were taking AccountTotal + address information
  now use the address + network information carried directly by
  AccountTotal.
- In cases where we don't have another way of inferring which network
  a given request is from, we assume it is from the UI's
  currently-selected network. The only exception is the internal
  Ethereum provider, which is currently hardcoded to always use mainnet.
- Keyring transaction signatures always reference the passed network
  base asset instead of hardcoding ETH.

The main decision worth flagging here is the hierarchy in the accounts
slice, which goes accounts -> `evm` key -> chainID -> address. This
structure matches the * -> `evm` -> chainID structure of other places
where we're handling multi-network data, and allows for future abstractions
that might descend into such hierarchies in consistent ways, as well as
provide fallbacks for various properties throughout that hierarchy (for
example, for address names).

## Testing

Note that this PR has not been thoroughly tested, so taking it through
its paces is a good idea. Things to look out for:

- Issues with the mainnet fork flag. Mainnet forks have a different
  chain id, so things could get weird.
- Any issues with adding or removing accounts.
- Any issues with editing account names.
- Any issues with balance updates.
- Any issues with the overview of all addresses (including  address
  counts).


background/main.ts
```diff
@@ -977,20 +1017,28 @@ export default class Main extends BaseService<never> {
...
   async connectKeyringService(): Promise<void> {
-    // TODO support other networks
-    const network = ETHEREUM
-
     this.keyringService.emitter.on("keyrings", (keyrings) => {
       this.store.dispatch(updateKeyrings(keyrings))
     })
 
     this.keyringService.emitter.on("address", (address) => {
+      // FIXME Should be .selectedNetwork once that exists.
```
0xDaedalus, line 1025: Are you thinking `selectedNetwork` is a pre-L2-support thing or a post-L2-support thing?  I think we can get away with only using `AddressOnNetwork` it until we support non-EVM chains.

background/redux-slices/accounts.ts
```diff
@@ -138,64 +143,79 @@ const accountSlice = createSlice({
...
+      immerState,
+      { payload: { address, network } }: { payload: AddressOnNetwork }
+    ) => {
+      const normalizedAddress = normalizeEVMAddress(address)
+      if (
+        immerState.accountsData.evm[normalizedAddress]?.[network.chainID] !==
+        undefined
+      ) {
+        // If the account data already exists, the account is already loaded.
+        return
```
0xDaedalus, line 156: I think we still need to return the `immerState` here

background/redux-slices/accounts.ts
```diff
@@ -138,64 +143,79 @@ const accountSlice = createSlice({
...
-      }
-      // Immutably remove the account passed in
-      // eslint-disable-next-line @typescript-eslint/naming-convention
-      const { [keyToRemove]: _, ...withoutAccountToRemove } = state.accountsData
-      return {
-        ...state,
-        accountsData: withoutAccountToRemove,
+      if (immerState.accountsData.evm[normalizedAddress]?.[network.chainID]) {
+        // If the account data does not exist, deleting is a noop.
+        return
```
0xDaedalus, line 172: same as above

background/redux-slices/accounts.ts
```diff
@@ -229,53 +249,67 @@ const accountSlice = createSlice({
...
-      // TODO Refactor when accounts are also keyed per network.
-      const accountKey = normalizeEVMAddress(addressNetworkName.address)
+      const normalizedAddress = normalizeEVMAddress(address)
 
-      // No entry means this ENS name isn't being tracked here.
-      if (immerState.accountsData[accountKey] === undefined) {
+      // No entry means this name doesn't correspond to an account we are
+      // tracking.
+      if (
+        immerState.accountsData.evm[normalizedAddress]?.[network.chainID] ===
```
0xDaedalus, line 260: I think if we segment by chainID here we will have to re-fetch all cached ENS & UNS  names whenever we switch networks right?  Would it be better just to check for `immerState.accountsData.evm[normalizedAddress]` ?

background/redux-slices/accounts.ts
Shadowfiend: Yeah basically you can but don't have to; however, the types demand that you choose one of the two consistently in a given action handler.

background/redux-slices/accounts.ts
Shadowfiend: The question of how we handle names across chain ids is probably going to require an extra layer of logic. I expect the move will be having the name service emit events for all the networks a given name is valid for. 

Otherwise we end up with issues with EVM compatible networks that use chain ids but may have a different name/address derivation strategy.

background/main.ts
Shadowfiend: I think we should have it regardless right now, tbh. Even if two networks are compatible, we need to be able to distinguish between them, otherwise we run the risk of trying to do something on one network when it's meant for another. 

We do need to develop some concept of an address that can be used on more than one network, but I think that's a layer on top of the network tracking rather than trying to continue without tracking a network.

background/redux-slices/accounts.ts
mhluongo: The hardest part for all of this isn't the code, it's the fact that the name services haven't actually figured this stuff out... and that L2s and users have given up on network-specific address derivation. There are only a couple EVMs I can think of that have different address derivation (Rootstock and Celo IIRC)

Shadowfiend: Thinking down this rabbit hole just popped this fellow into my head:

```typescript
function forNetwork<T>(container: NetworkContainer<T>, network: Network): T | undefined
```

That abstracts providerForNetwork to all network-containing types. I'm gonna have to swap chain ids and addresses in this PR hehe.

background/redux-slices/accounts.ts
```diff
@@ -229,53 +249,67 @@ const accountSlice = createSlice({
...
         return
       }
 
       const baseAccountData = getOrCreateAccountData(
-        immerState.accountsData[accountKey],
-        accountKey,
-        addressNetworkName.network,
-        Object.keys(immerState.accountsData).filter((key) => key !== accountKey)
-          .length
+        // TODO Figure out the best way to handle default name assignment
```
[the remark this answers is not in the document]
VladUXUI: I wonder if that is going to look strange/confusing in account selection page? Having shortened address twice (once as name and once as eth address).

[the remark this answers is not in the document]
Shadowfiend: If we choose to go down this path, we can skip the second display of the address if there's no name—just a question of what you want it to look like, basically. We can distinguish and adjust the presentation based on whether there is a name or not.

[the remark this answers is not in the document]
VladUXUI: I think it's good to keep it consistent. So everything is displayed the same, this way it's easier to scan and make decisions.

We have a few solutions around guiding users to get a useful name, either do it through onboarding/use more general names with number (address 1) to make users more prone to changing.

But i would like to table this for now and come back to it when we have a bit more time to properly figure out the UX around it.

background/redux-slices/accounts.ts
```diff
@@ -313,29 +353,42 @@ export const addAddressNetwork = createBackgroundAsyncThunk(
...
+    addressOnNetwork: AddressOnNetwork,
+    { getState, dispatch, extra: { main } }
+  ) => {
+    const normalizedAddress = normalizeEVMAddress(addressOnNetwork.address)
+
+    await dispatch(accountSlice.actions.deleteAccount(addressOnNetwork))
+
+    const state = getState() as { accounts: AccountState }
+
+    // Remove the corresponding keyring account iff there are no networks left
```
0xDaedalus, line 384: `iff` => `if`

background/redux-slices/accounts.ts
```diff
@@ -313,29 +353,42 @@ export const addAddressNetwork = createBackgroundAsyncThunk(
...
-    main.removeAccount(address, { type: "keyring" })
+  async (
+    addressOnNetwork: AddressOnNetwork,
+    { getState, dispatch, extra: { main } }
+  ) => {
+    const normalizedAddress = normalizeEVMAddress(addressOnNetwork.address)
+
+    await dispatch(accountSlice.actions.deleteAccount(addressOnNetwork))
+
+    const state = getState() as { accounts: AccountState }
```
0xDaedalus, line 382: I think we can destructure here

background/redux-slices/selectors/uiSelectors.ts
```diff
@@ -39,6 +39,11 @@ export const selectCurrentAddressNetwork = createSelector(
   (selectedAccount) => selectedAccount
 )
 
+export const selectCurrentNetwork = createSelector(
```
0xDaedalus, line 42: What are we memoizing here with this selector?

background/services/signing/index.ts
```diff
@@ -195,24 +189,36 @@ export default class SigningService extends BaseService<Events> {
...
     typedData: EIP712TypedData
-    account: HexString
+    account: AddressOnNetwork
     signingMethod: SigningMethod
   }): Promise<string> {
     try {
       let signedData: string
+      if (
+        typedData.domain.chainId !== undefined &&
+        // Let parseInt infer radix by prefix; chainID can be hex or decimal,
```
0xDaedalus, line 199: 🌶️

0xDaedalus (commented):
Looks great to me! - nothing blocking - will test this AM.

background/redux-slices/accounts.ts
Shadowfiend: I meant [iff](https://en.wikipedia.org/wiki/If_and_only_if), but open question whether my usage is correct 😁

0xDaedalus: ✅ wasn't familiar with that - TIL

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -93,7 +103,10 @@ const computeCombinedAssetAmountsData = (
 
 const getAccountState = (state: RootState) => state.account
 const getCurrentAccountState = (state: RootState) => {
-  return state.account.accountsData[state.ui.selectedAccount.address]
+  const { address, network } = state.ui.selectedAccount
+  return state.account.accountsData.evm[network.chainID][
```
mhluongo, line 107: This is erroring at startup with a fresh install. Doing some quick debugging...

`address` is `""`
`network` appears to be `ETHEREUM`
`state.account.accountsData.evm` is `{}`

mhluongo: It bubbles up to this instead of the onboarding flow

Screenshot from 2022-05-12 08-56-40
Screenshot from 2022-05-12 08-57-26

Shadowfiend: Appreciate the extra digging! Will see if I can give this a gander today. I think I ended up juggling too many branches with too little bandwidth last week.

background/redux-slices/selectors/uiSelectors.ts
Shadowfiend: If the network doesn't change, this selector won't update. So if the user changes the account without changing the network, we shouldn't see dependents of this selector trigger rerenders.

Shadowfiend:
Pushed an adjustment to the fix you had @0xDaedalus that's mostly type-related but actually should have fixed the issue @mhluongo saw. Everything seems to work with this commit in place regardless—tried both read-only and regular addresses, dApp and send ETH interactions, and historical activity.

background/redux-slices/selectors/uiSelectors.ts
0xDaedalus: I don't think thats the case - each AddressOnNetwork has its own `network` property right?  Even if the networks are deeply equal they will not be strictly equal since they are distinct objects.

0xDaedalus:
Oh good catch - I'm surprised that the ts compiler allows `[]` as an argument to `Object.entries` - whoopsie on my part thanks for that catch.

0xDaedalus:
Tested - things look great 🎉 🎉 🎉 .  I think just this last thing to bang out, merge conflicts, and we're good-to-gooooo.

✅ Issues with the mainnet fork flag. Mainnet forks have a different chain id, so things could get weird.
❌ Any issues with adding or removing accounts. - Unable to remove accounts
✅ Any issues with editing account names.
✅ Any issues with balance updates - both ETH and ERC20 sends / receives look good.
✅ Any issues with the overview of all addresses (including address counts).

Shadowfiend:
> Oh good catch - I'm surprised that the ts compiler allows [] as an argument to Object.entries - whoopsie on my part thanks for that catch.

It actually works the same as `Object.entries({})`, since an array is ~just an object with numerical keys (which get stringified under `Object.entries`, just for fun) so an empty array is an object with no keys.

0xDaedalus:
Fixing conflicts so I can branch off of this guy

mhluongo:
Testing has led to some interesting behavior. Below, I added 3 addresses on Ethereum, then switched each 1 by 1 to Polygon. Then I switched to Arbitrum... and got a new address I hadn't seen before. Then I added another one there that I hadn't seen before. Finally, deleting the most recently added on Arbitrum reset the selected account to one that wasn't yet under the account dropdown for that network.

Here's mainnet / chain ID 1

Screenshot from 2022-05-17 15-48-30

Here's Polygon. Note the different accounts / being out of order which I think would surprise users.

Screenshot from 2022-05-17 15-48-19

... and here's Arbitrum before and after deleting the last account

Screenshot from 2022-05-17 15-48-06

Note the avatar in the top right. We have a `selectedAddressOnNetwork` or similar that the dropdown doesn't appear to know about.

Screenshot from 2022-05-17 15-48-55

Screenshot from 2022-05-17 15-49-07

mhluongo:
Even more illuminating, it appears the keyrings and tracked accounts are getting out of sync.

Here's what I see when I keep the same extension state, recompile with `MULTI_NETWORK` as `false`, and click the accounts menu.

Screenshot from 2022-05-17 15-54-45

background/redux-slices/accounts.ts
```diff
@@ -165,19 +165,31 @@ const accountSlice = createSlice({
...
     ) => {
       const normalizedAddress = normalizeEVMAddress(address)
 
-      if (immerState.accountsData.evm[network.chainID]?.[normalizedAddress]) {
-        // If the account data does not exist, deleting is a noop.
+      const { evm } = immerState.accountsData
+
+      if (
+        // One of the chains
+        !Object.keys(evm ?? {}).some((chainID) =>
```
Shadowfiend, line 176: Just a minor note that this could be a little clearer as `Object.values(evm ?? {}).flat().some(...)`.

mhluongo:
... and here's what happens when I choose a new account (switching back to Ethereum, as the feature flag isn't set) and hit "add address".

Screenshot from 2022-05-17 15-56-30

Funny enough, *somehow* I also have a read-only account from a previous install... feeling sus

Screenshot from 2022-05-17 15-56-30

mhluongo (approved):
I vote we merge here despite the odd selector behavior as it's only with the feature flag. Leaving it to @0xDaedalus to press the green button

0xDaedalus (approved):

PR #1412, opened 2022-05-09
FF store updates: update Dockerfile and build steps for ff

greg-nagy:
> This is just to document the exact steps I took to submit the current v0.13.3 build to ff and to support @henryboldi 
> - The sqllite workaround should be removed when the dependency is fixed
> - The build steps should be automated if they work
> But I kept them as is, because not sure that they will work. They should but we had that thought a couple of time already... 

I have a hunch that because of sqlite3
install failure the postinstall script
is not run and our patches are not applied.
So just removed the remotedev-server from
the background package for build as quick
and hacky workaround until we solve the
issue properly. We don't need it for build
so it's a safe corner to cut here.

Also wanted to record the exact steps
to prepare the source and the output
for submission.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,14 @@ FROM node:16-alpine
 
 WORKDIR /tally-ho
 
-COPY . . 
+COPY . .
 
 RUN mv .env.prod .env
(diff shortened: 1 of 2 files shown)


greg-nagy:
Stale pr

PR #1414, opened 2022-05-09
Don't send an empty array when calling alchemy_getTokenBalances.

0xDaedalus:
Sending an empty array results in the following error: 
`Error: invalid 2nd argument: contract_addresses was not a valid contract address array or string literal 'DEFAULT_TOKENS'`

### To Test
- [ ] Balances should show up on adding a new wallet.


mhluongo (approved):
Thank you! :blush:

PR #1417, opened 2022-05-09
Properly log objects in exported logs.

0xDaedalus:
Before: 

```
chrome-extension://licmhmlaminmlpcdkobcjofpnchdlacg/background.js:61926:63
Error fetching, validating, and saving token list https://yearn.science/static/tokenlist.json,[object Object]
    at chrome-extension://licmhmlaminmlpcdkobcjofpnchdlacg/background.js:61926:63 
```

Note the `[object Object]`

After:

```
chrome-extension://licmhmlaminmlpcdkobcjofpnchdlacg/background.js:61939:63
Error fetching, validating, and saving token list https://yearn.science/static/tokenlist.json,{"foo":"bar"}
    at chrome-extension://licmhmlaminmlpcdkobcjofpnchdlacg/background.js:61939:63 
```

As a result of the following line of code: 

```
logger.error(`Error fetching, validating, and saving token list ${url}`, { foo: "bar" })
 ```
 
 This will make logs that users export for us a lot more useful - especially when looking at internal ethereum provider logs.

### DIFF
--- a/background/lib/logger.ts
+++ b/background/lib/logger.ts
@@ -74,11 +74,24 @@ function saveLog(
   input: unknown[],
   stackTrace: string[] | undefined
 ) {
+  const formattedInput = input
+  for (let i = 0; i < input.length; i += 1) {
+    const log = input[i]
+    if (log instanceof Object) {
+      try {
+        formattedInput[i] = JSON.stringify(log)
+      } catch (e) {
+        // if we can't stringify thats OK
+      }
+    }
+  }
   localStorage.setItem(
     "logs",
     `${(localStorage.getItem("logs") ?? "").slice(
       -50000
-    )}${purgeSensitiveFailSafe(`${logLabel}\n${input}\n${stackTrace}`)}\n\n`
+    )}${purgeSensitiveFailSafe(
+      `${logLabel}\n${formattedInput}\n${stackTrace}`
+    )}\n\n`
   )
 }
 
--- a/background/lib/logger.ts
+++ b/background/lib/logger.ts
@@ -77,11 +77,11 @@ function saveLog(
   const formattedInput = input
   for (let i = 0; i < input.length; i += 1) {
     const log = input[i]
-    if (log instanceof Object) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: background/lib/logger.ts [hunk 1]
```diff
@@ -74,11 +74,24 @@ function saveLog(
   input: unknown[],
   stackTrace: string[] | undefined
 ) {
+  const formattedInput = input
+  for (let i = 0; i < input.length; i += 1) {
+    const log = input[i]
+    if (log instanceof Object) {
```
Shadowfiend, line 80: Can probably hit more with typeof === object fwiw.

background/lib/logger.ts
```diff
@@ -74,11 +74,24 @@ function saveLog(
...
+      try {
+        formattedInput[i] = JSON.stringify(log)
+      } catch (e) {
+        // if we can't stringify thats OK
```
Shadowfiend, line 84: We should toString or assign the object directly in that case though right?

Shadowfiend (commented):
Glorious. Left one note and need to test drive so not approving yet.

background/lib/logger.ts
0xDaedalus: Should we?  toString is what gives us `[object Object]` in the first place I think.  Not sure what you mean by "assign the object directly" 😓

Shadowfiend: Yeah but if JSON.stringify fails and we don't set the index, I'm guessing we'll see `null`, which is misleading. Better to know there was an object IMO.

0xDaedalus: Oh if you mean we should add

`formattedInput[i] = input[i]` 

thats already handled when we initialize `formattedInput`

Shadowfiend: 🙈 lawwd. 

You're right, of course 😅

henryboldi (approved):
The code looks good and it appears to work nicely! Great job 👑👑👑

PR #1418, opened 2022-05-10
⛽️ Custom gas, miner fee, max base

henryboldi:
This adds a new gas picker interface which displays more information. Specifically, it displays a miner fee (`maxPriorityFee`) and max base (`maxFee - maxPriorityFee`) next to each option. Moreover, it adds the ability to pick a custom miner fee and max base.

- [x] **Displays max base and miner fee for each option**
A max base is somewhat unusual, but makes it clear for users. Typically, there's `maxPriorityFee` and `maxFee`. `maxPriorityFee` is included in the `maxFee` which can be confusing, so max base slips the base fee and the `maxPriorityFee`. A max priority fee is what we're calling the miner fee. While its naming implies max, visually always all is spent.
- [x] **Adds support for custom max base and miner fee**
Essentially this adds a custom gas option to Redux, which is then editable from the gas fee pick UI.
- [x] **Behind a feature flag. Set `CUSTOM_GAS_SELECT=true` to test**
Because the new gas feature is includes a slew of changes, I've put it behind a feature flag so we can merge changes into main incrementally. Due to the amount of changes to the settings, I've added `NetworkSettingsSelectDeprecated.tsx` which contains, aside from tweaks for compatibility with our updated store, the interface from the previous version.

_However, it's missing:_

- A migration for the custom gas field
Will add this in another PR. However, when testing make sure to give it a fresh install, and fresh `npm start`.
- Proper implementation of the design
I'd like to get the logic here approved separately from making it pretty.

custom

🤘


0xDaedalus:
> A migration for the custom gas field
Will add this in another PR. However, when testing make sure to give it a fresh install, and fresh npm start.

This makes me a bit nervous - I feel like migrations are one of the most fragile parts of our application and while I do agree that this PR already has quite a large diff I do think that our implicit convention of changes to the redux store requiring a corresponding migration is a useful convention to keep.

background/constants/network-fees.ts
```diff
@@ -2,21 +2,35 @@ export const ESTIMATED_FEE_MULTIPLIERS: { [confidence: number]: bigint } = {
...
+  0: 20n,
+}
+
+export const ESTIMATED_SPEED_IN_READABLE_FORMAT_RELATIVE_TO_CONFIDENCE_LEVEL: {
+  [confidence: number]: string
+} = {
+  70: "~5 min",
+  95: "~1 min",
+  99: "~30 sec",
+  0: "-",
```
0xDaedalus, line 30: We should be able to calculate estimated speed based on if the custom fee set right?  If the user is setting  the fee above whatever we predict `99` is we can say `<30 seconds` no?  Same with `> 5 mins` for less than our prediction for `70`.

background/constants/network-fees.ts
```diff
@@ -2,21 +2,35 @@ export const ESTIMATED_FEE_MULTIPLIERS: { [confidence: number]: bigint } = {
...
+  express: 13n,
+  instant: 18n,
+  custom: 20n,
+}
 
 export const MAX_FEE_MULTIPLIER: { [confidence: number]: bigint } = {
   70: 13n,
   95: 15n,
   99: 20n,
+  0: 20n,
```
0xDaedalus, line 21: The `0` as a key for custom confidence feels a bit awkward here and maybe is an indicator that we could rework some of the data structures around fees (although I completely understand why you've chosen it) - probably out of the scope of this PR though.

background/redux-slices/transaction-construction.ts
```diff
@@ -57,16 +59,24 @@ export type TransactionConstruction = {
...
+  regular?: BlockEstimate
+  custom?: BlockEstimate
 }
 
 export const initialState: TransactionConstruction = {
   status: TransactionConstructionStatus.Idle,
   feeTypeSelected: NetworkFeeTypeChosen.Regular,
-  estimatedFeesPerGas: undefined,
+  estimatedFeesPerGas: {
+    custom: {
```
0xDaedalus, line 73: Why do we add this to `intialState`?

background/redux-slices/transaction-construction.ts
```diff
@@ -168,57 +178,68 @@ const transactionSlice = createSlice({
...
-                (el) => el.confidence === REGULAR
-              )?.maxPriorityFeePerGas ?? 0n,
-            price:
-              estimatedFeesPerGas.estimatedPrices.find(
-                (el) => el.confidence === REGULAR
-              )?.price ?? 0n,
-          },
+      immerState.estimatedFeesPerGas = {
+        ...immerState.estimatedFeesPerGas,
+        baseFeePerGas: estimatedFeesPerGas.baseFeePerGas,
```
0xDaedalus, line 183: I think this block might benefit from extracting the construction of the `instant`, `express,` and `regular` blocks into a function.

Something like
```
const makeBlockEstimate = (type, estimatedFeesPerGas):  BlockEstimate => {
return {
          maxFeePerGas:
            (estimatedFeesPerGas.baseFeePerGas * MAX_FEE_MULTIPLIER[type]) /
            10n,
          confidence: type,
          maxPriorityFeePerGas:
            estimatedFeesPerGas.estimatedPrices.find(
              (el) => el.confidence === type
            )?.maxPriorityFeePerGas ?? 0n,
          price:
            estimatedFeesPerGas.estimatedPrices.find(
              (el) => el.confidence === type
            )?.price ?? 0n,
        }
}
```

background/redux-slices/transaction-construction.ts
```diff
@@ -168,57 +178,68 @@ const transactionSlice = createSlice({
...
+        payload: { maxPriorityFeePerGas, maxFeePerGas },
+      }: { payload: { maxPriorityFeePerGas: bigint; maxFeePerGas: bigint } }
+    ) => {
+      immerState.estimatedFeesPerGas = {
+        ...immerState.estimatedFeesPerGas,
+        custom: {
+          maxFeePerGas,
+          confidence: CUSTOM,
+          maxPriorityFeePerGas,
+          price: 41000000000n,
```
0xDaedalus, line 241: We may want to comment why we hard-code to this number

ui/components/NetworkFees/FeeSettingsText.tsx
```diff
@@ -51,14 +51,7 @@ export default function FeeSettingsText(): ReactElement {
...
     typeof selectedFeeType !== "undefined"
-      ? truncateDecimalAmount(
-          weiToGwei(
-            (estimatedFeesPerGas?.baseFeePerGas *
-              ESTIMATED_FEE_MULTIPLIERS_BY_TYPE[selectedFeeType]) /
-              10n
-          ),
-          0
-        )
+      ? truncateDecimalAmount(weiToGwei(networkSettings.values.maxFeePerGas), 0)
```
0xDaedalus, line 54: We can remove the `ESTIMATED_FEE_MULTIPLIERS_BY_TYPE` import from this file

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -159,25 +178,28 @@ export default function NetworkSettingsSelect({
...
         networkSettings.gasLimit ?? networkSettings.suggestedGasLimit
 
-      if (
-        typeof instant !== "undefined" &&
-        typeof express !== "undefined" &&
-        typeof regular !== "undefined"
-      ) {
-        const basePrices = [regular, express, instant]
+      if (typeof instant !== "undefined") {
+        const basePrices = [regular, express, instant, custom]
```
0xDaedalus, line 186: I think `baseFees` might be a better variable name here to match line 181 (since we're extracting the variables from `estimatedFeesPerGas`)

ui/components/NetworkFees/NetworkSettingsSelectDeprecated.tsx
```diff
@@ -0,0 +1,343 @@
+import { BlockEstimate } from "@tallyho/tally-background/networks"
```
0xDaedalus, line 1: Should this be called `Deprecated`?  Its a new component right?

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -204,29 +226,95 @@ export default function NetworkSettingsSelect({
...
+                          gweiFloatToWei(parseFloat(value)), // @TODO Replace
+                          option.maxPriorityFeePerGas
+                        )
+                      }}
+                    />
+                  </div>
+                </div>
+              </button>
+            ) : (
+              <button
```
0xDaedalus, line 299: Should this and the above `custom` implementation be factored out into standalone components?

ui/components/NetworkFees/NetworkSettingsSelectDeprecated.tsx
```diff
@@ -0,0 +1,343 @@
...
+        )
+      })}
+      <div className="info">
+        <div className="limit">
+          <SharedTypedInput
+            id="gasLimit"
+            value={networkSettings.gasLimit?.toString() ?? ""}
+            placeholder={networkSettings.suggestedGasLimit?.toString() ?? ""}
+            onChange={setGasLimit}
+            parseAndValidate={(value) => {
```
0xDaedalus, line 239: I think we can avoid defining this function inline.

ui/components/NetworkFees/NetworkSettingsSelectDeprecated.tsx
```diff
@@ -0,0 +1,343 @@
...
+
+  return (
+    <div className="fees standard_width">
+      <div className="title">Network Fees</div>
+
+      <EstimateRefreshCountdownDivider />
+
+      {gasOptions.map((option, i) => {
+        return (
+          <button
```
0xDaedalus, line 215: Another component that can be factored-out into its own file I think.

0xDaedalus (changes requested):
🎉 🎉 🎉  I am so frickin excited for this to land!!

background/redux-slices/transaction-construction.ts
henryboldi: In `NetworkSettingsSelect` we use display the buttons using a map on the `gasOptions` object. If custom isn't present here, that option will not display. I added it to the initial state in particular because `custom` doesn't need to be grabbed from a service.

background/redux-slices/transaction-construction.ts
henryboldi: Good point!

ui/components/NetworkFees/NetworkSettingsSelectDeprecated.tsx
henryboldi: This is essentially the old gas settings component.

ui/components/NetworkFees/NetworkSettingsSelectDeprecated.tsx
henryboldi: I agree, but this is the old/deprecated version of the gas settings

background/constants/network-fees.ts
henryboldi: 100% agree! But me agreeing is custom so it's 0% 😜

ui/components/NetworkFees/NetworkSettingsSelectDeprecated.tsx
0xDaedalus: I see - then I am confused by [this block of code](https://github.com/tallycash/extension/pull/1418/files#diff-332672ebfc50afcbf72043e6fb2f9c62c622ade6859df2e34e7e27aeed07a858R246).

My understanding here is custom gas select is net_new right?  The code makes it seem that if we have the CUSTOM_GAS_SELECT feature flag turned on we show the deprecated component - am i having a brain fart here?

background/constants/network-fees.ts
henryboldi: Estimated speed for custom values would be awesome! That said, it's actually not in the design.
gas_select

cc @VladUXUI

ui/components/NetworkFees/NetworkSettingsSelectDeprecated.tsx
henryboldi: Crap, it was meant to be `if (!CUSTOM_GAS_SELECT) {`! 😂

background/constants/network-fees.ts
mhluongo: I'm not sure those times makes sense with the behavior we've seen with EIP-1559 on mainnet... more likely we'd see something like

Instant, ~15 sec
Express, ~30 seconds
Regular, > 30 seconds


... seeing as increasing the max base fee doesn't make a tx land any faster, it makes a tx land more reliably... and increasing the priority should lead to it landing faster.
  
I don't have the data (we should collect it), but raising that before we remove the flag

VladUXUI: >Estimated speed for custom values would be awesome!

Didn't think we can calculate this tbh, we can definitely add time for custom as well. 
I expect/assume most users will use custom for higher values, for these values we can probably show <30 sec
For lower values we can calculate based on our estimations.

That being said, i don't see this as a big gain for users, but if it makes sense to spend the dev time. I'm all for it

VladUXUI: >seeing as increasing the max base fee doesn't make a tx land any faster

I mean i guess it depends on network conditions right? 
If network base fee goes higher than currently selected max base, then you have to potentially wait for the basefee to come down back again in order for your tx to be added/mined. Thus increasing the waiting times.

VladUXUI: >Instant, ~15 sec
>Express, ~30 seconds
>Regular, > 30 seconds

My numbers were not based on anything tbh, so i'm good with anything here.
Just keeping in mind that it's best for the users to be pleasantly surprised that it's mined faster than feeling like they were cheated out of some gas.

Let's go with your values and we can adjust later if need be.
CC @henryboldi

background/constants/network-fees.ts
VladUXUI: Uncertain if this assumption is actually correct. But i think users that use custom are going to set higher fees rather than lower ones.
So i guess the question is, should we also have pre-filed values for custom? Maybe Instant + 20%?

@henryboldi @mhluongo

Shadowfiend: You can just add a "custom" key, no?

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -274,11 +360,15 @@ export default function NetworkSettingsSelect({
             align-items: center;
             background: #002522;
             box-sizing: border-box;
-            padding: 12px;
+            padding: 15px;
```
VladUXUI, line 363: Small thing. This is probably 16px, as we always use multiple of 2

mhluongo:
Is this ready for another look? And if we're still missing the Redux migration... is there an associated issue?

henryboldi:
Yes, it's ready for another look! 🙇‍♂️

https://github.com/tallycash/extension/issues/1495

background/redux-slices/transaction-construction.ts
```diff
@@ -57,16 +59,23 @@ export type TransactionConstruction = {
...
 export type EstimatedFeesPerGas = {
-  baseFeePerGas: bigint
-  instant: BlockEstimate | undefined
-  express: BlockEstimate | undefined
-  regular: BlockEstimate | undefined
+  baseFeePerGas?: bigint
+  instant?: BlockEstimate
+  express?: BlockEstimate
+  regular?: BlockEstimate
+  custom?: BlockEstimate
```
Shadowfiend, line 66: Bothered by having a `custom` entry in an estimated fees type with a BlockEstimate structure when it's not actually an estimate.

But mostly in a "not willing to put in the time to suggest a better approach" way 😅

Shadowfiend (commented):
Quick note that the migration should only be a separate PR if the redux store only requires them with the feature flag on---if the store will break on upgrade even with the flag off, we shouldn't merge until there's a migration.

henryboldi:
> if the store will break on upgrade even with the flag off, we shouldn't merge until there's a migration.

Yes, I believe the store will not break. It'll work as normal but without the custom option.

background/redux-slices/transaction-construction.ts
henryboldi: I feel you! 😬

Hmm, would it be possible to solve this with some renaming?

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -61,18 +61,23 @@ const gasOptionFromEstimate = (
...
     estimatedGwei: weiToGwei(
       (baseFeePerGas * ESTIMATED_FEE_MULTIPLIERS[confidence]) / 10n
     ).split(".")[0],
+    maxPriorityGwei: weiToGwei(maxPriorityFeePerGas),
     maxGwei: weiToGwei(maxFeePerGas).split(".")[0],
     dollarValue: dollarValue ? `$${dollarValue}` : "-",
     estimatedFeePerGas:
       (baseFeePerGas * ESTIMATED_FEE_MULTIPLIERS[confidence]) / 10n,
-    price,
+    baseMaxFeePerGas: BigInt(maxFeePerGas) - BigInt(maxPriorityFeePerGas),
```
0xDaedalus, line 79: Do we need to mix units here?  Based on the other properties in this object it seems like it would be more consistent for this to be in `gwei`

ui/components/NetworkFees/NetworkSettingsSelectDeprecated.tsx
```diff
@@ -0,0 +1,342 @@
...
+}
+
+type GasOption = {
+  confidence: string
+  type: NetworkFeeTypeChosen
+  estimatedGwei: string
+  maxGwei: string
+  dollarValue: string
+  estimatedFeePerGas: bigint
+  maxFeePerGas: bigint
```
0xDaedalus, line 32: I think comments here that the units here are `wei` might be helpful to future contributors.

background/redux-slices/transaction-construction.ts
```diff
@@ -82,8 +91,40 @@ export type Events = {
...
+  type: NetworkFeeTypeChosen
+  estimatedGwei: string
+  maxPriorityGwei: string
+  maxGwei: string
+  dollarValue: string
+  estimatedFeePerGas: bigint
+  baseMaxFeePerGas: bigint
+  baseMaxGwei: string
+  maxFeePerGas: bigint
+  maxPriorityFeePerGas: bigint
```
0xDaedalus, line 106: I think comments here that these are in `wei` might help future contributors.

background/redux-slices/transaction-construction.ts
```diff
@@ -82,8 +91,40 @@ export type Events = {
   broadcastSignedTransaction: SignedEVMTransaction
 }
 
+export type GasOption = {
+  confidence: string
+  estimatedSpeed: string
+  type: NetworkFeeTypeChosen
+  estimatedGwei: string
+  maxPriorityGwei: string
+  maxGwei: string
```
0xDaedalus, line 100: Should we consistently make the units in we use in here `wei` and then use `weiToGwei` where we need to display `Gwei` to the user?

henryboldi: That's a good idea. How would you feel about that happening in a follow-up PR?

0xDaedalus: I feel 💯 about that!

ui/components/Shared/SharedInput.tsx
```diff
@@ -30,6 +31,7 @@ export function SharedTypedInput<T = string>(props: Props<T>): ReactElement {
     onFocus,
     value: currentValue,
     errorMessage,
+    step = undefined,
```
jagodarybacka, line 34: `step` will be `undefined` anyway, no need to set it here I guess 😛

ui/components/NetworkFees/NetworkSettingsSelectOptionButtons.tsx
```diff
@@ -0,0 +1,211 @@
...
+        Max Base:
+        <div className="price">{`${option.baseMaxGwei}`}</div>
+      </div>
+      <style jsx>{`
+        .option {
+          width: 100%;
+          display: flex;
+          justify-content: space-between;
+          align-items: center;
+          background: #002522;
```
jagodarybacka, line 42: we already have this color in 2 other places, maybe it is worth adding to CSS variables?

0xDaedalus (approved):
What an absolute unit of a PR!  There is a small visual problem that I've made a [separate issue](https://github.com/tallycash/extension/issues/1505) but considering that the visual portion is feature flagged this guys is more-than-ready to be merged.  Bravo ser.

PR #1421, opened 2022-05-10
Properly cleanup event listeners on the Internal Provider.

0xDaedalus:
We had a small memory leak that became especially apparent when the app was disconnected and attempting to reconnect and call `eth_chainID` over long periods of time.

This issue _might_ be related to https://github.com/tallycash/extension/issues/1158

### DIFF
--- a/background/redux-slices/utils/contract-utils.ts
+++ b/background/redux-slices/utils/contract-utils.ts
@@ -19,7 +19,7 @@ export const internalProviderPort = {
     this.listeners.push(listener)
   },
   removeEventListener(toRemove: (message: any) => unknown): void {
-    this.listeners.filter((listener) => listener !== toRemove)
+    this.listeners = this.listeners.filter((listener) => listener !== toRemove)
   },
   origin: window.location.origin,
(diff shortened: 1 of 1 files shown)


mhluongo (approved):
Looks good.

background/redux-slices/utils/contract-utils.ts
```diff
@@ -19,7 +19,7 @@ export const internalProviderPort = {
     this.listeners.push(listener)
   },
   removeEventListener(toRemove: (message: any) => unknown): void {
-    this.listeners.filter((listener) => listener !== toRemove)
+    this.listeners = this.listeners.filter((listener) => listener !== toRemove)
```
Shadowfiend, line 22: Oof.

mhluongo: What am I missing here? This is an intentional state mutation...?

mhluongo: Are we looking for a `splice()` or something else that will preserve the reference? And if so, why?

PR #1422, opened 2022-05-10
Refresh the base asset along with tokens when refreshing token balances.

0xDaedalus:
This change makes it so that ETH balance update on our 1-minute balance update alarm.

### To Test
- disconnect from internet
- clean install the extension
- import an account
- Stay disconnected for ~1-2 minutes
- Connect to internet
- [ ] ETH balance should show up in the wallet within about 1 minute of reconnecting.

This is related to https://github.com/tallycash/extension/issues/1158


mhluongo (approved):
These PRs are so smol :heart:

Shadowfiend:
We should rename the alarm and the method to reflect that it's checking balances in general.

PR #1424, opened 2022-05-10
Explicit error logging inside of chainService.addAccountToTrack

0xDaedalus:
Since we invoke some async functions without awaiting them inside of chain service's addAccountToTrack - any errors in those functions are swallowed without being logged. 

Adding these catch's at least lets us log the errors - although unfortunately we do not get a relevant stack trace so extra labeling of the logs is necessary.

### To Test
- disconnect from internet
- install tally
- import a new wallet
- [ ] You should see the three new errors in the background log since we expect those functions to fail.


mhluongo (approved):
:ok_hand:

mhluongo (approved):
:rocket: :clock11:

PR #1425, opened 2022-05-10
Fix async control flow for chainService's `internalStartService`.

0xDaedalus:
Related to https://github.com/tallycash/extension/issues/1158

The best way to grock this PR is probably to go through the commits sequentially.

This PR started by noticing that [the code as written did not match the comment describing the code](https://github.com/tallycash/extension/blob/main/background/services/chain/index.ts#L249-L250) - which in turn illuminated that we were swallowing tons of error messages by not awaiting promises.

Some questions that this PR raises:
- Are we OK with allowing modern `for ... of` syntax?  The alternative here is using `.map` with `Promise.all` but that seems to be clunky to me personally.
- There are at least two [other](https://github.com/tallycash/extension/blob/main/background/services/provider-bridge/index.ts#L198) [places](https://github.com/tallycash/extension/blob/main/background/servicesindexing/index.ts#L232) where we use `async` inside of a `forEach` loop that makes it unclear if we want to `await` the function or not.  These are other places where we can easily have errors swallowed up.
- It seems that we favor the pattern of not awaiting promises quite heavily which is fine but we need a way to ensure visibility of their rejections in case there is a failure.  Do we like adding `.catch` everywhere or is there a better way?

### To Test
Ensure that wallet balances are correctly loaded when importing and that new transactions are successfully subscribed to.


Shadowfiend:
> Are we OK with allowing modern for ... of syntax? The alternative here is using .map with Promise.all but that seems to be clunky to me personally.

Not really in this case, no. For/of with an await serializes subscriptions in a way that .map with Promise.all does not, and slows down startup in a hypothetical multi-network setup.

In general, I'm mixed. I prefer to stick with the linter rules we've got, and I like the clarity of forEach vs map to indicate side effects vs caring about return values. For/of makes it unclear what we're going for, but it does let us serialize certain promise flows that are more convoluted to construct with simple functional calls (though we can probably pattern this out with a helper). 

> There are at least two other places where we use async inside of a forEach loop that makes it unclear if we want to await the function or not. These are other places where we can easily have errors swallowed up.

We shouldn't think of this as unclear--if it's `forEach`, we're explicitly saying we don't want to await (ie the async is only there to support the internal usage of await). That might not be *correct* in some situations, and we should *probably* not silently discard the failures (or we should leave a note as to why), but the pattern of "`forEach` exists to parallelize promises whose outcomes we're not interested in" is worth absorbing. Put differently, `forEach` is a good signal that we're doing the internals strictly for side effects. 

> It seems that we favor the pattern of not awaiting promises quite heavily which is fine but we need a way to ensure visibility of their rejections in case there is a failure. Do we like adding .catch everywhere or is there a better way?

I wouldn't say we favor it but rather that we've done it in a lot of places. It's a form of laziness from our early work that's inertiaed itself throughout the code. We should def consider our usage of promises, especially non-void promises, and make sure our control flow makes sense if un-awaited promises fail. 

Agree on the logging, I wonder if there's a helper we can use to wrap these consistently, and then a linter rule we can find that warns us about unawaited promises. It feels like the typescript eslint rules package I've looked into in the past probably has this, just need to dig it up.

0xDaedalus:
> Not really in this case, no. For/of with an await serializes subscriptions in a way that .map with Promise.all does not, and slows down startup in a hypothetical multi-network setup.

The comment in the code suggests that we _do_ want to slow down startup though - right?  Consider the code as it was prior to this PR:
```
    this.supportedNetworks.forEach(async (network) => {
      const provider = this.providerForNetwork(network)
      if (provider) {
        const promises = Promise.all([
          provider.getBlockNumber().then(async (n) => {
            const result = await provider.getBlock(n)
            const block = blockFromEthersBlock(network, result)
            await this.db.addBlock(block)
          }),

          this.subscribeToNewHeads(network),
        ])
        if (network.chainID !== ETHEREUM.chainID) {
          // only block start to get Ethereum data
          await promises
        }
      } else {
        logger.error(`Couldn't find provider for supported network ${network}`)
      }
    })
```

That `await` isn't actually blocking anything.  There is nothing else to execute in the async function after the promise is awaited - so even if we do `await` nothing is actually being delayed right?  Seems like what we really want here is a `map` with an `allSettled`.

>  I like the clarity of forEach vs map to indicate side effects vs caring about return values.

I feel the same way about the clarity of `for ... of` vs `map` 😅 

> We shouldn't think of this as unclear--if it's forEach, we're explicitly saying we don't want to await.

To me it was unclear because the functions linked above are defined as async when they do not need to return a promise.  Maybe a step-forward in terms of clarity here is to fix those typings.

0xDaedalus:
So I think the outstanding question here is do we want to block on ETH subscription to new heads or should we keep the current behavior and delete the comment?

Shadowfiend:
> > I like the clarity of forEach vs map to indicate side effects vs caring about return values.
> I feel the same way about the clarity of for ... of vs map 😅

I think for/of and forEach def indicate that, but for/of says nothing about parallelism, and forEach is allowed by our current linter rules. That's reason enough for me unless we're specifically trying to serialize--that's really the only place I'm a bit torn, and tbh it's rare enough that we can probably do a linter exception for that sliver of cases. 

> The comment in the code suggests that we do want to slow down startup though - right? 

The meat of the problem! I think I had thought the comment was new and you were referring to the comment above the block. I'm not... Entirely clear what blocking startup here gets us to be honest. If the user brings up the UI before startup is unblocked and Ethereum is "connected", do we lose redux actions to the background script? Queue them up? What happens to the UI? Does it start in an incomplete state?

If we look at the previous state of the code, I had left a note because of some ineffective `Promise.all`s:

https://github.com/tallycash/extension/blob/cdada009edb0df5c88109f28ddbc82e226e7c4bf/background/services/chain/index.ts#L231

I believe @mhluongo made a second attempt at part of that in his commit adding multiple networks: 

https://github.com/tallycash/extension/commit/13298f54c73c040ac5fe5b7b2488a16a1e8a4a9a#diff-91b61b0f030f6297dc4a42d91331e187c77d1ca989cb2f277a62c4aacb4245b2L228

But this PR makes it clear that didn't work either. At this point it's clear this has never worked this way... So I'm leaning towards not making it work this way unless we specifically identify a problem that it will fix. 

In general I would rather let the rest of the system work on the possibility of the chain not having connected "fully" (ie with head tracking) yet, since that means it handles the possibility of being out of sync later as well. Alternatively, we could have the providers transparently handle the critical subscriptions from caller perspectives. Either way, I think we drop the await, log errors, and call it a day. My 2c.

background/services/chain/index.ts
```diff
@@ -238,34 +238,37 @@ export default class ChainService extends BaseService<Events> {
     // get the latest blocks and subscribe for all support networks
     // TODO revisit whether we actually want to subscribe to new heads
     // if a user isn't tracking a relevant addressOnNetwork
+    // eslint-disable-next-line no-restricted-syntax
```
Shadowfiend, line 241: We can lose this now!

0xDaedalus: whoopsie

background/services/chain/index.ts
```diff
@@ -238,34 +238,37 @@ export default class ChainService extends BaseService<Events> {
...
 
           this.subscribeToNewHeads(network),
-        ])
-        if (network.chainID !== ETHEREUM.chainID) {
-          // only block start to get Ethereum data
-          await promises
-        }
+        ]).catch((e) => {
+          logger.error("Error getting block number or new head", e)
+        })
```
Shadowfiend, line 255: Wondering if we drop `Promise.all` and individually log the two promises (failing to grab the block number and failing to subscribe to new heads). Alternatively, pushing the logging for new heads into the method makes sense to me as well.

background/services/chain/index.ts
```diff
@@ -238,34 +238,37 @@ export default class ChainService extends BaseService<Events> {
...
       } else {
         logger.error(`Couldn't find provider for supported network ${network}`)
       }
     })
 
-    Promise.all(
+    Promise.allSettled(
```
Shadowfiend, line 261: This doesn't make a difference either way, right? I would expect if we use `allSettled` that we follow up with a `forEach` that handles the logging for any failed promises, rather than individually logging in each one.

Shadowfiend (approved):
Couple of pretty minor notes. I need to test drive but am out of time for now.

background/services/chain/index.ts
0xDaedalus: I guess I wanted to use `allSettled` here to emphasize that we expect the possibility that some of the promises in the array fail rather than using `all` which implicitly rejects if even one of the promises fail.  And yeah in that context I definitely agree that a `forEach` to log errors is a win here.

background/services/chain/index.ts
0xDaedalus: love it

jagodarybacka (approved):
Works fine, are we waiting for anything else here @0xDaedalus? To me it seems to be ready for a merge

greg-nagy:
> Entirely clear what blocking startup here gets us to be honest. If the user brings up the UI before startup is unblocked and Ethereum is "connected", do we lose redux actions to the background script? Queue them up? What happens to the UI? Does it start in an incomplete state?

I think nothing "bad" happens. The services are waiting for each other, so that part is good, event subscriptions happen etc. 

There can be a brief period when the newest block data is not yet in the db and the UI shows stale data. This can be problematic on first install but I doubt that it will ever cause any problem.

0xDaedalus:
> Works fine, are we waiting for anything else here @0xDaedalus? To me it seems to be ready for a merge

I think we're ready yeah

greg-nagy (approved):
This looks good, let's 🚚

PR #1428, opened 2022-05-11
🧮 Use 0x quote for pre-transaction gas limit setting

henryboldi:
This PR adds the ability to see the gas limit for a swap on the swaps page
before creating the transaction. Because gas setting are visible before the
sign transactions page, and the gas limit info comes from a distractions, we're
using an early gas quote here.

To test,
- Select two assets in swap
- Check if the gas limit appears


0xDaedalus (approved):
👑 🙇 👑

PR #1433, opened 2022-05-12
Move the referral bridge to first claim step

wojciech-turek:
We used to set the referral on each visit to the claim container
component, which caused the referrer to be re-set after we come
back from signing delegation. Now unless users go back to the
first step the referred variable will stay removed.

https://user-images.githubusercontent.com/61434499/168087395-a50c17e2-e48b-40c8-b369-7ee0e45f15ec.mov


0xDaedalus (approved):
👨‍🍳

PR #1435, opened 2022-05-12
fix hover area on asset select dropdown button

### LINKED ISSUE #1087: Swap from button has incorrect hover trigger area
### Discord Discussion Link

_No response_

Chrome

Select swap from asset.

### What did not work?

Hover effect is triggered when mouse is not over the Select token button
Screenshot 2022-03-04 at 13 23 30

### Version

_No response_

### Relevant log output

_No response_

rootial:
close #1087
Per comments in  #1087, i think we don't need the hover bound from the label. So just remove it simply and it works now.

@henryboldi


ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -388,16 +388,7 @@ export default function SharedAssetInput<T extends AnyAsset>(
 
   return (
     <>
-      <label
-        className="label"
-        htmlFor={
-          typeof selectedAsset === "undefined"
-            ? "asset_selector"
-            : "asset_amount_input"
-        }
```
henryboldi, line 397: I'm worried this would impact accessibility. What do you think?

henryboldi:
I love to see PRs from you, @rootial ❤️🙌

ui/components/Shared/SharedAssetInput.tsx
Shadowfiend: Yes, it will. We need to see if we can handle this with styling rather than changing the `for` attribute.

I'm actually not convinced the `for` should change depending on the component state, but if we removed that behavior we'd probably still want it to always point to the selector, not the amount input. That part requires more research.

Anyway, I wonder if we can do something like `label:hover ~ #asset_selector:hover` to override the button's default hover styles.

greg-nagy:
@henryboldi what should be the next step with this PR? Based on the comments I would expect a change request, but it's not clear to me if this is the case.

PR #1436, opened 2022-05-12
🎨  Redesign of "new wallet" onboarding flow

### LINKED ISSUE #905: Seed gen - Design hasn't been fully implemented
### Discord Discussion Link

_No response_

Chrome

n/a

### What did not work?

See design here: https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=5474%3A154623

Specifically the 'reveal' design is missing. Not sure if we need or want that?

Also the presentation of the seed just doesn't seem up to our standards for visual fidelity. It doesn't copy and paste well either... we don't want people copying and pasting the seed, but some users may want to save in a text file offline

Screen Shot 2022-01-27 at 6 00 24 PM

### Version

_No response_

### Relevant log output

_No response_

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1338
Resolves #905

### What

Onboarding with new wallet creation has been redesigned to warn users about the importance of saving the seed. Also, the verification step has better UX - after failed verification user can try one more time without a need to regenerate the wallet. 

image

### Testing

1. Reinstall the extension and choose an option to create a new wallet during onboarding
2. go through seed creation and fail on word selection (seed verification) step
3. go through the verification step successfully


0xDaedalus (commented):
Works wonderfully!  Aside from some small nits I did notice some odd scrolling behavior.  I see that you've correctly implemented the figma design but just wanted to cc @VladUXUI that this is what we want.  

I especially found the 2nd screen jarring for a split-second - until I figured out that I had to scroll down to get to the "I wrote it down" button.

https://user-images.githubusercontent.com/94649004/168312266-ca01a347-18c5-4799-a5f4-bf25d64c4e23.mov

ui/pages/Onboarding/VerifySeed/VerifySeedError.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <span>Wrong order</span>
+        </div>
+        <p>
+          We are sorry, the recovery phrase you entered did not match.You can
+          try to re-order them, but make sure you have them written down.
+        </p>
+        <p>
+          If you prefer you can{" "}
+          <Link to="/">
+            <span className="link">start a new wallet</span>
```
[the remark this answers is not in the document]
jagodarybacka: I don't have any strong preferences, what do you think @VladUXUI?

jagodarybacka:
@0xDaedalus I don't really like how it looks with these scrolls, I want to meet with @VladUXUI and discuss what we can do better. And I'm not sure how we want to approach the issue with the "TallyHo" watermark on top (content is overlapping with it when scrolled).

ui/pages/Onboarding/VerifySeed/VerifySeedError.tsx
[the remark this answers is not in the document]
VladUXUI: Create a new wallet sounds good to me, it's a bit more clear than start over, that might mean start the process again.
Also talk to Jagoda about it, and the try again button will take you to your seed to double check what you wrote down

henryboldi:
We're waiting on an update to the scroll behavior before taking another look, right?

jagodarybacka:
@henryboldi no, a bunch of changes have been done after Vlad took a look, it is ready for review 😄

henryboldi:
Ah, excellent! I'll take a look today 😊

ui/pages/Onboarding/OnboardingSaveSeed.tsx
```diff
@@ -1,27 +1,72 @@
...
+  OnboardingHeader,
+  OnboardingSubheader,
+  OnboardingContainer,
+  OnboardingMessageHeader,
+} from "./styles"
 
-export default function OnboardingSaveSeed(): ReactElement {
+function WarningMessage({ onAccept }: { onAccept: () => void }): ReactElement {
+  return (
+    <>
```
henryboldi, line 19: Does it need to be wrapped in `<></>`? I often do this, but I know it has been something we prefer to avoid if possible 😅

ui/pages/Onboarding/OnboardingSaveSeed.tsx
```diff
@@ -1,27 +1,72 @@
...
+      <div className="onboarding_box">
+        <div className="message_header">
+          <img
+            className="message_icon"
+            src="./images/message_warning.png"
```
henryboldi, line 24: Why not `@2x`? Good to signify support for retina/4k displays, and these images are small file sizes anyway

ui/pages/Onboarding/styles.ts
```diff
@@ -0,0 +1,38 @@
...
+    border-radius: 8px;
+    font-weight: 600;
+    color: var(--green-20);
+    padding: 24px;
+    margin: 15px 0 24px;
+    display: flex;
+    flex-direction: column;
+    width: 280px;
+`
+export const OnboardingContainer = `
```
henryboldi, line 12: We should do this more often :)

henryboldi (approved):
Excellent! :)

Giving this a double check before merging

henryboldi (commented):
Nit: It appears that the padding between the logo and the step indicator on the verify page is taller versus on the create and save steps.

henryboldi:
@SJQuality This is going to be an important one to verify 🙏

ui/pages/Onboarding/OnboardingSaveSeed.tsx
jagodarybacka: hmm, I don't like unnecassary elements in the filename but that is a good point. But as long as we don't have two versions of the same image I feel like it is not that important, what do you think?

PR #1437, opened 2022-05-12
Richness not Flatness: Don't flatten recipients in internal log recipient lookups

### LINKED ISSUE #1430: We are making an enormous amount of requests to UNS every ~1 minute
It looks like we're trying to look up each character of a given address somewhere.

image

Shadowfiend:
To preserve uniqueness, the transfer log recipient addresses are
extracted and then passed through a Set. Due to a small error in how
the recipient addresses were being fed to the Set, the addresses were
being fully flattened into individual characters instead of addresses,
leading to very incorrect and extra-repetitive naming lookups.

Closes #1430 .

### DIFF
--- a/background/services/enrichment/index.ts
+++ b/background/services/enrichment/index.ts
@@ -270,11 +270,11 @@ export default class EnrichmentService extends BaseService<Events> {
         (
           await Promise.allSettled(
             [
-              ...new Set(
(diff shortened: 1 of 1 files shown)


0xDaedalus (approved):
👨‍🍳

PR #1440, opened 2022-05-13
Small cleanup of typing of synchronous functions.

0xDaedalus:
`notifyEnrichedTransaction` and `notifyContentScriptsAboutAddressChange` are actually synchronous functions that fire off unawaited promises outside of our control flow.  This PR updates their typing to match their behavior.


Shadowfiend (approved):
Looks good, will try to give this a test run when I get a chance but if someone beats me, :shipit:

Shadowfiend:
Both paths look good as expected here. Let's do eet.

PR #1441, opened 2022-05-13
Support UNS Name Resolution when using Polygon

0xDaedalus:
### To Test
- Importing a read-only address via UNS name should be possible when the `MULTI_NETWORK` flag is set to true.


jagodarybacka:
@0xDaedalus after extension reinstall it seems that network is set to Polygon, is it expected?
image

jagodarybacka:
What addresses are you using to test it? I've tried with official test addresses given by UNS and it works fine

0xDaedalus:
> after extension reinstall it seems that network is set to Polygon, is it expected?

Yes - if MULTI_NETWORK is set to true network will be set to Polygon.

> What addresses are you using to test it? I've tried with official test addresses given by UNS and it works fine

I use brad.crypto usually

PR #1442, opened 2022-05-13
Remove UNS Feature Flag

0xDaedalus:
Since this is now a fully-supported feature - we do not need the feature flag anymore.

### To Test
- UNS names should correctly resolve when importing read-only wallets and sending assets.


jagodarybacka (approved):
Works fine 💫

PR #1447, opened 2022-05-16
Add Polygon to Selectable Networks

0xDaedalus:
This PR adds Polygon to the list of selectable networks in our top menu protocol selector.

### To Test
- Set MULTI_NETWORK=true in your .env file.
- Compile & load the app.
- The network dropdown menu should contain polygon as an options.

**Note that selecting the network does not actually do anything yet**

image


jagodarybacka (approved):
it works 🔥

PR #1448, opened 2022-05-16
Remove unused addOrEditAddressName event in ui redux slice.

0xDaedalus:
This seems to be a holdover from a previous implementation attempt - discarded in favor of the [addOrEditAddressName](https://github.com/tallycash/extension/blob/main/background/redux-slices/accounts.ts#L325) function.


jagodarybacka (approved):
Works fine

PR #1449, opened 2022-05-16
Add the .vscode folder to .gitignore.

0xDaedalus:
Reasoning here is that we should allow developers to customize their development environment while developing without forcing a particular set of settings onto anyone or cluttering up the filesystem of developers who do not use vscode.


mhluongo (approved):
:+1:

PR #1477, opened 2022-05-16
Ξ -> ♾️  Switch to Polygon.

0xDaedalus:
This PR allows users to use the top menu protocol switcher to switch between Ethereum & Polygon and see their network balance on the corresponding network.

https://user-images.githubusercontent.com/94649004/168637792-048443f3-1e20-4371-b428-f49c497590b1.mov

### To Test
- Compile the extension with MULTI_NETWORK=true in your `.env` file.
- Import a wallet to the extension.
- Note balance on Polygon (import should default to this chain when MULTI_NETWORK=true).
- Use top switcher to switch to Ethereum.
- Your ETH balance should load quickly and your token balances should load over time.
- Switching back and forth between Polygon and Ethereum should correctly show your balance on the respective chain.

### DIFF
--- a/background/constants/networks.ts
+++ b/background/constants/networks.ts
@@ -18,7 +18,7 @@ export const POLYGON: EVMNetwork = {
 }
 
 export const ARBITRUM_ONE: EVMNetwork = {
-  name: "Arbitrum One",
+  name: "Arbitrum",
   baseAsset: ETH,
   chainID: "42161",
   family: "EVM",
--- a/background/redux-slices/ui.ts
+++ b/background/redux-slices/ui.ts
@@ -67,8 +67,17 @@ const uiSlice = createSlice({
       ...state,
       showingActivityDetailID: transactionID,
     }),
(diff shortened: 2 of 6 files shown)


jagodarybacka (approved):
Switcher works fine.

Improvements we could do:
- add `cursor: pointer` and change color of the network name on mouse hover (to match the arrow)
- maybe close the switcher after user has selected network? it seems weird to me that it is not auto-closing, I don't have anything else to do in this switcher after the network has been selected ;)
- it seems weird that account name and avatar is changing on network change
It's the same address:
image
image
- switching to another address is changing the network back to Ethereum
- I'm not sure if it is working correctly on the chain fork - have you added $MATIC to the accounts?

But as this PR is supposed to be merged to the feature branch, not main, I think it is good to go and we can address these issues later.

jagodarybacka:
> No - although once we re-fork we'll have MATIC in the test wallet since I put some into the mainnet account yesterday.

@0xDaedalus what is weird is that there are MATICs on the fork, the same amount as $ETH so I suppose that something is messed up there 

image
image

PR #1483, opened 2022-05-16
🦆🦆 Porygon2 - Sending and Swapping Assets on Polygon

### LINKED ISSUE #1446: L2 Spike Week of 5/16/22
Broad Goals (Polygon):
- [x] Switching between accounts via header menu correctly changes asset state.
- [x] Swaps working.
- [x] Sends of base asset working.

0xDaedalus:
image
> Porygon2's evolution from [Porygon](https://bulbapedia.bulbagarden.net/wiki/Porygon_(Pok%C3%A9mon)) was the result of it being upgraded from the most cutting-edge technology available. It was upgraded to work in space for planetary devolvement reasons, although its inability to fly limits this ability.

This PR adds the ability to send and swap assets on polygon!

Closes #1446 

There is however - a caveat.  Gas estimation is quite poor (since we are not using blocknative for polygon) so if your transaction cost is under 32 gwei it will fail.  Waiting on #1418 to land before diving back into fixing the gas issues.

### To Test
- [ ] mainnet swap assets should populate when MULTI_NETWORK is set to false;
- [ ] mainnet swap assets should populate when MULTI_NETWORK is set to true;
- [ ] polygon swap assets should populate when MULTI_NETWORK is set to true;
- [ ] polygon swap and mainnet swap should work when MULTI_NETWORK is set to true.
- [ ] mainnet swaps should still work when MULTI_NETWORK is set to false.
- [ ] sending ETH and ERC20 tokens on mainnet should work when MULTI_NETWORK is set to false.
- [ ] sending MATIC and ERC20 tokens on polygon and mainnet should work when MULTI_NETWORK is set to true .


background/lib/gas.ts
```diff
@@ -5,12 +5,72 @@ import Blocknative, {
...
+type PolygonGasResponse = {
+  safeLow: PolygonFeeDetails
+  standard: PolygonFeeDetails
+  fast: PolygonFeeDetails
+  estimatedBaseFee: number // gwei
+  blockTime: number
+  blockNumber: number
+}
+
+// Not perfect but works most of the time.  Our fallback method does not work at all for polygon.
```
mhluongo, line 29: A little more detail here would be nice

background/lib/gas.ts
```diff
@@ -5,12 +5,72 @@ import Blocknative, {
...
+const getPolygonGasPrices = async (price: bigint): Promise<BlockPrices> => {
+  const res = await fetch("https://gasstation-mainnet.matic.network/v2", {
```
mhluongo, line 31: Not blocking: Any early thoughts on how we abstract this away? This is a weird spot for a magic string. Also wondering if this `method` is required (isn't `GET` default?) and whether we can use the `fetchJSON` util we use elsewhere.

background/lib/gas.ts
```diff
@@ -5,12 +5,72 @@ import Blocknative, {
...
+    method: "get",
+  })
+
+  const gasEstimates = (await res.json()) as PolygonGasResponse
```
mhluongo, line 35: We also need a TODO for parsing and validation here, at a minimum. Also would be good to see some logging here... gas prices not showing up is lame, and we've struggled with it in the past.

background/lib/gas.ts
```diff
@@ -54,20 +126,20 @@ export default async function getBlockPrices(
       estimatedPrices: [
         {
           confidence: 99,
-          maxPriorityFeePerGas: 2500000000n,
-          maxFeePerGas: baseFeePerGas * 2n + 2500000000n,
+          maxPriorityFeePerGas: 2_500_000_000n,
```
mhluongo, line 129: nice

background/redux-slices/0x-swap.ts
```diff
@@ -103,12 +105,27 @@ export default swapSlice.reducer
...
+  // Use gated features if there is an API key available in the build.
+  const prefix =
+    typeof process.env.ZEROX_API_KEY !== "undefined" &&
+    process.env.ZEROX_API_KEY.trim() !== ""
+      ? "gated."
+      : ""
+
+  const base = chainIdTo0xApiBase[network.chainID]
+  if (!base) {
+    throw new Error(`Swaps not supported on ${network.name}`)
```
mhluongo, line 123: It feels like this should return a `null` base and log rather than `throw`

background/redux-slices/0x-swap.ts
```diff
@@ -128,9 +145,12 @@ const gatedHeaders: { [header: string]: string } =
 function build0xUrlFromSwapRequest(
   requestPath: string,
   { assets, amount, slippageTolerance, gasPrice }: SwapQuoteRequest,
+  network: EVMNetwork,
   additionalParameters: Record<string, string>
 ): URL {
-  const requestUrl = new URL(`https://${zeroXApiBase}/swap/v1${requestPath}`)
+  const requestUrl = new URL(
+    `https://${get0xApiBase(network)}/swap/v1${requestPath}`
```
mhluongo, line 152: All of this URL building without a lib is bugging me, but not a huge issue now

background/redux-slices/ui.ts
```diff
@@ -148,6 +150,14 @@ export const setSelectedNetwork = createBackgroundAsyncThunk(
     const state = getState() as { ui: UIState; account: AccountState }
     const { ui, account } = state
     dispatch(setNewSelectedAccount({ ...ui.selectedAccount, network }))
+    dispatch(clearSwapQuote())
+    const provider = getProvider()
+    // dogfood our switchEthereumChain handler
+    provider.send("wallet_switchEthereumChain", [
+      {
+        chainId: `0x${Number(network.chainID).toString(16)}`,
```
mhluongo, line 158: We need a util for this, it's going to come up a lot

background/services/internal-ethereum-provider/index.ts
```diff
@@ -116,8 +128,7 @@ export default class InternalEthereumProviderService extends BaseService<Events>
         return this.signTypedData({
           account: {
             address: params[0] as string,
-            // TODO Support variable network.
-            network: ETHEREUM,
+            network: this.activeNetwork,
```
mhluongo, line 131: Nice intermediate step

background/services/internal-ethereum-provider/index.ts
```diff
@@ -126,7 +137,7 @@ export default class InternalEthereumProviderService extends BaseService<Events>
         // allowed to have an RPC call made to it. Ideally this would be based
         // on a user's idea of a dApp connection rather than a network-specific
         // modality, requiring it to be constantly "switched"
-        return `0x${BigInt(ETHEREUM.chainID).toString(16)}`
+        return `0x${BigInt(this.activeNetwork.chainID).toString(16)}`
```
mhluongo, line 140: Yep, `utils.networks.hexChainID(network)` or something

background/services/internal-ethereum-provider/index.ts
```diff
@@ -198,6 +209,19 @@ export default class InternalEthereumProviderService extends BaseService<Events>
...
+        const newNetwork = EVM_MAIN_NETWORKS.find(
+          (network) =>
+            network.chainID === newChainId ||
+            `0x${Number(network.chainID).toString(16)}` === newChainId
+        )
+        if (newNetwork) {
+          this.activeNetwork = newNetwork
+          return null
+        }
+        throw new Error("Unsupported Chain")
```
mhluongo, line 223: I think you want something similar to this

```
 throw new EIP1193Error(EIP1193_ERROR_CODES.unsupportedMethod)
```

I don't know [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) well, but code `4901` seems reasonable...

ui/components/Wallet/WalletActivityDetails.tsx
```diff
@@ -109,11 +111,18 @@ export default function WalletActivityDetails(
...
   const { activityItem } = props
 
+  const network = useBackgroundSelector(selectCurrentNetwork)
+
   const openExplorer = useCallback(() => {
     window
-      .open(`https://etherscan.io/tx/${activityItem.hash}`, "_blank")
+      .open(
+        `https://${
+          network.name === "Polygon" ? "polygonscan.com" : "etherscan.io"
```
mhluongo, line 120: We should pull this out, we're going to need to make this call a lot

ui/pages/Swap.tsx
```diff
@@ -137,17 +143,40 @@ export default function Swap(): ReactElement {
...
-          (isFungibleAsset(asset) && asset.symbol === "ETH")) &&
-        asset.symbol !== sellAsset?.symbol
+      (asset): asset is SmartContractFungibleAsset | FungibleAsset => {
+        // We don't want to buy the same asset we're selling.
+        if (asset.symbol === sellAsset?.symbol) {
+          return false
+        }
+
+        if (isSmartContractFungibleAsset(asset)) {
+          // No cross-chain swaps (yet!!)
```
mhluongo, line 165: :sunglasses:

mhluongo (commented):
So excited for this! I left a few comments but haven't been able to test and want to give it a deeper review.

background/lib/gas.ts
```diff
@@ -31,14 +91,26 @@ export default async function getBlockPrices(
...
     provider.getBlock("latest"),
     provider.getFeeData(),
   ])
+
   const baseFeePerGas = currentBlock?.baseFeePerGas?.toBigInt()
 
+  if (network.name === "Polygon") {
+    try {
+      return await getPolygonGasPrices(
+        feeData?.gasPrice?.toBigInt() ||
```
Shadowfiend, line 104: Believe you're looking for `??`.

background/redux-slices/0x-swap.ts
```diff
@@ -103,12 +105,27 @@ export default swapSlice.reducer
...
 
-// Use gated features if there is an API key available in the build.
-const zeroXApiBase =
-  typeof process.env.ZEROX_API_KEY !== "undefined" &&
-  process.env.ZEROX_API_KEY.trim() !== ""
-    ? "gated.api.0x.org"
-    : "api.0x.org"
+const chainIdTo0xApiBase: { [chainID: string]: string | undefined } = {
+  "1": "api.0x.org", // Ethereum
+  "137": "polygon.api.0x.org", // Polygon
```
Shadowfiend, line 110:
```suggestion
  [ETHEREUM.chainID]: "api.0x.org", // Ethereum
  [POLYGON.chainID]: "polygon.api.0x.org", // Polygon
```

background/redux-slices/0x-swap.ts
Shadowfiend: [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) is effectively a lib, fwiw. It will parse, validate and decompose the URL, and allow us to attach querystring params safely like we do below.

For example:

```
> new URL("lol")
< TypeError: Type error
> new URL("lol.com")
< TypeError: Type error
> new URL("https://lol.com")
< URL {href: "https://lol.com/", origin: "https://lol.com", protocol: "https:", username: "", password: "", …}
> new URL("https://lol")
< URL {href: "https://lol/", origin: "https://lol", protocol: "https:", username: "", password: "", …}
```

We could switch to using the relative/base constructor if we wanted to add clarity:

```typescript
  const requestUrl = new URL(
    `/swap/v1${requestPath}`,
    `https://${get0xApiBase(network)}`
  )
```

It doesn't provide ways to append path components together other than string concatenation, but I thought about that when I went down this path and it seems a small thing to go pull in or write a library for tbh.

background/redux-slices/0x-swap.ts
```diff
@@ -186,14 +206,21 @@ function build0xUrlFromSwapRequest(
...
-  async (quoteRequest: SwapQuoteRequest, { dispatch }) => {
+  async (quoteRequest: SwapQuoteRequest, { getState, dispatch }) => {
     const signer = getProvider().getSigner()
     const tradeAddress = await signer.getAddress()
 
-    const requestUrl = build0xUrlFromSwapRequest("/quote", quoteRequest, {
-      intentOnFilling: "true",
-      takerAddress: tradeAddress,
-    })
+    const { ui } = getState() as { ui: UIState }
```
Shadowfiend, line 213: I strongly recommend feeding this into the `SwapQuoteRequest` via the current network selector. We should avoid `getState` whenever possible IMO, as using it requires us to end-run around the type system.

background/redux-slices/transaction-construction.ts
```diff
@@ -166,59 +170,64 @@ const transactionSlice = createSlice({
...
-            confidence: REGULAR,
-            maxPriorityFeePerGas:
-              estimatedFeesPerGas.estimatedPrices.find(
-                (el) => el.confidence === REGULAR
-              )?.maxPriorityFeePerGas ?? 0n,
-            price:
-              estimatedFeesPerGas.estimatedPrices.find(
-                (el) => el.confidence === REGULAR
-              )?.price ?? 0n,
+          ...immerState.estimatedFeesPerGas,
```
Shadowfiend, line 180:
```suggestion
          ...(immerState.estimatedFeesPerGas ?? {}),
```
Some historical wallets could be in a weird state if the previously-default `undefined` value was still set in their state.

background/redux-slices/transaction-construction.ts
```diff
@@ -278,33 +287,31 @@ export const rejectTransactionSignature = createBackgroundAsyncThunk(
...
-    transactionConstruction,
-  }: {
-    transactionConstruction: TransactionConstruction
-  }) => ({
-    feeType: transactionConstruction.feeTypeSelected,
-    selectedFeesPerGas:
-      transactionConstruction.estimatedFeesPerGas?.[
+  (state: { transactionConstruction: TransactionConstruction }) =>
+    state.transactionConstruction,
+  (state: { ui: UIState }) => state.ui.selectedAccount.network,
```
Shadowfiend, line 292: https://github.com/tallycash/extension/blob/main/background/redux-slices/selectors/uiSelectors.ts#L42-L45 is useful here. You'll get a circular import warning I expect, which is a good sign that it's time to push a lot of these selectors into `selectors/transactionConstructionSelectors` or similar.

background/redux-slices/ui.ts
Shadowfiend: `getProvider` can extend the provider with a `switchChain` method that takes a `Network` pretty easily.

background/redux-slices/ui.ts
```diff
@@ -148,6 +150,14 @@ export const setSelectedNetwork = createBackgroundAsyncThunk(
     const state = getState() as { ui: UIState; account: AccountState }
     const { ui, account } = state
     dispatch(setNewSelectedAccount({ ...ui.selectedAccount, network }))
+    dispatch(clearSwapQuote())
```
Shadowfiend, line 153: This feels like a serious single responsibility breach. Feels like Main should handle clearing the swap quote on `newSelectedAccount` event no?

background/redux-slices/ui.ts
Shadowfiend: Also—I think this is probably also escaping single responsibility land. Main should probably notify the internal Ethereum provider directly about the latest active chain. This is dogfooding, but it's in a place where we're already doing internal/non-dAppy stuff, so it feels like it's muddying the waters a bit.

background/lib/gas.ts
```diff
@@ -31,14 +91,26 @@ export default async function getBlockPrices(
...
   const [currentBlock, feeData] = await Promise.all([
     provider.getBlock("latest"),
     provider.getFeeData(),
   ])
+
   const baseFeePerGas = currentBlock?.baseFeePerGas?.toBigInt()
 
+  if (network.name === "Polygon") {
+    try {
+      return await getPolygonGasPrices(
```
Shadowfiend, line 103: It looks like the correct way to do this and the previous type of query is to use `eth_feeHistory` to get the percentile tips for the past couple of blocks and do some sort of simple guesstimate based on those (assuming more complex estimates are handled by Blocknative). `eth_feeHistory` seems to be supported by Alchemy for both mainnet and Polygon networks, Infura on both, and Pokt on mainnet (Ethers doesn't support it on Polygon, yet?).

Thoughts on separately switching to that for a common approach?

background/redux-slices/0x-swap.ts
mhluongo: I'm mostly thinking about what happens if I put empty strings everywhere, etc etc. If you think `URL` is enough I'm good

background/redux-slices/ui.ts
mhluongo: Yay, I'm not the only one who's unsure where to use the provider!

background/lib/gas.ts
0xDaedalus: `fetchJSON` is definitely the way to go here :+1: 

I it might be nice to have a set of `resolvers` similar to what we have for name resolution here?

So for a given network we could have N ways to fetch gas (2 is probably enough - 1 preferred and 1 fallback) so for Ethereum we could prefer blocknative and fallback to our own gas estimate, for polygon prefer this api endpoint and fallback to our own estimates, etc...

Ideally I want something that will make it easy to swap systems of estimating in & out while having a "priority list" of sorts.  What do you think?

background/redux-slices/0x-swap.ts
0xDaedalus: May I ask why?  I feel like this adds a log and pushes the error away from where it originates (to where we actually make the request) rather than explicitly having the error propagate from where it (conceptually) originates from.

mhluongo: When we have more networks, I suspect the caller of the func will want to know whether a network is supported. Calling this and getting `null` is a great way to tell... calling it and getting a nondescript error doesn't get you there.

Open to being wrong ofc!

0xDaedalus: Great point and good intermediate solution I think.  I do wonder if we'll want a separate function to determine wether a swap is allowed (do we have the api information for the network? - is there a product or security reason to turn off swaps on the network?) when we move to multi network.

background/lib/gas.ts
0xDaedalus: Love it!

mhluongo: Follow up issue https://github.com/tallycash/extension/issues/1506

background/redux-slices/ui.ts
0xDaedalus: Sounds good to me - I've created a [separate issue](https://github.com/tallycash/extension/issues) for this comment that we can tackle in a subsequent PR if thats gravy with you guys.

background/lib/gas.ts
```diff
@@ -1,16 +1,79 @@
...
+  blockTime: number
+  blockNumber: number
+}
+
+// Not perfect but works most of the time.  Our fallback method does not work at all for polygon.
+// This is because the baseFeePerGas on polygon can be so small (oftentimes sub 100 wei (not gwei)) that
+// estimating maxFeePerGas as a function of baseFeePerGas almost always results in a "transaction underpriced"
+// being returned from alchemy because our maxFeePerGas is below its acceptance threshhold.
+const getPolygonGasPrices = async (price: bigint): Promise<BlockPrices> => {
+  // @TODO Validate this response using ajv
```
greg-nagy, line 35: Not a blocker but would be nice to have this handled in this PR.

0xDaedalus: Might be a moot point in the context of https://github.com/tallycash/extension/issues/1506

background/lib/gas.ts
greg-nagy: ditto

background/services/internal-ethereum-provider/index.ts
```diff
@@ -198,6 +210,19 @@ export default class InternalEthereumProviderService extends BaseService<Events>
           hexData: params[0] as string,
           account: params[1] as string,
         })
+      case "wallet_switchEthereumChain": {
```
greg-nagy, line 213: Out of scope for this PR and non blocker at all.

My mental model is that our wallet is connected to all networks at all times. Put it differently we should use different default chain for different dapps (and internnaly in the wallet) So having a global `active chain` enforces the `active account` account concept further. 

I see why this is needed though. Will ponder about this a bit more, just wanted to leave a note.

greg-nagy: Until it's global, it probably should be persisted in the preference service.

background/main.ts
```diff
@@ -71,11 +71,12 @@ import {
   transactionRequest,
   updateTransactionOptions,
   clearTransactionState,
-  selectDefaultNetworkFeeSettings,
   TransactionConstructionStatus,
   rejectTransactionSignature,
   transactionSigned,
+  EstimatedFeesPerGas,
```
greg-nagy, line 77: This can be removed

greg-nagy (approved):
Impressive 👏

Ran these tests
- [x] mainnet swap assets should populate when MULTI_NETWORK is set to false;
- [x] mainnet swaps should still work when MULTI_NETWORK is set to false.
- [x] sending ETH and ERC20 tokens on mainnet should work when MULTI_NETWORK is set to false.

- [x] mainnet swap assets should populate when MULTI_NETWORK is set to true;
- [x] polygon swap assets should populate when MULTI_NETWORK is set to true;
- [x] polygon swap and mainnet swap should work when MULTI_NETWORK is set to true.
- [x] sending MATIC and ERC20 tokens on polygon and mainnet should work when MULTI_NETWORK is set to true .
  - > Didn't test the ERC20 send on polygon because the asset discovery — probably token list issue — is not really working for me. 

Found an issue with adding wallets, creating a new ticket for it.

PR #1485, opened 2022-05-17
Remove unused getNonce function.

0xDaedalus:
Nonce logic should automatically be handled by the provider or inside of chain service.  This function is unused and can potentially be harmful if accidentally used to manually set a nonce.

### To Test
- compile the extension.

### DIFF
--- a/background/redux-slices/utils/contract-utils.ts
+++ b/background/redux-slices/utils/contract-utils.ts
@@ -53,14 +53,6 @@ export const getSignerAddress = async (): Promise<string> => {
   return signerAddress
 }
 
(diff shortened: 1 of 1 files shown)


background/redux-slices/utils/contract-utils.ts
```diff
@@ -53,14 +53,6 @@ export const getSignerAddress = async (): Promise<string> => {
   return signerAddress
 }
 
-export const getNonce = async (): Promise<number> => {
```
mhluongo, line 56: :ballot_box_with_check:

mhluongo (approved):
Snip :scissors:

PR #1491, opened 2022-05-18
🦊  Impersonate MetaMask on OpenSea when TallyHo is set as default wallet

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/790
- Allow OpenSea to connect with TallyHo after click on Metamask icon if TallyHo is set as the default wallet. 
- If TallyHo is not the default wallet clicking on Metamask on OpenSea will open Metamask or its download page.

image

### Testing

- set TallyHo as default wallet
- try to connect to [OpenSea](https://opensea.io/) via MetaMask icon
- switch off default wallet switcher
- reload OpenSea
- click on MetaMask icon should not open TallyHo

It would be nice to do some regression tests with different dapps we know are working correctly (Uniswap etc)


mhluongo:
@mr-michael @SJQuality could use both of your eyes here.

On the product side, the big question is whether being this aggressive is okay when Tally Ho and MetaMask are both installed alongside eachother. Do we need another setting, or is "Use as default wallet" enough? We don't want people uninstalling because we overstepped and they want to use both... and if that's a risk, maybe we need an explicit "Impersonate MetaMask" setting.

On the QA side, we need to see if this breaks the dApps like Curve that have enabled explicit support for Tally Ho. If so, we're picking up OpenSea support and losing a bunch of hard-won biz dev.

mhluongo (commented):
Mixed product / engineering response here.

PR #1500, opened 2022-05-20
🗑  Fix selecting other address if current one is removed

### LINKED ISSUE #1497: Removing currently selected account is not switching to any valid account
### Reproduction steps
1. Add a couple of accounts to the wallet
  image
2. Remove selected account
  image

Result: selected account is still on the removed address and it is in the invalid state
image

jagodarybacka:
Resolves #1497

### Testing

1. Add a few addresses 
2. Remove the currently selected address

**Expected result:** extension will switch selected account to another address


0xDaedalus (approved):
A w e s o m e ! 🎉

PR #1509, opened 2022-05-23
✏️  Claim flow - add input for custom user or DAO address

### LINKED ISSUE #1461: Custom address field on DAO bonus page
Screen Shot

jagodarybacka:
Resolves #1461 

### What
* Added input under the list of DAOs so users are able to choose their own referrer by typing its name or address.
* Improved resolving names and avatars for these custom referrers
* Added "change" button for referrer so user can easily change it

image
image


background/services/signing/index.ts
```diff
@@ -200,7 +200,9 @@ export default class SigningService extends BaseService<Events> {
         // Let parseInt infer radix by prefix; chainID can be hex or decimal,
         // though it should generally be hex.
         // eslint-disable-next-line radix
-        typedData.domain.chainId !== parseInt(account.network.chainID)
+        parseInt(account.network.chainID) !==
+          // eslint-disable-next-line radix
+          parseInt(`${typedData.domain.chainId}`)
```
jagodarybacka, line 205: for some reason I couldn't sign delegation because `typedData.domain.chainId` was a string but it is expected to be a number 🤔

henryboldi:
Looking good!! 👑

If I remember correctly, isn't this custom DAO address here not really intended for DAO addresses? Well, I see you've implemented it to work with any address, but, in that case, we're perhaps massaging regular addresses onto the DAO type and name in a way that can get messy.

 cc @VladUXUI

jagodarybacka:
@henryboldi Yeah so the PR's title is misleading, as I understand it can be any address that user wants to share the bonus with so it's more like... choosing the referrer? I'm getting lost in who it referrer and delegate 😂 

But it's the same thing, given address will receive the bonus, no matter if it is a dao or custom address, right? I couldn't put this address as the `referrer` in the redux state as this would mess up the UI (switching to referrer style interface of this page). 

Maybe we shouldn't save the `address + name` from this page as a DAO, maybe we can rename it? I don't know the right word that would fit here 😞

henryboldi:
> Maybe we shouldn't save the address + name from this page as a DAO, maybe we can rename it? I don't know the right word that would fit here

How about `bonusOrigin` or `bonusBy`?

jagodarybacka:
@henryboldi hmm, but it is at the same time `bonusBy` and `bonusFor` 😄 as right now this field in redux state is named `selectedDAO` I think we can name it `selectedForBonus`

greg-nagy (changes requested):
This PR is looking good and works really well! Well done 🤩

There is the typing issue and a minor renaming that felt to be a blocker / should be explored and fixed ASAP.

background/services/signing/index.ts
jagodarybacka: I will revert this change and I will fix it as a separate issue to not block this PR

greg-nagy (approved):
let's ⛴  this beauty ✨

PR #1511, opened 2022-05-23
Fix Inverted `eth_sign` Parameters

nlordell:
This PR fixes `eth_sign` parameter order.

Specifically, `eth_sign` expects `[account, hexData]`. Here is the example from the [Ethereum JSON RPC documentation](https://eth.wiki/json-rpc/API#eth-sign):

```
// Request
curl -X POST --data '{"jsonrpc":"2.0","method":"eth_sign","params":["0x9b2055d370f73ec7d8a03e965129118dc8f5bf83", "0xdeadbeaf"],"id":1}'

// Result
{
  "id":1,
  "jsonrpc": "2.0",
  "result": "0xa3f20717a250c2b0b729b7e5becbff67fdaef7e0699da4de7ca5895b02a170a12d887fd3b17bfdce3481f10bea41f45ba9f709d39ce8325427b57afcfc994cee1b"
}
```

This differs from `personal_sign` which expects parameters in the order: `[hexData, account]`.


mhluongo:
Thanks for this @nlordell! Have a good dApp in mind for us to QA this against?

nlordell:
> Thanks for this @nlordell! Have a good dApp in mind for us to QA this against?

Hmm, can't think of any. I think it will be a hard find to actually encounter one where you can test this in the wild (I actually stumbled upon this as I was browsing the code looking at the state of `eth_signTypedData` support).

Specifically, [MetaMask implements `eth_sign` differently than the standard](https://docs.metamask.io/guide/signing-data.html#a-brief-history) for [historical reasons](https://github.com/MetaMask/metamask-extension/issues/9957#issuecomment-754826067):
> In particular, the method `eth_sign` is an open-ended signing method that allows signing an arbitrary hash, which means it can be used to sign transactions, or any other data, making it a dangerous phishing risk.

This makes it so most dApps already unconditionally use `personal_sign`. This is further facilitated by libraries like Ethers.js [was changed](https://github.com/ethers-io/ethers.js/commit/8947fd405e3aea07f6db958d89a3ad39abe3a25a) to [use `personal_sign`](https://github.com/ethers-io/ethers.js/blob/e85e32b8f74d6d1f809e3ab016170ff6b4bfb8dd/packages/providers/src.ts/json-rpc-provider.ts#L291-L296) for you under the hood for all wallets when signing a message:
```ts
    async signMessage(message: Bytes | string): Promise<string> {
        const data = ((typeof(message) === "string") ? toUtf8Bytes(message): message);
        const address = await this.getAddress();

        return await this.provider.send("personal_sign", [ hexlify(data), address.toLowerCase() ]);
    }
```

`eth_sign` has a juicy history in wallets 😛.

greg-nagy (approved):
this looks correct, thank you for this!

The comments, links, explanations and the demo are all highly appreciated 🙏

let's 🚢

PR #1512, opened 2022-05-23
Emigration: Fix account structure migration and move migrations to their own module

### LINKED ISSUE #1499: 0.13.4->0.13.5 migration is producing incorrect redux store structure
### **Build Info:**
0.13.4 --> 0.13.5

### **Related Ticket/PR**
https://github.com/tallycash/extension/pull/1496

### **Reproduction Steps:**

1. Have 0.13.4 installed
2. Upgrade to 0.13.5 by following these steps - (replacing relevant build versions) https://drive.google.com/file/d/1CnWuFYpcgmpTehbt5f1N1eaoxZyywr5p/view

### **Expected Result:**
0.13.5 works normally

### **Actual Result:**

image

https://www.awesomescreenshot.com/video/8995155?key=836c1ffafa780b84c64493e39a782f38
[logs_v0_13_4__2022-05-19T14_52_54-04_00.txt](https://github.com/tallycash/extension/files/8733033/logs_v0_13_4__2022-05-19T14_52_54-04_00.txt)

Shadowfiend:
In addition to the core fix, which is to make the account structure
migration land with a chain ID -> address structure instead of vice
versa, this PR moves migrations out of main.ts, gives each migration
its own file, and rolls up the migration process into a top-level
`background/redux-slices/migrations/index.ts` export for consumption
by `main.ts`.

Closes #1499.


background/redux-slices/migrations/to-6.ts
```diff
@@ -0,0 +1,35 @@
+// This migration transitions ledger device data to always track arbitrary data
+// signing state. It is defaulted to false for existing devices for safety, and
+// the flag will be updated when the Ledger is next connected.
+
+import { LedgerState } from "../ledger"
```
mhluongo, line 5: We should open an issue for this, it's going to break down the road

mhluongo: The types need to be frozen

background/redux-slices/migrations/index.ts
```diff
@@ -0,0 +1,62 @@
...
+      .sort()
+      .filter(([version]) => parseInt(version, 10) > resolvedVersion)
+      .map(([, migration]) => migration)
+    migratedState = outstandingMigrations.reduce(
+      (state: Record<string, unknown>, migration: Migration) => {
+        return migration(state)
+      },
+      migratedState
+    )
+  }
```
0xDaedalus, line 57: Man I would love a typeguard here that asserts that `migratedState` conforms to `RootState` to help us avoid typos when working with `unknown` and `as` but I don't know how we'd do it without either adding even more developer friction to making redux state changes or an outside dependency for transpiling typescript types to ajv schemas.

background/redux-slices/migrations/to-9.ts
```diff
@@ -0,0 +1,52 @@
+// This migration transitions the transaction construction slice's gas estimate
+// structure to support custom gas estimates.
+
+import { BlockEstimate } from "../../networks"
```
mhluongo, line 4: Here's another one that should've been frozen. Not for this PR but deserves a follow-up issue

mhluongo (approved):
Love it but haven't been able to test it. Looking for a second review here

background/redux-slices/migrations/index.ts
```diff
@@ -0,0 +1,62 @@
...
+const allMigrations: { [targetVersion: string]: Migration } = {
+  2: to2,
+  3: to3,
+  4: to4,
+  5: to5,
+  6: to6,
+  7: to7,
+  8: to8,
+  9: to9,
+}
```
0xDaedalus, line 33: Very soft suggestion that we change up the directory structure slightly and move this constant.
```
migrations/index.ts
- REDUX_STATE_VERSION
- migrateReduxState
migrations/versions/index.ts
- allMigrations
migrations/versions/to-N.ts
- migration logic
```

Just thinking that this file will get pretty big once we're at to-40 for example.  Might be nice to have the long list of imports in another file

Shadowfiend: I'm down to wait until it becomes a problem tbh. It's easy enough to collapse these in most modern editors, and I expect the index file will require relatively little maintenance other than the current state version (which is defined up top). Could be wrong about that though :grin:

background/redux-slices/migrations/index.ts
mhluongo: So if it doesn't conform, what do we do? Do we need to be able to roll these back? :sweat_smile:

0xDaedalus: I don't think so - but we get an explicit error at application load time rather than a potentially ambiguous error at application run time which makes this class of error quite a bit easier to debug.

[This is the specific class of error](https://github.com/tallycash/extension/commit/2397d75c42616f76430c444d9cc729950e12d0b3) I'm thinking of in this context - there's no guarantee that an exception will happen before a specific button is clicked or page is viewed.

0xDaedalus (approved):
🎉

background/redux-slices/migrations/index.ts
Shadowfiend: Eh, poked it a little, not as easy as I thought. There's a way where we require the last migration in the list to have return type `RootState` instead of `Record<string, unknown>`, but it's not clear what the downstream effects on new migrations might be. I think what we ideally want someday is a migration generator that handles transitioning the latest migration to the n-1st one somehow.

PR #1517, opened 2022-05-24
Notify the internal ethereum provider directly about network changes.

### LINKED ISSUE #1508: Refactor use of provider.send("wallet_switchEthereumChain") in redux-slices/ui.ts
getProvider can extend the provider with a switchChain method that takes a Network pretty easily.

Also—I think this is probably also escaping single responsibility land. Main should probably notify the internal Ethereum provider directly about the latest active chain. This is dogfooding, but it's in a place where we're already doing internal/non-dAppy stuff, so it feels like it's muddying the waters a bit.

_Originally posted by @Shadowfiend in https://github.com/tallycash/extension/pull/1483#discussion_r877565681_

0xDaedalus:
Closes #1508 

Do not merge until #1483 makes it into main.

### To Test
- [ ] switching from ethereum to polygon and back still works.

Also - I'm not sure of the utility of the `switchChain` extension of the provider if we notify directly through main - would you mind commenting @Shadowfiend

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1009,6 +1009,13 @@ export default class Main extends BaseService<never> {
         signingSliceEmitter.on("signatureRejected", rejectAndClear)
       }
     )
+
+    uiSliceEmitter.on("newSelectedNetwork", (network) => {
+      this.internalEthereumProviderService.routeSafeRPCRequest(
+        "wallet_switchEthereumChain",
+        [{ chainId: network.chainID }]
+      )
+    })
   }
 
   async connectProviderBridgeService(): Promise<void> {
--- a/background/tally-provider.ts
+++ b/background/tally-provider.ts
@@ -1,0 +1,13 @@
+import { Web3Provider } from "@ethersproject/providers"
+import { toHexChainID, EVMNetwork } from "./networks"
+
+// Not sure the best place to put this in but it feels like it deserves its own file.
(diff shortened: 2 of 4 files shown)


Shadowfiend:
I'll try and reload the mental state that made me suggest this in the first place today and give some more details on my thoughts.

background/tally-provider.ts
```diff
@@ -0,0 +1,13 @@
+import { Web3Provider } from "@ethersproject/providers"
+import { toHexChainID, EVMNetwork } from "./networks"
+
+// Not sure the best place to put this in but it feels like it deserves its own file.
```
greg-nagy, line 4: Because it has Tally in the name ... :)

background/main.ts
```diff
@@ -1009,6 +1009,13 @@ export default class Main extends BaseService<never> {
         signingSliceEmitter.on("signatureRejected", rejectAndClear)
       }
     )
+
+    uiSliceEmitter.on("newSelectedNetwork", (network) => {
+      this.internalEthereumProviderService.routeSafeRPCRequest(
+        "wallet_switchEthereumChain",
```
greg-nagy, line 1015: woooo :) I love this ... on the hand waving and all that level

greg-nagy (approved):
it's alive ⚡️

PR #1520, opened 2022-05-25
🐶  Treat DOGGO token as special kind of asset

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1476
### What

The DOGGO asset should be treated as a special kind of asset. It has different icons - rewards program and link to Earn tab. It also has a "locked" badge with a tooltip that explains what it means that DOGGO is locked.

image

### DIFF
--- a/ui/components/Shared/SharedTooltip.tsx
+++ b/ui/components/Shared/SharedTooltip.tsx
@@ -5,27 +5,38 @@ type VerticalPosition = "top" | "bottom"
 interface Props {
   verticalPosition?: VerticalPosition
   width: number
+  height?: number
   children: React.ReactNode
-  IconComponent?: () => ReactElement
+  IconComponent?: ({
+    isShowingTooltip,
+  }: {
+    isShowingTooltip: boolean
+  }) => ReactElement
 }
 
 function getHorizontalPosition(width: number) {
   return `right: -${width / 2 + 4}px;`
 }
 
-function getVerticalPosition(vertical: VerticalPosition) {
+function getVerticalPosition(vertical: VerticalPosition, height: number) {
   switch (vertical) {
     case "bottom":
-      return "top: 20px; margin-top: 5px;"
+      return `top: ${height}px; margin-top: 5px;`
     case "top":
-      return "bottom: 20px; margin-bottom: 5px;"
+      return `bottom: ${height}px; margin-bottom: 5px;`
     default:
       return ""
   }
 }
 
 export default function SharedTooltip(props: Props): ReactElement {
-  const { children, verticalPosition = "bottom", width, IconComponent } = props
+  const {
+    children,
+    verticalPosition = "bottom",
+    width,
+    height = 20,
+    IconComponent,
+  } = props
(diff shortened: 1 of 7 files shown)


ui/components/Wallet/AssetListItem/DoggoAssetListItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+                    width={18}
+                    color="var(--trophy-gold)"
+                  />
+                  locked
+                </div>
+              )}
+            >
+              <div>DOGGO tokens are locked.</div>
+              <div>
+                Untill DOGGO is unlocked you will not be able to transfer or
```
0xDaedalus, line 51: Until*

ui/components/Wallet/AssetListItem/DoggoAssetListItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+                sell your tokens. But you can earn more by sharing referral link
```
0xDaedalus, line 52: `tokens, but you can earn more by sharing your referral link`

ui/components/Wallet/AssetListItem/DoggoAssetListItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+                  <SharedIcon
+                    icon="lock@2x.png"
+                    width={18}
+                    color="var(--trophy-gold)"
+                  />
+                  locked
+                </div>
+              )}
+            >
+              <div>DOGGO tokens are locked.</div>
```
0xDaedalus, line 49: I think if we're gonna give DOGGO its own component we probably should programmatically determine if it is locked or not.  It is considered unlocked when the [unpauser](https://github.com/tallycash/contracts/blob/main/contracts/governance/Doggo.sol#L31) is equal to address(0)

jagodarybacka: We are missing the task for the "unlocked" DOGGO state because it will require more changes in the UI. I think we can do it in a separate task

ui/components/Shared/SharedTooltip.tsx
```diff
@@ -5,27 +5,38 @@ type VerticalPosition = "top" | "bottom"
...
   verticalPosition?: VerticalPosition
   width: number
+  height?: number
   children: React.ReactNode
-  IconComponent?: () => ReactElement
+  IconComponent?: ({
+    isShowingTooltip,
+  }: {
+    isShowingTooltip: boolean
+  }) => ReactElement
```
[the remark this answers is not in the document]
jagodarybacka: I agree that it is not the best solution but we will have to refactor/improve tooltip at some point anyway so it can support different positions and will take care of overflowing the window. I will leave a comment so we won't forget about it.

I think the problem here (at least for me) is that we are using `styled-jsx` that has very little flexibility over styling component's children in comparison to for example `styled-components` (with styled components it would be so much easier to reuse styles and style children components depending on the context).

0xDaedalus (approved):
👑 🚢 👑

PR #1416, opened 2022-05-09
Update remote devtools to @redux-devtools packages

Shadowfiend:
These are newer and actually maintained. This also fixes an issue with
installing the older tool on macOS due to precompiled SQLite packages
not being available for the sqlite version being referenced in the
unmaintained package.

Closes #1401 .

PR #1521, opened 2022-05-25
😵 Let me be clear: Adding sqlite3 as an explicit devDependency

greg-nagy:
@redux-devtools depends on sqlite3 but when
installing as a dependency of that package it
wants to compile from source through node-gyp
which fails.

But if sqlite3 is declared as a standalone
devDependency then it installs without any issue.

### DIFF
--- a/background/package.json
+++ b/background/package.json
@@ -64,6 +64,7 @@
     "mockzilla": "^0.12.0",
     "mockzilla-webextension": "^0.13.0",
     "redux": "^4.1.1",
+    "sqlite3": "^5.0.8",
     "ts-node": "^10.4.0",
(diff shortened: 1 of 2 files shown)


Shadowfiend:
Hmmm… I did not have this issue after #1416, wonder why/if this fix will be more consistent 🤔

greg-nagy:
it's seems that it's just me :(

greg-nagy:
Note to self: Not sure why this was happening but sqlite3 install did not work on osx with node v16.10.0 (which is selected in .nvmrc)

To solve the problem I had to install node v16.15.0 and after a failing `yarn install` an `npm rebuild` fixed the `yarn start`

PR #1522, opened 2022-05-25
⚠️ Warning for when max fee is less than recent base

henryboldi:
Provides a warning on if the max fee might be too low depending on the current
base fee. To test, flip on custom gas and then enter a low max fee.

The styling and copy revisions will be made in #1551.

Low warning

🎸🤘


ui/components/NetworkFees/NetworkSettingsSelectOptionButtons.tsx
```diff
@@ -109,6 +110,10 @@ export function NetworkSettingsSelectOptionButtonCustom({
     customMaxPriorityFeePerGas: bigint
   ) => void
 }): ReactElement {
+  const [errorMessage, setErrorMessage] = useState("")
+  const baseGasFee = useBackgroundSelector(
+    (state) => state.networks.evm[1].baseFeePerGas
```
0xDaedalus, line 115: Lets make this multi-network

0xDaedalus (commented):
I think this change needs to be coupled with a more frequent gas polling interval.  Base fee can change by >100 gwei in under 2 minutes  (our gas polling time) when the market is moving which means we could definitely be showing this error message when we don't need to.

henryboldi:
> when the market is moving which means we could definitely be showing this error message when we don't need to.

Thoughts and making this a warning message instead? Something to the effect of,
"⚠️ This base fee may be too low"

greg-nagy:
I think reducing the gas fee interval should not have any significant side effects.

https://docs.blocknative.com/gas-platform
> Each endpoint is limited to provide updated results each second for paid tier apikeys, or every 5 seconds for free tier apikeys. Polling faster than these rates may deliver stale data.

But I would also try to lean into the BN estimations to prevent this problem altogether. 

My current understanding is that BN gives us an estimation of gas prices for the next 5 blocks but we are not using that information anywhere.  In the [makeBlockEstimate](https://github.com/tallycash/extension/blob/b46acbe068a818f1e3d3a09f126bed48716ac200/background/redux-slices/transaction-construction.ts#L122) we use only the first element with the confidence that we are looking for. It's practical because it's the price of the next block, but we also have the estimation already there for the next 5 blocks. 
[BN gas api example](https://docs.blocknative.com/gas-platform#example-response-payload) so we should have estimations for the next minute or so. 

My ideal solution would probably be using the estimations based on the time that has passed and also lower the polling freq to 30-60s

ui/components/NetworkFees/NetworkSettingsSelectOptionButtons.tsx
henryboldi: I believe multi-network is now addressed here ✅

ui/components/NetworkFees/NetworkSettingsSelectOptionButtons.tsx
```diff
@@ -109,6 +110,11 @@ export function NetworkSettingsSelectOptionButtonCustom({
     customMaxPriorityFeePerGas: bigint
   ) => void
 }): ReactElement {
+  const [warningMessage, setWarningMessage] = useState("")
+  const baseGasFee = useBackgroundSelector(
+    (state) =>
+      state.networks.evm[state.ui.selectedAccount.network.chainID].baseFeePerGas
```
0xDaedalus, line 116: Can we use the `selectedNetwork` selector in lieu of `state.ui.selectedAccount.network.chainID` please?  Bit of a nit but I think we're gonna move away from coupling the selectedAccount and the selectedNetwork sooner or later and using the selector consistently will make refactoring that a lot more straight-forward.

0xDaedalus (approved):
LGTM 🚀

PR #1523, opened 2022-05-25
⬇️ Update gas limit validation to use transaction minimum

### LINKED ISSUE #1368: Swap- Setting Gas Limit doesn't do anything
### **Build Info:**
Build #2881

### **Reproduction Steps:**

1. Perform a Swap action
2. Click "Transaction Settings"
3. Change Gas Limit
4. Get Final Quote
5. Confirm Swap
6. Sign

### **Expected Result**

We should show an error notification that gas limit is bellow minimum, very similar to how we show gas price is too low

### **Actual Result:**
The wallet acts like nothing happened. 

Video: https://www.awesomescreenshot.com/video/8607384?key=42958d40417323c90eb355955e27ed21

May be worked on in https://github.com/tallycash/extension/issues/1359

henryboldi:
As discussed, leaving this hardcoded to 21000 (consistent with MM) w/ a TODO
comment, because I don't believe it'll impact the networks we're adding in the
short/mid-term.

Gas min

Closes #1368


ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -277,9 +277,9 @@ export default function NetworkSettingsSelect({
               }
               try {
                 const parsed = BigInt(value)
-                if (parsed < 0n) {
+                if (parsed < 21000n) {
```
Shadowfiend, line 280: This is an Ethereum minimum, not a universal minimum, fwiw. As we merge here we should ask ourselves where we might model this fix in practice.

ui/components/NetworkFees/NetworkSettingsSelect.tsx
```diff
@@ -277,9 +277,9 @@ export default function NetworkSettingsSelect({
...
                   return {
-                    error: "Gas Limit must be greater than 0",
+                    error: "Gas limit too low",
```
Shadowfiend, line 282: “too low” isn't particularly informative. I'm guessing listing the full number runs into wrapping issues? Is there something we can do to be clearer on what “too low” means? cc @VladUXUI

Shadowfiend (commented):
Couple of nonblocking comments.

ui/components/NetworkFees/NetworkSettingsSelect.tsx
mhluongo: I have a feeling this will break Poly/Opti/Arbi..?

henryboldi: Yeah true. Perhaps in the networks slice we can include values like this there

PR #1541, opened 2022-05-29
Allow string chainId for EIP712TypedData

beemeeupnow:
There is a need to allow typedData.domain.chainId to be a string.

"Why? Doesn't EIP-712 indicate this should be uint256?" you might ask. :smile: 

Yes, but certain programming languages have limitations that keep them from reaching anywhere near the max value:

```
> JSON.parse('{"chainId":9007199254740993}').chainId
9007199254740992
```

This also benefits us on the pawdnerships side. (e.g. it allows us to work on CowSwap)

For this PR I only made minimal changes to allow the functionality on dapps that send such values.

Would using bignumber or some other solution here be better? (in the long term avoiding parseInt for risk of incompatible values)

### DIFF
--- a/background/services/signing/index.ts
+++ b/background/services/signing/index.ts
@@ -195,12 +195,17 @@ export default class SigningService extends BaseService<Events> {
   }): Promise<string> {
     try {
       let signedData: string
+      const chainId =
+        typeof typedData.domain.chainId === "string"
+          ? // eslint-disable-next-line radix
+            parseInt(typedData.domain.chainId)
+          : typedData.domain.chainId
       if (
         typedData.domain.chainId !== undefined &&
         // Let parseInt infer radix by prefix; chainID can be hex or decimal,
         // though it should generally be hex.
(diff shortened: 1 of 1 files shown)


jagodarybacka:
Hi @beemeeupnow, thank you for the contribution. 
This PR will resolve https://github.com/tallycash/extension/issues/1535. I've made the same change and then reverted here https://github.com/tallycash/extension/pull/1509#discussion_r882930330. @greg-nagy @Shadowfiend as this is now blocking issue can you take a look? This is a safe change to made I think but not sure if we want to handle it like that.

beemeeupnow:
Agree, this is something that we just have to handle because it is inevitable and required due to language constraints displayed in the description.

This is something a certain competitor wallet has also been through. (as a degradation where string was no longer allowed and they changed back to allow it again)

greg-nagy (dismissed):
This is good to go as is, so I am approving without merging, but hoping that the types will be fixed in this PR.

PR #1542, opened 2022-05-30
😵‍💫 Multinetwork dApp connections RFB

greg-nagy:
CleanShot 2022-05-30 at 11 10 24

# Why? 

Because we want to support multiple networks for dApp connections as well and they are hard.


rfb/rfb-3-multinetwork-dapp-connections
```diff
@@ -0,0 +1,99 @@
...
+
+## dApp Settings
+> How do we persist the network of a given dapp? (likely preference service - but maybe a new service?).  Also we’ll probably want > an in-memory store as well to avoid doing a bunch of i/o every time we get rpc requests.
+
+This should be broken down into 2 pieces: permission and current connection,
+
+We store these in redux since this will be needed in the UI as well and it will be changeable through the UI. And it will need to be synced between the UI and dapp connection continuously. 
+So we should refactor the `dapp-permission` slice to be more generic and store all the settings for dApps. Probably also rename it to be simply `dapp`.
+
+⚠️TBD: How to query the redux store from services without creating circular dependencies? We have 3 options
```
0xDaedalus, line 25: 4th option here is to make a new redux store specifically for dapp connections - could be an intermediate step between our current state and refactoring redux into its own service.

rfb/rfb-3-multinetwork-dapp-connections
```diff
@@ -0,0 +1,99 @@
+# RFB 3: Multinetwork dApp Connections
+
+# Mental Models 
+
+1) The extension is connected to every supported chain all the time — meaning it has a live connection simultaneously — but presents only a single connection to the dApps.
```
0xDaedalus, line 5: There's a voice in the back of my head that eventually we'll reach a number of websocket connections that users with slow internet speeds won't be able to support.  Not part of this RFB but I don't think I've ever voiced this concern.

rfb/rfb-3-multinetwork-dapp-connections
```diff
@@ -0,0 +1,99 @@
...
+### `eth_chainId`
+> Many dapps periodically ping eth_chainId - we’ll need to refactor our response there to send our persisted chainId for a given > dapp.
+
+We can handle that in the `InternalEthereumProviderService`, the same way we handle the current context lookup. 
+
+## `InternalEthereumProviderService` vs direct calls 
+> There’s gonna be complexity in how we handle parts of the code where we dogfood our internal ethereum provider versus parts where we skip that step and call our services directly via main
+
+- When we use internal dApps we should use our internal provider path. 
+- For extension specific calls it's ok to make the calls directly through main. See the direct calls we have now. None of these belongs to chain communication, but kind of like utility functionality
```
0xDaedalus, line 82: What about the following:
[`broadcastSignedTransaction`](https://github.com/tallycash/extension/blob/main/background/main.ts#L601) - direct chain communication
[`requestSignTypedData`](https://github.com/tallycash/extension/blob/main/background/main.ts#L623) - functionality seems to exist both in internal-ethereum-provider and main
[`requestSignData`](https://github.com/tallycash/extension/blob/main/background/main.ts#L647) - functionality seems to exist both in internal-ethereum-provider and main

rfb/rfb-3-multinetwork-dapp-connections
greg-nagy: Interesting thought, I love it!

rfb/rfb-3-multinetwork-dapp-connections
greg-nagy: 👻🫣 ...dis scary ... but might be good scary like ... maybe like Casper

rfb/rfb-3-multinetwork-dapp-connections
```diff
@@ -0,0 +1,99 @@
+# RFB 3: Multinetwork dApp Connections
```
jagodarybacka, line 1: Can we add `.md` file extension - github is not displaying it as a markdown and it is hard to read 😞

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+    accountAddress: string,
++   networkID: number
+  }
+```
+
+## Current Connection Context For dApp RPC Calls
+
+> We'll need get rid of activeChain and add network context to every rpc request (context that dapps will not be sending us). Things like eth_estimateGas, eth_getBalance, eth_blockNumber all need a way of being network aware
+
+The lookup of the current dapp context should happen in the `InternalEthereumProviderService` and provide it as a parameter to the chain service calls.
```
0xDaedalus, line 76: We also need to remember that apps will query this without necessarily being connected or permissioned - so we'll _probably_ want to default to the chain of the currently `selectedAccount` in those cases.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+
+> We might need an additional layer in the future just like with the `SigningService` but we should worry about that we want to support non EVM chains.
+
+### `eth_chainId`
+
+> Many dapps periodically ping eth_chainId - we’ll need to refactor our response there to send our persisted chainId for a given > dapp.
+
+We can handle that in the `InternalEthereumProviderService`, the same way we handle the current context lookup.
```
0xDaedalus, line 84: We also need to remember that apps will query this without necessarily being connected or permissioned - so we'll probably want to default to the chain of the currently selectedAccount in those cases.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+This changes when
+
+- the user makes the change through the UI
+- or the dApp uses the RPC methods eg. `wallet_switchEthereumChain`
+
+```
+{
+  permissionRequests: { [url: string]: PermissionRequest }
+  allowedPages: { [origin_accountAddress_networkID: string]: PermissionRequest }
+  activeConnections: { [origin: string]: { accountAddress: string, networkID: number } }
```
0xDaedalus, line 56: Can we have more than one active connection?  I think either the definition of an active connection or how we manage connections doesn't fit my current (probably incomplete) mental mode.

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: I am a bit torn here. The security-focused me says providing the info without permission is wrong.
User-centric me says "yeah we should have a sensible default to make it smooth."

A thing to note though: I think in the long run we need to get rid of the concept of `selectedAccount`. Probably not now, but it's helpful to keep this in the back of our heads. 

So the question is I think: How to provide a default value before permission is granted if we don't have the concept of `selectedAccount`?

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
+# RFB 3: Multinetwork dApp Connections
+
+# Mental Models
```
Shadowfiend, line 3: Format nit mode: there's only one top-level heading in a document.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+
+This changes when
+
+- the user makes the change through the UI
+- or the dApp uses the RPC methods eg. `wallet_switchEthereumChain`
+
+```
+{
+  permissionRequests: { [url: string]: PermissionRequest }
+  allowedPages: { [origin_accountAddress_networkID: string]: PermissionRequest }
```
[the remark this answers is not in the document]
greg-nagy: It's the origin of the dApps. I see it's not really named consistently. Probably `allowedOrigins` should be the key. What do you think?

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
+# RFB 3: Multinetwork dApp Connections
+
+# Mental Models
+
+1. The extension is connected to every supported chain all the time — meaning it has a live connection simultaneously — but presents only a single connection to the dApps.
+2. Redux store is a normal service for UI/app state/ settings. It can call other services “directly" — through async thunks —, it is persisted, has its migrations, and other services can also query it.
```
Shadowfiend, line 6: > other services can also query it

This should not be a given—it's certainly not the way it currently works. Main queries redux, but other services generally have a state that is propagated *from* redux but is never read from it. This is an intentional and careful separation that we should strive to maintain.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Proposal
+
+## dApp Settings
+
+> How do we persist the network of a given dapp? (likely preference service - but maybe a new service?). Also we’ll probably want > an in-memory store as well to avoid doing a bunch of i/o every time we get rpc requests.
+
+This should be broken down into 2 pieces: permission and current connection,
+
+We store these in redux since this will be needed in the UI as well and it will be changeable through the UI. And it will need to be synced between the UI and dapp connection continuously.
```
Shadowfiend, line 23: All of these things are true of everything in the UI, right? The model we've used is that data flows from the UI (submitting a change via async thunk) -> services (persistence) -> UI (updating redux)[^1]. Not sure if this sentence is suggesting a different approach.

[^1]: The canonical example here is permissions, since most of the other writes happen via our internal provider, which due to being a quick hack isn't factored quite the same way. We should probably look at how we can fix that.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+So we should refactor the `dapp-permission` slice to be more generic and store all the settings for dApps. Probably also rename it to be simply `dapp`.
```
Shadowfiend, line 24: What are “all the settings” in this context?

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+
+⚠️TBD: How to query the redux store from services without creating circular dependencies? We have 3 options
```
Shadowfiend, line 26: See my note above on how our data typically flows. Why do we need to query the redux store from services?

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: We should have 1 active connection for every dApp that has been granted permission. 

Eg uniswap is using `0xdeadbeef` on mainnet and curve is using `0x12341234` on polygon. No other dApp has been granted permission.

In this case we would have 
```
{
   ...
  activeConnections: { 
    "https://app.uniswap.org": { "accountAddress": "0xdeadbeef", "chainID": "1"},
    "https://curve.fi": { "accountAddress": "0x12341234", "chainID": "137"}     
  }
  ...
}
```

note: `networkID` > `chainID`

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Break out redux into it's own service.
+- Back and forth through events.
+- ~~Query the persisted store in local storage~~ // This is just for the record. We should not do this if we can avoid.
+
+### dApp Permissions
+
+In that redux slice the permissions are stored with url + network + address compound key.
+
+⚠️TBD: Do we separate permissions for accounts by network? So if a dApp has permission for 0xdeadbeef on mainnet then it will also have permission on polygon?
```
Shadowfiend, line 36: I'm not sure what “compound key” means in this context. Slices operate on JS objects, so we're not row oriented. In other network-aware areas we've been using a hierarchy that is `chainID` -> `address` -> object. Amongst other things, this is going to allow us to set up generic combining functions and fallback rules across consumers like the “polygon data falls back on mainnet”.

I'll write up either a prototype branch or RFB here this week.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+### Current Connection Per dApps
+
+[This issue](https://github.com/tallycash/extension/issues/1532#issuecomment-1139410588) belongs to this topic.
+
+This would be the other part of the redux slice: dApp URL <> active network, selected account.
+
+This changes when
+
+- the user makes the change through the UI
+- or the dApp uses the RPC methods eg. `wallet_switchEthereumChain`
```
Shadowfiend, line 50: I thought we were still debating whether we should try to keep tabs and the popover synced?

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+  export type PermissionRequest = {
+    key: string
+    origin: string
+    faviconUrl: string
+    title: string
+    state: "request" | "allow" | "deny"
+    accountAddress: string,
++   networkID: number
+  }
+```
```
Shadowfiend, line 70: Doc structure note: I would break these bits up so that the relevant fields are shown in the section that discusses them; e.g.:

<pre>
#### dApp Permissions

In that redux slice the permissions are stored with url + network + address
compound key:

```
  allowedPages: { [origin_accountAddress_networkID: string]: PermissionRequest }
```

:warning: TBD...

#### Current Connection per dApp

...

```
  activeConnections: { [origin: string]: { accountAddress: string, networkID: number } }
```
</pre>

Alternatively, an approach where the type snippet is shown after a brief explanation under “dApp Settings” with callouts and then explained below can also be a good way for readers to orient themselves; e.g.:

<pre>
```
{
  permissionRequests: { [url: string]: PermissionRequest }
  allowedPages: { [origin_accountAddress_networkID: string]: PermissionRequest } // See <1> below.
  activeConnections: { [origin: string]: { accountAddress: string, networkID: number } } // See <2> below.
}
```

1. In that redux slice the permissions are stored with url + network + address compound
   key. See “dApp Permissions” below.
2. The active connection is tracked per origin, based on information received
   from RPC and the extension. See “Current Connections per dApp”
   below.
#### dApp Permissions

...

#### Current Connections per dApp

...
</pre>

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+     - These are mostly the informational views
+   - Internal dApps: swaps, send, ...
+     - The common thing about them is that their functionality is tightly coupled with a network.
+     - These are mostly the parts that can make changes to the chain.
+
+# Proposal
+
+## dApp Settings
+
+> How do we persist the network of a given dapp? (likely preference service - but maybe a new service?). Also we’ll probably want > an in-memory store as well to avoid doing a bunch of i/o every time we get rpc requests.
```
Shadowfiend, line 19: > likely preference service

Though this isn't part of the current RFB, note that we already have a service that handles dApp context—the provider bridge service. It has a db for this kind of information, though it's limited to permissions atm. This RFB largely discusses expanding the permissions responsibility to dApp settings in general on the redux side, but expanding that  makes sense on the service side as well.

If we think we need to handle it at a more inner level (e.g. for internal dApps), the internal provider service can have a db just like the other services.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+  export type PermissionRequest = {
+    key: string
+    origin: string
+    faviconUrl: string
+    title: string
+    state: "request" | "allow" | "deny"
+    accountAddress: string,
++   networkID: number
```
Shadowfiend, line 68: We have a type for this! `AddressOnNetwork` is the right thing to use here. Same applies to `activeConnections` above.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,103 @@
...
+  }
+```
+
+## Current Connection Context For dApp RPC Calls
+
+> We'll need get rid of activeChain and add network context to every rpc request (context that dapps will not be sending us). Things like eth_estimateGas, eth_getBalance, eth_blockNumber all need a way of being network aware
```
Shadowfiend, line 74: Let's make an exhaustive list of what methods currently do and don't include `chainId`. For example, I believe `eth_estimateGas` does in fact include it, at least optionally (see [the ethers `TransactionRequest` type](https://github.com/ethers-io/ethers.js/blob/8b62aeff9cce44cbd16ff41f8fc01ebb101f8265/packages/abstract-provider/src.ts/index.ts#L28) and [the Ethers `hexlifyTransaction` function](https://github.com/ethers-io/ethers.js/blob/8b62aeff9cce44cbd16ff41f8fc01ebb101f8265/packages/providers/src.ts/json-rpc-provider.ts#L671), which is used [in gas estimation](https://github.com/ethers-io/ethers.js/blob/8b62aeff9cce44cbd16ff41f8fc01ebb101f8265/packages/providers/src.ts/json-rpc-provider.ts#L558-L560)).

Shadowfiend (commented):
First off, @greg-nagy thank you so much for putting together an RFB here—instrumental first step and it's helping me get some questions out that I've been struggling to structure properly elsewhere.

Left a bunch of comments: a couple on style/organization, most on mental model and approach.

Mostly my sense here is that we still don't have quite a deep enough understanding both of the problem and some of our UX decisions. We need to get solid on both of those before we can finalize our approach here.

Just from this doc so far, these all seem crucial and currently unclear:

- The question of whether permissions are network bound or not (the question here is: do we need to track different permissions by network, or only by address?).
- The question of whether the current network is synced between a dApp and the extension popover (the question here is: do internal dApps need the same model to handle this as external dApps?).
- The question of what calls current do or don't carry chain information (the questions here are: what is the delta between where the current RPC sits and where we would like it to in a perfect world where all calls carry chain ids? At what level do we need to track chain id?)

There's probably more lurking as well.

rfb/rfb-3-multinetwork-dapp-connections.md
0xDaedalus: I think that we don't have a choice here - there are plenty of dapps that will just display something akin to "Wrong Chain Please Connect to Avalanche" and be non-interactable if they get the wrong chainID back from the provider.

> So the question is I think: How to provide a default value before permission is granted if we don't have the concept of selectedAccount?

We will probably need to add `selectedNetwork` state if we get rid of `selectedAccount`

rfb/rfb-3-multinetwork-dapp-connections.md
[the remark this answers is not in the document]
0xDaedalus: This is though and I don't have a good answer.   Maybe just `allowed`?  The compound key makes this a bit sticky.

rfb/rfb-3-multinetwork-dapp-connections.md
0xDaedalus: I believe that we've decided that tabs will not sync with the popover so we can be connected to Ethereum in the extension while having a dapp that is connected to Polygon open - cc @VladUXUI to confirm.

rfb/rfb-3-multinetwork-dapp-connections.md
Shadowfiend: > A thing to note though: I think in the long run we need to get rid of the concept of selectedAccount. 

I feel like this has bubbled around but I've lost context. Are you expecting that concept will go away for dApps, or for the extension in general?

rfb/rfb-3-multinetwork-dapp-connections
Shadowfiend: I think the best way to frame this is that internal dApps should not have access to privileged actions except through a provider. The transaction flow does (it can authorize signing by the internal keychain and then push that info into the network), so it's not an internal dApp. In many ways, authorizing signing is the only privileged action in the system, though I haven't done a rigorous listing yet to make sure that's true.

In practice, I think a good heuristic is “if this is something I can do entirely through the internal provider, I should do it through the internal provider”. If we tried to implement any signing flow that way, we would need a separate path to sign without user confirmation—which should trigger spidey senses the multiverse over :)

rfb/rfb-3-multinetwork-dapp-connections.md
Shadowfiend: I've seen the preference for flat stores; I'm not actually convinced by the performance argument in the standard case, as you mention. Maybe experience will prove that wrong, but I expect we can derive more advantage from the finer-grained updates that nesting will get us (due to how the store is synced) than whatever performance disadvantage is typically a concern for edge cases.

rfb/rfb-3-multinetwork-dapp-connections.md
Shadowfiend: If that's the case, then it's fair to say that ”the user makes the change through the UI” isn't a use case that applies here, right? All changes will come through the provider.

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: The intention was to break this data flow and this constraint and make redux a full-fledged service, meaning other services should be able to call methods and read the info they need from it. 

`this.uiService.getDappPermissionsByOrigin(origin: string)`

This does not mean that services shouldn't have their own internal state or database, but it should not be necessary to duplicate the same data in 2 places and keep them in sync — which is always a pain.

To be more explicit: The thought was to keep dApp related information in the redux store and directly query it from the `ProviderBridgeService` or `InternalEthereumProviderService` when it's needed. 

We are kind of halfway there since we can call services directly from redux, I am arguing to go all in and also allow the read part.

How do you feel about this? @Shadowfiend

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: It does, and it depends on the discussion above. Let's make a decision there and I will update the rest accordingly.

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: * permissionRequests
* allowedPages
* currentConnections

The redux slice structure below.

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: This is actually more of a habbit that stuck with me from firebase times.

`chainID -> address -> object`  I am totally fine with this approach, it just wasn't my first choice.

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: > I believe that we've decided that tabs will not sync with the popover so we can be connected to Ethereum in the extension while having a dapp that is connected to Polygon open - cc @VladUXUI to confirm.

Started a flowdock thread to discuss this, but this definitely feels like a good catch. Will change this in a bit.

rfb/rfb-3-multinetwork-dapp-connections.md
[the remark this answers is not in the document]
greg-nagy: `allowedDApps`? I will think about it when updating this doc and if nothing better comes I will go with `allowed` probably.

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: >> So the question is I think: How to provide a default value before permission is granted if we don't have the concept of selectedAccount?

>> We will probably need to add selectedNetwork state if we get rid of selectedAccount

Hm, probably makes sense, as a changable default. This will probably be true for acounts as well. 

>> A thing to note though: I think in the long run we need to get rid of the concept of selectedAccount.

> I feel like this has bubbled around but I've lost context. Are you expecting that concept will go away for dApps, or for the extension in general?

Based on @0xDaedalus's thought above, my point might be invalid, but left the description of the original concept below. 

I think for the extension in general it will go away. 
More precisly most of our extension will be multinetwork by default 
and our internal dApps — swap, send etc. — will have a separate selector for account and network.
Just like what we expect to see in the external

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: ok, common understanding is clear: they are independent from each other

rfb/rfb-3-multinetwork-dapp-connections.md
Shadowfiend: > How do you feel about this?

I do not feel great about it :grin: It's special-casing dApp permissions in a way that we're not doing for anything else, and I haven't quite understood the win we get from doing that. We can make redux a full-fledged service in terms of structure and such without exposing internal redux data to external services, right?

rfb/rfb-3-multinetwork-dapp-connections
0xDaedalus: Would it?  A message bus wouldn't decrease our amount of network i/o - which is almost almost certainly where the performance problems would kick in first.

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: added as a TODO

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: #1560 This is the PR with the updates for the current architecture. 

Haven't played everything out properly yet, so it's in draft, but wanted to share as soon as it seemed to make sense

greg-nagy: Went ahead and merged #1560 since it feels that it's better to move ahead and don't expect any significant pushback 🤞

I also would like to close this RFB ASAP even if there are some things not laid out perfectly well. I feel that we are ok to figure out the rest in PRs

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -0,0 +1,176 @@
...
+  allowed: ...
+  activeConnections: { [origin: string]: AddressOnNetwork}
+}
+```
+
+#### InternalEthereumProviderService
+
+[This issue](https://github.com/tallycash/extension/issues/1532#issuecomment-1139410588) belongs to this topic.
+
+The current connections for the dApps will be stored in the `InternalEthereumProviderService` because the augmentation of current network will be necessary for our internal dApps as well.
```
0xDaedalus, line 128: I am a bit worried that we're tracking the same state in three places with this (in the eth provider, inside of indexdb, and in redux) but I think the best way forward is to try to implement and see where we get.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -12,92 +23,154 @@
...
+⚠️ [Pending product verification](https://www.flowdock.com/app/cardforcoin/tally-product-design/threads/-dXUTwSD3bXZ9enPRczVmEoDR1X) We are exploring whether our internal dApps should share a common selected/current address+network or they should be independent. In the meantime we will use the global account + network.
+
+##### Default network and account
+
+In this new paradigm we still need to be able to select an initial value to be used.
+
+The global current address and current network should be used as a default network and account.
 
-> We'll need get rid of activeChain and add network context to every rpc request (context that dapps will not be sending us). Things like eth_estimateGas, eth_getBalance, eth_blockNumber all need a way of being network aware
+⚠️ [Pending product verification](https://www.flowdock.com/app/cardforcoin/tally-product-design/threads/mFivf2mZ7YAhKm5OPQIfxoVVkoW) We are exploring the options on how to handle default network and account when the global selector is gone.
```
Shadowfiend, line 140: This sounds like it's forward-looking enough that it's not worth leaving in the RFB—this is what the “Future Work” section of the RFB is for.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -12,92 +23,154 @@
...
+When connecting to a dApp the chainID needs to be set on the window-provider. The default network should be used for this.
 
-#### `eth_chainId`
+When permission is granted the default address and chain should be used if given permission. If not, then the first the was granted.
 
-> Many dapps periodically ping eth_chainId - we’ll need to refactor our response there to send our persisted chainId for a given > dapp.
+❗️The user can change networks e.g. on uniswap before granting permission but there is no way for us to know what it is and the dApp follows what the wallet sets on window-provider. So we can use the default value as active connection when permission is granted.
 
-We can handle that in the `InternalEthereumProviderService`, the same way we handle the current context lookup.
+##### QnA
```
Shadowfiend, line 150: Structure note: an RFB is not a conversation, it's a distillation of the outcomes of conversations. Q&A isn't really a format we should employ; instead, we should make sure that any outstanding questions are addressed in the core of the RFB proposal.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -12,92 +23,154 @@
...
 
-### `InternalEthereumProviderService` vs direct calls
+- > We'll need get rid of activeChain and add network context to every rpc request (context that dapps will not be sending us). Things like eth_estimateGas, eth_getBalance, eth_blockNumber all need a way of being network aware
+  - The lookup of the current dapp context should happen in the `InternalEthereumProviderService` and provide it as a parameter to the chain service calls.
+- > Many dapps periodically ping eth_chainId - we’ll need to refactor our response there to send our persisted chainId for a given > dapp.
+  - We can handle that in the `InternalEthereumProviderService`, the same way we handle the current context lookup.
+- > There’s gonna be complexity in how we handle parts of the code where we dogfood our internal ethereum provider versus parts where we skip that step and call our services directly via main
+  - When we use internal dApps we should use our internal provider path.
+  - For extension specific calls it's ok to make the calls directly through main. See the direct calls we have now. None of these belongs to chain communication, but kind of like utility functionality
```
Shadowfiend, line 158: This is covered implicitly in the background section's categorization, I think: the multi-network skeleton should communicate through main, the internal dApps through the provider. If we find cases where this distinction isn't enough, we can discuss them, but my expectation is there will be very few, possibly even none, of these.

Regardless, this kind of clarification is out of the scope of the multi-network work; if we need it, we should put it somewhere in our primary docs.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -12,92 +23,154 @@
...
+    deriveLedgerAddress
+    connectLedger
+    getAccountEthBalanceUncached
+    resolveNameOnNetwork
+    connectPopupMonitor
+    onPopupDisconnected
+    ```
 
-> There’s gonna be complexity in how we handle parts of the code where we dogfood our internal ethereum provider versus parts where we skip that step and call our services directly via main
+## Work Breakdown
```
Shadowfiend, line 172: Structure note: this belongs in an issue rooted in the finalized RFB, rather than the RFB itself. The goal of the RFB here is to align everyone around the full approach, at which point work can be better broken down and estimated.

Separately, I would consider doing this behind the same feature flag that's gating all multi-network work, unless we're planning on releasing multi-network without dApp support.

Shadowfiend (commented):
This feels very close to ready for dev, but is still in relatively draft shape in RFB terms. Before we start heavy work here (but after we merge and possibly after we've broken the work out and assigned it), we should finish polishing it into something that is in true RFB format by:

- Following the common structure described in https://github.com/tallycash/extension/blob/main/rfb/rfb.adoc .
- Articulating the things we're deciding now, with things to be decided later left in a separate future work section.
- Moving references (issues, flowdock threads, etc) to a separate section.
- Reworking the overall format into a document that presents a single directed solution, rather than interspersing a proposal with responses to feedback about earlier versions of that proposal.

Anything that is prefixed with “Structure note” isn't blocking to merge here IMO. The other notes probably need a look, though don't necessarily have to be all the way resolved before merging this initial version.

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: I spent a lot of time thinking about this and come to the understanding that redux in this case is equivalent to view-model. So we are not tracking the same state in multiple places, but the redux follows the state of the service.

This separation might stay with us or we might decide that it's time to change it, but for now it's better and easier to follow the existing architecture and refactor when we know what we want exactly.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -12,92 +23,154 @@
...
 
-We store these in redux since this will be needed in the UI as well and it will be changeable through the UI. And it will need to be synced between the UI and dapp connection continuously.
-So we should refactor the `dapp-permission` slice to be more generic and store all the settings for dApps. Probably also rename it to be simply `dapp`.
+## TBD
 
-⚠️TBD: How to query the redux store from services without creating circular dependencies? We have 3 options
+- Refactoring the permission handling in the `ProviderBridgeService` to the nested structure should be out of scope for this RFB. The transformation of the data should happen in the main.
```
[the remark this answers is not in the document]
greg-nagy: We could store nested objects in indexedDB, though I would prefer not to. I thought you want that, that's why I wanted to make this conversation explicit. 


https://web.dev/indexeddb-best-practices/#keeping-your-app-performant
> While IndexedDB makes is possible to store large, nested objects as a single record (and doing so is admittedly quite convenient from a developer perspective), this practice should be avoided. The reason is because when IndexedDB stores an object, it must first create a structured clone of that object, and the structured cloning process happens on the main thread. The larger the object, the longer the blocking time will be.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -12,92 +23,154 @@
...
+#### Redux <> ProviderBridgeService communication flow
+
+No changes should be need.
+
+#### Batching permissions
+
+From the perspective of permissions multi-network or multi-account permission grant should
+be broken down, to multiple single permission grant or deny.
+
+Note: This code path will be used rarely and the amount of inserts won't be significant so we don't need to worry about indexedDB write speed here (for now).
```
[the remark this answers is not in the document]
greg-nagy: > I'm not completely clear on what this is saying. Is this referring to needing to unroll the data when it hits the provider service's IndexedDB, with one address-network-chainID triplet getting one row in the db?

Yes, updated.

> Let's be specific about the changes we'll need to make, because right now permissions don't include chain ID. If we add it, do we need to add an index? Do we need a migration? I think that kind of stuff is important to flag at this stage.

Added

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -3,7 +3,18 @@
 ## Mental Models
 
 1. The extension is connected to every supported chain all the time — meaning it has a live connection simultaneously — but presents only a single connection to the dApps.
-2. Redux store is a normal service for UI/app state/ settings. It can call other services “directly" — through async thunks —, it is persisted, has its migrations, and other services can also query it.
```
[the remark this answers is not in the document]
greg-nagy: Agree, but decided to keep all the concerns about this out of this RFB, because we will need a spike before we can say anything about this solution, and didn't want to block polygon with this.

We will need to find a way how to decouple the redux store and keep its ability to make method calls to other services. 

The current setup is good as it is for now.

rfb/rfb-3-multinetwork-dapp-connections.md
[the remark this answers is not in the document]
Shadowfiend: Sorry, what I'm saying is that I'm not seeing a need for a refactoring at all--indexeddb can remain one-entry-per-permission, and can be queried as such. I'm not clear on where we need to introduce nesting in the bridge service specifically?

rfb/rfb-3-multinetwork-dapp-connections
greg-nagy: Re: many rpc calls 
- [alchemy supports RPC batch calls](https://docs.alchemy.com/alchemy/documentation/batch-requests)
- combining this with a FIFO queue

Combining these two will open up really good possibilities for batching, tx cancelation, offline support etc. Fun times :)

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -12,92 +23,154 @@
...
-- or the dApp uses the RPC methods eg. `wallet_switchEthereumChain`
+In that redux slice the permissions are stored with in `chainID -> address -> object` nested object style.
 
 ```
 {
-  permissionRequests: { [url: string]: PermissionRequest }
-  allowedPages: { [origin_accountAddress_networkID: string]: PermissionRequest }
-  activeConnections: { [origin: string]: { accountAddress: string, networkID: number } }
+  permissionRequests: { [url: string]: PermissionRequest },
+  allowed: { [chainID: string | number]: {
```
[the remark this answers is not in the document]
Shadowfiend: For internal data, as we have here, I think we can and should canonicalize to one type so any code that deals with it can be confident in the data (`HexString` here). That PR as I understand it was about loosening the types we accepted from dapps, which are external--I think that's also reasonable, and in those cases we should then convert to our canonical types as soon as we receive that external data, again so all internal code can work with the same data shape. 

Does that seem reasonable?

rfb/rfb-3-multinetwork-dapp-connections.md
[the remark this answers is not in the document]
greg-nagy: We are on the same page :) Ty for clarifying.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -12,92 +23,154 @@
...
   export type PermissionRequest = {
-    key: string
+-    key: string
     origin: string
     faviconUrl: string
     title: string
     state: "request" | "allow" | "deny"
-    accountAddress: string,
-+   networkID: number
++    addressOnNetwork: AddressOnNetwork
```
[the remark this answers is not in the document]
greg-nagy: Removed it though, to make it clearer.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -23,91 +37,72 @@
...
 
-Note: This code path will be used rarely and the amount of inserts won't be significant so we don't need to worry about indexedDB write speed here (for now).
+When connecting to a dApp the chainID needs to be set on the window-provider. The default network should be used for this.
 
-### Current Connection Per dApps
+When permission is granted the default address and chain should be used if given permission. If not, then the first the was granted.
 
-- [ ] communication flow + events of account changing for a dApp
-- [ ] db structure for internal ethereum provider
+❗️The user can change networks e.g. on uniswap before granting permission but there is no way for us to know what it is and the dApp follows what the wallet sets on window-provider. So we can use the default value as active connection when permission is granted.
```
Shadowfiend, line 103: We haven't really defined where the defaults come from here, I think? It looks like that bit got deleted in the changes, but I think we still need to articulate “this is what the default will be”.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -121,56 +116,80 @@ This changes when the dApp uses the RPC methods eg. `wallet_switchEthereumChain`
...
-The global current address and current network should be used as a default network and account.
-
-⚠️ [Pending product verification](https://www.flowdock.com/app/cardforcoin/tally-product-design/threads/mFivf2mZ7YAhKm5OPQIfxoVVkoW) We are exploring the options on how to handle default network and account when the global selector is gone.
-
-##### Initial active connection
-
-When connecting to a dApp the chainID needs to be set on the window-provider. The default network should be used for this.
-
-When permission is granted the default address and chain should be used if given permission. If not, then the first the was granted.
+⚠️ Note: the selected account related solution is currently located in the `PreferenceService`. We need to move the logic to the `InternalEthereumProvider` and migrate the existing settings. We do this migration in the `main.ts` but we will need to clean up `PreferenceService` after this implementation is released.
```
Shadowfiend, line 125: I feel like there are two separate concepts here—one is the account/network the user has selected in the UI, and one is the account/network a dApp is using. The user's selection of account/network in the UI belongs in the preferences service—while the information about the account/network a dApp is using belongs in the internal provider.

If we're not decoupling the two concepts yet, I think the internal provider can query it from the preferences service. This sets us up to decouple properly in the future. Moving the user's choice in the UI to the internal provider is adding some unnecessary coupling though, I think.

Thoughts? Did I miss a moving part here? I know there's a lot of stuff in flux at the moment so it's very possible that's the case.

rfb/rfb-3-multinetwork-dapp-connections.md
```diff
@@ -121,56 +116,80 @@ This changes when the dApp uses the RPC methods eg. `wallet_switchEthereumChain`
...
+
+⚠️ Note: the [else here](https://github.com/tallycash/extension/blob/0c12499d711290a0de9f28898be44f87fe6d664f/background/main.ts#L1098) should be removed as part of this work.
+
+###### Initialization flow
```
Shadowfiend, line 129: Could use a quick summary of what this initialization flow is meant to do. e.g. migrate field X in object/db Y to field(s) A in object/db B or whatever. I'm having a little trouble following what the following bullets are trying to achieve.

(Pseudocode may also help here.)

Shadowfiend (approved):
I left a few notes—looking for a little bit of clarification around the reasoning for where current account/network info is being tracked, and of the initialization flow listed under the internal provider.

That said, I don't think that needs to block merge here—we can handle it in a separate PR. Will leave this green and let @0xDaedalus take merge honors if satisfied.

rfb/rfb-3-multinetwork-dapp-connections.md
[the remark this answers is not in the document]
Shadowfiend: Rereading my comment now, I think I understand exactly how it wasn't clear :grin:

rfb/rfb-3-multinetwork-dapp-connections.md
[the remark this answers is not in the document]
Shadowfiend: Ah, that makes sense! I thought it was a merge thing or something 🙈  I actually do like that idea, tbh, either that way or with a comment that explains the field has been added. No big deal though.

0xDaedalus (dismissed):
Beautiful 👑

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: This was a migration for the 1st time when we shipped this feature so we can set the value in the db to what the user has already been using. 

If the value has not been set in service/db it reads from redux and persists it. 

- reads `selectedAccount` from `PreferenceService` db
  -  db returns `undefined` if this property has not been set 
- fires `initializeSelectedAccount` with the value as payload
- if the payload is falsy — meaning this has never been set
  - read from redux the `ui.selectedAccount`
  - and persists it into `PreferenceService`
-

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: TL;DR: Short term I ok with anything that's quicker to implement. Long term: See thoughts below. 

Note: it's ok to change the long term vision, but I have written and thought about the RFB from this perspective.

I wanted to merge these concepts intentionally. 

So my mental model for the long term: 
- The extension consists of a multi-network skeleton + internal dApp(s) [^1]
- The internal dApp(s)[^1] are using the same data path as external dApps except for the permission layer
- We will remove the global account/network selector from most of our screens and add it to internal dApps (swap, earn) 
  - This will make the current account/network to be a property of our internal dApp(s) and not application setting.

So the idea here was to move the internal dApps into the same layer as external dApps — inside the `InternalEthereumProvider` and stop thinking about it as an application-wide preference. 

[^1]: We are still deciding if our internal dApps will function as a single dApp or multiple independent ones. Practically it might happen, that we want send to be on mainnet and when I click to swap it's on polygon. This is pending product decision.

0xDaedalus (approved):

rfb/rfb-3-multinetwork-dapp-connections.md
greg-nagy: > will we need a “current” network of any sort in the services

I don't think we will need it. That's why I would like to have this in `InternalEthereumProviderService` and not in `PreferenceService`

> In this scenario, wouldn't the internal dApp(s) have their current network tracked like any dApp, using the existing mapping of dApp origin to current network in the internal provider?

Yes, this is what I had in mind. 

It feels that we are sharing the same vision \o/

PR #1548, opened 2022-05-31
🪦 Remove `ESTIMATED_FEE_MULTIPLIERS_BY_TYPE`

henryboldi:
`ESTIMATED_FEE_MULTIPLIERS_BY_TYPE` was previously used, essentially
arbitrarily, with the idea that estimated network fees should change depending
on the max fee selected. This is only partly true. The part of the max fee that
impacts the estimated network fee is only the max priority fee, since it's
completely used in almost all cases. So we removed the multiplier from the
network fee display, but left the constant here. This PR is just housekeeping
to remove the constant and a PR with some context.

### DIFF
--- a/background/constants/network-fees.ts
+++ b/background/constants/network-fees.ts
@@ -5,15 +5,6 @@ export const ESTIMATED_FEE_MULTIPLIERS: { [confidence: number]: bigint } = {
   0: 20n,
 }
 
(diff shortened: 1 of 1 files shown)


greg-nagy (approved):
🧹

0xDaedalus (approved):
https://media.giphy.com/media/3o7aCWJavAgtBzLWrS/giphy-downsized-large.gif

PR #1550, opened 2022-05-31
🔁 Rejigger custom gas in store for conceptual alignment

henryboldi:
In the redux store, this moves custom gas outside the estimated prices
attribute. This is because the custom gas is set by the user, and doesn't need
to change its display based on the selected network.

In other words,
- Custom gas is not an estimate from BlockNative
- Custom gas doesn't need to be nested within a network id

To test, flip on custom gas and try editing the values.

🤘


0xDaedalus (approved):
🚢

PR #1558, opened 2022-06-01
📡 Fix populating past transaction activity list

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1439

### What
Past transactions are fetched in batches - we should fetch only 5 transactions at once even if we know more transaction hashes.

### Fix
Filtering transactions hashes list has been fixed because `filter` was not waiting for `async`, promisified arrow function.

### To test
1. Check given account's activity on Etherscan (to know how many transactions expect)
2. Add this account to the wallet (preferred fresh install)
3. Go to the activity list and wait 5-10min - depending on how old transactions are
4. Activity list should be slowly populated with transactions.

image


background/services/chain/index.ts
```diff
@@ -911,7 +911,7 @@ export default class ChainService extends BaseService<Events> {
 
     // Drop all transactions that weren't retrieved from the queue.
     this.transactionsToRetrieve = this.transactionsToRetrieve.filter(
```
0xDaedalus, line 913: Haven't had a chance to test this yet but initial thought here is we should probably also change the typing of the `handleQueuedTransactionAlarm` function to not be `async` and not return a `Promise`

jagodarybacka: I'm not sure, it seems like most methods in this service are async, even if they don't have to be, is it a rule for services we agreed on at some point?

background/services/chain/index.ts
```diff
@@ -911,7 +911,7 @@ export default class ChainService extends BaseService<Events> {
...
-      async ({ network, hash, firstSeen }) => {
+      ({ network, hash, firstSeen }) => {
```
Shadowfiend, line 914: Worms soundboard: “well that was pretty dumb”.

background/services/chain/index.ts
jagodarybacka: I feel like improving just this one function might not do much, maybe we should create an issue to clean this up?

0xDaedalus (approved):
🚢

PR #1559, opened 2022-06-01
(localize) distil strings from wallet Send and receive pages

gasolin:
related to #438

- Update readme
- distil strings from wallet Send and receive pages (includes the protocol swither)

截圖 2022-06-01 下午11 19 13

TODO
- will deal SharedAssetInput component when localize the swap UI
- Acitivities are not translated yet since some code condition depends on the label name
https://github.com/tallycash/extension/blob/main/ui/pages/SignTransaction.tsx#L82


ui/_locales/en/messages.json
```diff
@@ -31,10 +27,70 @@
...
+    "description": "in TopMenu > TopMenuProtocolList"
+  },
+  "ProtocolCompatibleChain": {
+    "message": "Ethereum-compatible blockchain",
+    "description": "in TopMenu > TopMenuProtocolList"
+  },
+  "ProtocolConnected": {
+    "message": "Connected",
+    "description": "in TopMenu > TopMenuProtocolListItem"
+  },
```
jagodarybacka, line 45: let's start with a lowercase letters so `"protocol___": { ...`

jagodarybacka (changes requested):
Looks good, just one minor thing:

ui/_locales/en/messages.json
gasolin: thanks, comment addressed

jagodarybacka (approved):
Works fine, let's 🚀

PR #1562, opened 2022-06-02
🔓 Don't ask for password with Ledger accounts

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1389

### What

Sometimes app was asking for keyring unlock even if the current account was Ledger based. As the app is not asking for password setup during Ledger account creation we shouldn't ask for password during the transaction
signing because sometimes there is no password set at all.

### Testing

One way:
1. Reinstall the extension
2. Add account with ledger
3. Try to sign transaction (in app or with dapp) - app shouldn't ask for a password

Other ways:
- unplug Ledger or wait 30min

We should test regular accounts too.

### DIFF
--- a/ui/pages/Popup.tsx
+++ b/ui/pages/Popup.tsx
@@ -15,7 +15,10 @@ import { isAllowedQueryParamPage } from "@tallyho/provider-bridge-shared"
 import { PERSIST_UI_LOCATION } from "@tallyho/tally-background/features"
 import { runtime } from "webextension-polyfill"
 import { popupMonitorPortName } from "@tallyho/tally-background/main"
-import { selectKeyringStatus } from "@tallyho/tally-background/redux-slices/selectors"
+import {
+  selectCurrentAccountSigningMethod,
+  selectKeyringStatus,
+} from "@tallyho/tally-background/redux-slices/selectors"
 import { selectIsTransactionPendingSignature } from "@tallyho/tally-background/redux-slices/selectors/transactionConstructionSelectors"
 import {
   useIsDappPopup,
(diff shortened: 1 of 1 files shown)


jagodarybacka:
@SJQuality could you test it? I'm not sure what other scenarios we should test that were broken 😄

ui/pages/Popup.tsx
```diff
@@ -44,7 +47,7 @@ const pagePreferences = Object.fromEntries(
 function transformLocation(
   inputLocation: Location,
   isTransactionPendingSignature: boolean,
-  keyringStatus: "locked" | "unlocked" | "uninitialized"
+  isLocked: boolean
```
Shadowfiend, line 50: My fourth edit after attempting to propose adjustments here: everything about this parameter gives me hives :laughing:

I'm wondering if we replace this parameter and `isTransactionPendingSignature` with a single `needsKeyringUnlock` boolean that is determined below as:

```typescript
    const needsKeyringUnlock =
      isTransactionPendingSignature &&
        signingMethod?.type === "keyring" &&
        keyringStatus !== "unlocked"
```

jagodarybacka: it is nice but I'm afraid we still need to pass `isTransactionPendingSignature` because of `const maybePage = params.get("page")` magic that is later overwritten by `sign-transaction` or `keyring/unlock` if `isTransactionPendingSignature = true`... I'm not really sure about all the cases in which `maybePage` will have some content, I'll investigate

Shadowfiend: I wouldn't dig too hard right now honestly... I think this whole approach probably needs reworking.

jagodarybacka:
@0xDaedalus I've just tried to make a swap and it didn't ask for a password 🤔 Was it a fresh install? Or changed branch from main?

0xDaedalus (approved):
Issues resolved - just needed to merge in `main` :+1:

SJQuality:
https://github.com/tallycash/extension/issues/1593

PR #1569, opened 2022-06-02
Fix Umbra Bridge Functionality

### LINKED ISSUE #1481: Unable to connect to and use Umbra Bridge.
Version: 0.13.4

Application: https://bridge.umbria.network/bridge/ethereum-polygon/matic

On page load Tally Ho is not able to connect to the application (this can be remedied by quickly setting `window.ethereum.isMetaMask = true` on refresh).  Once successfully connected however - attempting to bridge an asset from Ethereum to Polygon results in the following error in the console:

```
provider-bridge.js:250 
content: inpage > background: {"id":"4","target":"tally-provider-bridge","request":{"method":"eth_gasPrice","params":[]}}

provider-bridge.js:257  
content: background > inpage: {"id":"4","result":"0x6693eba8d"}

VM217:540 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'matic')
    at HTMLAnchorElement.<anonymous> (<anonymous>:540:84)
```

0xDaedalus:
Closes #1481 

This allows Tally Ho to work on https://bridge.umbria.network/bridge/ethereum-polygon/eth.

Test Bridge Transaction:
https://etherscan.io/tx/0xb42a061cb8c1074da094f08dfc59bfb5d8a2cb55088e93706687f1fc26f4b9cd
https://polygonscan.com/tx/0x44040ab6154ef7f6583233f1104c74eb40586a04259de7e414bb3be8ed9221ff

Simple but probably impactful fix.  Details are in the commit message.

### DIFF
--- a/ui/components/NetworkFees/FeeSettingsText.tsx
+++ b/ui/components/NetworkFees/FeeSettingsText.tsx
@@ -3,22 +3,27 @@ import {
   truncateDecimalAmount,
   weiToGwei,
 } from "@tallyho/tally-background/lib/utils"
+import { CUSTOM_GAS_SELECT } from "@tallyho/tally-background/features"
 import { NetworkFeeSettings } from "@tallyho/tally-background/redux-slices/transaction-construction"
 import {
   selectDefaultNetworkFeeSettings,
   selectEstimatedFeesPerGas,
   selectFeeType,
 } from "@tallyho/tally-background/redux-slices/selectors/transactionConstructionSelectors"
-import { selectMainCurrencyPricePoint } from "@tallyho/tally-background/redux-slices/selectors"
+import {
+  selectCurrentNetwork,
+  selectMainCurrencyPricePoint,
+} from "@tallyho/tally-background/redux-slices/selectors"
 import { enrichAssetAmountWithMainCurrencyValues } from "@tallyho/tally-background/redux-slices/utils/asset-utils"
 import { PricePoint } from "@tallyho/tally-background/assets"
 import { useBackgroundSelector } from "../../hooks"
 
 const getFeeDollarValue = (
   currencyPrice: PricePoint | undefined,
-  networkSettings: NetworkFeeSettings
+  networkSettings: NetworkFeeSettings,
+  baseFeePerGas?: bigint
 ): string | undefined => {
-  if (networkSettings.values?.baseFeePerGas) {
+  if (baseFeePerGas) {
     const gasLimit =
       networkSettings.gasLimit ?? networkSettings.suggestedGasLimit
 
@@ -29,12 +34,11 @@ const getFeeDollarValue = (
       enrichAssetAmountWithMainCurrencyValues(
         {
           asset,
-          amount: networkSettings.values?.baseFeePerGas * gasLimit,
+          amount: baseFeePerGas * gasLimit,
         },
         currencyPrice,
         2
       )
-
     return localizedMainCurrencyAmount
   }
   return undefined
@@ -45,10 +49,15 @@ export default function FeeSettingsText({
 }: {
   customNetworkSetting?: NetworkFeeSettings
 }): ReactElement {
+  const currentNetwork = useBackgroundSelector(selectCurrentNetwork)
   const estimatedFeesPerGas = useBackgroundSelector(selectEstimatedFeesPerGas)
   const selectedFeeType = useBackgroundSelector(selectFeeType)
   let networkSettings = useBackgroundSelector(selectDefaultNetworkFeeSettings)
   networkSettings = customNetworkSetting ?? networkSettings
+  const baseFeePerGas =
+    useBackgroundSelector((state) => {
+      return state.networks.evm[currentNetwork.chainID].baseFeePerGas
+    }) ?? networkSettings.values?.baseFeePerGas
 
   const mainCurrencyPricePoint = useBackgroundSelector(
     selectMainCurrencyPricePoint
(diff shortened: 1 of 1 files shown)


Shadowfiend:
I'm a little puzzled:
- The window provider reads defaultness from [the Tally config payload](https://github.com/tallycash/extension/blob/febc6b1fa460389e6979d8987fac4c53ee24db38/window-provider/index.ts#L74-L83).
- That payload reads the defaultness from [the preference service](https://github.com/tallycash/extension/blob/febc6b1fa460389e6979d8987fac4c53ee24db38/background/services/provider-bridge/index.ts#L126-L132).

So the default value in the preference service should in fact be reflected in the UI.

Secondly, MetaMask impersonation [should only be happening on OpenSea](https://github.com/tallycash/extension/blob/febc6b1fa460389e6979d8987fac4c53ee24db38/window-provider/index.ts#L19-L21), so this:

> but window.ethereum.isMetaMask is "false" 

Should be irrelevant to Umbra, right?

0xDaedalus:
> Secondly, MetaMask impersonation [should only be happening on OpenSea](https://github.com/tallycash/extension/blob/febc6b1fa460389e6979d8987fac4c53ee24db38/window-provider/index.ts#L19-L21), so this:

This is correct - i missed a file in the commit so you should now see that umbra is whitelisted. 

> I'm a little puzzled:

Everything you've said above is true - but it looks like there is no tally config payload sent initially on application load.  You'll see that when [preference service initializes](https://github.com/tallycash/extension/blob/main/background/services/preferences/index.ts#L70) it sends an `initializeDefaultWallet` event.  As far as I can tell that event does not interact with the window provider

background/services/preferences/defaults.ts
```diff
@@ -21,7 +21,7 @@ const defaultPreferences: Preferences = {
     ],
   },
   currency: USD,
-  defaultWallet: true,
+  defaultWallet: false,
```
greg-nagy, line 24: This change feels dangerous, but let's wait for product decision before we move forward with this PR 

[Product discussion thread](https://www.flowdock.com/app/cardforcoin/tally-product-design/threads/yg2ogQe4-fso7NQw6MZFancQaPi)

greg-nagy (changes requested):
I put this in "Request changes" state so it won't be accidentally merged before product decision. 

I feel that this is a very important piece and I am bit paranoid 🤷‍♂️

greg-nagy:
ok, this part is quite tricky , it's makes a lot of sense if it doesn't make sense :) 

It's absolutely possible that there is a bug that I am not aware of so I will just focus on explaining how the setup works and we can discuss after this if something is broken or not. 

So there are 2 data flows at play here and they are connected through the value of the `defaultWallet` property in `PreferenceService`
- window-provider (injected script into dApp)<>  provider bridge (content script) <> provider bridge service and the decision if we overwrite `window.ethereum` or not in the dApp
  - We create the `window.walletRouter` which [preserves any provider(s) that might have been occupying `window.ethereum` before we got here](https://github.com/tallycash/extension/blob/4207b3a9766fd24f51bf76c827d2006b6ba221c2/src/window-provider.ts#L33)
  - [We overwrite `window.ethereum`  with a trap object](https://github.com/tallycash/extension/blob/4207b3a9766fd24f51bf76c827d2006b6ba221c2/src/window-provider.ts#L88)  by default  in window-provider
  - Then during initialization of provider-bridge [we ask the `ProvderBridgeService` if we should be the default wallet](https://github.com/tallycash/extension/blob/1243a02b4038c35e89839c578edaa2e16dd6e0fb/provider-bridge/index.ts#L48)
  - The [`ProviderBridgeService` reads the `defaultWallet` value from the `PreferenceService`](https://github.com/tallycash/extension/blob/9b346745dd093d650a64197b2bd33673cb747004/background/services/provider-bridge/index.ts#L131) and sends back in the response 
  - In `provider-bridge` [we send the response from background to `window-provider`](https://github.com/tallycash/extension/blob/1243a02b4038c35e89839c578edaa2e16dd6e0fb/provider-bridge/index.ts#L38)
  - On response in `window-provider` [we check if we should be the default wallet and according to the response we set or reset the `window.ethereum` through the `window.walletRouter`](https://github.com/tallycash/extension/blob/b37e205c0e9e51841d193f9e2e7c3b70ef71b55b/window-provider/index.ts#L83)
  > In a previous implementation we asked the background for the value of `defaultWallet` first and then overwrote `window.ethereum` based on the actual user setting. But it resulted in a race condition because getting the value and setting `window.ethereum` happend after the dApps checked `widnow.ethereum` and as a result we weren't recognised by dApps.
- `PrereferencesService` <> main <> redux setup
  - On `PreferenceService` initialization 
    - [We set the `defaultWallet` value in indexedDB in the 2nd migration](https://github.com/tallycash/extension/blob/b94bdc9a5255bd35003c454cff8ac9b6bcfbe0ea/background/services/preferences/db.ts#L50) based on the [`defaultPreferences.defaultWallet`](https://github.com/tallycash/extension/blob/02e3a43bbd7fc7708008dc631656fd769b3865c8/background/services/preferences/defaults.ts#L23)
    - After db is initialized during service startup [we fire the `initializeDefaultWallet` based on `defaultWallet` in the db](https://github.com/tallycash/extension/blob/8f31861f0cca280a8b62bbe5adb0769712a15fe1/background/services/preferences/index.ts#L73)
    - In main [we listen to this event and fire `setDefaultWallet` action]
    - The `setDefaultWallet` action [overwrites the redux store value](https://github.com/tallycash/extension/blob/0c12499d711290a0de9f28898be44f87fe6d664f/background/redux-slices/ui.ts#L101)
    - When user changes the default wallet toggle [`setNewDefaultWalletValue` async thunk](https://github.com/tallycash/extension/blob/0c12499d711290a0de9f28898be44f87fe6d664f/background/redux-slices/ui.ts#L128) is fired, which emits `newDefaultWalletValue`
    - [in main.ts we listen to `newDefaultWalletValue`](https://github.com/tallycash/extension/blob/2e642bdeea6acc0971ef22ffea0b79c131547211/background/main.ts#L1129) and there we
      - [We update `PreferenceService`](https://github.com/tallycash/extension/blob/2e642bdeea6acc0971ef22ffea0b79c131547211/background/main.ts#L1131)
      - [And notify the connected `provider-bridges` (content scripts) through the `ProviderBridgeService`](https://github.com/tallycash/extension/blob/2e642bdeea6acc0971ef22ffea0b79c131547211/background/main.ts#L1135)
        - From here the event flow is the same as during initialization utilizing the `tally_getConfig` method.
        
        
Phew ... this a lot to unpack :) Let's see what happens next.

background/services/preferences/defaults.ts
0xDaedalus: So - I think regardless of the product decision - we revert that change for now - and merge this PR with just the Umbra whitelisting.

Thanks to your amazing comment - I see that I was just not grokking all the ways that metamask impersonations worked / was persisted - and now I'm confident that changing that variable is due to a misunderstanding of mine.

I think the thing that sent me down this rabbit hole was the mismatch between what I saw in the dapp:
image

and what I saw in the extension:
image

I guess this issue will get better as we whitelist more-and-more apps though.

0xDaedalus: To be clear - my intention with the `defaultWallet: false,` change was to address what I perceived was a mismatch between what we were presenting to the user (Tally Ho is the default wallet) and what I thought was happening (Tally Ho was not being set as the default wallet on initialization).  Since thats not the case - we can revert the change - and make another PR for whatever the outcome of the product discussion is.

greg-nagy (approved):
Let's 🚢

Shadowfiend:
A note on the commit history here:

Screen Shot 2022-06-03 at 11 13 12

With only this information in mind, what did this PR change?

PR #1571, opened 2022-06-03
💲 Add accounts to all EVM networks

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1501

### What
To show and track a new account properly we need to add it to each EVM network we support when the account is added by the user.

### Test

Fresh install
1. Reinstall extension
2. Add wallet with seed
3. Check if the first address is fine on Ethereum mainnet and Polygon (it is displayed on the accounts list)
4. Add some new addresses to the same wallet on Ethereum
5. Switch to Polygon - added addresses should be visible on the list (even without selecting them as active accounts)
6. Add some new addresses to the same wallet on Polygon
7. Switch to Ethereum  - they should be on the list too

Migration testing
1. Checkout `main`
2. Add accounts to Ethereum (2+ addresses), don't switch to Polygon
3. Checkout this branch
4. Switch to Polygon - all the addresses should be there as well

image

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -74,6 +74,7 @@ import {
   TransactionConstructionStatus,
   rejectTransactionSignature,
   transactionSigned,
+  clearCustomGas,
 } from "./redux-slices/transaction-construction"
 import { selectDefaultNetworkFeeSettings } from "./redux-slices/selectors/transactionConstructionSelectors"
 import { allAliases } from "./redux-slices/utils"
@@ -1010,6 +1011,7 @@ export default class Main extends BaseService<never> {
         "wallet_switchEthereumChain",
         [{ chainId: network.chainID }]
       )
+      this.store.dispatch(clearCustomGas())
     })
   }
 
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -896,7 +896,7 @@ export default class ChainService extends BaseService<Events> {
 
     // Drop all transactions that weren't retrieved from the queue.
     this.transactionsToRetrieve = this.transactionsToRetrieve.filter(
-      async ({ network, hash, firstSeen }) => {
+      ({ network, hash, firstSeen }) => {
         fetchedByNetwork[network.chainID] ??= 0
 
         if (
--- a/background/redux-slices/transaction-construction.ts
+++ b/background/redux-slices/transaction-construction.ts
@@ -6,7 +6,6 @@ import {
   INSTANT,
   MAX_FEE_MULTIPLIER,
   REGULAR,
-  CUSTOM,
 } from "../constants/network-fees"
 import { USE_MAINNET_FORK } from "../features"
 
@@ -72,15 +71,17 @@ export type EstimatedFeesPerGas = {
   custom?: BlockEstimate
 }
 
+const defaultCustomGas = {
+  maxFeePerGas: 0n,
+  maxPriorityFeePerGas: 0n,
+  confidence: 0,
+}
+
 export const initialState: TransactionConstruction = {
   status: TransactionConstructionStatus.Idle,
   feeTypeSelected: NetworkFeeTypeChosen.Regular,
   estimatedFeesPerGas: {},
-  customFeesPerGas: {
-    maxFeePerGas: 0n,
-    maxPriorityFeePerGas: 0n,
-    confidence: 0,
-  },
+  customFeesPerGas: defaultCustomGas,
   lastGasEstimatesRefreshed: Date.now(),
 }
 
@@ -182,11 +183,7 @@ const transactionSlice = createSlice({
           ]?.maxPriorityFeePerGas ?? transactionRequest.maxPriorityFeePerGas,
       },
       transactionLikelyFails,
-      customFeesPerGas: {
-        maxFeePerGas: 0n,
-        maxPriorityFeePerGas: 0n,
-        confidence: 0,
-      },
+      customFeesPerGas: defaultCustomGas,
     }),
     clearTransactionState: (
       state,
@@ -198,11 +195,7 @@ const transactionSlice = createSlice({
       feeTypeSelected: state.feeTypeSelected ?? NetworkFeeTypeChosen.Regular,
       broadcastOnSign: false,
       signedTransaction: undefined,
-      customFeesPerGas: {
-        maxFeePerGas: 0n,
-        maxPriorityFeePerGas: 0n,
-        confidence: 0,
-      },
+      customFeesPerGas: defaultCustomGas,
     }),
     setFeeType: (
       state,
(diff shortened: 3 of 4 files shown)


background/constants/networks.ts
```diff
@@ -74,6 +74,8 @@ export const FORK: EVMNetwork = {
 
 export const EVM_MAIN_NETWORKS = [ETHEREUM, ARBITRUM_ONE, OPTIMISM, POLYGON]
 
+export const EVM_MAIN_NETWORKS_SUPPORTED = [ETHEREUM, POLYGON]
```
0xDaedalus, line 77: I never knew how much I wanted this until this very moment :bow:

0xDaedalus (changes requested):
There's an edge case here we have to guard against - and I'm not sure of the best way to do it.

Consider the following steps

- checkout `main`
- import a wallet
- add 3 addresses to your wallet (do NOT switch to polygon)
image
- checkout your branch - refresh the extension
- switch to polygon (you'll only see one address)
- add another address
image

That address will be in polygon (and properly added to Ethereum as well) - but the addresses you added on Ethereum will not be visible on the Polygon chain.  Unfortunately I think this will happen to _everyone_ who has more than one address imported into their wallet for a given keyring when we make the update from single-chain to multi-chain.

What do you think the best way forward here is?  This definitely feels like some kind of migration or perhaps we could simplify it to `as part of keyring service init - ensure that all addresses are present on all supported chains`.  We should think about the best approach here because we'll likely encounter this problem each time we add a new chain.

background/constants/networks.ts
Shadowfiend: Any reason not to fix it so EVM_MAIN_NETWORKS only lists the ones we support? Feature flags can add or not like we do for name resolvers.

jagodarybacka:
@0xDaedalus Thank you, I haven't thought about this problem 😞 Will try to find a way to fix it for existing wallets

background/constants/networks.ts
jagodarybacka: Yeah, I'm not sure why do we even show Arbitrum and Optimism in the dropdown if we don't have any support for them 😄

Shadowfiend: Guessing the state of play may have changed since we had that conversation (in particular, the focus on getting Polygon out the door as a complete package before the other two).

Shadowfiend:
A migration for existing wallets is a good idea IMO. This is a one-time thing, and future behavior will work this way.

We've fixed adding an address here, but we'll need to fix deleting an address as well.

background/main.ts
```diff
@@ -791,23 +791,19 @@ export default class Main extends BaseService<never> {
...
+          })
+        )
 
-      // Mark as loading and wire things up.
-      this.store.dispatch(
-        loadAccount({
+        this.chainService.addAccountToTrack({
           address,
-          network: selectedNetwork,
+          network,
```
Shadowfiend, line 805: We probably need to do this for Ledger accounts as well, right?

One important thing to note as we look further into the future: EVM compatibility isn't enough to say an address/network pair should be added for a new address. Certain networks will be EVM-compatible but will derive addresses differently. Not an issue right now as I believe the 3 L2s we're looking at all use the same Ethereum addresses.

0xDaedalus: Whats an example of this type of network?  I'd love to start looking into this

background/constants/networks.ts
mhluongo: Yep! We maybe remove them and rename `USE_MULTINETWORK` or something?

I was working on Arbi and Option in parallel. Now that we know we're going to focus on Polygon first, the feature flags need a lil refactor

background/main.ts
mhluongo: RSK uses a different derivation path by default. If you're an RSK user you wouldn't expect the same address on Ethereum.

I think they should reconsider this, but there's a bunch of legacy wallets already 🤷‍♂️

Shadowfiend:
Perfectly down to do it separately, yeah. Just wanted to make sure that + Ledger didn't get missed before we un-feature-flag.

jagodarybacka:
This can be reviewed and tested.
Still to do (but can be done in the next PR):
- [ ] support Ledger with Polygon accounts
- [ ] fix the order of accounts on Polygon after the migration (they are the same but the order is different than on the Ethereum list)

jagodarybacka:
@0xDaedalus for Ledger I've added new issue, accounts order problem is right now only visible on this branch, either I will fix it tomorrow or we can merge it (if the rest will be working fine) and then create an issue (as this is minor problem anyway compared to not having accounts added to Polygon at all)

background/services/chain/index.ts
```diff
@@ -201,13 +190,9 @@ export default class ChainService extends BaseService<Events> {
...
     })
 
-    this.supportedNetworks = USE_MULTI_NETWORK
-      ? [ETHEREUM, ARBITRUM_ONE, OPTIMISM, POLYGON]
-      : [ETHEREUM]
-
     this.providers = {
       evm: Object.fromEntries(
-        this.supportedNetworks.map((network) => [
+        EVM_MAIN_NETWORKS.map((network) => [
```
Shadowfiend, line 195: Sorry for not noticing this earlier, but hard objection to moving from the chain service managing supported networks to having it be a global. The chain service is the right source of truth for this info, and all uses of it right now have access to the chain service. Strongly recommend that we keep the logic that sets the "main networks" in here as supported networks.

It won't break anything now, but by globalizing this responsibility we're inviting the leaking of decisions based on this info into strange places.

background/services/name/resolvers/uns.ts
```diff
@@ -82,7 +80,7 @@ export default function unsResolver(): NameResolver<"UNS"> {
     canAttemptAddressResolution({ name, network }: NameOnNetwork): boolean {
       return (
         isValidUNSDomainName(name) &&
-        supportedNetworks.some((supportedNetwork) =>
+        EVM_MAIN_NETWORKS.some((supportedNetwork) =>
```
Shadowfiend, line 83: I don't think this is correct behavior. UNS may not support a network even if we do.

Shadowfiend (commented):
Noticed something that makes me uncomfortable after merge, sorry. Lmk what you think here, I don't think it'll fundamentally change behavior on the user side right now.

background/services/chain/index.ts
0xDaedalus: We'll need to query `chainService.supportedNetworks` inside of `internal-ethereum-provider` to prevent a regression of the accidental uniswap polygon bug.

jagodarybacka: @Shadowfiend ok, reverted back to using `chainService.supportedNetworks` - but the value is still based on `EVM_MAIN_NETWORKS` which is changed by feature flags so right now the behaviour is the same. I guess you are mostly thinking ahead that we would like to make the logic of `supportedNetworks` more complex, right?

jagodarybacka: @0xDaedalus that file wasn't changed in this PR, can we make it a separate issue? Are you talking about [this line](https://github.com/tallycash/extension/blob/main/background/services/internal-ethereum-provider/index.ts#L219)?

PR #1572, opened 2022-06-03
Add Tokemak to list of sites to override Metamask

imthinhvu:
Adds Tokemak as one of the sites to override Metamask with


window-provider/index.ts
```diff
@@ -18,7 +18,7 @@ import { EventEmitter } from "events"
 
 // TODO: we don't want to impersonate MetaMask everywhere to not break existing integrations,
 //       so let's do this only on the websites that need this feature
-const impersonateMetamaskWhitelist = ["opensea.io", "bridge.umbria.network"]
+const impersonateMetamaskWhitelist = ["opensea.io", "bridge.umbria.network", "tokemak.xyz"]
```
jagodarybacka, line 21: please run `yarn lint-fix` to satisfy the linter 🤓

0xDaedalus:
I'm actually not sure that this is needed.  Tally Ho seems to connect to and work fine with tokemak without needing to override `window.ethereum.isMetaMask`.  Was there some specific functionality that was not working @imthinhvu ?

imthinhvu:
> I'm actually not sure that this is needed. Tally Ho seems to connect to and work fine with tokemak without needing to override `window.ethereum.isMetaMask`. Was there some specific functionality that was not working @imthinhvu ?

The connect doesn't work for me. When I try in Brave Browser with Shields down, I get the following error: 

Screen Shot 2022-06-07 at 12 40 39 PM

If I use Metamask instead, it works fine.

jagodarybacka:
It's not blocking for this PR and if we really have this problem on Brave maybe we should merge it but I'm wondering:

do we even support Brave right now? As it is chromium/V8 based browser most of the features should be working fine but maybe there are some minor problems and I'm pretty sure we are not testing extension on anything besides Chrome and Firefox.

greg-nagy (changes requested):
@imthinhvu Thank you very much for this PR. 

Tried on brave, but adding tokemak to the list did not make any difference.  In both cases clicking the `Metamask` button opens [this url](brave://wallet/crypto/onboarding).

Tested dApp connections on brave with uniswap and sushi and both worked well, so I assume the issue is with how tokemak is implemented dApp connection. 

Tried to search for the source for their UI, but couldn't find it.

Are you interested in digging deeper into the issue? We would be really happy to deliver a more robust impersonation feature.

imthinhvu:
Not at this time, but I can try to revisit this once I free up some time on my end. Can close this PR for now.

PR #1575, opened 2022-06-05
Fix Network Switcher styles

### LINKED ISSUE #1482: Cursor should change to `pointer` when hovering over network switcher.
The cursor should change to indicate that the TopMenuProtocolSwitcher is clickable.

image

### LINKED ISSUE #1487: The Multi-Network switcher should automatically close after selecting a new network.
After selecting a network - the `SharedSlideUpMenu` housing the network options should close.  `MULTI_NETWORK=true` needs to be set in the `.env` file when compiling the repo to access the menu.

A good place to start would be looking at `TopMenu.tsx`

image

ngrinkevich:
Closes #1482 and closes #1487


ui/components/TopMenu/TopMenu.tsx
```diff
@@ -101,7 +101,11 @@ export default function TopMenu(): ReactElement {
           setIsProtocolListOpen(false)
         }}
       >
-        <TopMenuProtocolList />
+        <TopMenuProtocolList
+          close={() => {
```
Shadowfiend, line 105: Out of curiosity, can we use the event bubble instead of having to dig a handler all the way through here?

ngrinkevich: ah just to put onClick on parent component? like this `<div className="standard_width_padded center_horizontal" onClick={() => {close()}}>` ?

Shadowfiend: Potentially. Not 100% sure if react messes with event propagation, or whether there are any cases where a click on a descendant shouldn't trigger this behavior.

ngrinkevich: ok just tried it works 🚀  but linter is now complaining -> we need to add these extra attributes to the parent: 
```
role="button"
tabIndex={0}
onKeyDown={() => {
    close()
}}
```

@henryboldi @greg-nagy do you have any objections? what are the "react way" / best practices these days?

jagodarybacka: we are already adding these attributes in some places so I think you can add them here too

Shadowfiend: I don't think we want to do this unfortunately:
- The container isn't a button—we're just attaching a common behavior to the clicking of child buttons.
- The container shouldn't get its own focus (the goal of the tab index)—only the buttons inside it should.
- The container probably shouldn't respond to key events (though this catches the fact that a child might be selected using a key press, which wouldn't be caught by the click handler).

As such, I actually take back my comment completely. I think passing an event handler through is the right thing, but we need to (a) follow our event naming style (`on*`) and (b) name the event for what it's doing (`onNetworkChange` or `onProtocolChange` or so).

ngrinkevich: yeah makes sense, i'll update 👍

greg-nagy (approved):
looks good, works well! 

let's 🚢

PR #1577, opened 2022-06-06
Flagging Flags: Drop a few unused feature flags

Shadowfiend:
> [**flagging** (adj)](https://en.wiktionary.org/wiki/flagging#Adjective)
> [1. Becoming tired or less dynamic; declining in strength; dwindling.](https://en.wiktionary.org/wiki/flagging#Adjective)

This PR kills 3 feature flags for features that have long since gone out.
It also adjusts an existing property that referenced feature flags but is
now a simple availability toggle.


jagodarybacka (approved):
Everything works fine ✨

PR #1580, opened 2022-06-06
Chain Specific dapp Permissions

### LINKED ISSUE #1581: Refactor `PermissionRequest` to be network-aware
See https://github.com/tallycash/extension/pull/1542 for details

0xDaedalus:
Closes #1581 

Refactors `dapp-permissions` redux store and the `provider-bridge` db to have network-aware `PermissionRequests`.

### To Test
-  checkout `main` 
-  import a wallet
- grant permission to uniswap
- checkout `chain-specific-permissions`
- update extension
- [ ] you should see a permission in uniswap with `chainID = 1` in the provider-bridge IndexDB.
- [ ] you should see a permission in uniswap with `chainID = 137` in the provider-bridge IndexDB.
- [ ] granting permissions to new dapps should work as before - and it should grant permission to all networks.
- [ ] revoking permission from a dapp should revoke permission on all networks (can be verified through indexDB).


background/redux-slices/dapp-permission.ts
```diff
@@ -25,7 +26,13 @@ export const emitter = new Emittery<Events>()
 export const grantPermission = createBackgroundAsyncThunk(
   "dapp-permission/permissionGrant",
   async (permission: PermissionRequest) => {
-    await emitter.emit("grantPermission", permission)
+    EVM_MAIN_NETWORKS.forEach((network) => {
```
jagodarybacka, line 29: In my PR there is no `EVM_MAIN_NETWORKS` as we wanted to store it in the chain service 😢  https://github.com/tallycash/extension/pull/1589

background/redux-slices/dapp-permission.ts
```diff
@@ -2,10 +2,11 @@ import { createSlice } from "@reduxjs/toolkit"
 import Emittery from "emittery"
 import { PermissionRequest } from "@tallyho/provider-bridge-shared"
 import { createBackgroundAsyncThunk } from "./utils"
+import { EVM_MAIN_NETWORKS } from "../constants"
 
 export type DAppPermissionState = {
   permissionRequests: { [url: string]: PermissionRequest }
-  allowedPages: { [origin_accountAddress: string]: PermissionRequest }
+  allowedPages: { [origin_accountAddress_chainId: string]: PermissionRequest }
```
greg-nagy, line 9: There are some parts that diverge from the RFB, and I think it's a good thing!

To have a clear list of what is the difference from the RFB can you please list them in https://github.com/tallycash/extension/issues/1604? 

No need to create issues for now, just a simple list about the diff

greg-nagy: actually went ahead, and listed what I see. Please verify and update the list as per your understanding

greg-nagy (changes requested):
I love this PR, and it's soooo good to see this coming to life 💖

I have found a couple of incosistencies that I felt should be handled in this PR to cover a unit of work.  Or would you like to merge the code as is, and ticket out the rest? 
I am fine with both, because at the end of the day it doesn't break functionality and all the updates belong to the polygon support. 

Please update `authorization.ts` with `chainID` handling, and also please validate that the `chainID` in the transaction payload has been granted permission for the dApp. 

Please update the migration to add permission for the new network to behave the same way when I grant permission after the network has been enabled

Please update the permission deletion logic to remove all granted permission for every chain on permission revoke.

greg-nagy (changes requested):
> I personally prefer smaller PRs to larger ones - but am happy to go with whichever path you think is best. I'll leave it up to you to merge or not and get started on the missing work here tomorrow morning regardless of which path we choose :)

I am down to work with this mindset, and frankly I do like it a lot. Ty.

The only thing that I think is important to do in this PR is updating the migration to create the needed new permissions.

Shadowfiend:
My understanding may not be correct here, but it sounds like we're refactoring and migrating the permission data in redux, but we're not at the final data shape that we expect to land on? If that's the case, I would err on the side of fewer migrations by getting the data shape correct now. Whether that's in this PR, or in a separate PR that this PR gets rebased on is up to y'all ofc.

If we had a good way to avoid the redux migration while the feature flag is off without blocking other dev in the meantime, I think the situation would be different if we could update the migration later to the correct data shape, but the way migrations currently work, I don't think that's going to be straightforward.

greg-nagy:
> it sounds like we're refactoring and migrating the permission data in redux, but we're not at the final data shape that we expect to land on? If that's the case, I would err on the side of fewer migrations by getting the data shape correct now. 

Yes and no :) 

Because the data is persisted in indexedDB and the migration happens there the redux store will be initialized with the correct data. This means that there is no need for redux migration in this particular case. 

I think this PR is a good step and we can do the redux restructure in a later PR without any difficulty.

greg-nagy (approved):
this is good, works well, let's 🚢

Shadowfiend:
> I think smaller PRs getting into master faster are safer and more productive than larger ones

Always agree with this fwiw. As we've done with most multichain stuff, we can decompose the problem here by first introducing the new structure and moving only Ethereum stuff to it and getting existing functionality working on the new data shape, then introducing the capabilities needed for additional networks.

> Because the data is persisted in indexedDB and the migration happens there the redux store will be initialized with the correct data.

Aha, looking closer it looks like we always replace the redux data on startup! Handy that this allows us to skip the migration :muscle:

greg-nagy:
> Always agree with this fwiw. As we've done with most multichain stuff, we can decompose the problem here by first introducing the new structure and moving only Ethereum stuff to it and getting existing functionality working on the new data shape, then introducing the capabilities needed for additional networks.

💯

Issue #1547, opened 2022-05-31
Polygon Tokenlist


0xDaedalus:
Ensure that top 200 Polygon tokens are visible in the wallet and swappable through our swap feature.

PR #1588, opened 2022-06-07
Clean polygon tokenlist response prior to validating.

0xDaedalus:
Unfortunately the polygon tokenlist does not adhere to the uniswap tokenlist schema - so we clean it up a little bit.

Partially addresses #1547 since it gets us the top 20 polygon tokens.  Will need to find another source to get to top 200.


background/lib/token-lists.ts
```diff
@@ -12,6 +12,18 @@ import {
   prioritizedAssetSimilarityKeys,
 } from "./asset-similarity"
 
+// We allow `any` here because we don't know what we'll get back from a 3rd party api.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const cleanTokenListResponse = (json: any, url: string) => {
```
greg-nagy, line 17: I don't like any here, but I don't like typing unknown either ... :(

greg-nagy (approved):
let's 🚢

background/lib/token-lists.ts
mhluongo: `unknown` is usually for unvalidated data though, right?

greg-nagy: My understanding is a bit different. `unknown` is for data that we have no idea what it is — which also implies it has not yet been validated. 

What unknown gives us is type safety over any, which is good. Otoh doing proper type validation for unknown objects is a pain and most of the time it does not worth it. 

The type guard @0xDaedalus has used is totally sufficient for this case. (Actually, it's not even necessary but I love to have it.)

0xDaedalus:
Here is an example of some of the errors:

```
0: {instancePath: '/tags', schemaPath: '#/definitions/TagIdentifier/maxLength', keyword: 'maxLength', params: {…}, message: 'must NOT have more than 10 characters', …}
1: {instancePath: '/tags', schemaPath: '#/properties/tags/propertyNames', keyword: 'propertyNames', params: {…}, message: 'property name must be valid'}
2: {instancePath: '/tags/customWithdrawEventSig/name', schemaPath: '#/definitions/TagDefinition/properties/name/maxLength', keyword: 'maxLength', params: {…}, message: 'must NOT have more than 20 characters'}
```

Shadowfiend:
Lame lol.

PR #1591, opened 2022-06-08
(localization) distill swap page strings

gasolin:
related to https://github.com/tallycash/extension/issues/438

distil swap page strings for translation


jagodarybacka (approved):
Looks good ✨

PR #1594, opened 2022-06-08
Require users to use strong(er) passwords

### LINKED ISSUE #1329: Password strength bar
Using a robust pattern-based tool like https://github.com/dropbox/zxcvbn. Combined with #1328, we'll have a strong mitigation for [cracking encrypted wallets](https://twitter.com/MetaMask/status/1515727239391809536).

0xDaedalus:
Closes #1329 

This PR requires users to have create a somewhat-strong password to use the extension.  

image

How strong of a password? 
`3` strong of a password!  
How strong is 3?
Well - the tool we're using has the following scale:

```
 // 0 - too guessable: risky password. (guesses < 10^3)
 // 1 - very guessable: protection from throttled online attacks. (guesses < 10^6)
 // 2 - somewhat guessable: protection from unthrottled online attacks. (guesses < 10^8)
 // 3 - safely unguessable: moderate protection from offline slow-hash scenario. (guesses < 10^10)
 // 4 - very unguessable: strong protection from offline slow-hash scenario. (guesses >= 10^10)
```

While we all want to be good security-conscious citizens of the web I do think that requiring users to have complex passwords of more than 16 characters (anecdotally what it seems to take to score 4) might cause a bit of friction.  @mhluongo @mr-michael what do you think?

### DIFF
--- a/ui/components/Keyring/KeyringSetPassword.tsx
+++ b/ui/components/Keyring/KeyringSetPassword.tsx
@@ -5,6 +5,7 @@ import {
   selectDefaultWallet,
 } from "@tallyho/tally-background/redux-slices/ui"
 import { useHistory } from "react-router-dom"
+import zxcvbn from "zxcvbn"
 import {
   useBackgroundDispatch,
   useAreKeyringsUnlocked,
@@ -35,12 +36,19 @@ export default function KeyringSetPassword(): ReactElement {
   }, [history, areKeyringsUnlocked])
 
   const validatePassword = (): boolean => {
-    if (password.length < 8) {
-      setPasswordErrorMessage("Must be at least 8 characters")
+    if (password.length < 12) {
+      // Less than 12 won't reliably give users a score of 3 or more.
+      setPasswordErrorMessage("Must be at least 12 characters")
       return false
     }
     if (password !== passwordConfirmation) {
       setPasswordErrorMessage("Passwords don’t match")
+      return false
+    }
+    const evaluation = zxcvbn(password)
+
+    if (evaluation.score < 3) {
+      setPasswordErrorMessage("Password too weak")
       return false
     }
     return true
--- a/ui/package.json
+++ b/ui/package.json
@@ -45,11 +45,13 @@
     "react-router-dom": "^5.3.0",
     "react-transition-group": "^4.4.2",
     "redux": "^4.1.0",
-    "webextension-polyfill": "^0.8.0"
+    "webextension-polyfill": "^0.8.0",
+    "zxcvbn": "^4.4.2"
   },
   "devDependencies": {
     "@types/react-dom": "^17.0.9",
     "@types/react-transition-group": "^4.4.4",
+    "@types/zxcvbn": "^4.4.1",
     "webext-redux": "^2.1.7"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3593,6 +3593,11 @@
   dependencies:
     "@types/yargs-parser" "*"
 
+"@types/zxcvbn@^4.4.1":
+  version "4.4.1"
+  resolved "https://registry.yarnpkg.com/@types/zxcvbn/-/zxcvbn-4.4.1.tgz#46e42cbdcee681b22181478feaf4af2bc4c1abd2"
+  integrity sha512-3NoqvZC2W5gAC5DZbTpCeJ251vGQmgcWIHQJGq2J240HY6ErQ9aWKkwfoKJlHLx+A83WPNTZ9+3cd2ILxbvr1w==
+
 "@typescript-eslint/eslint-plugin@^4.28.4":
   version "4.30.0"
   resolved "https://registry.yarnpkg.com/@typescript-eslint/eslint-plugin/-/eslint-plugin-4.30.0.tgz#4a0c1ae96b953f4e67435e20248d812bfa55e4fb"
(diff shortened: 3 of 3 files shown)


mhluongo:
Considering the key derivation function we're using is parallelizeable, `3` feels like the right balance..? Maybe we lower it to a `2` once we have Argon2, maybe we leave it where it is. `4` is too much for a consumer product

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -35,14 +36,21 @@ export default function KeyringSetPassword(): ReactElement {
   }, [history, areKeyringsUnlocked])
 
   const validatePassword = (): boolean => {
-    if (password.length < 8) {
-      setPasswordErrorMessage("Must be at least 8 characters")
+    if (password.length < 12) {
+      // Less than 12 won't reliably give users a score of 3 or more.
+      setPasswordErrorMessage("Must be at least 12 characters")
```
mhluongo, line 41: We should move to a password strength meter, though this might be okay for the start... @VladUXUI @mr-michael thoughts on whether a length error works in the interim?

Shadowfiend:
Ngl, getting "Password too weak" without any guidance on how to make it stronger would make me want to throw my laptop at the wall and sling Mjölnir after it lol. 

On the other hand I would probably use a password generator and not run into the problem in the first place so maybe the issue is imaginary 😅

jagodarybacka:
I agree with @Shadowfiend - we have to give user info about what we consider a strong password. Maybe a tooltip somewhere? cc @VladUXUI 

Second thing - it will drive me crazy during development where I'm reinstalling the extension very often to first copy the password and then to copy the seed. Can we consider allowing super easy passwords under the feature flag?

greg-nagy (approved):
I do like this lib, but does it really worth including 800 kb source to have this level of sophistication?

And I also wonder does it worth it for us to be this sophisticated? 

What were your thoughts @0xDaedalus? 
It's a non blocker, left an approve, but I would like to have a back and forth before merging.

0xDaedalus:
> I do like this lib, but does it really worth including 800 kb source to have this level of sophistication?

I thought of this as well - we're an extension not a website so I'm personally not as sensitive to bundle size (perhaps erroneously?)

> And I also wonder does it worth it for us to be this sophisticated?

I think the worth here comes from not only actual security (for example this helps mitigate users using passwords that are vulnerable to dictionary attacks) - but the perceived security as well.  I think there is a non-zero benefit to being able to clearly point to exactly why and where our wallet is more secure than others.
wq
> I agree with @Shadowfiend - we have to give user info about what we consider a strong password. Maybe a tooltip somewhere? cc @VladUXUI

We can do this but there are at least ~20 different reasons a password can fail right now (from playing around with it) - I intended this PR to be a quick win while we're pushing for polish and polygon so if we want to go that way I'd rather close it and then pick this up again once its properly been scoped.

Shadowfiend:
For context this is like 10% of our existing bundle size right?

I would consider making this a recommendation, not a requirement, which is how password strength is typically implemented. We should warn, but blocking specific passwords seems like a path to frustration, even more so if, again, we can't guide the user to something better. 

Lastly I'll softly suggest that making a change like this that throws up an explicit barrier to potentially completing our equivalent of "signup" should maybe *start* at checking in with product and design--even if the implementation itself is easy. Ultimately up to @VladUXUI/@mr-michael whether that is true though.

mhluongo:
Good point. I'd assumed the issue behind it (https://github.com/tallycash/extension/issues/1329) was already reviewed by product, or it wouldn't have made it into the shortlist for the sprint or in @mr-michael's polish ticket though...?

0xDaedalus:
> I would consider making this a recommendation, not a requirement, which is how password strength is typically implemented.

FWIW I think this _is_ generally a requirement and not a recommendation for financial applications (certainly for trading applications) which is a category that I think its fair to say that we fall into.

henryboldi:
> Ngl, getting "Password too weak" without any guidance on how to make it stronger would make me want to throw my laptop at the wall and sling Mjölnir after it lol.

Same 😬

> We can do this but there are at least ~20 different reasons a password can fail right now 

This makes it even more frustrating

mhluongo:
Let's solve the problem. What do you guys think of a password bar that directly corresponds to the levels we get back from the library? Something sort of like this

https://dribbble.com/shots/4069878-Password-Strength-Meter

... and in the meantime, maybe we set the strength lower?

henryboldi:
^ Big fan of this solution!

Shadowfiend:
Def comfortable with a password strength meter, though I think it's only a partial solve. 

My 2c on why I don't think this rises to the level of a trading account (incidentally these are usually implemented in terms of dumb but easily explained rules, at least the ones I've seen) in terms of requirements is that this isn't known to be a large account, and we support hardware wallets for higher security. 

We've briefly discussed in the past a suggested gradual ramp of security for users based on the funds they have... Keeping that in mind as a potentially interesting model, requiring somewhat-opaquely "high" security vs suggesting it seems wobbly to me. 

With that said, I'll quiet down on the issue and leave it to product and design and user input to guide things from here 😉

Shadowfiend:
For the individual user, paranoia is not the wrong strategy. In my view, however, what we're erring on the side of here is paternalism because we are removing user choice instead of guiding them to better ones with information.

It may still be the right trade off, but it makes me itch.

PR #1605, opened 2022-06-09
Remove Unneeded Polygon Defaults

0xDaedalus:
These defaults are cruft from a version of the extension that did not yet have a network switcher.

Now that the user can switch networks through the TopMenu there's no reason to have them.


greg-nagy (approved):
brave and smart ⚔️

PR #1606, opened 2022-06-09
🧑‍🤝‍🧑 My Best Friend - One active network per origin.

0xDaedalus:
This PR adds functionality to support 1 active chain per origin - allowing for a wonderful UX for people wanting to use mlutiple dapps on different chains at once.

Some key parts:
 - a given origin can only have 1 active chain (no chain-per-tab _yet_)
 - we have a special origin for our internal dapp
 
 Outstanding question - is this in-scope for Polygon launch?

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -116,6 +116,7 @@ import {
   migrateReduxState,
   REDUX_STATE_VERSION,
 } from "./redux-slices/migrations"
+import { TALLY_INTERNAL_ORIGIN } from "./services/internal-ethereum-provider/constants"
 
 // This sanitizer runs on store and action data before serializing for remote
 // redux devtools. The goal is to end up with an object that is directly
@@ -1009,7 +1010,8 @@ export default class Main extends BaseService<never> {
     uiSliceEmitter.on("newSelectedNetwork", (network) => {
       this.internalEthereumProviderService.routeSafeRPCRequest(
         "wallet_switchEthereumChain",
-        [{ chainId: network.chainID }]
+        [{ chainId: network.chainID }],
+        TALLY_INTERNAL_ORIGIN
       )
       this.store.dispatch(clearCustomGas())
     })
--- a/background/services/internal-ethereum-provider/db.ts
+++ b/background/services/internal-ethereum-provider/db.ts
@@ -1,29 +1,48 @@
 import Dexie from "dexie"
-import { ETHEREUM } from "../../constants"
+import { TALLY_INTERNAL_ORIGIN } from "./constants"
 
+type ActiveChainId = {
+  chainId: string
+  origin: string
+}
 export class InternalEtheremProviderDatabase extends Dexie {
-  private activeChainId!: Dexie.Table<{ chainId: string }, string>
+  private activeChainId!: Dexie.Table<ActiveChainId, string>
 
   constructor() {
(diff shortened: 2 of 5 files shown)


background/services/provider-bridge/index.ts
```diff
@@ -321,6 +322,7 @@ export default class ProviderBridgeService extends BaseService<Events> {
           return await this.routeSafeRequest(
             method,
             params,
+            enablingPermission.origin,
```
Shadowfiend, line 325: I would pass the request origin separately from the enabling permission. Though we currently don't implement this, there's no guarantee that the enabling permission's tracked origin is the same as the source request's origin, and semantically the latter is what we're trying to work with.

background/main.ts
```diff
@@ -1009,7 +1010,8 @@ export default class Main extends BaseService<never> {
     uiSliceEmitter.on("newSelectedNetwork", (network) => {
       this.internalEthereumProviderService.routeSafeRPCRequest(
         "wallet_switchEthereumChain",
-        [{ chainId: network.chainID }]
+        [{ chainId: network.chainID }],
+        TALLY_INTERNAL_ORIGIN
```
Shadowfiend, line 1014: Are we only using the origin for this call? If so, should we carry it as a param instead?

background/services/internal-ethereum-provider/db.ts
```diff
@@ -1,28 +1,48 @@
...
 export class InternalEtheremProviderDatabase extends Dexie {
-  private activeChainId!: Dexie.Table<{ chainId: string }, string>
+  private activeChainId!: Dexie.Table<ActiveChainId, string>
 
   constructor() {
     super("tally/internal-ethereum-provider")
 
     this.version(1).stores({
-      activeChainId: "&chainId",
+      activeChainId: "&origin,chainId",
```
Shadowfiend, line 15: Are you meaning to change the primary key here?

background/services/internal-ethereum-provider/db.ts
```diff
@@ -1,28 +1,48 @@
...
-  async getActiveChainId(): Promise<string> {
-    const [activeChainId] = await this.activeChainId.toArray()
-    return activeChainId.chainId
+  async getActiveChainIdForOrigin(origin: string): Promise<string> {
+    const activeChainId = await this.activeChainId.get({ origin })
+    if (!activeChainId) {
+      const defaultChainId = (await this.getInternalActiveChain()).chainId
+      return defaultChainId
+    }
+    return activeChainId?.chainId
```
Shadowfiend, line 45: What about implementing this in the service rather than the db? This would let us skip the circular-dependency avoidance with the single-value file, and it's not really clear to me that it's the db "layer" that should be doing this defaulting anyway.

Shadowfiend (commented):
Concept seems reasonable. Left a couple of notes. We also need to make sure to push to any connected RPC clients when the chain switches; in this case, that means making sure other tabs of the same dApp are aware that their underlying provider just got pushed to a different chain.

background/services/internal-ethereum-provider/db.ts
greg-nagy: If so it will require the same dance with migrations as in the provider bridge db

PR #1607, opened 2022-06-10
🧹 Add linting error for (and fix) unused imports

henryboldi:
The impetus here is me not removing a few unused imports lately 🙈. And the
title of the PR is slightly misleading because this is implemented by using the
`@typescript-eslint/no-unused-vars` rule. So **it covers all unused vars, not
just imports.**

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -92,6 +92,8 @@ module.exports = {
         ],
       },
     ],
+    "@typescript-eslint/no-unused-vars": "error",
+    "no-unused-vars": "off",
   },
   ignorePatterns: [
     "dist/",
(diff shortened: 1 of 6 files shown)


0xDaedalus:
image


One question - will the project still compile with unused imports / variables?  I'm hoping for the same functionality we have with console.log's where a project will compile if they exist but won't pass CI.

henryboldi:
> will the project still compile with unused imports / variables?

Yes! For example, https://github.com/tallycash/extension/runs/6832565803?check_suite_focus=true

jagodarybacka (approved):
🔥

PR #1608, opened 2022-06-10
Explicitly check chain id and if actually allowed when checking for permissions.

0xDaedalus:
### To Test

No great way to test the denial now (because we auth dapps for both polygon & eth  for now) but being able to sign a transaction (for example a quickswap swap) should be a good enough test that this works.

### DIFF
--- a/background/services/provider-bridge/authorization.ts
+++ b/background/services/provider-bridge/authorization.ts
@@ -5,13 +5,17 @@ import {
   PermissionRequest,
 } from "@tallyho/provider-bridge-shared"
 import { sameEVMAddress } from "../../lib/utils"
+import { toHexChainID } from "../../networks"
 import { HexString } from "../../types"
 
 export function checkPermissionSignTypedData(
   walletAddress: HexString,
   enablingPermission: PermissionRequest
 ): void {
-  if (!sameEVMAddress(walletAddress, enablingPermission.accountAddress)) {
+  if (
+    enablingPermission.state !== "allow" ||
+    !sameEVMAddress(walletAddress, enablingPermission.accountAddress)
+  ) {
     throw new EIP1193Error(EIP1193_ERROR_CODES.unauthorized)
   }
 }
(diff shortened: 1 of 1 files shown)


background/services/provider-bridge/authorization.ts
```diff
@@ -5,13 +5,17 @@ import {
...
+import { toHexChainID } from "../../networks"
 import { HexString } from "../../types"
 
 export function checkPermissionSignTypedData(
   walletAddress: HexString,
   enablingPermission: PermissionRequest
 ): void {
-  if (!sameEVMAddress(walletAddress, enablingPermission.accountAddress)) {
+  if (
+    enablingPermission.state !== "allow" ||
```
Shadowfiend, line 16: Should this be getting called if the enabling permission is not allowed?

0xDaedalus: In the current implementation it is not getting called if the enabling permission is not allowed (as a result of how we've structured provider-bridge service) - the idea here is that a functions prefixed with `checkPermission` should probably actually check if the permission is allowed.

0xDaedalus: > particularly for exported publicly-available functions.

Strong agree with that ^

>  Do we think these functions in their current state are useful?

I think in their _current_ state they are probably not very useful.   As far as I can tell we're using ~50 LOC to abstract away the equivalent of 3 `if` statements. I'm guessing we factored these out into an authorization.ts because we predicted they would be re-used elsewhere in the codebase - but so far it does not look like they have been.  I'm in favor of getting rid of these in a downstream PR if you are.  Alternatively we can wait a Providers Refactor RFC

greg-nagy (approved):
This is good as is for now. 

> I'm in favor of getting rid of these in a downstream PR if you are. 

I think this should be a good follow up

background/services/provider-bridge/authorization.ts
Shadowfiend: Either is fine with me.

PR #1609, opened 2022-06-10
Persist activeNetwork on disk

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/1532

This prevents a bug where if a user has polygon selected as the wallet's network and closes chrome.  Next time they open it the UI will show polygon as the active network but the extension will in fact be connected to Ethereum.

### To test
- open wallet
- select polygon as network
- quit chrome
- open chrome again
- [ ] verify through devtools that the activeNetwork is in fact polygon and not ethereum


jagodarybacka (changes requested):
It seems like the basic scenario with browser restart works fine. But I see that we are not supporting a case where selected and saved network is not available (Polygon flag has been turned off) this scenario was working previously and I think we should be able to turn off specific chain support and app should be working fine after that kind of update.

0xDaedalus:
> a case where selected and saved network is not available

I'm not sure I follow - can you outline a scenario or reproduction steps please?

jagodarybacka:
@0xDaedalus 
1. With `SUPPORT_POLYGON=true` change selected network to Polygon
2. Change `SUPPORT_POLYGON=false`
3. Reload background script
**Result:** extension is broken

Not sure how much we care about this scenario, do we even consider switching off any chain after we publish it's support.

greg-nagy (changes requested):
Please set up the db structure in a way that we expect in the RFB. There is no point in creating a lot of migrations when we know what we will need. 

OR

If we want to keep the PRs small, let's put things behind a feature flag, so the db is not created for our users before we release polygon support.

background/services/internal-ethereum-provider/db.ts
```diff
@@ -0,0 +1,31 @@
+import Dexie from "dexie"
+
+export class InternalEthereumProviderDatabase extends Dexie {
+  private activeChainId!: Dexie.Table<{ chainId: string }, string>
+
+  constructor() {
+    super("tally/internal-ethereum-provider")
+
+    this.version(1).stores({
+      activeChainId: "&chainId",
```
greg-nagy, line 10: This primary index will need to be changed. Would be good to do that without migrations.

background/services/internal-ethereum-provider/db.ts
```diff
@@ -0,0 +1,31 @@
...
+    })
+
+    this.activeChainId.put({
+      chainId: "1",
```
greg-nagy, line 14: The structure should be `[origin: string]: AddressOnNetwork`

background/services/internal-ethereum-provider/index.ts
```diff
@@ -275,18 +277,36 @@ export default class InternalEthereumProviderService extends BaseService<Events>
...
     return new Promise<SignedEVMTransaction>((resolve, reject) => {
-      this.emitter.emit("transactionSignatureRequest", {
-        payload: {
-          ...convertedRequest,
-          from,
-          network: this.activeNetwork,
-        },
-        resolver: resolve,
-        rejecter: reject,
+      this.getActiveNetwork().then((activeNetwork) => {
```
greg-nagy, line 280: Are we ok mixing await and then?

greg-nagy (commented):
left some more thoughts to the early feedback about this PR

background/services/internal-ethereum-provider/index.ts
0xDaedalus: I believe it is an antipattern for promise executors to be async functions

background/services/internal-ethereum-provider/db.ts
0xDaedalus: I don't believe thats a valid schema for an indexDB database.  Remember that the schema we're defining results in an array (not a map) so in order to derive the map that you've defined above the schema will need to be `{ origin, address, network }[]`

background/services/internal-ethereum-provider/db.ts
0xDaedalus: fixed by merging in 1606

0xDaedalus:
> Please set up the db structure in a way that we expect in the RFB. There is no point in creating a lot of migrations when we know what we will need.

To be honest with you I'm a bit hesitant to do this.  I think having just `origin: chainId` is a nice intermediate step between `origin: AddressOnNetwork` that will let us work out kinks with network-per-dapp without having to worry about the kinks of address-and-network-per-dapp.  Happy to press forward if you think its worth it though.

background/services/internal-ethereum-provider/index.ts
Shadowfiend: This is avoidable by just doing the active network lookup outside the promise executor with an `await` though, no? Limiting the executor strictly to the part that is executor-like seems like a good strategy.

background/services/internal-ethereum-provider/db.ts
greg-nagy: What I meant to say is that we also need to store the address, not just the `chainID`. 

I am not particularly concerned about the shape of the data, I think all the above options could work well.

Shadowfiend:
> If we want to keep the PRs small, let's put things behind a feature flag, so the db is not created for our users before we release polygon support.

This is also an option right? In general I think this is the _correct_ option regardless. We're adding functionality, it's related to a new feature, it doesn't really apply to the current state, and therefore it should be behind the relevant feature flag. If it feels like it's too complicated to if/then everywhere with the feature flag, it's not unreasonable to do a separate version of the service that is instantiated when the feature flag is on. The concern is usually that there may be drift, but I think the non-multi-network work on the internal provider is going to be minimal between now and when we flip the feature flag.

Regardless of the path forward though, I'll suggest we should probably use `Network` instead of a plain chain id. This will let us maintain fidelity in the types, stay aligned to other places we track network in IndexedDB, and make the transition to `AddressOnNetwork` just a matter of adding the `address` field, vs if we're tracking solely chain id we will have to move to a different structure (`network.chainId`).

0xDaedalus:
> Not sure how much we care about this scenario, do we even consider switching off any chain after we publish its support.

To me this feels like a dev-specific issue that doesn't cause that much friction (unless you think it will).  Ideally we'll have the SUPPORT_POLYGON flag switched off in a week or two so even the friction should be short lived.

Shadowfiend:
> Everything we build out for multi-network support should also work for single-network. I'd even argue that from a technical point of view what we're doing here is much more of a refactor than working on a new feature. 

It's true that there is a refactor involved in every area where we have to move from single to multi-network. However, the point of the feature flag is to reduce risk to existing, shipping functionality, so everything we do that isn't necessary to single-network unguarded by the flag reduces its value. It's a tradeoff evaluation against the cost of trying to make a change only when the feature flag is on. Given that all of the `activeNetwork` stuff is only relevant to multi-network, the tradeoff feels like it favors the feature flagged approach to me.

Anyway, not trying to throw up roadblocks here—mostly I want to move us towards a clearer mental model for feature flags. Multi-network *is* a feature. It *also* requires refactoring. This will be true of many if not most features. Splitting one from the other during ongoing dev is a bit of an art, no doubt, but it's good to keep the split in mind and try to minimize risk to production users overall. As a specific assessment,  if we create the db without the feature flag, and we then want to update the schema, the stakes for the migration are “the production extension breaks” instead of “I need to reset my local dev env”. Those are the stakes for most of our migrations right now, so in that sense it's not that different, but it's still a riskier play overall.

Explicitly asserting here that I am _not suggesting we block this PR on this note_ :) I will leave that up to y'all.

0xDaedalus:
Thank you for the detailed comment!  I agree with all of your points but I just think the weight of the constants in my `f(risk) = reward` algorithm are different.  c'est la vie!

I've already added a feature flag and reverted some of the activeNetwork refactor so I think the question of feature flagging vs not can be put to rest for this particular feature.  Will be mindful of feature flagging going forward.

Shadowfiend:
Definitely! And for what it's worth I'm currently actively trying to decouple my risk function from the team's risk function, since my attention is more split than it has been and there's no universe where I can be a blocker on risky decisions now 😁

greg-nagy (changes requested):
This PR is shaping up nicely and we are almost there! Bear with me — you know, because it's a bear market ... <slowly> he he he :) 

The last blocking problem: The db should not overwrite the user setting. 

Also noticed a strange bug (maybe?) where dApp connections default to ethereum regardless of the state of the internal network selector. 

Steps to reproduce: 
- I switch the extension to polygon 
- request a new permission with a dApp
- the dApp will be set to use mainnet, and not polygon by default

Saw this on cowswap and 1inch. We set the default network only on `wallet_switchEthereumChain` call, so the only reasonable explanation I can come up with is that the dApps are sending a request to change it back. But that doesn't really make any sense tbh.

0xDaedalus:
> Bear with me — you know, because it's a bear market ... slowly he he he :)

Too soon 🐻 🥲

background/services/internal-ethereum-provider/db.ts
```diff
@@ -0,0 +1,46 @@
...
+  private activeNetwork!: Dexie.Table<ActiveNetwork, string>
+
+  constructor() {
+    super("tally/internal-ethereum-provider")
+
+    this.version(1)
+      .stores({
+        activeNetwork: "&origin,chainId,network, address",
+      })
+      .upgrade((tx) => {
```
greg-nagy, line 21: non blocker: hm, I am surprised it worked on the first version, but I love this :) 

Another solution for the same problem
https://dexie.org/docs/Version/Version.upgrade()

greg-nagy (dismissed):
lookin' good, let's 🚀

A thought for the future, and not a blocker: As long as we are on trunk-based development we should consider feature flagging any new addition/changes by default. This will let us A/B test in the future, showcase new features to non-devs, and maybe do a gradual rollout. The reason I am writing all this is that to me in the "it's a refactor why feature flag" argument the purpose of the feature flags is greatly reduced and a lot of aspects are lost. 

This ofc does not mean, that we need to feature flag everything we do. But if it's about creating a new db, needing new migrations, or some functionality that is only needed for a specific feature that should be feature flagged. 
(With all that said the use of ff here is ingenious and deserves a h/t. Never used ff like this before)

This is especially true and important until we have proper testing and automated QA in place. 


From a pragmatic point of view: I think it's ok to not be too strict about it now, because of the near soft release date. 

Edit: Didn't realize [this has happened](https://github.com/tallycash/extension/blob/30d607a0c6e7005e9ea12b3db49cf76f8b83219e/background/services/internal-ethereum-provider/index.ts#L291) before I wrote the above. ❤️

greg-nagy (approved):
let's 🚀

PR #1610, opened 2022-06-10
🐛 Fix `maxFeePerGas` value set by `makeBlockEstimate`

henryboldi:


greg-nagy (approved):
love the readability refactor :) 

and also the fact that it's fixes the bug

PR #1616, opened 2022-06-13
🫥 Fix linter warnings

jagodarybacka:
Disable the linter `no-non-null-assertion` rule where it is needed.

### DIFF
--- a/src/window-provider.ts
+++ b/src/window-provider.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import {
   WindowListener,
   WindowRequestEvent,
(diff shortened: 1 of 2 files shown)


Shadowfiend:
All linter disables should come with an explanatory comment on why we are disabling. If you need some insight to fill in, we can ask the folks who originally introduced the behavior to see if they can share why they ignored the linter warning :)

In the case of this ignore:

https://github.com/tallycash/extension/blob/3dd12c5ebbcd7a2d41113d112fa0c92c8c4fecd8/window-provider/index.ts#L97

I would consider adding an if that checks for the non-nullness, and throw an exception with a clear error message if that invariant isn't true (e.g. “Expected window.tally to be set but it is not; Tally Ho provider configured incorrectly.”).

src/window-provider.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import {
```
jagodarybacka, line 2: @greg-nagy please let me know what comment should I add to the lines https://github.com/tallycash/extension/blob/3dd12c5ebbcd7a2d41113d112fa0c92c8c4fecd8/src/window-provider.ts#L91-L96 to explain non null assertion 😄

greg-nagy: We are defining the [window.walletRouter](https://github.com/tallycash/extension/blob/3dd12c5ebbcd7a2d41113d112fa0c92c8c4fecd8/src/window-provider.ts#L25) above in the same injected script, so it's safe to assume that it's there. I am just not sure how to inform typescript about this. 

It was important to have this as an [optional property on window](https://github.com/tallycash/extension/blob/d19915b29b629d762e9186085faf7d5e79a80bce/env.d.ts#L37) because it only exists in the dapp window context, not in other places.

jagodarybacka: Changed but the same check must be done inside the `Proxy.get`

Shadowfiend:
Stunning y'all! 🙌

PR #1617, opened 2022-06-13
🧚 Minor transaction and gas improvements

jagodarybacka:
### What has changed

- renamed `updateOptions` to `updateTransaction` as we are not only updating transaction's options (what options? network/gas options?) but this function is updating a whole set of transaction data
- removed duplicated gas settings redux selector - selector `selectDefaultNetworkFeeSettings` has been duplicated but only one function has been used. Removed unused one and moved the missing `baseFeePerGas` field to the other selector
- fixed showing gas price instead of `TBD` - If we know gas limit by suggested gas limit or selected by a user then we
should show the estimated gas value in the network settings selector button

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -69,7 +69,7 @@ import {
   estimatedFeesPerGas,
   emitter as transactionConstructionSliceEmitter,
   transactionRequest,
-  updateTransactionOptions,
+  updateTransactionData,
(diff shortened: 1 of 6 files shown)


henryboldi (approved):
I like the improved naming and clean up :)

👑👑👑

PR #1618, opened 2022-06-13
Grant permission to polygon apps even if polygon support is not enabled.

0xDaedalus:
Since we granted permission to polygon applications as part of [our latest provider-bridge db migration](https://github.com/tallycash/extension/pull/1580/commits/86639b0c1839d5fa735660456cc25f70d3ca9642) even if polygon support is not enabled - we should be consistent and grant permission to polygon applications when a user grants permission to a dapp even if polygon support is not enabled.


greg-nagy (approved):
ty 🙏

Issue #1604, opened 2022-06-09
multi-network dApp connections


greg-nagy:
- Feature pitch: https://coda.io/d/Minimum-Viable-Launch_dznBnP8Wtkc/Dapp-Connections-Update_su0P4#_luLeR
- Design
  - [dApp connection window](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=5242%3A123793)
  - [Inside extension](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=5242%3A123793)
- Guiding RFB: https://github.com/tallycash/extension/blob/main/rfb/rfb-3-multinetwork-dapp-connections.md

# dApp permissions
- [ ] refactor redux slice 
  - from flat structure to nested [see RFB](https://github.com/tallycash/extension/blob/284945f64ef704aba24d849dfd773ad8bf9e2d52/rfb/rfb-3-multinetwork-dapp-connections.md#L54)
  - rename slice to simply `dapp-permissions` > `dapp`
  - rename `allowedPages` > `allowed`
- [x] update authorization.ts with chainID handling
  - validate that the `chainID` in the transaction payload has been granted permission for the dApp.
- [x] update the migration to add permission for the new network to create permissions for polygon 
- [x] update permission deletion logic to remove all granted permission for every chain on permission revoke.

# Current Connections
...

greg-nagy:
- refactor redux slice 
  - from flat structure to nested [see RFB](https://github.com/tallycash/extension/blob/284945f64ef704aba24d849dfd773ad8bf9e2d52/rfb/rfb-3-multinetwork-dapp-connections.md#L54)
  - rename slice to simply `dapp-permissions` > `dapp`
  - ❌ blocked: handle multi network deletion

0xDaedalus:
confirmed - these are the two divergences (so far)

Shadowfiend:
Closing as the important bits are very very done.

PR #1619, opened 2022-06-13
Dapp Permission Redux Refactor

0xDaedalus:
Partially Addresses #1604 

This PR changes our dappPermission redux store to match the desired spec in the [Multi Network RFC](https://github.com/tallycash/extension/blob/main/rfb/rfb-3-multinetwork-dapp-connections.md#redux)

### To Test
- [ ] granting permission to dapps should work.
- [ ] revoking permission from dapps should work.
- [ ] the above should work both with SUPPORT_POLYGON=true and SUPPORT_POLYGON=false


jagodarybacka (commented):
hmm, uniswap is not working correctly but maybe that's my fault so can you check it? Sushi swap is fine, both connect and disconnect

jagodarybacka:
I've found a path to break it:
1. Connect and disconnect with dapp (sushiswap in my case)
2. reload the extension 
3. open extension -> blank screen

greg-nagy:
> hmm, uniswap is not working correctly but maybe that's my fault so can you check it? Sushi swap is fine, both connect and disconnect

it's not you, uni integration is broken for unknown reasons — will dig into it (gdi)

greg-nagy (changes requested):
I've noticed a few inconsistent behaviors: 
- In the redux slice we store permission only for the current network, but in indexedDB we store permission for every network 
- When revoking a permission only the the current network is removed from both redux slice and indexedDB
- When granting or revoking permissions, the `permissionRequests` list is not updated properly
- When reloading the extension the permissions from the indexedDB are overwriting the redux slice (this is the intended behavior, but important in this context)

I think some or all the above are related [to the bug](https://github.com/tallycash/extension/pull/1619#issuecomment-1158776446) @jagodarybacka has found

note: redux devtools usually does not work for me, but this script can be used from devtools > sources > snipetts (left hand side, narrow window tab)

```
console.clear()
    const state = JSON.parse(stateString)
    console.log("dapp.allowed.evm", JSON.stringify(state.dapp.allowed.evm, null, 2))
    console.log("dapp.permissionRequests", JSON.stringify(state.dapp.permissionRequests, null, 2))
})
```

background/redux-slices/selectors/dappSelectors.ts
```diff
@@ -26,15 +27,22 @@ export const selectCurrentPendingPermission = createSelector(
...
-          accountAddress.toLowerCase() === currentAccount.address.toLowerCase()
-      )
-    )
+  (allowed, currentAccount): PermissionRequest[] => {
+    // Return an array of all permissions corresponding to
+    // the currently selected account
+    const permissions: PermissionRequest[] = []
+
+    Object.keys(allowed.evm).forEach(([chainId]) => {
+      Object.keys(allowed.evm[chainId]).forEach((address) => {
```
greg-nagy, line 38: > I've found a path to break it:
> Connect and disconnect with dapp (sushiswap in my case)
> reload the extension
> open extension -> blank screen

CleanShot 2022-06-17 at 16 12 17

CleanShot 2022-06-17 at 16 13 51

I found these issues. Scratching my head why this is happening.

```
console.clear()

    function selector() {
        Object.keys(state.dapp.allowed.evm).forEach(([chainId]) => {
        console.log(chainId)    
        console.log(Object.keys(state.dapp.allowed.evm))
          Object.keys(state.dapp.allowed.evm[chainId]).forEach((address) => {
              console.log(address)
          })
        })
    }

    
    const state = JSON.parse(stateString)

    
    console.log("dapp.allowed.evm", JSON.stringify(state.dapp.allowed.evm, null, 2))
    console.log("dapp.permissionRequests", JSON.stringify(state.dapp.permissionRequests, null, 2))
    selector()
})
```

CleanShot 2022-06-17 at 16 26 07

```
['137'].forEach(([chainID]) => console.log(chainID))
VM348:1 1
```

background/redux-slices/selectors/dappSelectors.ts
```diff
@@ -26,15 +27,22 @@ export const selectCurrentPendingPermission = createSelector(
...
-        ([, { accountAddress }]) =>
-          accountAddress.toLowerCase() === currentAccount.address.toLowerCase()
-      )
-    )
+  (allowed, currentAccount): PermissionRequest[] => {
+    // Return an array of all permissions corresponding to
+    // the currently selected account
+    const permissions: PermissionRequest[] = []
+
+    Object.keys(allowed.evm).forEach(([chainId]) => {
```
greg-nagy, line 37: `Object.keys(allowed.evm).forEach(([chainId]) => {` > `Object.keys(allowed.evm).forEach(chainId => {`

greg-nagy (commented):
Found the problem \o/

background/redux-slices/selectors/dappSelectors.ts
0xDaedalus: oof. thanks - wonderful debugging

background/redux-slices/migrations/to-11.ts
```diff
@@ -0,0 +1,80 @@
...
+  origin: string
+  faviconUrl: string
+  chainID: string
+  title: string
+  state: "request" | "allow" | "deny"
+  accountAddress: string
+}
+
+type PermissionMap = {
+  evm: {
```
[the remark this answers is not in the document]
0xDaedalus: I think I want the extra typesafety of having an explicit `evm` key while we don't support others VMs.  Once we do we can move to a `[networkFamily: string]` type signature

background/redux-slices/selectors/dappSelectors.ts
Shadowfiend: > forEach(([chainId])

This should not have square brackets; we're deconstructing the first letter/number from the chain id instead of pulling the full chain id.

background/redux-slices/migrations/to-11.ts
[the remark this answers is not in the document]
Shadowfiend: Migrations should def be specific in terms of what we expect, and snapshot-in-time in terms of not trying to track the long term goals.

For future typing purposes though, `[networkFamily: NetworkFamily]` is the correct typing from Greg's note, as `NetworkFamily` is more restrictive than `string`.

background/services/provider-bridge/utils.ts
```diff
@@ -0,0 +1,25 @@
+import { PermissionRequest } from "@tallyho/provider-bridge-shared"
+
+export type PermissionMap = {
```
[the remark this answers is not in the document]
Shadowfiend: I would nuke this file completely and have these in /index tbh.

We should not reference the redux state to define the `PermissionMap` type; instead, we should reference the `PermissionMap` type in the redux slice.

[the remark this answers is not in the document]
0xDaedalus: Doesn't the above pattern break our rule of avoiding services depending on redux?

background/redux-slices/dapp.ts
```diff
@@ -0,0 +1,155 @@
...
+        grantPermission.fulfilled,
+        (
+          immerState,
+          { payload: permission }: { payload: PermissionRequest }
+        ) => {
+          const updatedPermissionRequests = { ...immerState.permissionRequests }
+          delete updatedPermissionRequests[permission.origin]
+
+          // Support Both networks regardless of which one initiated grant request
+          const permissions = [ETHEREUM, POLYGON].map((network) => ({
```
jagodarybacka, line 111: shouldn't we get suported networks from `chainService.supportedNetworks`?

background/redux-slices/dapp.ts
```diff
@@ -0,0 +1,155 @@
...
+            if (
+              immerState.allowed.evm[network.chainID]?.[
+                permission.accountAddress
+              ]
+            ) {
+              const { [permission.origin]: _, ...withoutOriginToRemove } =
+                immerState.allowed.evm[network.chainID][
+                  permission.accountAddress
+                ]
+              // chainID;
```
jagodarybacka, line 142:
```suggestion
```

background/redux-slices/dapp.ts
```diff
@@ -0,0 +1,155 @@
...
+import Emittery from "emittery"
+import { PermissionRequest } from "@tallyho/provider-bridge-shared"
+import { createBackgroundAsyncThunk } from "./utils"
+import { keyPermissionsByChainIdAddressOrigin } from "../services/provider-bridge/utils"
+import { ETHEREUM, POLYGON } from "../constants"
+import { SUPPORT_POLYGON } from "../features"
+
+export type DAppPermissionState = {
+  permissionRequests: { [origin: string]: PermissionRequest }
+  allowed: {
```
jagodarybacka, line 11: It confuses me to have `allowed` and `allowedPages` side by side, can you maybe leave a comment here what is allowed? Below we have the method `initializeAllowedPages` but we are actually setting `allowed` field. I don't get at the first sight what should I expect from each field 😞

jagodarybacka (commented):
Wow, so much work done 🥳

background/redux-slices/dapp.ts
0xDaedalus: We don't really ever want to pull services into redux - and since the all-or-nothing approach to granting & denying is temporary - i think this is probably an ok amount of tech debt to assume for the time being.

background/redux-slices/dapp.ts
0xDaedalus: 🙇

background/redux-slices/dapp.ts
0xDaedalus: Good catch and tyty - `allowedPages` was left over from the initial run through of the refactor when I was being extra careful about types.  Went ahead and renamed the method as well to be less confusing

background/redux-slices/dapp.ts
```diff
@@ -0,0 +1,152 @@
...
+    evm: {
+      // @TODO Re-key by origin-chainID-address
```
0xDaedalus, line 13: @greg-nagy @Shadowfiend What do you guys think about this deviation from the RFB? (in a downstream PR)

greg-nagy: I am good with it, but I will let @Shadowfiend have the final say here if he wants to.

0xDaedalus: To be clear this is not something that should block this PR from being merged

greg-nagy (approved):
let's 🚢

greg-nagy (approved):
🚢 is delayed, let's 🚀

PR #1621, opened 2022-06-14
📜 Remove padding to remove scroll from swap tab

### LINKED ISSUE #1611: Wallet and Swaps tabs should not have scroll
Scroll on Swaps should be fixed (page should not move up and down). We made need to decrease size of certain elements to make this possible.

Only the bottom section of Wallet should scroll. The top half of the page ("Assets... Activity" line and above) should stay static.

https://user-images.githubusercontent.com/7005061/173157644-c6a8cd3c-ab0a-49fe-9bdb-0f7376cfce0f.mov

henryboldi:
Closes #1611


greg-nagy (approved):
nice 💅

PR #1623, opened 2022-06-14
🛟 Display warning when there is not enough funds to pay for transaction

jagodarybacka:
### What is new
* calculate if a user will be able to pay for gas with base asset
* display a notification if there is not enough base asset to pay for gas

https://user-images.githubusercontent.com/20949277/173618928-d2c37e3d-c560-4bae-a958-7507f5770ed0.mov

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG.yml
+++ b/.github/ISSUE_TEMPLATE/BUG.yml
@@ -51,6 +51,7 @@ body:
       label: Version
       description: What version of the extension are you running?
       options:
+        - Community Edition v0.13.10
         - Community Edition v0.13.9
         - Community Edition v0.13.8
         - Community Edition v0.13.7
--- a/background/lib/utils/index.ts
+++ b/background/lib/utils/index.ts
@@ -122,7 +122,7 @@ export function getEthereumNetwork(): EVMNetwork {
   return ETHEREUM
 }
 
-export function isProbablyEVMAddress(str: string): str is HexString {
+export function isProbablyEVMAddress(str: string): boolean {
   if (normalizeHexAddress(str).startsWith("0x") && str.length === 42) {
     return true
   }
--- a/manifest/manifest.json
+++ b/manifest/manifest.json
@@ -1,6 +1,6 @@
 {
   "name": "Tally Ho",
-  "version": "0.13.9",
+  "version": "0.13.10",
   "description": "The community owned and operated Web3 wallet.",
   "homepage_url": "https://tally.cash",
   "author": "https://tally.cash",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tallyho/tally-extension",
   "private": true,
-  "version": "0.13.9",
+  "version": "0.13.10",
   "description": "Tally Ho, the community owned and operated Web3 wallet.",
   "main": "index.js",
   "repository": "git@github.com:thesis/tally-extension.git",
--- a/ui/_locales/en/messages.json
+++ b/ui/_locales/en/messages.json
@@ -152,7 +152,7 @@
     "description": "in Swap > SwapTransactionSettingsChooser"
   },
   "transactionSettingsDAOFee": {
-    "message": "Tally Ho fee for the DAO",
+    "message": "DAO fee",
     "description": "in Swap > SwapTransactionSettingsChooser"
   },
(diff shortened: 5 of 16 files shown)


background/main.ts
```diff
@@ -589,6 +596,26 @@ export default class Main extends BaseService<never> {
           annotation,
         }
 
+        const gasFee = BigInt(
```
jagodarybacka, line 599: @greg-nagy @Shadowfiend I'm wondering where can we move this code to not pollute this function too much - it is not really a selector, maybe to some utlis file? What do you think we can do to make it look clean?

background/utils/transaction.ts
```diff
@@ -0,0 +1,34 @@
...
+    transactionType === "asset-transfer" ||
+    transactionType === "asset-approval"
+  ) {
+    return annotation.assetAmount.asset.symbol === mainAssetSymbol
+      ? annotation.assetAmount.amount + gasFee
+      : gasFee
+  }
+
+  if (transactionType === "asset-swap") {
+    return annotation.fromAssetAmount.asset.symbol === mainAssetSymbol
```
jagodarybacka, line 28: Am I right that "asset-swap" is not possible from base asset (ETH/MATIC) - for base asset I'm always getting "contract-interaction" type and this type doesn't include any info about assets involved in the transaction 😞 Not sure where to get that info besides digging in the React components to pass that info to the background directly

background/main.ts
jagodarybacka: Ok I've moved it but let me know if you see a better place where it should belong

background/utils/transaction.ts
Shadowfiend: asset-swap isn't currently resolved by the enrichment service anywhere I believe. We want to be able to identify these but swaps are a bit gnarly to identify with certainty.

background/main.ts
Shadowfiend: This is a great hunch! The right way to handle this isn't actually in the UI portions at all, it's in the enrichment service. Transaction annotations already carry a [`warnings` field](https://github.com/tallycash/extension/blob/main/background/services/enrichment/types.ts#L33-L36), and that field can be used to verify and then carry this kind of info.

However, there's one more piece here! We don't and shouldn't try to detect whether the main asset is involved—any transaction whose max fee + value is > the account balance is a candidate for this warning, so all we need to do is compute ~`maxGasFee(transactionRequest) + transactionRequest.value` and compare it to the sender balance. The enrichment service has access to the chain service, so it can query latest balance directly.

background/main.ts
```diff
@@ -589,6 +596,26 @@ export default class Main extends BaseService<never> {
...
+          weiToGwei(
+            enrichedPopulatedRequest.gasLimit *
+              (enrichedPopulatedRequest.maxFeePerGas +
+                enrichedPopulatedRequest.maxPriorityFeePerGas)
+          ).split(".")[0]
```
Shadowfiend: This utility should almost never be used, and definitely not outside of redux/the UI.

`gasLimit`, `maxFeePerGas`, and `maxPriorityFeePerGas` should all be directly multipliable and yield a value denominated in wei (i.e., with 18 decimals). We shouldn't need to convert further between precisions; if we needed to, we should use the fixed-point utilities we have (e.g. `convertFixedPoint`).

Dealing with any asset values as strings in scenarios involving math is something we should be able to avoid at all times. Dealing with any asset values as non-`BigInt` should be similarly avoidable, even for intermediate states. There is a *small* set of scenarios where we'll deal with `BigNumber` due to using ethers, and that's about the only time we should be doing asset math without `BigInt`.

background/main.ts
jagodarybacka: @Shadowfiend 
> any transaction whose max fee + value is > the account balance is a candidate for this warning

But do I understand it correctly - gas will be always paid with base asset, so we can't just add it to the `transactionRequest.value` as this value can be paid with any asset. We have to check base asset balance to know if there is enough to pay for gas and also balance of the asset that is involved in the transaction (it can be any token or base asset).

Shadowfiend: `.value` is definitely denominated in the base asset in Ethereum. I would guess EVM-compatible L2s also mostly work that way, though that will require verification.

greg-nagy (approved):
works well, code looks good, ty 

let's 🚢

PR #1625, opened 2022-06-15
🖍 Debounce ENS/UNS name lookup

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1356

### What

If a user is typing UNS or ENS name we should allow them to finish typing before showing an error message - let's debounce by 500ms name validation to create better UX.

https://user-images.githubusercontent.com/20949277/173837618-045ed616-ee24-4293-be90-cce97fcd7101.mov

### DIFF
--- a/background/lib/utils/index.ts
+++ b/background/lib/utils/index.ts
@@ -122,7 +122,7 @@ export function getEthereumNetwork(): EVMNetwork {
   return ETHEREUM
 }
 
-export function isProbablyEVMAddress(str: string): str is HexString {
+export function isProbablyEVMAddress(str: string): boolean {
   if (normalizeHexAddress(str).startsWith("0x") && str.length === 42) {
     return true
   }
--- a/ui/components/Claim/ClaimReferral.tsx
+++ b/ui/components/Claim/ClaimReferral.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useState } from "react"
+import React, { ReactElement, useCallback, useState } from "react"
 import {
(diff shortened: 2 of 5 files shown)


henryboldi (changes requested):
Love this! However, I don't think it closes https://github.com/tallycash/extension/issues/1356 because

> I think we should hold off on domain resolution until an input value looks like a domain.

background/lib/utils/index.ts
```diff
@@ -122,7 +122,7 @@ export function getEthereumNetwork(): EVMNetwork {
   return ETHEREUM
 }
 
-export function isProbablyEVMAddress(str: string): str is HexString {
+export function isProbablyEVMAddress(str: string): boolean {
```
0xDaedalus, line 125: Why loosen our typesafety here?

Shadowfiend:
> I think we should hold off on domain resolution until an input value looks like a domain.

I can reply in the issue. This is a non-goal, the lookups will not necessarily just be for domains (eg address book entries).

henryboldi (approved):
> This is a non-goal, the lookups will not necessarily just be for domains (eg address book entries).

Then we're good to go here :)

background/lib/utils/index.ts
henryboldi: I thought that `boolean` is more correct here?

jagodarybacka: @0xDaedalus because this function is returning `boolean` right? I've had type errors on the string passed as an argument to the function,

PR #1626, opened 2022-06-15
🦑 Squid Game: Improvements to swap state handling

### LINKED ISSUE #1587: Swaps failing for unknown reason
### **Build Info:**
0.13.7 and 0.13.9

### **Reproduction Steps:**

1. Swap
2. add tokens
3. Get quote

I couldn't get this to reproduce. I asked where geographically the user was and they said "China". I tried to connect via VPN to Hong Kong (closet I could get) and I noticed that some of the tokens weren't loading on the "Wallet" page. I deleted/reinstalled Tally and did NOT use my VPN. All of the tokens loaded, but when I tried to Swap, I couldn't get a quote... Maybe something is cached from when the VPN was connected? 

### **Expected Result:**

Quote should be displayed

### **Actual Result:**

Spinner spinning
Discord: https://discord.com/channels/808358975287722045/888500174685614090/983687282768900106

Logs:
[logs_v0_13_7__2022-06-07T13_22_16-04_00.txt](https://github.com/tallycash/extension/files/8855305/logs_v0_13_7__2022-06-07T13_22_16-04_00.txt)

Video: https://www.awesomescreenshot.com/video/9325687?key=1be6bc595b7056043ef79e4ac7838fc4

greg-nagy:
> Let's win the swap game or die! - Greg

This PR:
- updates the disable state of the final quote button
- restricts quote updates to happen only when we have all the input data
- fix a weird behaviour that is happening because of `webext-redux`
- ~~start every swap with a clean slate~~ keep the persisted state functionality

# Test

Try to swap, and it should work as expected. 

~~When `Approve asset` happens, the swap page will NOT be repopulated with the previous settings.~~

Closes #1587

### DIFF
--- a/ui/pages/Swap.tsx
+++ b/ui/pages/Swap.tsx
@@ -129,21 +129,24 @@ export default function Swap(): ReactElement {
     assets: { sellAsset: locationAsset },
   }
 
-  const [sellAsset, setSellAsset] = useState<
-    FungibleAsset | SmartContractFungibleAsset | undefined
-  >(undefined)
-  const [buyAsset, setBuyAsset] = useState<
-    FungibleAsset | SmartContractFungibleAsset | undefined
-  >(undefined)
-  const [sellAmount, setSellAmount] = useState("")
-  const [buyAmount, setBuyAmount] = useState("")
+  const [sellAsset, setSellAsset] = useState(savedSellAsset)
+  const [buyAsset, setBuyAsset] = useState(savedBuyAsset)
+  const [sellAmount, setSellAmount] = useState(
+    typeof savedSwapAmount !== "undefined" && "sellAmount" in savedSwapAmount
+      ? savedSwapAmount.sellAmount
+      : ""
+  )
+  const [buyAmount, setBuyAmount] = useState(
+    typeof savedSwapAmount !== "undefined" && "buyAmount" in savedSwapAmount
+      ? savedSwapAmount.buyAmount
+      : ""
+  )
   const [needsApproval, setNeedsApproval] = useState(false)
   const [approvalTarget, setApprovalTarget] = useState<string | undefined>(
     undefined
   )
 
   useEffect(() => {
-    dispatch(clearSwapQuote())
     setSellAsset(undefined)
     setBuyAsset(undefined)
     setSellAmount("")
--- a/ui/pages/Swap.tsx
+++ b/ui/pages/Swap.tsx
@@ -408,20 +408,26 @@ export default function Swap(): ReactElement {
   const updateSellAsset = useCallback(
     (asset: SmartContractFungibleAsset | FungibleAsset) => {
       setSellAsset(asset)
-      // Updating the sell asset quotes the new sell asset against the existing
-      // buy amount.
-      updateSwapData("buy", buyAmount, asset)
+
+      if (buyAsset && buyAmount !== "") {
+        // Updating the sell asset quotes the new sell asset against the existing
+        // buy amount.
+        updateSwapData("buy", buyAmount, asset)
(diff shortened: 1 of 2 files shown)


ui/pages/Swap.tsx
```diff
@@ -129,29 +129,26 @@ export default function Swap(): ReactElement {
...
+  >(undefined)
+  const [sellAmount, setSellAmount] = useState("")
+  const [buyAmount, setBuyAmount] = useState("")
   const [needsApproval, setNeedsApproval] = useState(false)
   const [approvalTarget, setApprovalTarget] = useState<string | undefined>(
     undefined
   )
 
   useEffect(() => {
+    dispatch(clearSwapQuote())
```
Shadowfiend, line 146: Do we need the below changes if these are here?

IMO we should still feature flag this whole `useEffect`, or the inclusion of the chain id in its dependency array if we want the clearing behavior regardless.

ui/pages/Swap.tsx
```diff
@@ -129,29 +129,26 @@ export default function Swap(): ReactElement {
...
-      : ""
-  )
+  const [sellAsset, setSellAsset] = useState<
+    FungibleAsset | SmartContractFungibleAsset | undefined
+  >(undefined)
+  const [buyAsset, setBuyAsset] = useState<
+    FungibleAsset | SmartContractFungibleAsset | undefined
+  >(undefined)
+  const [sellAmount, setSellAmount] = useState("")
+  const [buyAmount, setBuyAmount] = useState("")
```
Shadowfiend, line 139: Does this also eliminate the preset of the asset when arriving from an asset page's swap button?

ui/pages/Swap.tsx
```diff
@@ -408,20 +405,26 @@ export default function Swap(): ReactElement {
...
-      updateSwapData("sell", sellAmount, asset)
+
+      if (sellAsset && sellAmount !== "") {
+        // Updating the buy asset quotes the new buy asset against the existing
+        // sell amount.
+        updateSwapData("sell", sellAmount, asset)
+      }
     },
-    [sellAmount, updateSwapData]
+    [sellAmount, sellAsset, updateSwapData]
```
Shadowfiend, line 427: This feels reasonable enough that I am wondering why it wasn't there lol.

Shadowfiend (commented):
Left a couple of notes around intended outcomes, but the code looks fine. Will let product/design chime in on whether the changes in behavior are okay tradeoffs.

0xDaedalus:
Tested locally and functionality works as intended but it is _very_ jarring to have your assets and amounts dissappear after approving

ui/pages/Swap.tsx
greg-nagy: it was a wrong direction, reverted the whole approach

ui/pages/Swap.tsx
greg-nagy: release crunch™️

0xDaedalus (approved):
Definitely fixes the bug although introduces some jitter which I think we can fix in a downstream PR - notice the flashing buy amount at ~0:07 and ~0:27

https://user-images.githubusercontent.com/94649004/173900397-7a6b138f-4537-43ad-b6d6-f195944a511b.mov

Shadowfiend:
Note that the jitter probably isn't straightforward to fix—it's because we refresh the quote when the page reopens and then when the approval completes, so that the quote is accurate. The 0:27s one is particularly important for that, since at times the approval won't change.

What's happening could definitely be clearer though, and I believe we have some designs lying around for a loading state other than “lol I disabled the inputs”, which is what I have currently implemented 😅

PR #1633, opened 2022-06-15
Fix hover area for Swap buttons

ngrinkevich:
Closes https://github.com/tallycash/extension/issues/1087
This one doesn't affect accessibility, more info [here](https://github.com/tallycash/extension/pull/1435/files#r872526312)

### DIFF
--- a/ui/components/Shared/SharedAssetInput.tsx
+++ b/ui/components/Shared/SharedAssetInput.tsx
@@ -328,6 +328,9 @@ export default function SharedAssetInput<T extends AnyAsset>(
 
   const [openAssetMenu, setOpenAssetMenu] = useState(false)
 
+  // TODO: use https://reactjs.org/docs/hooks-reference.html#useid once we update to version 18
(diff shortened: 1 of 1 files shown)


ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -514,6 +517,14 @@ export default function SharedAssetInput<T extends AnyAsset>(
             box-sizing: border-box;
             position: relative;
           }
+          // Using :global() to target child component
```
jagodarybacka, line 520: this is why I don't like `styled-jsx` 😜 I hope we will move to `styled-components` some day 🫣

jagodarybacka (approved):
Looks good to me, works fine 👑  
@henryboldi In my opinion accessibility is not perfect anyway as we don't highlight the dropdown even if it is selected (I'm testing moving selection with `Tab` key). But I will let you take a look before we merge it 😃

PR #1639, opened 2022-06-16
🔄 Fix styles of switch assets button

jagodarybacka:
### What

Fix "switch assets" button border.
Before:
image
After:
image


greg-nagy (approved):
took me a while :)

PR #1646, opened 2022-06-20
Countersignature:  Replace SigningMethod with single AccountSigner type

Shadowfiend:
First step towards implementing RFB 2.

These two were subtly different data structures that carried roughly the
same information. Rather than keep them split, unify them as one type,
the `AccountSigner`. The old `AccountSigner` type had an `accountID` for
all signers; the new structure looks more like `SigningMethod` in that each signing
type has its own properties to identify the signer uniquely for its
service.

`SigningMethod` subtypes for ledger and keyring were defined inline in
the same file; `AccountSigner` subtypes are instead defined in the
specific signer slices with the information needed to identify the way
to sign for an address with that signer, and then unioned into the top-
level `AccountSigner` type in the signing slice.

This change was implemented in two parts: redefining `SigningMethod`
as an alias for `AccountSigner` and leaving consuming code otherwise
unchanged, followed by removing `SigningMethod` entirely and renaming
any related fields and functions to refer to `AccountSigner`.

## Testing

Though the underlying code is mostly unchanged, these changes merit
checking:

- [x] Transaction signing with built-in and ledger accounts, from internal
      and external dApps.
- [x] Data signing with built-in and ledger accounts.
- [x] Typed data signing with built-in and ledger accounts.


background/redux-slices/ledger.ts
```diff
@@ -222,7 +223,7 @@ export const fetchAddress = createBackgroundAsyncThunk(
   ) => {
     try {
       dispatch(ledgerSlice.actions.setFetchingAddress({ deviceID, path }))
-      const address = await main.deriveLedgerAddress(path) // FIXME: deviceID is ignored
+      const address = await main.deriveLedgerAddress(deviceID, path) // FIXME: deviceID is ignored
```
jagodarybacka, line 226: Do we need this comment if we have one in the ledger service?

jagodarybacka (approved):
LGTM but I will give other folks a chance to review it ✨ I will make changes for https://github.com/tallycash/extension/issues/1630 on top of these changes to avoid conflicts

greg-nagy:
@SJQuality this refactor shouldn't break anything but has the potential to break a lot of important things. Can you please test this PR thoroughly?

background/redux-slices/ledger.ts
Shadowfiend: Yeah I was of two kinds on that and ended up deciding to leave it as a marker when the call gets made that might help someone wondering why the device id isn't doing anything.

SJQuality:
@greg-nagy looks good

jagodarybacka:
All good, lets 🚀

PR #1654, opened 2022-06-21
Properly initialize internal active network.

0xDaedalus:
@greg-nagy's [comment](https://github.com/tallycash/extension/pull/1609#discussion_r901790735) turned out to be prescient.  The current implementation causes inconsistent behavior when SUPPORT_POLYGON is set to true.

### To Test

Attempt to buy a polygon NFT (for example[ this one](https://opensea.io/assets/matic/0x66194b1abcbfbedd83841775404b245c8f9e4183/210624583337114373395836055367340864637790190801098222508621988249)).  You should be able to do so successfully.

### DIFF
--- a/background/services/internal-ethereum-provider/db.ts
+++ b/background/services/internal-ethereum-provider/db.ts
@@ -18,16 +18,11 @@ export class InternalEthereumProviderDatabase extends Dexie {
       activeNetwork: "&origin,chainId,network, address",
     })
 
-    this.initialize()
-  }
-
-  async initialize(): Promise<void> {
-    const existingActiveNetwork = await this.activeNetwork.get({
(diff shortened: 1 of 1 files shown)


background/services/internal-ethereum-provider/db.ts
```diff
@@ -14,17 +14,20 @@ export class InternalEthereumProviderDatabase extends Dexie {
...
-        })
-      })
+    this.version(1).stores({
+      activeNetwork: "&origin,chainId,network, address",
+    })
+
+    this.initialize()
+  }
+
+  async initialize(): Promise<void> {
```
greg-nagy, line 24: non-blocker: Subscribing to the populate event would be better here I think.

PR #1655, opened 2022-06-22
💢 Remove read-only and Ledger accounts

### LINKED ISSUE #1598: Can't add the same Ledger account the second time
To reproduce:
1. Add account with Ledger, import one address
2. After import is completed - remove the address from the wallet
3. Try to import the same account from the Ledger one more time

Result: account isn't imported, no error message visible
Expected: user is able to remove and import the same account multiple times

### LINKED ISSUE #1630: Ledger and Read Only Accounts removed in Tally don't stay removed after app refresh
### **Build Info:**

Main Build #3482 https://github.com/tallycash/extension/actions/runs/2504535746

### **Related Ticket/PR/Commit***
Commit:
https://github.com/tallycash/extension/commit/6b7e0013d2d7f30b86b147939e2b11cfdf773ca4
6b7e0013d2d7f30b86b147939e2b11cfdf773ca4

### **Reproduction Steps:**

v3 (reported by me)

1. Have Ledger and/or Read-Only wallet(s) connected to Tally
2. Remove Wallets
3. Close Tally
4. Refresh Tally on the Extensions page
5. Open Tally.

https://www.awesomescreenshot.com/video/8951615?key=bb5733a79aaf293a3b4dd5ec9507ac3e

Expected Result:
All removed accounts (wallets) should stay removed

Actual Result:
They come back

jagodarybacka:
Resolves #1630
Resolves #1598

### What

Handle removing read-only accounts and accounts added with Ledger.

Changes:
* pass correct signer type for an account that is supposed to be deleted
* implement removing ledger accounts from the database
* fix selecting new account after current one has been removed

### Testing

Add and remove all types of accounts (seed phrase, Ledger, read-only). Try different combinations and reload the extension from time to time. All removed accounts should stay removed.


0xDaedalus (commented):
It looks like read-only accounts are still not getting cleared the way we want them to:

https://user-images.githubusercontent.com/94649004/176025307-b858c497-a84a-45f5-9b60-4877e20521c5.mov

jagodarybacka:
@0xDaedalus Ready for review, I can't believe how stupid was that mistake 🫠

ui/components/AccountItem/AccountItemRemovalConfirm.tsx
```diff
@@ -92,14 +94,19 @@ export default function AccountItemRemovalConfirm({
...
+                })
+              )
               if (sameEVMAddress(selectedAddress, address)) {
                 const newAddress = Object.keys(
                   accountsData.evm[network.chainID]
                 ).find((accountAddress) => accountAddress !== address)
                 if (newAddress) {
                   dispatch(
-                    setSelectedAccount({
+                    setNewSelectedAccount({
```
greg-nagy, line 109: I know it's not in the scope of this PR but it would be good to somehow make the naming more obvious. 

The best would be to remove the `setSelectedAccount` from the export, but it's used in main.

greg-nagy (approved):
Code looks good, can't test now, so I won't be mergeing this.

ui/components/AccountItem/AccountItemRemovalConfirm.tsx
jagodarybacka: yeah, I'm not sure what will be the best name to avoid confusion in the future - maybe `setSelectedAccount 
 -> setSelectedAccountField` to inform that it won't change selected account properly?

PR #1662, opened 2022-06-23
⚙️ Saving slippage tolerance

jagodarybacka:
### What
Save selected slippage tolerance in the state so user don't have to change it every time during swap.


jagodarybacka:
@VladUXUI @mr-michael As it wasn't specified anywhere - do we want to save selected slippage tolerance? I feel like it is better UX than selecting custom value every time

henryboldi (approved):
Clean 🙌

PR #1663, opened 2022-06-23
💯 Fix "0.00 ETH" spend when sending a low amount

### LINKED ISSUE #1570: Sending low amount of an Asset results in an incorrect summary on the sign transaction screen
Attempting to send 0.001 ETH results in a confirmation screen asking the user to confirm sending 0 ETH:

image

henryboldi:
low_amount

Closes #1570


ui/components/SignTransaction/SignTransactionTransferInfoProvider.tsx
```diff
@@ -35,6 +36,14 @@ export default function SignTransactionTransferInfoProvider({
     enrichAssetAmountWithMainCurrencyValues(assetAmount, assetPricePoint, 2)
       .localizedMainCurrencyAmount ?? "-"
 
+  const readableDecimalAmount = fromFixedPoint(
+    assetAmount.amount,
+    assetAmount.asset.decimals,
+    20
+  )
+    .toFixed(20)
```
Shadowfiend, line 44: Can you use the fixed-point stringification function here or is there something wonky it does? This approach will lose precision.

ui/components/SignTransaction/SignTransactionTransferInfoProvider.tsx
```diff
@@ -31,10 +33,21 @@ export default function SignTransactionTransferInfoProvider({
     assetAmount.asset.symbol,
     mainCurrencySymbol
   )
+  // @TODO Remove this wonky massaging
```
greg-nagy, line 36: I think you need [isFungibleAssetAmount()](https://github.com/tallycash/extension/blob/a639aefcefb1bb52e242234e9e77867307aaa9e7/background/assets.ts#L199) and some error handling here. 

You can probably use this in the ternary below.

greg-nagy (changes requested):
The `wonky massaging` is really wonky. :) 

Let's use a type guard instead!

ui/components/SignTransaction/SignTransactionTransferInfoProvider.tsx
henryboldi: Updated, my co-author :)

0xDaedalus (approved):
👨‍🍳 💋

PR #1666, opened 2022-06-24
Fix sign data displayed message

### LINKED ISSUE #1590: Some weird looking message when signing a transaction on utopialabs.com icw Ledger. Transaction also fails
### **Build Info:**
0.13.17

### **Reproduction Steps:**

1. Sign transaction on utopialabs.com  icw Ledger

### **Expected Result:**

https://user-images.githubusercontent.com/103908695/172630926-dfd98a1b-2886-41c5-aee5-6fce58dfaaba.mov

### **Actual Result:**

https://user-images.githubusercontent.com/103908695/172630967-038f325d-7a57-4348-869d-19d3adc4f642.mov

jagodarybacka:
Resolves #1590

### What
Some dApps are sending sign data in plain text, not hex as expected. We should check if a sign message input is actually in hex format before decoding.

### Testing
Try to sign EIP191 message - for example, sign in to https://passport.gitcoin.co/#/

### DIFF
--- a/background/services/internal-ethereum-provider/index.ts
+++ b/background/services/internal-ethereum-provider/index.ts
@@ -229,7 +229,7 @@ export default class InternalEthereumProviderService extends BaseService<Events>
       case "eth_sign": // --- important wallet methods ---
         return this.signData(
           {
-            hexData: params[1] as string,
+            input: params[1] as string,
             account: params[0] as string,
           },
           origin
@@ -237,7 +237,7 @@ export default class InternalEthereumProviderService extends BaseService<Events>
       case "personal_sign":
         return this.signData(
           {
-            hexData: params[0] as string,
+            input: params[0] as string,
             account: params[1] as string,
           },
           origin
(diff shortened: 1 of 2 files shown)


background/services/internal-ethereum-provider/index.ts
```diff
@@ -355,15 +355,15 @@ export default class InternalEthereumProviderService extends BaseService<Events>
...
       account,
     }: {
-      hexData: string
+      input: string
       account: string
     },
     origin: string
   ) {
-    const asciiData = hexToAscii(hexData)
+    const asciiData = input.startsWith("0x") ? hexToAscii(input) : input
```
Shadowfiend, line 366: A small note for cleanup: we should replace `hexToAscii` everywhere with the Ethers [`toUtf8String`](https://docs.ethers.io/v5/api/utils/strings/#utils-toUtf8String). All else aside, `hexToAscii` is misnamed since it handles all UTF-8 codepoints rather than just ASCII :grimacing:

PR #1667, opened 2022-06-24
Initialize ledger accounts for all supported networks.

0xDaedalus:


background/main.ts
```diff
@@ -494,15 +494,17 @@ export default class Main extends BaseService<never> {
...
-        await this.chainService.addAccountToTrack(addressNetwork)
-        this.store.dispatch(loadAccount(addressNetwork))
-        this.store.dispatch(setNewSelectedAccount(addressNetwork))
+        await Promise.all(
+          this.chainService.supportedNetworks.map(async (network) => {
+            const addressNetwork = {
+              address,
+              network,
+            }
+            await this.chainService.addAccountToTrack(addressNetwork)
```
mhluongo, line 503: Hm, this means indexing across all networks we support, even if the user never looks at it... are we sure that's desirable? I guess it's necessary for the overview page, but it could significantly slow down the wallet

background/main.ts
```diff
@@ -494,15 +494,17 @@ export default class Main extends BaseService<never> {
...
+            this.store.dispatch(loadAccount(addressNetwork))
+            this.store.dispatch(setNewSelectedAccount(addressNetwork))
```
mhluongo, line 505: I don't think this is the behavior we want. Could lead to screen flashes from network to network, right?

Instead, let's just do this once outside the loop, defaulting to `ETHEREUM` or whatever the currently selected network is?

mhluongo (commented):
Quick thoughts!

background/main.ts
0xDaedalus: I thought about this while making the change and in practice there is no flashing but you're right that under slow network conditions it could definitely lead to it ✅ .

background/main.ts
0xDaedalus: I'm not quite sure if its desirable but that is the current functionality for accounts imported via seed.  I think the decision to not pre-optimize this was made when first discussing our multi network implementation.

Will this scale to 10 networks?  Almost certainly not.  But we can address when we get there I think.

mhluongo: 👍

jagodarybacka (approved):
Works fine 🥇

background/main.ts
Shadowfiend: I am realllly twitchy about having this logic in Main. I feel like it's almost the job of the chain service to handle this when it sees a ledger import...

Will think some more about the wiring here for the future.

PR #1669, opened 2022-06-24
🌅 Fix old fee preview w/ a deprecated fee settings text component

henryboldi:
This PR adds an updated version of the old fee settings text. Our fee settings
text got updated to work with the new gas selection screen, but left the old
screen behind. This is added to make sure the feature flag provides a working
experience.


0xDaedalus:
I think unfortunately it is still broken:

https://user-images.githubusercontent.com/94649004/175719100-4ed4593a-f80f-4106-8b72-1963b5602744.mov

Issue #1586, opened 2022-06-07
Removing an address fails if wallet is locked.


0xDaedalus:
If the wallet is locked - entering your password after selecting an address to remove does not remove it - and you have to remove it again (once the wallet is in an unlocked state).

https://user-images.githubusercontent.com/94649004/172430227-cdca9e22-3f0a-4383-b6cb-5da52ec6a9cd.mov

beemeeupnow:
still relevant in current version

Issue #1564, opened 2022-06-02
Performance improvements


greg-nagy:

henryboldi:
There's a performance issue is mentioned here: https://github.com/tallycash/extension/issues/1742#issuecomment-1176375994

And discussion in [flowdock](https://www.flowdock.com/app/cardforcoin/tech/threads/tZtrk7Ye89GhvY51NDoXJfpMBUo)

From a cursory look, `getCachedAssets` is merging assets I think twice per call. And `getCachedAssets` is running hundreds of times in unison, particularity noticeable at startup or when adding an address

greg-nagy:
Just a quick notes from my personal list. Not all of these needs to happen though.
The most prominent is floodgate, blocks and json serialization

* try patching BigInt so we don't have to check every property if it's bigint [here](https://github.com/tallycash/extension/blob/831676b8807e61b59a99821f0cc2ba80992b8f3f/background/lib/utils/index.ts#L77)  but use only a simple `JSON.stringify()` (ugly hack, only worth it if the performance gain is significant)
* prune unused data from redux and/or implement paging [flowdock thread] (https://www.flowdock.com/app/cardforcoin/tech/threads/mBOx5TsYPZ6JNqtUg0dJ42cRiJg) (activities, logs, transaction inputs)
* try to mitigate "blocks" event emits firing rapidly in a short period of time 
* finish #1296 to make persisting state reasonable
* check if there is a better way to generate the cachekey for `memoizedMergeAssets`
* rewrite store persistance so it's not persisting data redundantly but grabs what it can from services on startup and only persist the rest

greg-nagy:
Closing in favor of #1350

Issue #1492, opened 2022-05-18
Read-only accounts persisting through account removal on chrome restart.


0xDaedalus:
In this setup:
![image](https://user-images.githubusercontent.com/20949277/169055347-e294b4b3-0bcc-4621-a5c0-f04688235381.png)
if I remove `Sirius` and reload the extension this account will be back on the list.

Otherwise, code looks good ✅ And the flow reported in the issue works fine

_Originally posted by @jagodarybacka in https://github.com/tallycash/extension/issues/1490#issuecomment-1130043107_

PR #1670, opened 2022-06-24
⚠️ Prevent `undefined reading toLowerCase` in uiSelector

henryboldi:
An error here was reported when experiencing a green screen of death. This
doesn't address the root cause but hopefully is an immediate remedy. I haven't
been able to reproduce the issue.

Here's some context,
message one
message two
message three
console

https://discord.com/channels/808358975287722045/888500174685614090/989593132045848596

### DIFF
--- a/background/redux-slices/selectors/uiSelectors.ts
+++ b/background/redux-slices/selectors/uiSelectors.ts
@@ -30,7 +30,7 @@ export const selectCurrentAccount = createSelector(
   ({ address, network }) => ({
     address,
     network,
-    truncatedAddress: address.toLowerCase().slice(0, 7),
+    truncatedAddress: address?.toLowerCase().slice(0, 7),
   })
 )
 


mhluongo:
Not sure we should merge this — feels like bug whackamole, and maybe giving up an opportunity to dig deeper? Could be off here.

Thoughts @Shadowfiend?

henryboldi:
I don't follow how it's giving up an opportunity to dig deeper 🤔

mhluongo:
It's a hazy idea, but I'm wondering if we can use this bug as a tracer to get to a full reproduction. Fixing it this way but leaving the root cause is going to make finding the root cause more difficult down the road.

Eg maybe we put together a build that saves path history, or snapshots the Redux state, or logs better, and triggers an easy export when this happens? Spitballing.

Shadowfiend:
My main concern isn't the tracing approach, it's that this is not the only place we assume `selectedAccount` has a defined `address` and `network`—so rather than remedy the issue, it's likely it just kicks the problem down to the next place in the code that expects the selected account, possibly rendering the UI with missing data and undefined behavior when we try to perform actions.

If we want to fix this, I think what we need is a migration that checks for a missing `selectedAccount` and reinitializes it to the latest value in the db (assuming that isn't also cleared!). But to do that, we'll need to first track down what is clearing the selected account. My best guess after 5 minutes of tracing the paths that affect it is that something is up with account removal, though I can't be sure: https://github.com/tallycash/extension/blame/a7a7c1148025e27f522db201e49a0cb711fef9cd/ui/components/AccountItem/AccountItemRemovalConfirm.tsx#L95-L108 . Incidentally I believe we've got a few other long-tail issues with account removal (e.g. #1586, #1492, etc) and Jagoda's opened #1655 to refactor a bit of how this works.

Overall, this feels like an amazing starting point, but I would suggest:
- Exercising the paths that set the selected account to see if we can reproduce the undefined value being percolated,
- Checking if Jagoda's PR fixes it, if we can reproduce,
- Writing the migration to fill the info back in for installs that have lost it, and,
- Trying to figure out where we leaked type information that allowed an undefined to sneak in so we can patch that hole as well.

mhluongo:
> Checking if Jagoda's PR fixes it, if we can reproduce

We haven't been able to reproduce it yet, and it appears to be the same bug that's blocking us getting a release out for Chrome.

greg-nagy:
I don't think the cause of the issue is that we lose the selected account in the store. I think the proxy store is not yet initialized properly — not yet sure how or why — before the UI is shown. I managed to partially reproduce the issue with CPU throttling enabled, after installation. Can't yet reproduce reliably though. This makes me think that event bursting is also playing a role here (`updateBalance`, `newPricePoint`). 

If my theory is correct, my mitigation plan:
- Finishing #1296 
- Implementing a global loading / error fallback screen [discusses here](https://www.flowdock.com/app/cardforcoin/tally-product-design/threads/mNy4oziTTvDs0GCM_n31ZRW8pwH)
  - make sure the proxy store is properly initalized and show this screen before that happens
- doing #1564 as a safety measure

Notes: 
- I suspect that chrome store did some optimization recently and now the initial extension tests are done on low-resource machines
  - other extensions also started having issues
- the extension was working fine — install + onboard — for the discord user, then it broke 
  - I did a quick check on the code paths and we don't remove the selected account in any of them. We set it, or reset it
  - if the selected account has been set once it's persisted, and initialized from there (then overwritten based on the content of the service)
  - after 2 initial rejections we were told that everything works fine

greg-nagy:
I could reproduce the issue somewhat consistently with CPU throttling on. Usually takes a couple of tries. 
If I open up the UI console and hit cmd+R to reload the UI thread, it fixes the problem. 
I don't see any logs on the UI console and the errors on the background console are related to token lists. 


https://share.cleanshot.com/joWCpd
https://share.cleanshot.com/p523AF

CleanShot 2022-06-27 at 09 45 51

greg-nagy:
https://github.com/tallycash/extension/pull/1672 - this PR mitigates the issue reasonably well so we can probably can be unblocked on chrome store side. 

Will dig a bit more to check why `proxyStore.ready()` is not working in certain situations.

greg-nagy:
@henryboldi I don't think this change should be merged. Can you please close the PR if you agree?

PR #1671, opened 2022-06-24
Telos Feature Request

nathanduft44:
Support Telos EVM


nathanduft44:
Yo what's missing here?

greg-nagy:
Hey @nathanduft44!

Thanks for this contribution! 
Before we go ahead with these changes, we should start a discussion around this feature on discord? https://chat.tally.cash

This is a bigger engineering lift, and it's not currently on our roadmap. Because of this, I don't see clearly how this feature will be finished to a releasable state and it would be a shame to leave it half finished.

PR #1672, opened 2022-06-27
🏴‍☠️🛠  Green screen of death: reload the UI page if it's not rendered in 1s

greg-nagy:
In certain situations the `proxyStore.ready()` is not fired,
and in those cases a green screen is shown forever.
This effect can be reproduced with enabling network +
CPU throttling for the background script in devtools
and a couple of reload>UI open cycle. The blocked
state can be resolved if the UI thread is reloaded.

The solution waits for 1s for the script to start running and
render the first top level element. If this does not happen
then reloads the UI thread.


jagodarybacka (approved):
This change looks good but I can't reproduce it at all on `main` using throttling

greg-nagy:
Background to the problem (as I see it now)

-  the UI rendering is blocked, becuase `proxyStore.ready()` is not firing
    - [ui init](https://github.com/tallycash/extension/blob/87d5fa375b71950bfdcb0a075db276eb282be30f/src/ui.ts#L12)
    - [initial render](https://github.com/tallycash/extension/blob/17b27b03d0afce97ac9f26abc25ce03e4a069b3e/ui/index.ts#L21)
    - [waiting for `proxyStore.ready()`](https://github.com/tallycash/extension/blob/4173caf6720042951204f8d700406210b9a74933/background/index.ts#L33)
- in webext-redux `Store.js` (which is the `proxyStore` in tally) it's fired 2 places in the background
  - [safetyHandler](https://github.com/greg-nagy/webext-redux/blob/8afeefcb4813ded3ed613ff5f1794a9788ca1320/src/store/Store.js#L234)
    - this is fired as a last step for [wrapStore initialization](https://github.com/greg-nagy/webext-redux/blob/8afeefcb4813ded3ed613ff5f1794a9788ca1320/src/wrap-store/wrapStore.js#L200) to notify existing content scripts
    -  This is of little importance to us. There should not be any existing content scripts unless the ledger tab is open. But for the above error it's irrelevant. 
  - [STATE_TYPE message](https://github.com/greg-nagy/webext-redux/blob/8afeefcb4813ded3ed613ff5f1794a9788ca1320/src/store/Store.js#L70)
    - this happens when the whole store is sent through for the first time and it needs to be set (not patched)
  - the [STATE_TYPE is fired](https://github.com/greg-nagy/webext-redux/blob/36cc80936feb10e34e2dd82be8a8c27ff518ffac/src/wrap-store/wrapStore.js#L126) in the [browserAPI.runtime.onConnect listener](https://github.com/greg-nagy/webext-redux/blob/36cc80936feb10e34e2dd82be8a8c27ff518ffac/src/wrap-store/wrapStore.js#L152)
- but when the problem presents itself the onconnect listener is never run 
  - see the [debug setup for webext-redux](https://github.com/tshaddix/webext-redux/pull/285)
    > sorry for the reformat, it makes it hard to see what'shappeningg, but the intersting parts are the console.logs


After some debugging my current understanding is, that in situations where the green screen of death is shown the wrapStore initialization happens correctly, but slower than the UI opening > the [port listener in the background](https://github.com/tshaddix/webext-redux/blob/master/src/wrap-store/wrapStore.js#L152) is not yet registered > the initialization process is not starting at all. 

The below gif shows the problem

Shadowfiend:
Phew that's a deep one 😮‍💨 

> slower than the UI connecting to the port > the [port listener in the background](https://github.com/tshaddix/webext-redux/blob/master/src/wrap-store/wrapStore.js#L152) is not yet registered

Armed with this info, I see exactly how this can happen. The issue is that we don't run `wrapStore` until [the Main service constructor](https://github.com/tallycash/extension/blob/main/background/main.ts#L366), and the Main service constructor won't run [until the remaining services are fully initialized](https://github.com/tallycash/extension/blob/main/background/main.ts#L271-L285)—e.g. dbs are initialized, etc.

I think we can fix the underlying cause instead of patch it by:
- Running `initializeStore` and `wrapStore` in `create` instead of the constructor,
- Passing the initialized store (`ReduxStoreType`) to the constructor.
- Running `this.initializeRedux` normally in the constructor (this would be the only action other than `super` in the constructor).

We can probably reproduce this issue by adding this line to `create`:
```typescript
```

If I were in there I might also move all the redux stuff to a separate file (maybe even `redux-slices/index` tbh), though that's more annoying.

Shadowfiend:
> instead of patch it 

This doesn’t really have to be an instead of, as a side note, particularly if we’re not confident the above fixes the whole issue. Merging this as-is if that’s what y’all prefer is 👍

greg-nagy:
I kept digging, and came back with surprising news! 

But let's start with the not-so-surprising part. 
@Shadowfiend 's above suggestion will fix the issue, and we will do that in a later PR. 

Now the surprising part ([see the measurement commit](https://github.com/tallycash/extension/commit/e1e248664069a22769cfeb0f15f3945eef3d6f82)):
- The main startup time slows down from 600-800ms to 1800-1900ms if I open the UI quickly after extension restart
- This happens because of the `await db["migrate"]` calls in [chain db](https://github.com/tallycash/extension/blob/e1e248664069a22769cfeb0f15f3945eef3d6f82/background/services/chain/db.ts#L326) and in [indexing db](https://github.com/tallycash/extension/blob/e1e248664069a22769cfeb0f15f3945eef3d6f82/background/services/indexing/db.ts#L303)
  - they run parallel but each of them blocks ~400ms and ~1400ms 
  - the service dependency propagates this to the overall startup time 
- the other long-running step during startup is the reading + decoding of the redux state from local storage
  - this step takes ~380ms for read and ~100ms for decode
  - this does not get longer when I open the UI quickly

greg-nagy:
Yeah, so about the self-merge thingy ... :) 

I decided to merge because 
- I got excited and misinterpreted the meaning of the approval
- it's not a costly safety net which is not bad to have independent of the current problem
- wanted to tackle chrome store releases quickly

greg-nagy:
> Suspecting this is prone to an infinite reload loop

I am using a `setTiimeout()` once, when the UI loads and don't restart it.

Otoh I was thinking about a more robust solution where this check is running continuously. 
With some considerations, eg the restart has been done 2 times then let's show an error screen on 3rd. 

This is something that's called watchdog in electric engineering and is used as a fail-safe in case 
anything goes off.

How would you feel about a solution like this? @henryboldi @Shadowfiend 

This could be the error screen, and the button would reload the whole extension like the refresh button.

CleanShot 2022-06-29 at 10 00 59

henryboldi:
> I am using a setTiimeout() once, when the UI loads and don't restart it.

Interesting, wouldn't this create a new timeout on the reload though?

greg-nagy:
> Interesting, wouldn't this create a new timeout on the reload though?

I see what you mean. Yes, it would keep reloading infinitely or until the app is rendered properly.

Partially preventing this would be the purpose of #1734  + some added guards to prevent this error from happening during the whole time when the UI is open, and not just on initial load.

PR #1673, opened 2022-06-27
WIP: use i18next instead of chrome extension locailization

### LINKED ISSUE #1543: Use i18next for internationalization
Related issue https://github.com/tallycash/extension/issues/1498
Right now we are using the built-in browser's i18n API but it won't allow us much control over the languages.

### What
Setup and use [i18next](https://www.i18next.com/) instead of built-in i18n module. Currently we have some strings prepared for translation [here](https://github.com/tallycash/extension/tree/main/ui/_locales).
No visual change is expected.

gasolin:
to fix #1543


gasolin:
test on Brave and Edge works fine 😅
will install Chrome and check

gasolin:
@jagodarybacka installed a new chrome and loaded without issue with `yarn start --config-name chrome`

jagodarybacka:
Hmmm, was it the first time you created a build for chrome? Maybe it's bugged for me because I had already chrome build created and it didn't update manifest properly. If that is a case then the question is if it won't cause any problems for users that have Tally installed and they are doing update 🤔 @greg-nagy you have a lot of experience with Chrome Store, what do you think?

jagodarybacka (dismissed):
Ok, now everything works fine 🎉
@greg-nagy please confirm that the manifest will update correctly and we can merge it

0xDaedalus (approved):
LGTM as well 👍

ui/_locales/i18n.ts
```diff
@@ -0,0 +1,26 @@
...
+  },
+}
+
+i18n
+  .use(initReactI18next) // passes i18n down to react-i18next
+  .init({
+    resources,
+    lng: process.env.DEFAULT_LANG || "en",
+    fallbackLng: "en",
+    debug: true,
```
greg-nagy, line 17: This should probably be turned off when in prod mode.

greg-nagy (changes requested):
The PR looks great, ty 🙏

Please turn off the debug flag at least in prod mode and resolve the conflicts. 
If these are done feel free to dismiss my review and merge it. 

@jagodarybacka Tested the manifest upgrade problem, and it won't cause any problem as far as I can tell. 

And again, this is awesome work!

gasolin:
set debug: false by default and solved conflict, thanks

greg-nagy (approved):
debug: false, merge: true, conflicts: 0

🙇‍♂️

ui/_locales/en/messages.json
```diff
@@ -1,214 +1,55 @@
...
-  },
-  "settingsExportOptionLogTitle": {
-    "message": "2. Export logs",
-    "description": "in pages > Settings > Report bug"
-  },
-  "settingsExportOptionLogDesc": {
-    "message": "To better understand your bug, we need you to export your logs.",
-    "description": "in pages > Settings > Report bug"
-  },
-  "settingsExportOptionLogBtn": {
```
henryboldi, line 210: This broke the settings screen. `settingsExportOptionLogBtn` got overwritten with `menuExportOptionLogBtn`.

gasolin: ouch will change string tomorrow based on yesterday's change https://github.com/tallycash/extension/commit/ca73504124a134cf9b6bc06ddd365d218bcb0419

or @henryboldi would you like take care of it?

PR #1679, opened 2022-06-28
🎩 Max character limit on custom miner/max base

### LINKED ISSUE #1644: Gas Control update #1302 |  Estimated network fee Settings - Need Max Character Limit on Custom Miner/Max Base
### **Build Info/PR/Commit:**
https://github.com/tallycash/extension/pull/1641

https://github.com/tallycash/extension/commit/f29b2f20220abc9b49db00ac3d33e138d55e943e

### **Related Ticket:***
https://github.com/tallycash/extension/issues/1302

### **Reproduction Steps:**

1. Send Asset
2. Estimated network fee Settings
3. Add numbers to Custom Miner/Max Base fields

### **Expected Result:**
Prevent Users from entering too many characters. 

### **Actual Result:**
image

image

henryboldi:
Closes #1644


ui/components/NetworkFees/NetworkSettingsSelectOptionButtons.tsx
```diff
@@ -176,6 +176,7 @@ export function NetworkSettingsSelectOptionButtonCustom({
                   gweiFloatToWei(parseFloat(value))
                 )
               }}
+              maxLength={3}
```
jagodarybacka, line 179: I would allow 4 signs here to support `1.05` for example

Also there is a problem with `1.1`
image

jagodarybacka (dismissed):
Looks better but please take a look at the `1.1` case. Also for me it is still not saving custom values but I guess it is not in the scope of this task.

Maybe gas limit should have max length too to avoid:
image

ui/components/NetworkFees/NetworkSettingsSelectOptionButtons.tsx
henryboldi: > I would allow 4 signs here to support 1.05 for example

Great point! Updated 🙇✅


> Also there is a problem with 1.1

I believe that one is fixed in https://github.com/tallycash/extension/pull/1660/commits/d97b6b99a2aa9868b889f6f209dd1fd2a4d34b36

0xDaedalus (approved):
👍  Looks good to me

Created https://github.com/tallycash/extension/issues/1704 as a follow-on

greg-nagy (approved):
let's 🚢

PR #1683, opened 2022-06-29
Exit claim check early if HIDE_EARN_PAGE is on

0xDaedalus:
This will get rid of the `Async thunk failed Error: call revert exception (method="isClaimed(uint256)",` error thats all over the console logs.


background/redux-slices/claim.ts
```diff
@@ -202,6 +202,9 @@ export default claimingSlice.reducer
 export const checkAlreadyClaimed = createBackgroundAsyncThunk(
   "claim/checkAlreadyClaimed",
   async ({ claimState }: { claimState: ClaimingState }, { dispatch }) => {
+    if (HIDE_EARN_PAGE) {
```
jagodarybacka, line 205: I would use `HIDE_TOKEN_FEATURES` flag

PR #1685, opened 2022-06-29
Network-specific asset-transfers and block fetching.

0xDaedalus:
This PR fixes a race condition where fetching the latest Ethereum block would sometimes fetch the latest Polygon block.

It also addresses a similar issue where we were sometimes looking for the newest/oldest account transfer on the wrong network.

### To Test

Load up a fresh wallet 
- wait.
- Ethereum activity should populate
- wait even longer (5-10m should be good)
- Polygon activity should populate


greg-nagy (approved):
ty sir, this feels like something that you have brought back with yourself from a very deep rabbit hole 🕳🐇👑

PR #1686, opened 2022-06-29
fix selectShowingActivityDetail

greg-nagy:
current account address was another layer in the data structure
which wasn't represented in the selector


background/redux-slices/selectors/uiSelectors.ts
```diff
@@ -40,6 +18,19 @@ export const selectCurrentAccount = createSelector(
   })
 )
 
+export const selectShowingActivityDetail = createSelector(
+  (state: RootState) => state.activities,
+  selectCurrentAccount,
+  (state: RootState) => state.ui.showingActivityDetailID,
+  (activities, currentAccountOnNetwork, showingActivityDetailID) => {
+    if (!showingActivityDetailID) return null
```
0xDaedalus, line 26:
```suggestion
    if (!showingActivityDetailID) {
      return null
    }
```

0xDaedalus (approved):

PR #1707, opened 2022-07-01
Sort dropped transactions directly after a successful transaction with the same nonce.

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/998

image

## To Test

- Copy/paste the following code between lines 97 and 96 [here](https://github.com/tallycash/extension/blob/main/background/redux-slices/activities.ts#L97) to simulate 2 dropped transactions for every succesfull one
```
          activitiesAdapter.upsertOne(
            immerState[address][activityItem.network.chainID],
            {
              ...activityItem,
              status: 0,
              blockHash: null,
              blockHeight: null,
              hash: `${activityItem.hash}1`,
            }
          )

          activitiesAdapter.upsertOne(
            immerState[address][activityItem.network.chainID],
            {
              ...activityItem,
              status: 0,
              blockHash: null,
              blockHeight: null,
              hash: `${activityItem.hash}2`,
            }
          )
```
- Dropped transactions should show up directly after their corresponding successful one


background/redux-slices/activities.ts
```diff
@@ -18,6 +18,15 @@ const activitiesAdapter = createEntityAdapter<ActivityItem>({
...
     ) {
+      // Sort dropped transactions after their corresponding successful ones.
+      if (a.nonce === b.nonce) {
+        if (a.blockHash === null) {
+          return 1
+        }
+        if (b.blockHash === null) {
+          return -1
+        }
+      }
```
Shadowfiend, line 29: We consistently use `blockHeight` rather than hash elsewhere in the function—I suggest we stick to that here.

jagodarybacka (commented):
For me they are showing like this, is it correct? It's the reverse order looking at your screen. I've added your snippet line below `else` https://github.com/tallycash/extension/blob/sort-dropped-tx-2/background/redux-slices/activities.ts#L105
image

0xDaedalus:
Whoops sorry about that - I did not update the testing steps after adding https://github.com/tallycash/extension/pull/1707/commits/b1927f51079cb2c9e0f678fa4a8c03425d324747.  Crucially the copypasta should also have `blockHeight: null` as part of the dropped tx construction.  I've fixed it in the original comment so you should be good to test with it.

henryboldi (approved):
Lookin' good! 👏👑

dropped order

PR #1716, opened 2022-07-02
pass param correctly with i18next

gasolin:
- fix #1673 by update the change from https://github.com/tallycash/extension/commit/ca73504124a134cf9b6bc06ddd365d218bcb0419
- pass the `$2` param correctly


henryboldi (approved):
Looks good! 👑 Thank you for the quick fix

PR #1718, opened 2022-07-02
Fix overview address count

0xDaedalus:
Fixes https://github.com/tallycash/extension/issues/1700

To Test:
- Ensure that the overview address count is not double the number of addresses in your address list.
- Adding an address should increase the address count by 1
- Removing an address should decrease the address count by 1


Shadowfiend:
Ah, the old “what is an account” problem, solved in a rather awkward way, resurfaces :grin:

mhluongo (approved):
Easy peasy, tested and approved :raised_hands:

PR #1720, opened 2022-07-02
Bugfix: Swapping tokens w/ same name & symbol but different networks no longer allowed.

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/1703.

image

## To Test
- Open SWAP menu on ethereum network
- searching for ETH should no longer show `Ether - PoS` in the buy token list

### DIFF
--- a/background/redux-slices/selectors/accountsSelectors.ts
+++ b/background/redux-slices/selectors/accountsSelectors.ts
@@ -8,6 +8,7 @@ import {
   enrichAssetAmountWithMainCurrencyValues,
   formatCurrencyAmount,
   heuristicDesiredDecimalsForUnitPrice,
+  isNetworkBaseAsset,
 } from "../utils/asset-utils"
 import {
   AnyAsset,
@@ -31,7 +32,7 @@ import {
   selectSourcesByAddress,
 } from "./keyringsSelectors"
 import { AddressOnNetwork } from "../../accounts"
-import { EVMNetwork, sameNetwork } from "../../networks"
+import { EVMNetwork, NetworkBaseAsset, sameNetwork } from "../../networks"
 import { BASE_ASSETS_BY_SYMBOL } from "../../constants"
 import { DOGGO } from "../../constants/assets"
 import { HIDE_TOKEN_FEATURES } from "../../features"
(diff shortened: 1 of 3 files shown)


background/redux-slices/utils/asset-utils.ts
```diff
@@ -35,6 +35,10 @@ export type AssetDecimalAmount = {
   localizedDecimalAmount: string
 }
 
+function isBaseAsset(asset: AnyAsset): asset is NetworkBaseAsset {
+  return !("homeNetwork" in asset)
```
Shadowfiend, line 39: I think this needs to be a presence check on `coinType` rather than an absence check on `homeNetwork`, because the `NetworkBaseAsset` type doesn't check for absence of fields in TypeScript, just presence.

background/redux-slices/utils/asset-utils.ts
```diff
@@ -48,12 +52,12 @@ export type AssetDecimalAmount = {
...
+): asset is NetworkBaseAsset {
   return (
-    !("homeNetwork" in asset) &&
-    "family" in network &&
-    network.family === "EVM" &&
-    asset.symbol === network.baseAsset.symbol
+    isBaseAsset(asset) &&
+    asset.symbol === network.baseAsset.symbol &&
+    asset.coinType === network.baseAsset.coinType &&
+    asset.name === network.baseAsset.name
```
Shadowfiend, line 60: Also worth noting: base assets may have a `homeNetwork` field or not have a `coinType` if they come in through a pipeline other than our constants. Currently that isn't the case anywhere I don't believe, but we may want to safeguard by checking that if there *is* a `homeNetwork`, it matches the network we're checking against. I also *think* that invalidates the usage of this function as a type guard vs a simple Boolean filter, though I haven't thought it all the way through and could be wrong. 

All that said, obv I wrote the first version of this code and missed these corner cases, so this probably isn't blocking.

0xDaedalus: > if they come in through a pipeline other than our constants.

This seems like something we should prevent.  Could this behavior be whats leading to https://github.com/tallycash/extension/issues/1715 ?

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -43,6 +44,21 @@ const desiredDecimals = 2
...
 const userValueDustThreshold = 2
 
+const shouldForciblyDisplayAsset = (
+  assetAmount: CompleteAssetAmount<AnyAsset>,
+  network: EVMNetwork,
+  baseAsset?: NetworkBaseAsset
+) => {
+  if (!baseAsset) {
+    return false
+  }
```
jagodarybacka, line 54: I'm not sure in what case `baseAsset` won't be provided but looking at the old version of `isForciblyDisplayed` this changed the behaviour, we previously were displaying DOGGO regardless of `baseAsset`

jagodarybacka (approved):
Works fine overall but please take a look at these two things before the merge 😄

background/redux-slices/selectors/accountsSelectors.ts
0xDaedalus: Yup - since DOGGO is not a thing yet and may not be for months - I've taken it out for now.  Should be quite straightforward to add it back in.

background/redux-slices/utils/asset-utils.ts
```diff
@@ -35,6 +35,10 @@ export type AssetDecimalAmount = {
   localizedDecimalAmount: string
 }
 
+function isBaseAsset(asset: AnyAsset): asset is NetworkBaseAsset {
```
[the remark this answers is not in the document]
0xDaedalus: Correct - only our defined-in-code baseASsets have the `coinType` field

PR #1722, opened 2022-07-03
Deeply Diffed: Move webext-redux updatesto deep-diff strategy

Shadowfiend:
This strategy only sends the actual changed areas across the port, but
more importantly it only changes the objects that need to be changed in
the proxy store. This leads to less rerendering in the extension popup.

This is something I noticed while debugging a network-switching-related
issue, where rerenders were triggering certain `useEffect`s rather
unexpectedly, because of how objects are compared. This should make object
comparisons and resulting invalidation of any dependencies work a little
more predictably/like we might expect.


greg-nagy:
I was convinced that this will be a big lift based on past experiences. Should have verified ... 
Happy to see that this works so well :)

PR #1723, opened 2022-07-03
Operation Unifier: Make transactions follow the SignOperation<T> contract

Shadowfiend:
> [**Operation Unifier**, also known as **Canadian Armed Forces Joint Task Force-Ukraine**, is the Canadian Armed Forces contribution to the security of Ukraine in coordination with the Ukrainian Armed Forces. It also includes a small Swedish contingent. It was begun in light of fomentation by separatist sentiments in the Donetsk and Luhansk and Crimean regions of Ukraine after the 2014 Ukrainian revolution and the 2014 pro-Russian unrest in Ukraine.](https://en.wikipedia.org/wiki/Operation_Unifier)

Transaction signing requests were of type `SignatureRequest`, which
predates the `SignOperation<T>` type that unifies typed data and message
signing requests. To unify all signing and make progress towards moving
all signing management into the signing slice, signing operations are
now all represented by `SignOperation<T>`, and the T type parameter is
restricted to the known types of signing requests the wallet can handle.

The changes are minimal, since `SignatureRequest` was very close in
structure already to `SignOperation`.

## Testing

- [ ] Test transaction signing for internal dApps (send ETH, send tokens, swaps).
- [ ] Test transaction signing for external dApps.
- [ ] Test other types of signing (data, typed data), though they should not be on any of the updated paths here.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -632,13 +632,10 @@ export default class Main extends BaseService<never> {
 
     transactionConstructionSliceEmitter.on(
       "requestSignature",
-      async ({ transaction, accountSigner }) => {
+      async ({ request, accountSigner }) => {
         try {
           const signedTransactionResult =
-            await this.signingService.signTransaction(
-              transaction,
-              accountSigner
-            )
+            await this.signingService.signTransaction(request, accountSigner)
           await this.store.dispatch(transactionSigned(signedTransactionResult))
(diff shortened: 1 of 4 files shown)


mhluongo (approved):
I've tested swaps as well as dApp tx and data signatures. Excited for the eventual implications here!

PR #1724, opened 2022-07-03
Look Don't Sign: Introduce a first-class ReadOnlyAccountSigner

Shadowfiend:
Until now, null or undefined as used to represent a read-only account,
but as we move towards a transaction signing flow parametrized by the
AccountSigner, it becomes useful to explicitly treat read-only as signer
objects with a discriminant `type` field.

## Testing

- [ ] Ensure adding accounts of all types (read-only, Ledger, import, generated) continues to work.
- [ ] Ensure the account selector continues to work correctly for all account types.
- [ ] Test attempting transaction, data, and typed data signatures with read-only, Ledger, imported, and locally-generated accounts.


mhluongo (approved):
Excited about this :raised_hands:

PR #1726, opened 2022-07-04
🤗 Fix uniswap connection

### LINKED ISSUE #1647: dApp connection is not working on Uniswap

greg-nagy:
Uniswap has migrated to web3-react v8 and during this process, they broke our integration. 

For the exact problems see the commit messages. 

# Test
- [ ] check if live uniswap dApp connection is working

Closes #1647

### DIFF
--- a/background/services/provider-bridge/index.ts
+++ b/background/services/provider-bridge/index.ts
@@ -145,6 +145,10 @@ export default class ProviderBridgeService extends BaseService<Events> {
       this.emitter.emit("setClaimReferrer", String(referrer))
 
       response.result = null
+    } else if (event.request.method === "eth_chainId") {
+      // we need to send back the chainId independent of dApp permission if we want to be compliant with MM and web3-react
+      const { network } = await this.preferenceService.getSelectedAccount()
+      response.result = network.chainID
     } else if (typeof originPermission !== "undefined") {
       // if it's not internal but dapp has permission to communicate we proxy the request
       // TODO: here comes format validation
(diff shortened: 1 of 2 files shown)


jagodarybacka:
So does it mean that we don't need our own connector in web3-react and we can close PRs in Uniswap and web3-react repositories?

greg-nagy:
> So does it mean that we don't need our own connector in web3-react and we can close PRs in Uniswap and web3-react repositories?

After this, we don't **need** it anymore, but it would still be preferable. This solution only works for `window.ethereum` and thus we need to fight for it > back to default wallet or not. 

If we were to have our own connectors, then we could use `window.tally` which would allow using tally and MM side by side. I wouldn't close them, because it would be better for us and frankly everybody if we would have that.

background/services/provider-bridge/index.ts
```diff
@@ -145,6 +145,10 @@ export default class ProviderBridgeService extends BaseService<Events> {
       this.emitter.emit("setClaimReferrer", String(referrer))
 
       response.result = null
+    } else if (event.request.method === "eth_chainId") {
+      // we need to send back the chainId independent of dApp permission if we want to be compliant with MM and web3-react
+      const { network } = await this.preferenceService.getSelectedAccount()
```
0xDaedalus, line 150: this needs to be `ethereumService.activeNetwork` since e.g. we can be connected to ethereum in our wallet but connected to polygon in a given dapp.

greg-nagy: Don't think this is the correct implementation, but the catch is a good  🙏

0xDaedalus (approved):
Glorious 🥲 

image

PR #1732, opened 2022-07-05
🌡 Password strength bar

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1329

### What

Create the password strength bar component and use it to evaluate how strong is the password that user wants to create.

https://user-images.githubusercontent.com/20949277/177314535-0c8de87d-565a-44bc-8819-6ef748229299.mov

### DIFF
--- a/ui/components/Shared/SharedPasswordBar.tsx
+++ b/ui/components/Shared/SharedPasswordBar.tsx
@@ -1,31 +1,42 @@
 import classNames from "classnames"
-import React, { ReactElement } from "react"
+import React, { ReactElement, useEffect, useState } from "react"
+import zxcvbn from "zxcvbn"
 import SharedTooltip from "./SharedTooltip"
 
 type PasswordBarProps = {
-  strength?: PasswordStrength
+  password: string
 }
 
-export type PasswordStrength = "Weak" | "Average" | "Strong"
+function getDescription(evaluation: number) {
+  if (evaluation < 3) return "Weak"
+  if (evaluation === 3) return "Average"
+  return "Strong"
+}
 
 export default function SharedPasswordBar(
   props: PasswordBarProps
 ): ReactElement {
-  const { strength } = props
+  const { password } = props
+
+  const [evaluation, setEvaluation] = useState(0)
+
+  useEffect(() => setEvaluation(zxcvbn(password).score), [password])
 
   return (
     <div
       className={classNames("bar_wrap", {
-        bar_weak: strength === "Weak",
-        bar_average: strength === "Average",
-        bar_strong: strength === "Strong",
+        bar_weak: password.length && evaluation < 3,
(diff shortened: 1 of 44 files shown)


ui/components/Shared/SharedPasswordBar.tsx
```diff
@@ -0,0 +1,99 @@
...
+  password: string
+}
+
+function getDescription(evaluation: number) {
+  if (evaluation < 3) return "Weak"
+  if (evaluation === 3) return "Average"
+  return "Strong"
+}
+
+export default function SharedPasswordBar(
```
mhluongo, line 16: A more descriptive name might be `SharedPasswordStrengthBar`

ui/components/Shared/SharedPasswordBar.tsx
```diff
@@ -0,0 +1,99 @@
...
+      })}
+    >
+      <div className="bar_background bar_shape">
+        <div className="bar_fill bar_shape" />
+      </div>
+      <div className="bar_description">
+        {password.length ? getDescription(evaluation) : "Strength"}
+      </div>
+      <SharedTooltip width={150}>
+        For a strong password use a mix of: Lowecase, Uppercase, symbols and
```
mhluongo, line 40: I know this copy was in Figma, but let's tighten it up.

```
For a strong password, consider a mix of lower and uppercase characters, symbols, and numbers. Passwords should be over 8 characters.
```

Also wondering if we should be dropping new copy into our locales file rather than directly in JSX?

mhluongo (changes requested):
Works well locally. Left a couple thoughts on copy + naming!

ui/components/Shared/SharedPasswordBar.tsx
Shadowfiend: Should this be shared at all? Are we expecting it will be used anywhere other than that one page?

jagodarybacka: I think maybe on the "change password" page it should be used as well when we will create that page in the future

Shadowfiend: Not really looking to waste much time on it, but I think to me the Shared namespace implies sharing across fairly unrelated use cases. If we have a password creation and change flow I'd probably expect a `components/Password/PasswordCreate` and `components/Password/PasswordChange` type of structure that would allow the strength bar to live alongside both. 

All of that can be handled another time tbh, just sharing some of my thinking here.

henryboldi: > I think to me the Shared namespace implies sharing across fairly unrelated use cases. If we have a password creation and change flow I'd probably expect a components/Password/PasswordCreate and components/Password/PasswordChange type of structure that would allow the strength bar to live alongside both.

100%

PR #1738, opened 2022-07-05
Fallback to sendAsync inside of window-provider's send method

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/1736

Evidence of bridge working: 
Approval: https://etherscan.io/tx/0xae3dd5f15036b1359faf712e5c4c0272d444927eeaef92d605c625b0353f619e
Bridge Transaction: https://etherscan.io/tx/0x8c4a3701d38085ecd06d3a4b8fab161eac59924d89a2ee2dc0fb0fe7c84d716b

So it appears what was happening was that https://wallet.polygon.technology/bridge/ was calling `send` but expecting the response to their call to be in the shape of a `sendAsync` response.

Question to @greg-nagy or @Shadowfiend  - do we think this breaks existing functionality - or is the conditional branch inside of the `if` statement on line 142 something that has never worked?

If we think it will break existing functionality - I fear we may need some sort of whitelist for dapps that require the response to be 1 way vs the other.

### DIFF
--- a/window-provider/index.ts
+++ b/window-provider/index.ts
@@ -140,10 +140,7 @@ export default class TallyWindowProvider extends EventEmitter {
     }
 
     if (isObject(methodOrRequest) && typeof paramsOrCallback === "function") {
-      return this.request(methodOrRequest).then(
-        (response) => paramsOrCallback(null, response),
-        (error) => paramsOrCallback(error, null)
-      )
+      return this.sendAsync(methodOrRequest, paramsOrCallback)
     }
 
     return Promise.reject(new Error("Unsupported function parameters"))


Shadowfiend:
Can we instead have a try/catch type fallback when we call the callback? Where if it fails we try again with a sendAsync-shaped response?

Regardless this is… Not great heh. It may also be worth it to dig into how other wallets are handling it atm—in particular whether they've had to introduce special cases or not.

mhluongo:
I'm pretty sure MetaMask "is the spec" here :grimacing:

0xDaedalus:
`Can we instead have a try/catch type fallback when we call the callback? Where if it fails we try again with a sendAsync-shaped response?`

I don't think this will work - the exceptions we've been dealing with here, in opensea, etc.. happen inside of the code handling the response from `send`/`sendAsync`.  Our window provider processes everything fine but the calling code is what ultimately fails.

0xDaedalus:
As I feared - it does not seem that adding a try/catch block around the callback is sufficient.  I think it's probably reasonable to assume that the passed in callback may not be on the stack when an eventual error is thrown in the dapp.

henryboldi (approved):
> My take after digging in is that this change is actually a bug fix to the original incorrect implementation.
So merging makes sense.

Appears to work!

works2

Issue #13, opened 2021-03-31
Versioned extension state migration


mhluongo:
Developing fast means we need a simple way to upgrade user state every new extension version. How do we handle state schema changes?

mhluongo:
Closed by #753

PR #1744, opened 2022-07-06
Show network specific asset in activity card

### LINKED ISSUE #1731: Polygon - Txn page updates
Amount is displayed in ETH! :)

I imagine max fee, gas price, gas, etc. may need also to be tweaked

Screen Shot 2022-07-04 at 10 50 29 PM

jagodarybacka:
Resolves #1731

TO DO:

- [ ] migrate previous activities

### What

Show correct currency symbol for Polygon activities. Other rows are fine, same as on Polygonscan.

image

### DIFF
--- a/background/redux-slices/utils/activity-utils.ts
+++ b/background/redux-slices/utils/activity-utils.ts
@@ -1,6 +1,34 @@
 import { convertToEth, weiToGwei } from "../../lib/utils"
+import { ConfirmedEVMTransaction } from "../../networks"
 import { EnrichedEVMTransaction } from "../../services/enrichment"
 import { HexString } from "../../types"
+
+type FieldAdapter = {
+  readableName: string
+  transformer: (tx: EnrichedEVMTransaction) => string
+  detailTransformer: (tx: EnrichedEVMTransaction) => string
+}
+
+type UIFields = keyof (EnrichedEVMTransaction &
+  Pick<ConfirmedEVMTransaction, "gasUsed">)
+
+type UIAdaptationMap = {
+  [P in UIFields]?: FieldAdapter
+}
+
+export type ActivityItem = EnrichedEVMTransaction & {
+  localizedDecimalValue: string
+  blockHeight: number | null
+  fromTruncated: string
+  toTruncated: string
+  infoRows: {
+    [name: string]: {
+      label: string
+      value: string
+      valueDetail: string
+    }
+  }
+}
 
 export function getRecipient(activityItem: ActivityItem): {
   address: HexString | undefined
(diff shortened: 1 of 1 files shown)


background/redux-slices/migrations/index.ts
```diff
@@ -36,6 +37,7 @@ const allMigrations: { [targetVersion: string]: Migration } = {
   10: to10,
   11: to11,
   12: to12,
```
jagodarybacka, line 39: @greg-nagy there is a problem with migration #12 because `REDUX_STORE_VERSION` wasn't updated while that migration was added. After I've updated `REDUX_STORE_VERSION` to `13` so my migration would run it seems like migration `12` is not correct and throws errors. Can you take a look at it? That was your migration as I suspect from the commits

I've tested migration #13 without #12 and it seems fine

greg-nagy: That was a merge conflict resolution and me editing somebody else's PR – shame –, but I will take a look.

greg-nagy: TL;DR: We can remove/replace the `to12` migration.

The flow behind this is: 
- we handled the migration in indexedDB here https://github.com/tallycash/extension/blob/aeab7812d633a8d63fcab30d37928cf0f274d84b/background/services/provider-bridge/db.ts#L73-L113
- we read the migrated data on `ProviderBridgeService.internalStartService` https://github.com/tallycash/extension/blob/aeab7812d633a8d63fcab30d37928cf0f274d84b/background/services/provider-bridge/index.ts#L97
  - and dispatch `initializeAllowedPages`
- we listen to this in main and dispatch `initializePermissions` https://github.com/tallycash/extension/blob/aeab7812d633a8d63fcab30d37928cf0f274d84b/background/main.ts#L1050
- `initializePermissions` overwrite/sets the content of the redux store https://github.com/tallycash/extension/blob/aeab7812d633a8d63fcab30d37928cf0f274d84b/background/redux-slices/dapp.ts#L60-L74
  - it is using the new naming here https://github.com/tallycash/extension/blob/aeab7812d633a8d63fcab30d37928cf0f274d84b/background/redux-slices/dapp.ts#L76

A bit more context
- the change has been introduced in this PR https://github.com/tallycash/extension/pull/1619 and has been released in v0.13.13
- if a user was on 0.13.7 and upgrades to your branch everything works 
- if a user was on 0.13.13 — which is highly likely — the error presents itself. Because the migration has been done 
  - but the user will still have 
- the slice was also renamed to `dapp` from `dappPermissions`. Because of this 
  - the `dappPermissions` was in the persisted state, and was used to set the initial store state
    - it didn’t have a slice with this name, so I assume redux throw it away
  - the dapp slice was automatically created 
  - the slice was set to the payload of the `initializePermissions`
  - only the new slice was persisted

background/redux-slices/migrations/to-13.ts
```diff
@@ -0,0 +1,75 @@
...
+export default (prevState: Record<string, unknown>): State => {
+  const oldState = prevState as State
+
+  const newActivities: ActivitiesState = Object.fromEntries(
+    Object.entries(oldState.activities).map(([address, activitiesByChain]) => [
+      address,
+      Object.fromEntries(
+        Object.entries(activitiesByChain).map(
+          ([chainId, activitiesOnChain]) => {
+            if (chainId !== POLYGON.chainID) return [chainId, activitiesOnChain]
```
Shadowfiend, line 34: If this migration only applies to polygon, and we've never done a release with polygon enabled, why risk an unexpected corner case to migrate? Feels like we should just skip the migration, no?

PR #1746, opened 2022-07-06
Fix asset transfers

### LINKED ISSUE #1158: Wallet Balances Not Updating

greg-nagy:
# What

Created an alarm to check for transactions every minute. 
There is a ~5 minutes delay in transactions that we are picking up now, which is not ideal.

Also fixed the asset flow to add the assets from the history to the `assetsToTrack` list.

# Test
- revert #1755
- import a new wallet with a history 
- `tally/indexing > assetsToTrack` should be populated with all the assets
  - note: the UI will pick up the changes in the next minute or so 

🤔 Why is the UI not populated with all the assets on the first run?

Ref #1158

### DIFF
--- a/background/lib/alchemy.ts
+++ b/background/lib/alchemy.ts
@@ -108,7 +108,7 @@ export async function getAssetTransfers(
         return null
       }
 
-      const asset = !transfer.rawContract.address
+      const asset = transfer.rawContract.address
         ? {
             contractAddress: transfer.rawContract.address,
             decimals: Number(BigInt(transfer.rawContract.decimal)),
--- a/background/lib/alchemy.ts
+++ b/background/lib/alchemy.ts
@@ -51,13 +51,13 @@ export async function getAssetTransfers(
 
   // Default Ethereum Mainnet categories per the documentation:
   // https://docs.alchemy.com/alchemy/enhanced-apis/transfers-api#alchemy_getassettransfers-ethereum-mainnet
-  let category = ["external", "internal", "token"]
+  let category = ["external", "internal", "erc20"]
 
   if (addressOnNetwork.network.name !== "Ethereum") {
     // Unfortunately even though "token" is supposed the default category for this API call - if the `category` property is omitted
     // the api returns an error about the category "iternal" not being supported
     // https://docs.alchemy.com/alchemy/enhanced-apis/transfers-api#alchemy_getassettransfers-testnets-and-layer-2s
-    category = ["token"]
+    category = ["erc20"]
   }
   // TODO handle partial failure
   const rpcResponses = await Promise.all([
--- a/background/lib/alchemy.ts
+++ b/background/lib/alchemy.ts
@@ -49,15 +49,14 @@ export async function getAssetTransfers(
     // excludeZeroValue: false,
   }
 
-  // Default Ethereum Mainnet categories per the documentation:
+  // Categories that are most important to us, supported both on Ethereum Mainnet and polygon
   // https://docs.alchemy.com/alchemy/enhanced-apis/transfers-api#alchemy_getassettransfers-ethereum-mainnet
-  let category = ["external", "internal", "erc20"]
+  const category = ["external", "erc20"]
 
-  if (addressOnNetwork.network.name !== "Ethereum") {
-    // Unfortunately even though "token" is supposed the default category for this API call - if the `category` property is omitted
-    // the api returns an error about the category "iternal" not being supported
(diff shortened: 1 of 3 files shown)


background/services/chain/index.ts
```diff
@@ -768,9 +776,7 @@ export default class ChainService extends BaseService<Events> {
   private async loadRecentAssetTransfers(
     addressNetwork: AddressOnNetwork
   ): Promise<void> {
-    const blockHeight =
-      (await this.getBlockHeight(addressNetwork.network)) -
-      BLOCKS_TO_SKIP_FOR_TRANSACTION_HISTORY
```
Shadowfiend, line 773: This is important, and leaving it off will likely result in errors. See the comment for `BLOCKS_TO_SKIP_FOR_TRANSACTION_HISTORY`. The nodes will throw 500s if the requested block is too close.

This is also a reasonable flag in terms of trying to use this to fix very recent asset transfer lookups—we would need to wait ~20 blocks ~= 5 minutes before picking stuff up this way unfortunately.

background/lib/alchemy.ts
```diff
@@ -51,13 +51,13 @@ export async function getAssetTransfers(
...
   // https://docs.alchemy.com/alchemy/enhanced-apis/transfers-api#alchemy_getassettransfers-ethereum-mainnet
-  let category = ["external", "internal", "token"]
+  let category = ["external", "internal", "erc20"]
 
   if (addressOnNetwork.network.name !== "Ethereum") {
     // Unfortunately even though "token" is supposed the default category for this API call - if the `category` property is omitted
     // the api returns an error about the category "iternal" not being supported
     // https://docs.alchemy.com/alchemy/enhanced-apis/transfers-api#alchemy_getassettransfers-testnets-and-layer-2s
-    category = ["token"]
+    category = ["erc20"]
```
0xDaedalus, line 60: Neat!!  This is a very recent update to the alchemy api.  Lets add "external" here as well because it seems they support it now too.

background/services/doggo/index.ts
```diff
@@ -59,30 +60,32 @@ export default class DoggoService extends BaseService<Events> {
...
-    })
-
-    // Track referrals for all added accounts and any new ones that are added
-    // after load.
-    this.chainService.emitter.on("newAccountToTrack", (addressOnNetwork) => {
-      this.trackReferrals(addressOnNetwork)
-    })
-    ;(await this.chainService.getAccountsToTrack()).forEach(
-      (addressOnNetwork) => {
+    if (!HIDE_TOKEN_FEATURES) {
```
0xDaedalus, line 63: 👏

background/lib/alchemy.ts
```diff
@@ -108,7 +108,7 @@ export async function getAssetTransfers(
         return null
       }
 
-      const asset = !transfer.rawContract.address
+      const asset = transfer.rawContract.address
```
greg-nagy, line 111: This sneaky little fella ...

mhluongo: 🎉🎉🎉

mhluongo:
So now that we've fixed the asset discovery here... it's making all the terrible spam tokens show up in my wallet! We should probably only add these assets **if and only if** the `from` is us, otherwise anyone can aidrdop scam tokens on us without much recourse.

mhluongo:
Eh, but that doesn't cover eg sending a token from another wallet of yours. Damn. I suppose we'll just need to double-down on metadata and the trusted / untrusted token UI

henryboldi:
Perhaps can exclude aidrdop scam tokens by checking the price?

mhluongo:
Unfortunately, that would also exclude LP tokens and other new stuff.

Shadowfiend:
If there's a way to do it cleanly, we should probably only add tokens from asset transfers that are already in the token list until we can add a better UI for managing a denylist and trusted/untrusted stuff.

Shadowfiend:
(Notably that limitation should ideally not exist for transactions that are not explicitly from our address.)

mhluongo:
I think we should stick with what we have for now and solve the real problem of trust. Looking at my wallet, there are some good tokens in there that will never be in a token list... but that I want to add to my own internal list. Better to show those *and* the spam for a short while I think, and solve auto-discovered tokens ASAP.

PR #1753, opened 2022-07-06
Allow wallet_addEthereumChain to switch a given origin's active chain

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/1748

Only switches the chain if the `chainId` passed into `wallet_addEthereumChain` is known to Tally

### DIFF
--- a/background/redux-slices/activities.ts
+++ b/background/redux-slices/activities.ts
@@ -18,6 +18,15 @@ const activitiesAdapter = createEntityAdapter<ActivityItem>({
         a.blockHeight === b.blockHeight) &&
       a.network.name === b.network.name
     ) {
+      // Sort dropped transactions after their corresponding successful ones.
+      if (a.nonce === b.nonce) {
+        if (a.blockHeight === null) {
+          return 1
+        }
+        if (b.blockHeight === null) {
+          return -1
+        }
+      }
       // Sort by nonce if a block height is missing or equal between two
       // transactions, as long as the two activities are on the same network;
       // otherwise, sort as before.


henryboldi (changes requested):
On a fresh install, I'm still seeing the same bug in the ticket's video. After disconnecting and reconnecting from both `polygon.` and `.` it then works properly.

0xDaedalus:
Are you positive that you're seeing the same bug?  The page now automatically refreshes (triggered by code in the dapp) rather than needing a manual refresh - which is the best we can do for dapps using `metamask-utils`. 

To avoid a manual refresh we would need to mimic metamask's `metamask_chainChanged` event on `wallet_switchEthereumChain` and `wallet_addEthereumChain` which is not in scope.

jagodarybacka (approved):
It works fine in my opinion, if user has selected Ethereum in wallet while connecting to Polygon in dapp then dapp asks to switch chain inside the wallet and it reloads curve site. It was not necessary to reload curve page manually

henryboldi:
> if user has selected Ethereum in wallet while connecting to Polygon in dapp then dapp asks to switch chain inside the wallet and it reloads curve site.

Even though that's how the bug is described, the actual steps in the video are:
1) Connect to https://curve.fi
2) Change networks to Polygon on Curve
3) This error
error2

henryboldi:
Reproduced:

https://user-images.githubusercontent.com/1918798/177782258-e9604dca-68fe-4f39-bc5e-9ab0764aefd8.mov

PR #1754, opened 2022-07-07
Use correct chainID when looking up permissions.

0xDaedalus:
This fixes an issue where if a user had `Ethereum` selected in the wallet and was trying to sign a transaction on `Polygon` (or vice versa) the call would fail because there would be a mismatch between the chainID of the permission and the chainID of the requested transaction.

### DIFF
--- a/background/services/provider-bridge/index.ts
+++ b/background/services/provider-bridge/index.ts
@@ -123,9 +123,10 @@ export default class ProviderBridgeService extends BaseService<Events> {
 
     const response: PortResponseEvent = { id: event.id, result: [] }
 
-    const {
-      network: { chainID },
-    } = await this.preferenceService.getSelectedAccount()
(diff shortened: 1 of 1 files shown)


jagodarybacka:
Hmm, is this correct?
- I've connected to Uniswap on Ethereum
- switched to Polygon inside Uniswap
- tried to swap, result:

image

It allowed to setup transaction but took MATIC as ETH. It would be scary to see that you want to swap 1MATIC but wallet thinks that you are swapping 1ETH :joy: But I'm not sure if this is in the scope of this task. What is the expected result in this scenario?

0xDaedalus:
This is _definitely_ not correct but also out of the scope of this issue - I'll create a follow-on

jagodarybacka (approved):
Works fine because now we at least see the confirmation popup

Issue #1082, opened 2022-03-03
WETH and USDC balances not showing up In wallet balances


Hennessy231:
### Discord Discussion Link

https://discord.com/channels/808358975287722045/944337346684391474/949055034665091102

### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

I have sent ETH to Tally wallet and the ETH balance shows up in Tally fine. I did the same transactions but instead sent USDC and WETH in separate transactions.  Sending from MM to Tally. Now the USDC and WETH balances have not shown up in Tally. All of the transactions have been confirmed in Etherscan. 

Transaction hash for USDC Send:  0xf186272baaad2ba0d7f01f16379268294ddd152135ebac1785fd7419e1065134
Transaction hash for WETH Send:  0xeb29551085fd46f41bcf0be8b79d7dc59a531aae0653a10f421b9d1730bb5c76

### What did not work?

Both the balances for USDC and WETH never appeared in Tally wallet even when they left my account in MM. I have checked Tally wallet on etherscan and the transactions have not appeared yet.

Tally Wallet address:  0x7d7b0a482509e81fe5170410ad1c1f4f418845da

greg-nagy:
Closed with #1746

PR #1755, opened 2022-07-07
🕵️‍♂️ Show me the money: Query all tokens on `tokenRefreshes` alarm

### LINKED ISSUE #1158: Wallet Balances Not Updating

greg-nagy:
The asset discovery should happen on new account addition
through the `IndexingService.retrieveTokenBalances()` or
the `ChainService.loadAssetTransfers()`.

At the moment the `ChainService.loadAssetTransfers()` is broken
for an unknown reason and this result in us not picking up any
new tokens, that were not present at the time of adding the new
address.

Querying for all the tokens should not be a costly addition to
out api usage — haven't checked though — but will provide
us a solution for now and a safety net for the future.

## Next steps
- fix the `loadAssetTransfers` flow
- implement [event listener on provider](https://docs.ethers.io/v5/api/providers/provider/#Provider--events) (and move away from polling)

## Test
- import a wallet and send an untracked asset to it from another wallet/address
- wait max 1 minute
- see the new asset appear

Closes #1158 #1082


Shadowfiend:
This is going to explicitly miss assets the user has (that we've detected through regular transaction enrichment) that aren't on Alchemy's top 100 list (see https://docs.alchemy.com/alchemy/enhanced-apis/token-api/alchemy_gettokenbalances#parameters for what happens if default tokens are used). If we need to go down this path, we need to do active assets AND top-100 assets every time.

greg-nagy:
> explicitly miss assets the user has (that we've detected through regular transaction enrichment)

Currently, this list has 0 elements because of [this bug](https://github.com/tallycash/extension/pull/1746#discussion_r915805696)

I am wondering if it's worth the resources to run 2 API calls and merge their results so we can have a fairly stable safety net.

Shadowfiend:
That bug, though absolutely infuriating, doesn't apply to any transactions initiated from our wallet (swaps or otherwise) that carry ERC20 transfer events picked up by the enrichment service. Though it does explain why we ended up having to add that enrichment pathway to get proper token balances for long tail tokens we were actually broadcasting transactions for 🙃

greg-nagy:
> On the question of whether to do the two calls or something else--I haven't gone deep enough to know for sure. It seems likely we need both if we want decent experience, until we do event/bloom-based detection of all transfers.

I love this so much :) Ty. I was thinking about the same problem, and arrived at the same conclusion!

mhluongo:
I believe this is outdated / good to close —- @greg-nagy can you confirm?

greg-nagy:
The high level idea is still solid although we will take another path on this.

PR #1758, opened 2022-07-07
Correctly display & calculate price of a network's base asset when signing a transaction on that network.

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/1757

https://user-images.githubusercontent.com/94649004/177788907-b1474f63-45ee-4cd1-9b7f-e4ead239ce04.mov


jagodarybacka (approved):
👑

PR #1759, opened 2022-07-07
🤏 Fix enrichment service's insufficient funds warning calc

henryboldi:
We should not do `(maxFeePerGas + maxPriorityFeePerGas)` because `maxPriorityFeePerGas` is included in `maxFeePerGas`.

This also fixes the dollar calc in `NetworkSettingsSelect`, although I don't believe that's used anywhere at the moment.


jagodarybacka (approved):
👍 

Was it possible to get that information somewhere from the codebase? I wonder how could I prevent that mistake, where should I check for it

PR #1760, opened 2022-07-07
Emit a chainChanged event after successfully switching chains

0xDaedalus:
### To Test

- Connect to https://wallet.polygon.technology/bridge/
- Switch back and forth between polygon and ethereum using the in-app switcher
- The dapp should detect the changes and properly switch the text to Polygon / Ethereum - respectively.
- Switching the network inside of the wallet should not switch the network in the dapp.

### DIFF
--- a/window-provider/index.ts
+++ b/window-provider/index.ts
@@ -232,7 +232,10 @@ export default class TallyWindowProvider extends EventEmitter {
           this.emit("connect", { chainId: this.chainId })
         }
 
-        if (sentMethod === "wallet_switchEthereumChain") {
+        if (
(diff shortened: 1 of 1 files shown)


greg-nagy:
I got this when I tried to do this 

CleanShot 2022-07-08 at 17 13 10

0xDaedalus:
Would you mind re-testing after clearing site data on wallet.polygon.technology?  I think that you might be in a bad state from our earlier testing

greg-nagy:
My extension was in a bad state ... sorry for that, strange though.

greg-nagy (approved):
love this solution!

PR #1762, opened 2022-07-07
Update Tally Tokenlist

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/1730

### To Test
- Reinstall Tally
- There should be more than 10 polygon swap pairs
- the ipfs uri should match the [`build token lists for main`](https://github.com/tallycash/token-list/runs/7240322358?check_suite_focus=true) output in the token-list repo.

### DIFF
--- a/background/services/preferences/db.ts
+++ b/background/services/preferences/db.ts
@@ -105,6 +105,36 @@ export class PreferenceDatabase extends Dexie {
           })
       })
 
+    // Add the new default token list
+    this.version(6)
+      .stores({
+        preferences: "++id",
+      })
+      .upgrade((tx) => {
+        return tx
+          .table("preferences")
+          .toCollection()
+          .modify((storedPreferences: Preferences) => {
+            // Get rid of old tally URL
+            const newURLs = storedPreferences.tokenLists.urls.filter(
+              (url) =>
+                !url.includes(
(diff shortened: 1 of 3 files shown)


mhluongo:
Let's say you don't reinstall Tally, and instead just upgrade in place. Will you still see the new tokens? I suspect you won't without some sort of upgrade.

0xDaedalus:
> Let's say you don't reinstall Tally, and instead just upgrade in place. Will you still see the new tokens? I suspect you won't without some sort of upgrade.

It should be refreshed after a minute as part of `handleTokenAlarm`

mhluongo:
Wallets before this PR will have the old token list URL in their default preferences, though, not the new one. I'll test to confirm

0xDaedalus:
Ah if thats the case I must have misunderstood @Shadowfiend's comment in slack - happy to test and fix if we find that an upgrade is needed

mhluongo:
Did a build from `main`, loaded an account, waited, then screenshotted the token list preferences. Then rebuilt and refreshed the extension. Screenshots below — not they both include the old URL.

Screenshot from 2022-07-08 17-48-50
Screenshot from 2022-07-08 17-44-19

mhluongo (approved):
Very nice 👌

PR #1774, opened 2022-07-08
Fix meshswap dapp connection

### LINKED ISSUE #1761: Polygon | Meshswap - Can't connect wallet (us or them?)
### **Build Info/PR/Commit:**
https://github.com/tallycash/extension/tree/polygon-qa-3

### **Reproduction Steps:**

1. Attempt to connect Tally using Meta Mask (no inject or Tally option)

### **Expected Result:**
Connected! 

### **Actual Result:**
image

greg-nagy:
- adding isWeb3 flag to `window-provider`
- saving the chainId on the `window-provider` in case of `wallet_switchEthereumChain` || `wallet_addEthereumChain`

Note: on the initial connection, a `wrong network` error will be shown. 
Fixing this would require white-labeling meshswap which is probably not a good idea. 
This behavior is identical with MM. 

Closes #1761

### DIFF
--- a/window-provider/index.ts
+++ b/window-provider/index.ts
@@ -240,6 +240,9 @@ export default class TallyWindowProvider extends EventEmitter {
         ) {
           // null result indicates successful chain change https://eips.ethereum.org/EIPS/eip-3326#specification
           if (result === null) {
+            this.chainId = (
+              sendData.request.params[0] as { chainId: string }
+            ).chainId
             this.emit(
               "chainChanged",
               (sendData.request.params[0] as { chainId: string }).chainId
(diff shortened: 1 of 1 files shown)


window-provider/index.ts
```diff
@@ -33,6 +33,8 @@ export default class TallyWindowProvider extends EventEmitter {
 
   isMetaMask = false
 
+  isWeb3 = true
```
henryboldi, line 36: > for some reason meshswap checks for this
flag before willing to work with any injected
provider

Curious!

henryboldi (approved):
Looks good 💪

PR #1775, opened 2022-07-08
👻 Remove gas settings before sign step

henryboldi:
Swap
Send

⚠️ Would like to add an explanation behind this choice to the PR description, but feel free to review now if we need it quickly

### DIFF
--- a/ui/_locales/en/messages.json
+++ b/ui/_locales/en/messages.json
@@ -25,8 +25,7 @@
     "sendAsset": "Send Asset",
     "assetAmount": "Asset / Amount",
     "sendTo": "Send To:",
-    "estimatedFee": "Estimated network fee",
-    "sendButton": "Send",
+    "sendButton": "Continue",
     "receiveAddress": "Receive address",
     "receiveAddressDesc": "Only send Ethereum Mainnet compatible assets to this address."
   },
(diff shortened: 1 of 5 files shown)


ui/pages/Send.tsx
```diff
@@ -71,17 +63,14 @@ export default function Send(): ReactElement {
     string | undefined
   >(undefined)
   const [amount, setAmount] = useState("")
-  const [gasLimit, setGasLimit] = useState<bigint | undefined>(undefined)
+  const [gasLimit] = useState<bigint | undefined>(undefined)
```
0xDaedalus, line 66: I think we can remove this - its just undefined forever right?

ui/_locales/en/messages.json
```diff
@@ -25,8 +25,7 @@
     "sendAsset": "Send Asset",
     "assetAmount": "Asset / Amount",
     "sendTo": "Send To:",
-    "estimatedFee": "Estimated network fee",
-    "sendButton": "Send",
+    "sendButton": "Continue",
```
0xDaedalus, line 28: ❤️

ui/pages/Send.tsx
henryboldi: Done!

0xDaedalus (approved):
Looks great - swaps work - rock & roll

PR #1787, opened 2022-07-11
Revert "Fix asset transfers"

greg-nagy:
Reverts tallycash/extension#1746


Shadowfiend:
I propose we use #1788 instead.

Issue #1021, opened 2022-02-23
Infer metadata from tokens found in local transactions


mhluongo:

VladUXUI:
Going to pick the second bullet up tomorrow. I have a couple of solutions for the 4 in mind but i don't want to rush it.

Can you say a bit more about "Add a way to see every asset on Etherscan"?

mhluongo:
Yep! In particular, for an ERC-20 I want to see my balance + see what the contract is to make sure it's the token I want.

VladUXUI:
Hopefully i understood this correctly.
Basically you will see a warning that something is up with this asset. On the click of the asset we will let you know the name/contract address and what TX this is coming from. This way you have several ways to verify.

This is for the first implementation, with two entry points to opening the slide-up. From asset page and asset details page.

![image](https://user-images.githubusercontent.com/80846886/157210147-ee08fe73-dc76-41d5-aa8e-fcf48428d00b.png)

VladUXUI:
As a second implementation for _Eventually add a way for users to say "trust this token" or "ignore token" on the asset page for all auto-inferred tokens_ we would only need to add two buttons to the slide-up. (This version requires a bit of cleaning, but sharing for visibility)

![image](https://user-images.githubusercontent.com/80846886/157210421-595dcb42-ee4c-4b1b-81ed-67b65986b014.png)

mhluongo:
Love love love this @VladUXUI.

mhluongo:
I think adding a link out to EtherScan for every asset page is the only thing this is missing design-wise.

mr-michael:
+ Toggle: "Display untrusted tokens" (default OFF)

puppycodes:
To breakdown into newer tasks

Shadowfiend:
Shipped in #3424 .

PR #1788, opened 2022-07-11
Oil Filter: Add baseline trust to detected tokens, ignore untrusted

Shadowfiend:
New tokens are only detected through enriched EVM transaction ERC20
transfer events. Currently this means any transaction we detect with
tokens will add that token, as long as the sender or recipient of the
token are tracked.

However, some scams rely on sending fake tokens that look legit to a
user. Auto-detecting these tokens and presenting them in the UI is
dangerous without flagging for the user that the tokens may be
untrusted.

This commit introduces a concept of baseline trust when deciding to
check for balances in ERC20 transfers. Baseline trust is currently
defined as a token that was SENT BY an address that is being tracked
(i.e., an account the user has added to the wallet) OR that is already
known (typically via a token list).

Currently tokens without baseline trust are ignored (i.e. they are
neither tracked nor balance-checked), but this is the core function that
will allow presenting the user with choices about these tokens.

See #1021.


0xDaedalus (approved):
Tested & passing as part of https://github.com/tallycash/extension/pull/1800

0xDaedalus (approved):
🚢

PR #1790, opened 2022-07-11
Fix on Sign Data page how messages are displayed

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1653

Fix on the Sign Data page how messages are displayed

image


Shadowfiend:
Cutting data off like this feels like a bad solution. The point of presenting the data is to give the user an opportunity to review.

In the short term this is probably better than what was before, but I wonder if we can come up with a better strategy for displaying longer data? Perhaps an expand button or similar.

jagodarybacka:
@Shadowfiend I've added browser's title tooltip to this field to allow users to see the whole thing but I agree that expand button would be the best. @VladUXUI what do you think?

VladUXUI:
Expand or copy should work, but not a priority right now. The tooltip you added to see the entire data should be enough for now.
We can circle back to it later

0xDaedalus (approved):
:tada:

PR #1793, opened 2022-07-12
fix the default wallet toggle functionality

greg-nagy:
When caching the proxy object we also need to cache the current provider at the time of
creating a cache. This way we can recognise when the underlying provider is changed
and we can recreate the proxy object / invalidate the cache accordingly.


0xDaedalus:
Love it - I was pretty sure there was something along these lines I was missing in #1792.  Appreciate You.

0xDaedalus (approved):

PR #1795, opened 2022-07-12
Fix window provider dapp chainid

greg-nagy:


background/services/provider-bridge/index.ts
```diff
@@ -191,13 +185,7 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
 
       const { address: accountAddress } = selectedAccount
-      let { network } = selectedAccount
-
-      // @TODO Quickswaps expects to be able to switch network to polygon before
-      // requesting accounts so it always assumes its connecting to Polygon
-      if (origin.includes("quickswap.exchange")) {
-        network = POLYGON
-      }
+      const { network } = selectedAccount
```
0xDaedalus, line 188: We need a TODO here because if we have `Ethereum` selected in the wallet but are requesting permission to an app that has switched its chain to `Polygon` - the wallet will ask for permission on the `Ethereum` chain.

This is not a problem yet because of our current all-or-none-per-origin permission model but we'll run into an issue here when we move away from that model or if we ever want to display information about which network a specific permission request corresponds to.

0xDaedalus: Actually - it might be as simple as using 

```
await this.internalEthereumProviderService.getActiveOrDefaultNetwork(
        origin
      )
```

to get the network - but I haven't tested this

window-provider/index.ts
```diff
@@ -78,6 +74,10 @@ export default class TallyWindowProvider extends EventEmitter {
       }
 
       if (isTallyConfigPayload(result)) {
+        if (result.chainId && result.chainId !== this.chainId) {
```
0xDaedalus, line 77: ❤️ 🧠

window-provider/index.ts
```diff
@@ -284,6 +278,12 @@ export default class TallyWindowProvider extends EventEmitter {
     })
   }
 
+  handleChainIdChange(chainId: string) {
```
0xDaedalus, line 281:
```suggestion
  handleChainIdChange(chainId: string): void {
```

background/services/provider-bridge/index.ts
greg-nagy: same idea, a bit different implementation ... 🙏

greg-nagy:
We probably need to fire an `eth_requestAccounts` before the `wallet_addEthereumChain` in the window provider, because quickswap reloads as soon as it gets a response for the `wallet_addEthereumChain` request. (MM does internal req queing it seems, and if the 2 are fired rapidly they grab the user input one after the other before they send a response to the dapp)

When this is done we can also remove the [exception for the `wallet_addEthereumChain` in the `ProviderBridgeService`](https://github.com/tallycash/extension/blob/a3301f32452b69f1c12947dd073c7ee8612efcdb/background/services/provider-bridge/index.ts#L174-L181)

This is the last piece of the puzzle imo

0xDaedalus:
> We need to fire an eth_requestAccounts before the wallet_addEthereumChain in the window provider,

You want to do this for all dapps?  Do you think that as an alternative it may make sense to accept this tech debt and implement proper request queueing down the line?

greg-nagy:
> You want to do this for all dapps? Do you think that as an alternative it may make sense to accept this tech debt and implement proper request queueing down the line?

This is not something that we should need to implement ... Otoh I think you are correct here and we will probably have to. 
I think quickswap behavior can be good enough for now. 

Anyways this PR is ready for merge, the rest can be done in follow up PRs

PR #1800, opened 2022-07-12
🐾 Update `isTrackingAsset` to match intent

henryboldi:
We track more than custom assets, but custom assets were the only assets being
checked by `isTrackingAsset`. Although all custom assets are tracked, this can
exclude non-custom tracked assets. This PR updates `isTrackingAsset` to
match the expected behavior of returning `true` on all tracked assets.

### DIFF
--- a/background/services/indexing/db.ts
+++ b/background/services/indexing/db.ts
@@ -219,6 +219,16 @@ export class IndexingDatabase extends Dexie {
       .first()
   }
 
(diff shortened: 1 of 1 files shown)


0xDaedalus (approved):
Tested sending DAI to a brand new wallet and seems to be working as intended 👍 

Also spam tokens still don't show up 🎉

PR #1804, opened 2022-07-12
Fix Double Dapp Connect Bug

0xDaedalus:
### To Test

Connect wallet to uniswap 
You should be able to connect without needing to refresh

### DIFF
--- a/background/services/provider-bridge/index.ts
+++ b/background/services/provider-bridge/index.ts
@@ -186,12 +186,14 @@ export default class ProviderBridgeService extends BaseService<Events> {
 
       const { address: accountAddress } = selectedAccount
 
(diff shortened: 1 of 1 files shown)


0xDaedalus:
Sure - the symptom was that connecting to any dapp would seem to fail initially (dapps would display "Connection Failed") and then a refresh of the page would load the dapp with the wallet properly connected.

greg-nagy:
ty. This feels like an issue with how the permission is verified/propagated somewhere at the end of the code path

SJQuality:
I was seeing this issue on almost every dApp.. the 0.14.0 build doesn't have that issue.. now I know why!

PR #1806, opened 2022-07-13
Fix read-only accounts

### LINKED ISSUE #1799: Read-only marked as a Tally Ho account
read-only

jagodarybacka:
Resolves #1799

### What
Fix problems where read-only accounts were allowed to try to sign transactions and other regressions caused by #1724

### Testing

Please test this side by side with our last release to make sure read-only accounts are handled the same way they were before.

image


Shadowfiend (commented):
Yikes, lot of misses on my part 🙃

0xDaedalus (approved):
👨‍🍳

PR #1811, opened 2022-07-13
Add translation JSON files to prettier ignore.

0xDaedalus:
This will make merging translations from `weblate` significantly smoother and should allow us to merge https://github.com/tallycash/extension/pull/1756 after merging this change into the branch.


jagodarybacka (approved):
I wanted to avoid that but it's ok I guess, these files will be mostly edited by weblate with exception of English translation as this is something we will be updating manually

PR #1814, opened 2022-07-13
Preference Service Tests

0xDaedalus:
This PR lays the groundwork for testing codepaths that rely on indexedDB - and adds some basic preferenceService tests to serve as an example of how we may want to test a service going forward.

### DIFF
--- a/background/services/name/resolvers/address-book.ts
+++ b/background/services/name/resolvers/address-book.ts
@@ -1,3 +1,4 @@
+import { AddressOnNetwork, NameOnNetwork } from "../../../accounts"
 import PreferenceService from "../../preferences"
 import { NameResolver } from "../name-resolver"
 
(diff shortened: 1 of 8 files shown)


greg-nagy (approved):
CleanShot 2022-07-14 at 16 30 25

that's all I can say ... other than spectacular 🚀

background/services/name/resolvers/address-book.ts
```diff
@@ -17,12 +18,20 @@ export default function addressBookResolverFor(
...
     },
 
-    lookUpAddressForName:
-      preferenceService.lookUpAddressForName.bind(preferenceService),
+    async lookUpAddressForName(nameOnNetwork: NameOnNetwork) {
+      return preferenceService.lookUpAddressForName.call(
+        preferenceService,
+        nameOnNetwork
+      )
+    },
```
Shadowfiend, line 26: If you're making this a full on method you don't need to use `call`. `preferenceService.lookUpAddressForName(nameOnNetwork)` will bind `this` without extra ceremony.

background/services/preferences/index.ts
```diff
@@ -102,19 +102,19 @@ export default class PreferenceService extends BaseService<Events> {
     this.emitter.emit("addressBookEntryModified", newEntry)
   }
 
-  async lookUpAddressForName({
+  lookUpAddressForName({
     name,
     network,
-  }: NameOnNetwork): Promise<AddressOnNetwork | undefined> {
+  }: NameOnNetwork): AddressOnNetwork | undefined {
```
Shadowfiend, line 108: Thisss was written as a promise with the expectation that the address book was going to move to IndexedDB fwiw.

Shadowfiend: (I realize I did not say that, or write it, and therefore there was no reason to know it lol. Just clarifying.)

greg-nagy: This is part of the Indiana Jones vibes :)

PR #1824, opened 2022-07-14
Remove `call` code cruft

0xDaedalus:
No need for a `call` here since changing `lookUpAddressForName` and `lookUpNameForAddress` to functions binds `this` for us.


jagodarybacka (approved):
👑

PR #1826, opened 2022-07-15
🖌 Fix sign message text from dApps

jagodarybacka:
Fixing displayed message - https://github.com/tallycash/extension/issues/1763

### What
Change text input from dapps to hex before assigning to `rawSigningData`.

### Testing

Check how signing message is looking:
- at sushiswap with sushiguard on - try to swap (will fail but message should show hex value instead of weird symbols)
- at gitcoin passport - sign in, message should be displayed as a long text
- other apps that needs to sign messages.... 🤔

### DIFF
--- a/ui/components/Signing/index.tsx
+++ b/ui/components/Signing/index.tsx
@@ -1,0 +1,23 @@
+import { SignOperationType } from "@tallyho/tally-background/redux-slices/signing"
+import { AccountSigner } from "@tallyho/tally-background/services/signing"
+import React, { ReactElement } from "react"
+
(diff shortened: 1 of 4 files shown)


0xDaedalus:
This seems like a fantastic candidate for a unit test :)

jagodarybacka:
I agree we can write tests for this thing but I'm not sure if we want to write tests ad hoc, without planning it. To me it seems like a way to miss important places and spend time on less important ones.

greg-nagy (approved):
let's 🚢

PR #1828, opened 2022-07-15
Signing Service Test Environment setup & Tests

0xDaedalus:
Some more foundational test stuff laid down in this PR coupled with a small illustrative set of tests.  Nothing set in stone just experimenting here

Highlights are
- Factories for creating services so devs can easily create and reuse them.
- Examples of how to create a service and test its dependencies on another service.
- Some more progress in mocking globals in `jest` so we can compile our code in our test environment.
- a new `setupJest` script to accomplish the above. (lets think about name and location of this file).


background/services/signing/tests/index.unit.test.ts
```diff
@@ -0,0 +1,71 @@
+import "fake-indexeddb/auto"
+import sinon from "sinon"
+import {
+  createKeyringService,
+  createLedgerService,
+  createSigningService,
+} from "../../../tests/factories"
+import SigningService from "../index"
+
+describe.skip("Signing Service Unit", () => {
```
jagodarybacka, line 10: Why are tests skipped? Are they not working yet?

background/services/signing/tests/index.unit.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  let signingService: SigningService
+  const sandbox = sinon.createSandbox()
+
+  beforeEach(async () => {
+    signingService = await createSigningService()
```
greg-nagy, line 15: I think this should be removed since it's not used.

greg-nagy (dismissed):
I love the approach. 

At first the factories didn't make much sense, but reading through the code I think this approach is useful. 

Let's fix the cleanup then it's good to merge.

background/services/signing/tests/index.unit.test.ts
0xDaedalus: We do actually use it on line 65 - but I agree that with how few tests there are in this file (for now!) we can use a beforeEach if and when we find we're manually creating and signing too much

background/services/signing/tests/index.unit.test.ts
0xDaedalus: 🧠 💨

background/services/signing/tests/index.unit.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      expect(deriveAddressStub.called).toBe(true)
+    })
+
+    it("should error when trying to derive from a read-only account", () => {
+      expect(
+        signingService.deriveAddress({
+          type: "read-only",
+        })
+      ).rejects.toBeTruthy()
```
mhluongo, line 70: Love a simple test that covers a big ass future bug :)

mhluongo (approved):
I'm really happy with this start!

PR #1829, opened 2022-07-15
Monster Paths: Allow for paths that don't have page entries

Shadowfiend:
persistOnClose lookup assumes the normalized path will always have a
page entry, but there's nothing about the normalization process that
guarantees this. Allow for undefineds to deal with this problem.


Shadowfiend:
I believe this should fix the issue reported at https://discord.com/channels/808358975287722045/888500174685614090/997580044455129109 .

0xDaedalus (approved):

PR #1837, opened 2022-07-19
📒Create a memoized version of the mergeAssets

greg-nagy:
this function call is extremply expensive cpu wise
and it's also a pure function in the practical sense
so we can optimize by running it only if the input
list changes

we use a shortcut for calculating the cache key
for similarity definition. We are not checking for
all the contract addresses to be the same, but
we use the number of all the elements in the list
as an indicator that the incoming lists are the same
This is a reasonable compromise, given we are dealing
with token lists here.

### DIFF
--- a/background/lib/token-lists.ts
+++ b/background/lib/token-lists.ts
@@ -1,5 +1,6 @@
 import { TokenList } from "@uniswap/token-lists"
 
+import { memoize } from "lodash"
 import {
   FungibleAsset,
   SmartContractFungibleAsset,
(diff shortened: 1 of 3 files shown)


background/lib/token-lists.ts
```diff
@@ -145,6 +146,10 @@ export function mergeAssets<T extends FungibleAsset>(
   )
 }
 
+export const memoizedMergeAssets = memoize(mergeAssets, (...assetLists) => {
```
mhluongo, line 149: This deserves a comment... it feels like a decision that could work out and be fine forever, or could be a source of unknown and surprising bugs later.

mhluongo (approved):
Tested and it works well!

PR #1838, opened 2022-07-25
💬Memoize 2/n: use the memoized version of `mergeAssets` in `networkAssetsFromLists` 

greg-nagy:
networkAssetsFromLists is using mergeAssets
and it's called every time getCachedAssets is called

so it also worth memoizing

the effect of this optimization is not that significant
as it was for the `getCachedAssets` but this still
makes this to the top of the list in terms of cpu usage


mhluongo (dismissed):
Tested and looks good!

jagodarybacka (approved):
Not sure why it is not auto merging, here goes another approve ✅

PR #1839, opened 2022-07-25
Fix incorrect balance on Ledger onboarding 

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1827

### What

- Fix how we save balances on unimported Ledger accounts so they are saved correctly per chain.
See the issue above for reproduction steps

- Fix link to scan website to match with currently selected network


background/redux-slices/ledger.ts
```diff
@@ -175,14 +176,21 @@ const ledgerSlice = createSlice({
...
+          network: EVMNetwork
+        }
+      }
     ) => {
       const device = immerState.devices[deviceID]
       if (!device) return
       const account = device.accounts[path]
       if (!account) return
-      if (account.balance === null) account.balance = balance
+      account.balance[network.chainID] ??= balance
```
mhluongo, line 193: Doesn't `??` check for undefined, while before we were strictly checking for `null`? Would `||=` make sense here?

jagodarybacka: But there should be no possibility of `null` here if I'm being correct - after migration only `undefined`

ui/pages/Ledger/LedgerImportAccounts.tsx
```diff
@@ -173,7 +174,9 @@ function LedgerAccountList({
                         onClick={() => {
                           window
                             .open(
-                              `https://etherscan.io/address/${address}`,
+                              `${
+                                scanWebsite[selectedNetwork.chainID].url
```
mhluongo, line 178: This would make for a good util function — we're doing it a couple times in the UI

mhluongo: Maybe you caught them all, but I'd thought we used `/address/` on EtherScan in other spots.

mhluongo:
Screenshot from 2022-07-26 15-51-14

I got this error testing

1. I added a Ledger with account A on Ethereum
2. I rebuilt the extension with this PR to test the migration
3. I refreshed the extension, and everything looked good
4. I added the same Ledger with account A on Polygon

Everything looked good, but wanted to report the above as it looks like a deeper backend issue around cross-chain Ledger stuff.

mhluongo (approved):
The code looks good, the migration works, and everything was smooth! I found one issue in the logs that I believe cropped up outside this PR, but it shouldn't block merge.

ui/pages/Ledger/LedgerImportAccounts.tsx
jagodarybacka: Yeah, we can refactor it at some point to a function

PR #1840, opened 2022-07-25
:eye: Trust but verify

mhluongo:
Add a warning in the wallet asset list when interacting with assets whose metadata has been inferred from transaction history.

Screenshot from 2022-08-06 21-30-32
Screenshot from 2022-08-06 22-20-45

To test

- [ ] Load a read-only account with tokens that aren't on our lists
- [ ] As historical activities are loaded, additional token balances should be discovered
- [ ] Tokens that aren't on our lists should show a warning

Refs #1021, #1893

Latest build: [extension-builds-1840](https://github.com/tallyhowallet/extension/suites/10339714339/artifacts/509199402) (as of Thu, 12 Jan 2023 19:06:15 GMT).


mhluongo:
@Shadowfiend would love your thoughts on how to make this consistent with your plans in https://github.com/tallycash/extension/pull/1788

I'd like to add an "internal" token list for each user in a future PR, and use that to track custom assets going forward. But in the meantime, this PR should be enough to show discovered assets and warn people of scams... except they won't currently show up at all with after this change

https://github.com/tallycash/extension/pull/1788/files#diff-b0e4a7036b74ef74bb2061264a7470aa9dc941d5bb0b326556464bb28480fc80L328

Shadowfiend:
We already have the internal list--that is what the custom asset stuff is I believe. This todo explains a reasonable path to adding the relevant metadata there:

https://github.com/tallycash/extension/blob/6a69598bdab150ee1c0cebc3b81e05d6369a0fa6/background/services/indexing/index.ts#L294-L299

greg-nagy:
I think we have started using custom assets for assets with 0 balance at some point.

mhluongo:
Taking a different tack: this PR shows a warning for any asset in the asset list that isn't on a token list.

It appears that these assets are no longer auto-discovered from the asset transfer call as of #1788... given that we have a warning now, should I [revert part of that PR](https://github.com/tallycash/extension/pull/1788/files#diff-b0e4a7036b74ef74bb2061264a7470aa9dc941d5bb0b326556464bb28480fc80L328) here?

greg-nagy:
Yes, I think this is the way. 

Nit: I wouldn't revert that change but reintroduce a modified version [according to this comment](https://github.com/tallycash/extension/blob/main/background/services/indexing/index.ts#L294-L299)

Shadowfiend:
I think two paths we can take here and it's ultimately a product decision. We can nuke the baseline trust concept, always display the warning for non-token list assets without a way to mark them trusted, then separately add a warning dismissal path that adds the token to the custom token list. Or we can land this, invisible because those assets are never added, and then fill out the trust concept as per the TODO and attach the warnings to that. 

Either way works for me.

mhluongo:
I think the former is the shorter path, but will confirm.

Getting to more token balances in the wallet sooner is going to be a big win for users.

mhluongo:
@mr-michael calling your attention here: merging this PR means we'll pick up all / nearly all token balances at the risk of including scam tokens, though they'll have clear and obvious warnings.

My thought is we merge this, then do the next bit (allowing tokens to be flagged as trusted or ignored) in a future PR / release based on other priorities. I think that's the right call because it immediately improves the UX for **new** wallets. It can degrade UX a bit for older wallets like mine with a bunch of spam tokens, but most users seem to be onboarding with new accounts.

mhluongo:
@Shadowfiend @greg-nagy I opened https://github.com/tallycash/extension/issues/1893 to continue the asset trust discussion and unify the two approaches.

ui/components/Wallet/AssetListItem/CommonAssetListItem.tsx
```diff
@@ -6,16 +6,30 @@ import SharedLoadingSpinner from "../../Shared/SharedLoadingSpinner"
...
+  } = props
   const isMissingLocalizedUserValue =
     typeof assetAmount.localizedMainCurrencyAmount === "undefined"
 
+  // NB: non-base assets that don't have any token lists are considered
+  // untrusted. Reifying base assets clearly will improve this check down the
+  // road. Eventually, assets can be flagged as trusted by adding them to an
+  // "internal" token list that users can export and share.
+  const numTokenLists = assetAmount?.asset?.metadata?.tokenLists?.length ?? 0
+  const baseAsset = !("homeNetwork" in assetAmount?.asset)
```
jagodarybacka, line 31: Have you tested it with both ETH and MATIC? I've seen cases where `homeNetwork` was missing in base asset 😞

mhluongo: I have! I'd love another look though if you have a minute.

In the case of `MATIC`, there's an ERC-20 with a `homeNetwork` and a base asset without one... could that be the confusion @jagodarybacka?

jagodarybacka: hmmmm, yeah, that is the mistake I made probably 😂

0xDaedalus (commented):
Code looks good to me - waiting on product input before merging

greg-nagy (approved):
lookin' good, let's ship

ui/components/Wallet/AssetWarningSlideUp.tsx
```diff
@@ -0,0 +1,124 @@
...
+import SharedSlideUpMenu from "../Shared/SharedSlideUpMenu"
+import SharedBanner from "../Shared/SharedBanner"
+import SharedButton from "../Shared/SharedButton"
+import SharedAddress from "../Shared/SharedAddress"
+
+type TitledSlideUpProps = Parameters<typeof SharedSlideUpMenu>[0] & {
+  title: string
+}
+
+function TitledSlideUpMenu(props: TitledSlideUpProps): ReactElement {
```
jagodarybacka, line 14: I think either we can include it in the `AssetWarningSlideUp` or we can add an option of having a title to `SharedSlideUpMenu`. Or make it a separate component?

jagodarybacka (changes requested):
- let's add translations
- let's find a better position for this icon, maybe on the right side with send and swap icons? cc @VladUXUI 
image

ui/components/Wallet/AssetWarningSlideUp.tsx
```diff
@@ -0,0 +1,124 @@
...
+          <div className="right">
+            <SharedAddress
+              address={
+                asset && "contractAddress" in asset ? asset.contractAddress : ""
+              }
+            />
+          </div>
+        </li>
+      </ul>
+      <SharedButton size="medium" type="secondary" id="close_asset_warning">
```
jagodarybacka, line 119: There is no "onClick" on this button

jagodarybacka:
Let's add a feature flag here too

ui/components/Wallet/AssetWarningSlideUp.tsx
jagodarybacka: Done in a different PR by @kkosiorowska, will be unified - depends on which PR will be merged first

background/services/indexing/index.ts
```diff
@@ -412,6 +412,24 @@ export default class IndexingService extends BaseService<Events> {
...
+    // listen for assetTransfers, and if we find them, track those tokens
+    // TODO update for NFTs
+    this.chainService.emitter.on(
+      "assetTransfers",
+      async ({ addressNetwork, assetTransfers }) => {
+        assetTransfers.forEach((transfer) => {
+          const fungibleAsset = transfer.assetAmount
+            .asset as SmartContractFungibleAsset
+          if (fungibleAsset.contractAddress && fungibleAsset.decimals) {
+            this.addTokenToTrackByContract(
```
jagodarybacka, line 424: I've tested it a little bit more and we have a problem here because even for smaller wallets this method is causing WAY too many requests, it is causing UNS to reject requests, it is throwing an "insufficient resources" chrome error.

To reproduce:
- add a console log here 
- add a wallet, (tested on `berrry.eth` and `mhluongo.eth`) open Polygon network 
- in the background script console you will see the avalanche of errors after ~1min or less

jagodarybacka (changes requested):
This needs more discussion about the approach as the current solution is draining the network resources

0xDaedalus:
Found the issue (took forever!), fixed the issue, added regression tests.  Should be ready for another look @jagodarybacka ✅

jagodarybacka (changes requested):
It's no longer draining resources but I see there is some problem with base assets as they are sometimes marked as unverified, noticed it for ETH on Optimism and Matic on Polygon 
image

0xDaedalus:
Should be good for another look - note that AVAX on polygon is displayed as untrusted because it is indeed not on the official polygon tokenlist that we use.  Otherwise, network base assets erroneously showing as untrusted should be fixed.

jagodarybacka:
Hmm 😞 
image

jagodarybacka:
For reference - reproduction steps:

https://user-images.githubusercontent.com/20949277/212067677-8278e920-e1bf-4464-94e8-1ca59df03d10.mov

After account is fully loaded it seems to be fixed:
image

Problem here is because `MATIC` token doesn't contain `coinType` property and because of it it is not considered base asset.

PR #1843, opened 2022-07-25
fix for github CI memory issue

greg-nagy:
github actions started to kill the build process
with error message of 137 which signals out of
memory

Couldn't find any way to set it manually without
self-hosting the builder, so I went with a shortcut
of building only for browsers that we are currently
shipping for.

This decreased the memory usage to it's third
8394096640 > 3120578560

to measure on osx: /usr/bin/time -l yarn build


greg-nagy:
Cutting a corner here, merging into a release branch. 

It's probably ok because it's a CI fix, and it's easier to test it here than on main.
(And we don't know yet whether this will be enough or not)

PR #1844, opened 2022-07-26
Put balance reloader button behind flag

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1816

### What

The balance reloader button can be now toggled with the feature flag `USE_BALANCE_RELOADER`
image


mhluongo (approved):
LFG :rocket:

PR #1850, opened 2022-07-27
🌠 NFTs prototype part one

henryboldi:
Wallet
Overview NFT
Preview


background/redux-slices/nfts.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const NFTsSlice = createSlice({
+  name: "nfts",
+  initialState,
+  reducers: {
+    updateNFTs: (immerState, { payload: { address, NFTs, network } }) => {
+      const normalizedAddress = address
+      immerState.evm[network.chainID] ??= {}
+      immerState.evm[network.chainID][normalizedAddress] ??= []
+      immerState.evm[network.chainID][normalizedAddress] = NFTs
```
jagodarybacka, line 30: it's an override anyway so we probably can commit `??= []` and go straight to assignment. But I wonder if we want to override whole thing every time

background/constants/networks.ts
```diff
@@ -69,3 +69,15 @@ export const FORK: EVMNetwork = {
   chainID: process.env.MAINNET_FORK_CHAIN_ID ?? "1337",
   family: "EVM",
 }
+
+export const NETWORK_BY_CHAIN_ID = {
```
0xDaedalus, line 73: ❤️

background/redux-slices/nfts.ts
```diff
@@ -0,0 +1,78 @@
...
+    tokenId: string
+  }
+  contract: { address: string }
+  title: string
+  chainID?: number
+}
+
+export type NFTsState = {
+  evm: {
+    [chainID: string]: {
```
0xDaedalus, line 17: Want to inject some food-for-thought here for us all to think about.  I would the structure of evm => chainId => address => item very cumbersome to work with and requiring somewhat opaque selectors when implementing cross-chain dapp permissions.  Does that resonate with you at all?  I wonder if evm => address => chainId => item would be better (considering our whole application is address-centric).

background/redux-slices/nfts.ts
```diff
@@ -0,0 +1,78 @@
...
+      immerState.evm[network.chainID][normalizedAddress] = NFTs
+    },
+  },
+})
+
+export const { updateNFTs } = NFTsSlice.actions
+
+export default NFTsSlice.reducer
+
+async function fetchNFTs(address: string, currentNetwork: EVMNetwork) {
```
0xDaedalus, line 44: Do you think that `fetchNFTsByNetwork` could be a more explicit name here?

background/redux-slices/nfts.ts
```diff
@@ -0,0 +1,78 @@
...
+      address: string
+      currentNetwork: EVMNetwork
+    },
+    { dispatch }
+  ) => {
+    const { address, currentNetwork } = payload
+    const ownedNFTs = await fetchNFTs(address, currentNetwork)
+
+    await dispatch(
+      NFTsSlice.actions.updateNFTs({
```
0xDaedalus, line 71: I think we can probably get rid of `NFTsSlice.actions.` here (just because I think thats the pattern we tend to favor in the other reducers)

background/redux-slices/selectors/nftsSelectors.ts
```diff
@@ -0,0 +1,33 @@
+import { createSelector } from "@reduxjs/toolkit"
+import { RootState } from ".."
+import { NFTsState } from "../nfts"
+
+// Adds chainID to each NFT for convenience in frontend
+const selectNFTs = createSelector(
```
0xDaedalus, line 6: We might be well served by extending our type for NFTs and adding the `chainID` inside of the `fetchNFTs` function since at that point we know what network we're fetching them for.  If we do that we can just set-it-and-forget-it

0xDaedalus (commented):
Took a quick look at the code - none of my comments are blocking - this is so exciting!!! 🎉 🎉

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -440,17 +440,29 @@ export const selectCurrentAccountTotal = createSelector(
...
+  ]
 )
 
-export const getNetworkCount = createSelector(
+export const getAddressCount = createSelector(
+  getAllAddresses,
+  (allAddresses) => allAddresses.length
+)
+
+export const getAllNetworks = createSelector(
```
0xDaedalus, line 459: I'm a bit concerned about performance here and in `getAllAddresses` - if you find that there are some you might want to consider passing a custom memoization function to these selectors.

jagodarybacka: ah yes, that's the thing I was struggling with today - but I haven't found any good solutions, do you have some example how to make`createSelector` remember that array?

background/redux-slices/nfts.ts
henryboldi: 100%

background/redux-slices/selectors/nftsSelectors.ts
henryboldi: I'd be nervous about storing `chainID` in two places. What if the `chainID` on the NFT gets out of sync with the   `chainID` used as the key in the store?

background/redux-slices/nfts.ts
greg-nagy: I feel that we are moving back and forth between chainID or address focused. It feels to me that they craete a unique key together and it's probably not a good idea to put one above the other. See above comment.

background/redux-slices/nfts.ts
```diff
@@ -0,0 +1,78 @@
+import { createSlice } from "@reduxjs/toolkit"
+import { createBackgroundAsyncThunk } from "./utils"
+import { EVMNetwork } from "../networks"
+
+export type NFTItem = {
+  media: { gateway?: string }[]
```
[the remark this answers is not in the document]
greg-nagy: Partially the reason for this comment is that I think we should start shifting our mindset when we think about redux from something defined by "this is what's in the background" into "hey redux is actually a view model, let UI define the shape it needs"

ui/components/NFTs/NFTsOverview.tsx
```diff
@@ -0,0 +1,44 @@
...
+import NFTsEmpty from "./NFTsEmpty"
+
+export default function NFTsOverview(): ReactElement {
+  const NFTs = useBackgroundSelector(selectNFTs)
+  const allNetworks = useBackgroundSelector(getAllNetworks)
+  const allAddresses = useBackgroundSelector(getAllAddresses)
+  const dispatch = useBackgroundDispatch()
+
+  useEffect(() => {
+    allAddresses.forEach((address) =>
```
greg-nagy, line 19: I wanted alchemy to support batched requests sooooo much. But it doesn't ... :(

ui/components/NFTs/NFTsWallet.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { ReactElement, useEffect } from "react"
```
greg-nagy, line 1: Love the separation ❤️

greg-nagy (changes requested):
Wonderful work ✨

The api error handling is an absolute must with user notification as well. (It can create really hard to catch bugs later and we should get error handling into our DNA) 

We should make a quick decision about the slice data structure as well. I can accept the current shape if that's the consensus, but feel that the proposed version would serve us better even at the cost of refactoring now.

henryboldi:
> The api error handling is an absolute must with user notification as well. 

@greg-nagy, We're not flipping the feature flag yet, so I don't believe everything needs to be completed in this "part one" PR

greg-nagy:
>> The api error handling is an absolute must with user notification as well.
>  We're not flipping the feature flag yet, so I don't believe everything needs to be completed in this "part one" PR

Yes, you are correct and this thought changes the data structure issue from a blocker to a TBD and follow up PR. 

But I feel strongly about the api error handling and user notification. So much so, that I would like us to get to a point that if we write api call and not handle the errors we feel something is off. (The user notification can be a console.log, or simply an error message [like for the swaps](https://github.com/tallycash/extension/blob/860ce3a61af9f753a0bb64d74678f325c3c3152b/background/redux-slices/0x-swap.ts#L283))

background/redux-slices/nfts.ts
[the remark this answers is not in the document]
henryboldi: Here are a few reasons this change shouldn't be blocking:

1. The data structure of NFTs here is simply following the pattern in the accounts slice. Changing the structure will create an inconsistency.

2. Until we stop fetching NFTs every time the tab is opened, there would be no need for a migration.

3. The goal of this iteration is to put a prototype in the hands of users asap. It's not clear that giving NFTs their own slice like this, separate from assets, is even ideal anyway.

henryboldi:
Proper error handling here is enough code to justify a separate PR/discussion. I totally agree that it's important, but this PR is already too big. We're aiming to make PRs more task oriented rather than feature by feature

background/redux-slices/nfts.ts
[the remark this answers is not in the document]
jagodarybacka: Another problem I see with first suggested approach is that `contractID + tokenID` is not unique key (I've tried to use it as a React list item key and there were warnings).

I love flat structure but then why to we need `evm`? We won't have any other VM supported for a long time I guess.

Also we have to take into consideration tight deadline for this feature, I would love to upgrade our codebase but I feel like we don't have much room for innovation right now 😞

background/redux-slices/nfts.ts
```diff
@@ -0,0 +1,77 @@
...
+
+async function fetchNFTs(address: string, currentNetwork: EVMNetwork) {
+  // @TODO: Move to alchemy.ts, remove hardcoded polygon or eth logic
+  const result = await (
+    await fetch(
+      `https://${
+        currentNetwork.name === "Polygon" ? "polygon-mainnet.g" : "eth-mainnet"
+      }.alchemyapi.io/nft/v2/${
+        process.env.ALCHEMY_KEY
+      }/getNFTs/?owner=${address}`
```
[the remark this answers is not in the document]
jagodarybacka: about pagination - in the next PR we have added lists virtualization so we can possibly implement endless scroll, paginated fetching NFTs could be nice but I think it can be done a little bit later 😄

jagodarybacka:
@greg-nagy's comments addressed in the https://github.com/tallycash/extension/pull/1866

greg-nagy (approved):
thank you all, this is beautiful ✨🙇‍♂️

greg-nagy:
A couple of thoughts here because there was tension around this.

@jagodarybacka 's solution was perfect. (I intended to do this myself, sorry for not getting to it.) 
I was looking for that 5 lines of code here. Nothing else. 

The reason I blocked this is that we tend to handle error handling as a separate task, and this is really dangerous, and not from an engineering perspective but from a product and user experience point. 

PR is big: that's true, but not related to the 5 lines of code for error handling. Let's split up PRs into smaller chunks next time so it won't cause extra tension if something needs to be changed. 

But at the same time, I do appreciate that this was an experiment that grow to be big, and you wanted to get it under control by merging the current state. I felt bad for hindering this effort but felt it was necessary. 

I also appreciate the fact that we haven't talked about error handling and this change in general beforehand. It wasn't the best communication on my part to do it by blocking this PR. But again, even though it was not ideal at all, I felt that it's better to start this sooner than wait for the ideal situation later. 

Anyhow, ty @henryboldi @jagodarybacka! You are truly the MVPs for this sprint 🚀

PR #1853, opened 2022-07-27
feat(signing): ensure ChainID does not overflow

sambacha:
use `MAX_SAFE_CHAIN_ID`

This PR is INCOMPLETE, as I am unsure of where to actually enforce this check in the signing process. I have observed that Tally uses hex-encoded ChainID's (as is the proper way). So this may not be entirely relevant. However, if users are sharing private keys between wallets or some other contrived example this may cause an issue depending on Tally's dependency usage.


0xDaedalus:
Hey Sam,
Thanks for the contribution!  I've been on vacation for the past 2 weeks and am just now catching up on PRs 🌴.  This is an excellent check and I think we'll need to use it once we implement letting users add their own custom RPC endpoints rather than just the ones we have whitelisted.

sambacha:
> This PR has been added to the Deferred Community Contributions milestone - we would like to revisit it later. For now, let's close it as we don't plan to continue work in the near future.

Thanks!

PR #1874, opened 2022-08-02
🗞 Handle read-only mode

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1813
### What 
* display read-only badge on swap and send pages
* disable action buttons on swap and send pages

image


mhluongo (approved):
Looks good, works well. :rocket:

PR #1875, opened 2022-08-02
🪶Reduce activity size: transform tx data between ui and background

greg-nagy:
## Why?
with our current architecture a heavy redux is the worst
thing we can have for performance reasons

## What?
this commit introduces a facade layer between
background and ui which can be used to propagate
only those properties to the UI that are actually shown

you can think of this as poor man's graphql query

the activities slices still can be heavy, 220kb+ on
the dev account so this is a big step but pagination
and more improvements will be necessary

this change reduced the activities slice size from 1.8mb+
to 223kb which is 12% of the original size

## Follow up work

* #1885
* #1886

## Context

⚠️ It takes the extension ~1 hour to load all historical transaction data. After 1 hour, this is the store size. It's still a lot which makes #1885 essential to do as follow-up work.

```
2022-08-05T07:38:17.048Z state size [kb]: 930.4462890625
redux size:6 	assets size [kb]: 372.0537109375
redux size:6 	activities size [kb]: 432.7861328125
redux size:9 		0x208e94d5661a73360d9387d3ca169e5c130090cd size [kb]: 432.740234375
```

Script to measure store size
```
    const state = JSON.parse(stateString)
    console.log(`${new Date(Date.now()).toISOString()} state size [kb]: ${stateString.length/1024}`)
    Object.keys(state).filter(k => k === "assets" || k === "activities").map(sliceKey => {
        const slice = state[sliceKey]
        console.log(`\t${sliceKey} size [kb]: ${JSON.stringify(slice).length/1024}`)
        if (!Array.isArray(slice)) {
            Object.keys(slice).map(key => {
                console.log(`\t\t${key} size [kb]: ${JSON.stringify(slice[key]).length/1024}`)
            })
        }
    })
})
```

With this store size our CPU usage is still visible in activity monitor but a lot better then it was. 
CleanShot 2022-08-05 at 11 41 07

CleanShot 2022-08-05 at 11 41 43

Our cpu usage can grow up to 86% 

## Test

* test the changes 
  * create a fresh install from main and start a dev server
  * import `enroll...` 
  * wait a bit ~5-10 mins or so 
  * the wallet > actitiy tab should work as normal 
  * the wallet > activity > activity detail tab should work as normal
* testing the migrations
  * create a fresh install from main and start a dev server
  * import `enroll...` 
  * wait a bit
    * to fully grab all the activities, it can take ~1 hour or so (I haven't checked precisely) 
      * at this point, the `activities` slice will be ~1.8 mb and the whole store will be ~2.2 mb
    * after a couple of minutes, the size growth will be visible 
    * it's ok not to wait for all the activities arrive 
  * check the size of the store and the activities slice (script below)
  * checkout this branch 
  * recheck the size of the store and activities slice 

## Scripts

Store and slice size

```
    const state = JSON.parse(stateString)
    console.log(`${new Date(Date.now()).toISOString()} state size [kb]: ${stateString.length/1024}`)
    Object.keys(state).filter(k => k === "activities").map(sliceKey => {
        const slice = state[sliceKey]
        console.log(`\t${sliceKey} size [kb]: ${JSON.stringify(slice).length/1024}`)
        if (!Array.isArray(slice)) {
            Object.keys(slice).map(key => {
                console.log(`\t\t${key} size [kb]: ${JSON.stringify(slice[key]).length/1024}`)
            })
        }
    })
})
```

Inspect slice

```
    const state = JSON.parse(stateString)

    const addressOnNetwork = state.activities["0x208e94d5661a73360d9387d3ca169e5c130090cd"]["1"]
    const txHash = addressOnNetwork.ids[0]
    console.log(JSON.stringify(addressOnNetwork.entities[txHash],null,2))
    console.log(JSON.stringify(state.activities,null,2))

})
```


background/utils/view-model-transformer.ts
```diff
@@ -0,0 +1,48 @@
...
+]
+
+export function filterTransactionPropsForUI<T>(transactionData: {
+  transaction: T
+  forAccounts: string[]
+}): { transaction: T; forAccounts: string[] } {
+  return {
+    transaction: pick(transactionData.transaction, transactionPropertiesForUI),
+    forAccounts: transactionData.forAccounts,
+  } as { transaction: T; forAccounts: string[] }
```
greg-nagy, line 47: We are cheating with the types a little bit here. We can't guarantee that the generic type will not be modified. Otoh the [`ActivityItem`](https://github.com/tallycash/extension/blob/4f2be1e04ce3f2ecaacb74c5c4e916a35d70295c/background/redux-slices/utils/activity-utils.ts#L19) used going forward will be fulfilled. But here we are between 2 worlds background and UI and it's probably ok to cheat a little bit with the typings.

jagodarybacka:
Ok, I've tested migration and it looks fine but we are losing this field. I'm going to fix this
image

background/utils/view-model-transformer.ts
Shadowfiend: Will it be fulfilled? Is this still a valid `EnrichedEVMTransaction` without eg the `input` field? Or are you saying you're planning on modifying the `ActivityItem` type to be reflective of the fields we're actually pulling here?

jagodarybacka:
Fixed 
image

background/utils/view-model-transformer.ts
greg-nagy: I missread the types. Will fix later.

jagodarybacka: @greg-nagy this is non-blocking right?

jagodarybacka (approved):
I've tested fresh install as well as migrations, all seems fine. Redux state size is growing as expected. I think we can merge it 😅

background/utils/view-model-transformer.ts
jagodarybacka: ok, I assume it's not 😃 Merging this so I can test the whole releasable code

greg-nagy: It is not, but requires follow up work .

PR #1877, opened 2022-08-03
✏️ NFT's post QA improvements

jagodarybacka:
### What

Various minor fixes for bugs that were found during QA.

- Change the empty NFT list component's text to not mention Looks Rare 
image
- fix selecting NFTs for read-only accounts that were imported using raw address (not ENS or UNS names)


greg-nagy (approved):
👑

PR #1878, opened 2022-08-03
💣 Filter out NFTs with errors

jagodarybacka:
### What
If NFT has an error then maybe better to hide it from the list
image


greg-nagy (approved):
let's 🚢

PR #1880, opened 2022-08-03
Hide "max" button for base asset

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1825
### What 
As it is not possible to send the full amount of base asset we shouldn't suggest users to do it by having "max" button next to the asset amount input

image

### DIFF
--- a/ui/components/Shared/SharedAssetInput.tsx
+++ b/ui/components/Shared/SharedAssetInput.tsx
@@ -13,6 +13,7 @@ import {
   fixedPointNumberToString,
   parseToFixedPointNumber,
 } from "@tallyho/tally-background/lib/fixed-point"
+import { selectCurrentNetwork } from "@tallyho/tally-background/redux-slices/selectors"
 import SharedButton from "./SharedButton"
 import SharedSlideUpMenu from "./SharedSlideUpMenu"
 import SharedAssetItem, {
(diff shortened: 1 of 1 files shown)


ui/components/Shared/SharedAssetInput.tsx
```diff
@@ -352,6 +355,10 @@ export default function SharedAssetInput<T extends AnyAsset>(
     [onAssetSelect]
   )
 
+  const isMaxButtonVisible =
+    showMaxButton &&
+    selectedAssetAndAmount?.asset.symbol !== currentNetwork.baseAsset.symbol
```
jagodarybacka, line 360: I wanted to use [`isNetworkBaseAsset()`](https://github.com/tallycash/extension/blob/hide-max-for-base-asset/background/redux-slices/utils/asset-utils.ts#L52) here but there is a problem because MATIC is not treated as a base asset by this function. I believe [this comment](https://github.com/tallycash/extension/blob/hide-max-for-base-asset/ui/pages/Swap.tsx#L104-L105) is referring to the same problem. 

After some time of debugging I think it is outside of the scope of this task so let's use simple solution and fix underlying problem in a separate issue.

greg-nagy (approved):
ty 🙏

PR #1883, opened 2022-08-03
🗣️ Add language switch in Settings

gasolin:
- Add language select UI in Settings
- Add `SUPPORT_MULTIPLE_LANGUAGES=false` to env.default
- Use localStorage to store lang selection

~~Check the source and there's no existing menu selector element to use (the closest is the ShareAssetInput)~~

~~to make test multi lang easier, let developer able to switch languages by add `DEFAULT_LANGUAGE="zh_tw"` or `DEFAULT_LANGUAGE="pt_br"` in .env file~~

~~show current language in Settings when DEFAULT_LANGUAGE env is set~~

截圖 2022-08-03 下午10 46 00


jagodarybacka:
This is looking great so far! ✨ 

Don't you want to implement the whole dropdown as designed here: https://github.com/tallycash/extension/issues/1544#issuecomment-1147138639? So we can have it without the `DEFAULT_LANGUAGE` flag.

Or you can do it in another PR if you prefer 😄

gasolin:
@jagodarybacka as I mentioned there's no existing menu selector component and will take quite an effort to align with the design.

The possible approach is to distill the MenuSelector element from ShareAssetInput and reuse it for the language switch, but it will not include in this PR

jagodarybacka:
@gasolin It's `SharedSelect` 😄 
Here is snippet to start:
```javascript
      <SharedSelect
        width={194}
        options={[{ value: "en", label: "English" }]}
        onChange={() => {}}
      />
```
image

gasolin:
@jagodarybacka thanks! 
Updated to the language select UI and add `SUPPORT_MULTIPLE_LANGUAGES=false` to env.default

What's missing is to get/store the user-select language to `tally-background/redux-slices/ui`
could be a separate patch before multi-lang goes to the production

mhluongo:
So excited for this :smile:

.env.defaults
```diff
@@ -21,4 +21,5 @@ SUPPORT_OPTIMISM=false
 CUSTOM_GAS_SELECT=true
 USE_UPDATED_SIGNING_UI=false
 USE_BALANCE_RELOADER=false
-SUPPORT_NFTS=true
\ No newline at end of file
+SUPPORT_NFTS=false
```
jagodarybacka, line 24:
```suggestion
SUPPORT_NFTS=true
```
😂

ui/_locales/index.ts
```diff
@@ -1,11 +1,49 @@
 import EN from "./en/messages.json"
+import zhHant from "./zh_Hant/messages.json"
+import ptBR from "./pt_BR/messages.json"
+
+interface ILang {
```
jagodarybacka, line 5: we are not really following that convention with prefixing interface names with `I` but it's up to you. And we are mostly using `type` instead of `interface` for simple types like `ILang` and `ILangOptions` here

.env.defaults
```diff
@@ -21,4 +21,5 @@ SUPPORT_OPTIMISM=false
...
+SUPPORT_MULTIPLE_LANGUAGES=false
```
jagodarybacka, line 25: this feature flag is not working correctly because you have to add it to the https://github.com/tallycash/extension/blob/main/background/features.ts

jagodarybacka (changes requested):
Amazing progress, so excited to see this! 🔥 🔥 🔥 

Some things I've noticed:
1. Maybe we can rename folder `_locales` to `locales` as I don't see any reason why underscore is needed here
2. Can you try to fix this empty space that is visible on the open dropdown? It feels like styles are broken here
image

gasolin:
> Amazing progress, so excited to see this! 🔥 🔥 🔥
> 
> Some things I've noticed:
> 
> 1. Maybe we can rename folder `_locales` to `locales` as I don't see any reason why underscore is needed here

was named after chrome localization convention, fine to rename the folder now (it will effect weblate setup though)

> 2. Can you try to fix this empty space that is visible on the open dropdown? It feels like styles are broken here
>    image

The switch UI could be fixed in a separate task since its a general issue

jagodarybacka:
> was named after chrome localization convention, fine to rename the folder now (it will effect weblate setup though)

if it will break weblate we can do it later if you want

> The switch UI could be fixed in a separate task since its a general issue

👍

jagodarybacka:
@gasolin Let me confirm with the rest of the team if we want to hide it behind the feature flag or not 😄

gasolin:
@jagodarybacka for a proper multi-lang release I think we still missed the translation of the onboarding/ledger support part, so I'd tend to hide it at this stage

jagodarybacka:
Good point but on the other hand we can use existing implementation to get more people involved in the translations 😃

jagodarybacka (approved):
@gasolin I've fixed dropdown placement and feature flag 😄 Thank you for this PR, it's awesome! 

Let's merge it with feature flag so I can create a QA build so others can take a look at this!

PR #1884, opened 2022-08-04
Patch packages before each dev build

jagodarybacka:
### What

Apply patches before dev build

### Why

If for some reason webext-redux package is not patched then we experience weird bugs without an easy way to debug them. As this is a recurring issue and we are wasting our time on it,
let's apply patches before dev builds.


greg-nagy (approved):
ty for the bughunt :)

PR #1889, opened 2022-08-05
Fixed regression with tx screen loader

jagodarybacka:
### What
If there are no transaction details yet we should show tx screen loader instead of an empty screen

https://user-images.githubusercontent.com/20949277/183091010-79ed5abc-0dae-410f-b7f4-f0ebca8b5bd1.mov


greg-nagy (approved):
Looks good let's 🛳

PR #1890, opened 2022-08-05
Fix selecting currency and asset from price pair

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1888
### What 
`isSmartContractFungibleAsset()` is not working correctly with base assets so let's select them by their symbols.

### DIFF
--- a/background/redux-slices/assets.ts
+++ b/background/redux-slices/assets.ts
@@ -15,6 +15,7 @@ import { getProvider } from "./utils/contract-utils"
 import { sameNetwork } from "../networks"
 import { ERC20_INTERFACE } from "../lib/erc20"
 import logger from "../lib/logger"
+import { FIAT_CURRENCIES_SYMBOL } from "../constants"
 
 type SingleAssetState = AnyAsset & {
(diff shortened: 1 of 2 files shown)


background/redux-slices/assets.ts
```diff
@@ -76,11 +77,11 @@ const assetsSlice = createSlice({
...
     ) => {
-      const fiatCurrency = pricePoint.pair.find(
-        (asset) => !isSmartContractFungibleAsset(asset)
+      const fiatCurrency = pricePoint.pair.find((asset) =>
+        FIAT_CURRENCIES_SYMBOL.includes(asset.symbol)
       )
-      const pricedAsset = pricePoint.pair.find((asset) =>
-        isSmartContractFungibleAsset(asset)
+      const [pricedAsset] = pricePoint.pair.filter(
+        (asset) => asset !== fiatCurrency
```
henryboldi, line 84: `findClosestAssetIndex` maybe could make sense here, but this solution works!

henryboldi (approved):
✨🚢

Issue #1234, opened 2022-03-23
Enrichment "admonitions"


mhluongo:
The `EnrichmentService` includes annotations about transaction requests and addresses. It should also include the idea of "admonitions" — annotations that the UI should show as warnings, perhaps even requiring a user to agree.

Some example admonitions.

- [ ] If a user is sending funds to an address that has no ETH balance and a `0` nonce, this is probably a mistake.
- [ ] If a user is sending funds to any contract directly — but especially an ERC-20 or NFT contract address — this is a mistake.
- [ ] If a user is interacting with a contract that no user has ever interacted with before, be extra cautious.

H/t @Shadowfiend for the name and offline discussion.

Issue #509, opened 2021-12-03
Warn user before signing a token approval to an EOA


mhluongo:
We can't reliably tell is an address is an EOA or a CREATE2 contract that hasn't been deployed, but there should be a "THIS CERT IS WRONG" style warning that needs dismissing to enable this behavior.

We've been [talking about in Discord](https://discord.com/channels/808358975287722045/875438141610270740/916156896392658985) due to the Badger hack, and @banteg summed the issue up well

https://twitter.com/bantg/status/1466724441866526726

mhluongo:
@VladUXUI do we have a dismissable warning for the signature page designed yet? Even a mockup would be helpful

mhluongo:
<img width="959" alt="image-2" src="https://user-images.githubusercontent.com/427505/144710958-1e167536-e1dd-4556-aa12-e0d414d5950e.png">

Issue #262, opened 2021-09-20
Warn before sending funds to a known contract


mhluongo:
Users rarely mean to directly send assets to a smart contract, as most of those systems are instead built on a pull-based payment model. We should warn users before accidentally losing funds by sending directly to a contract.

This is a generalization of #261, and will require quite a bit more work on the backend... either by indexing every smart contract deployed on the network, or bringing in a third party dataset.

mhluongo:
Closed by https://github.com/tallycash/extension/pull/1993

Issue #261, opened 2021-09-20
Warn before sending funds to a known ERC-20 contract


mhluongo:
This is a really easy safety win, and should be doable today without backend changes.

@VladUXUI this could happen sending ETH or ERC-20s. Ideas on how we might warn users off before they sign?

Shadowfiend:
@mhluongo @VladUXUI this is probably increasingly easy at the sign stage with the newly-hooked-up enrichment service, _but_ not at the “send” stage.

Just to clarify the interaction here, if we wanted to look this up when sending ETH, we would have to block/put the page in a loading state until we could look up whether the address in question looked like a contract or not. Not sure if we like that friction, and we could reduce it by caching the last X addresses that were looked up, but we could never all the way get rid of the delay without risking the user hitting send before we can warn them.

VladUXUI:
What waiting times are we talking about here @Shadowfiend?

mhluongo:
Less than 250 ms on most network connections @VladUXUI. I think the approach would be to load the page, then have the "send" button be blocked rather than the whole page.

VladUXUI:
Added to my to-do list

mhluongo:
Background script changes for this landed in #1516. Should now be easy to snipe

Shadowfiend:
Local state for the cache sounds reasonable. The functionality isn't completely implemented, I believe, but the framework for it is. 

Right now the warning that is implemented is for sending to the same token address that is being interacted with, where I believe this issue was also encompassing sending ETH to a contract address or any ERC20 to another ERC20 address. We can also probably start by only doing ERC20 contracts in our token lists before adding any extra lookup logic. 

I don't think the UI currently displays the send-to-token warning, so that's probably the biggest missing piece.

mhluongo:
@hyphenized are you still interested in taking this? I'm doing some related background work, happy to advise if you want to take the frontend work!

Note the design from https://github.com/tallycash/extension/issues/525 might make more sense as it covers

> So we do need a "i know what i'm doing" function.

cc @VladUXUI

mhluongo:
Closed by #1993

PR #1900, opened 2022-08-08
🍽️ Shared Enrichment: Enrich addresses to inform users of potentially malicious transactions

mhluongo:
Enrich addresses found in a transaction, transaction request, or transfer event, including...

* the sender of a tx
* the contract a tx interacts with
* the sender of an ERC-20
* the recipient of an ERC-20

... but not (yet) including any addresses that might be found in function args in other ABIs.

Each address is annotated with...

* the account nonce
* a resolved ENS or UNS name, if any
* a recent account balance
* whether the account has non-zero code (a deployed smart contract)

NB: This work isn't optimized against `NameService` or `ChainService` lookups, though it does some light caching inherited from the existing approach. It adds roughly 2 provider calls for every address that's enriched as part of a transaction, but doesn't add any open-ended polling. Adding optional cache-keys to both service lookups will work wonders here.

# To Test

- [x] Load a read-only account, and make sure wallet activities still include ENS names where appropriate
- [x] Sign In With Ethereum, confirming typed data signatures look right
- [x] Send an ERC-20 to an ENS name, confirming the name for the recipient is resolved in the review screen
- [x] Send ETH to an ENS name, confirming the name is resolved

Unblocks #1234 #509 #261 #262 #84


Shadowfiend:
High level approach looks real good---enriched address types, switching up fields accordingly, etc---don't expect I'll have time to go deep in review though.

background/services/enrichment/addresses.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+
+export async function enrichAddressOnNetwork(
+  chainService: ChainService,
+  nameService: NameService,
+  addressOnNetwork: AddressOnNetwork
+): Promise<EnrichedAddressOnNetwork> {
+  return {
+    ...addressOnNetwork,
```
jagodarybacka, line 40: Why do we need address and network on both levels - here and inside the annotation?

background/services/enrichment/index.ts
```diff
@@ -397,7 +102,10 @@ export default class EnrichmentService extends BaseService<Events> {
     const enrichedTxSignatureRequest = {
       ...transaction,
       network,
-      annotation: await this.resolveTransactionAnnotation(
+      annotation: await resolveTransactionAnnotation(
+        this.chainService,
+        this.indexingService,
+        this.nameService,
```
Shadowfiend, line 108: Not necessarily convinced this is an improvement. May make more sense to pass the whole enrichment service instance to the enrichers rather than double the parameter list. The dependencies are clearer, but the outcome feels more brittle. Could easily be convinced otherwise though. 

Also wondering if we call these resolve* functions enrich* instead so anything that adds enrichment data is simply an enrichment.

mhluongo: It's a pretty convoluted service otherwise... these functions have a lot of dense heuristics / knowledge that I think should be separated out for easier review. I do hate this though, and passing a single service arg does make sense.

Agreed on the `enrich*` name too. Just cargo-culted to get this together.

background/services/enrichment/addresses.ts
mhluongo: Hmm, no need to repeat that, taking a look

background/services/enrichment/index.ts
mhluongo: Ehhhhh actually... man this ask is a bunch of rework of existing code. Not excited about it, and would rather tackle it in a separate PR / next time we touch this part of the codebase.

background/services/enrichment/addresses.ts
mhluongo: It doesn't appear to be repeated, what am I missing? There's no `AddressOnNetwork` in the annotation.

It is annoying that the `NameOnNetwork` is in the annotation, but I actually think that's the correct type relative to what could happen (resolving a mainnet ENS name for a Polygon address/network)

background/services/enrichment/index.ts
Shadowfiend: That's fine, def not a blocking request. Also I was specifically referring to the explosion of parameters--factoring enrichment functions out as simple functions and pulling them into separate files is 💯💯💯 and already a huge improvement imo.

background/services/enrichment/addresses.ts
Shadowfiend: I think address and network are both at this level and in the annotation:

https://github.com/tallycash/extension/pull/1900/files#diff-f7c854a5832bd3939d021b1db87eaf2062a16e02da3e1bce4f832c9889cab854R40

https://github.com/tallycash/extension/pull/1900/files#diff-f7c854a5832bd3939d021b1db87eaf2062a16e02da3e1bce4f832c9889cab854R25

mhluongo: Oof, any way I can get the type system to have prevented that? Certainly not the intent

https://github.com/tallycash/extension/pull/1900/files#diff-8b87467ba85224abe33cbe9bc89de9060126857d46bd6f03ef71a7bdce3dba57R133

mhluongo: Fixing it in the meantime

Shadowfiend: Typescript is a minimal definition rather than a full one unless/until they implement exact types unfortunately. Excess property checks do happen in certain scenarios but not everywhere.

mhluongo:
This is ready for another look @jagodarybacka @Shadowfiend @0xDaedalus. Note that it will unblock a number of security features and proposed GitHub grants... I think we should merge it!

Shadowfiend:
PR author thinks PR should be merged... News at 11 😉

mhluongo:
It's okay, now there's a big ol' conflict from @0xDaedalus 's work :sob:

background/services/enrichment/types.ts
```diff
@@ -130,3 +129,33 @@ export type SignTypedDataAnnotation =
...
+  timestamp: UNIXTime
+  /**
+   * The latest nonce associated with the address / network.
+   */
+  nonce: bigint
+  /**
+   * Whether code was found at this address at the time of annotation
+   * resolution.
+   */
+  code: boolean
```
0xDaedalus, line 147: calling this property `hasCode` or `codeExists` might be a bit closer to what the comment is describing

0xDaedalus (changes requested):
Looks like the activity screen is broken on this branch:

image

[Looks like](https://github.com/tallycash/extension/compare/shared-enrichment...shared-enrichment-patch) we're not always setting `recipient`, `contractInfo`, and `sender` despite our type system thinking that we do. 

The above patch fixes the error but I think its probably better to figure out the root cause.

mhluongo:
Pretty sure your patch is all we need @0xDaedalus? `sender` and `recipient` are only supposed to show up on `AssetTransfer`s and `contractInfo` is only supposed to show on `contractInteraction` after all

Shadowfiend:
> but I think its probably better to figure out the root cause

I believe this is the sticking point. Why did the types not catch this? Any thoughts?

background/services/enrichment/transactions.ts
```diff
@@ -0,0 +1,342 @@
...
+    assetAmount: { amount: baseAssetBalance },
+  } = await chainService.getLatestBaseAccountBalance({
+    address: transaction.from,
+    network,
+  })
+
+  const { gasLimit, blockHash } = transaction
+
+  const gasFee: bigint = isEIP1559TransactionRequest(transaction)
+    ? transaction?.maxFeePerGas ?? 0n * (gasLimit ?? 0n)
```
jagodarybacka, line 157: `0n * (gasLimit ?? 0n)` isn't it always `0n`?

jagodarybacka, line 157: we want to do `maxFeePerGas * gasLimit` right? so to fix this calculation it should be:
```suggestion
    ? (transaction?.maxFeePerGas ?? 0n) * (gasLimit ?? 0n)
```

background/services/enrichment/transactions.ts
```diff
@@ -0,0 +1,342 @@
...
+    : (("gasPrice" in transaction && transaction?.gasPrice) || 0n) *
```
jagodarybacka, line 158: why not just:
```suggestion
    : (transaction?.gasPrice ?? 0n) *
```

background/services/enrichment/transactions.ts
mhluongo: Right you are!

background/services/enrichment/transactions.ts
mhluongo: Compiler error! `"gasPrice does not exist on..."`

mhluongo:
> I believe this is the sticking point. Why did the types not catch this? Any thoughts?

No idea. I'd expect that `EnrichedEVMTransaction["annotation"]` being a tagged union would cover this, but clearly it hasn't.

mhluongo:
I've found the offending code, but I can't for the life of me understand how the type system allowed this `undefined` through. Thoughts?

https://github.com/tallycash/extension/pull/1900/files#diff-0effdeeeba712d998d5baa16729996bd8cc9cc6815a2cb278994a6a76ffac716R92

0xDaedalus (changes requested):
Looks like there is an issue when sending ERC-20 tokens to an ENS name:

image

Reference: https://polygonscan.com/tx/0x45bd67f5c15215e49b39614b9fdedd06bbef95ee1e360746c39f0687cb393a15

mhluongo:
It's worse than that @0xDaedalus → recipients aren't showing up when they should, regardless of the ENS name. Debugging

mhluongo:
It looks like the data issue comes from [`filterTransactionPropsFromUI`](https://github.com/tallycash/extension/blob/f023871299b9a75a84aefa5fe299c7ea2606e2fa/background/main.ts#L703) — maybe you understand the type ramifications better than I do here @Shadowfiend?

mhluongo:
Ah. So we have an `as` in `filterTransactionPropsFromUI` that's casting the picked-over object it returns as an `EnrichedEVMTransaction`.

Do we still want / need `filterTransactionPropsFromUI`? It's going to be an issue every time we add new enrichment data.

mhluongo:
I believe I've fixed the underlying issue, but the type info is still being swallowed. Happy to address is here if y'all have any ideas, otherwise this is ready for another look!

0xDaedalus:
>Happy to address is here if y'all have any ideas

I am thinking the answer to our problems here is a new `TransactionForUI` type - hesitant to address it here though as this PR is already quite big

Took another look -

- sending ETH is still broken with `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'toBigInt')
    at enrichTransactionWithReceipt (utils.ts:240:41)` showing up in the console
- sending ERC-20s works but now shows the address rather than the resolved name on the tx receipt.

Shadowfiend:
Yep. Thought about recommending that in the original filtering PR, don't remember if I had time to post the comment though. A dedicated type is the right move here, even though it will also be annoying to maintain. 

I believe we already have an ActivityItem type that can be adjusted to this.

mhluongo:
> sending ETH is still broken with [Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'toBigInt') at enrichTransactionWithReceipt (utils.ts:240:41)](https://github.com/tallycash/extension/blob/main/background/services/chain/utils/index.ts#L240) showing up in the console

I can't reproduce this, ETH sends appear to work fine on mainnet and as an ERC-20 on Polygon :thinking:

> now shows the address rather than the resolved name on the tx receipt.

Looking into this now.

mhluongo:
Screenshot from 2022-08-17 14-14-55

Looks like that took care of the name display issue. Note this screenshot is sending ETH to myself with a Ledger on this branch @0xDaedalus — lmk if there are different reproduction steps I should be following.

Moving on to fix conflicts against `main`.

mhluongo:
Conflicts resolved if you want to give this another go @0xDaedalus. Note that I had to freehand merge the Optimism gas enrichment a bit — [this bit](https://github.com/tallycash/extension/pull/1900/commits/7af1096dccbad3ff72e1007888ae5749e1ba36a8#diff-0effdeeeba712d998d5baa16729996bd8cc9cc6815a2cb278994a6a76ffac716R172) deserves extra scrutiny.

background/services/enrichment/transactions.ts
```diff
@@ -168,10 +169,15 @@ export default async function resolveTransactionAnnotation(
 
   const { gasLimit, blockHash } = transaction
 
+  const additionalL1Gas = EVM_ROLLUP_CHAIN_IDS.has(network.chainID)
+    ? await chainService.estimateL1RollupFee(network, transaction.input)
+    : 0n
+
   const gasFee: bigint = isEIP1559TransactionRequest(transaction)
-    ? (transaction?.maxFeePerGas ?? 0n) * (gasLimit ?? 0n)
+    ? (transaction?.maxFeePerGas ?? 0n) * (gasLimit ?? 0n) + additionalL1Gas
```
0xDaedalus, line 177: no need for `+ additionalL1Gas` here since if its an EIP1559 tx we know its not an Optimism tx

mhluongo: I 'd rather leave it in case we add other L2s that support EIP-1559, and the logic should be good regardless.

0xDaedalus (approved):
Code looks good - tests pass - lets 🚢

PR #1904, opened 2022-08-08
🔧  Expect activities ids list to be `undefined`

jagodarybacka:
### What
Apparently, the wallet can be in a state where the ids list is not defined for a given account


greg-nagy (approved):
⛴

PR #1914, opened 2022-08-10
Thomas Flair: Add new signing UI skeleton

Shadowfiend:
Two primary components here:

- The signer frames for keyring and read-only signers are in place. Note that these don't render any transaction details (e.g., recipient, interpretation of the transaction's intent, warnings, etc). Instead all they do is render the footer buttons (sign/reject) and handle the workflow of signing (for keyrings, this includes unlocking behavior).
- The anchor point for signing data rendering. Right now no data is actually rendered (see above), but the path is now present to add that kind of interpretation and rendering.

Also included is internal handling the signing loader in the new UI structure. The same component is used to present loading as in the `SignTransaction` component.

## Testing

- [ ] All of this should be behind a feature flag. Ensure signing is unchanged with the feature flag disabled. There is a slight touch to `SignTransaction` to handle the signing loader that makes it worth doing some testing here.
- [ ] Enabling the feature flag, ensure that keyring transactions can be signed---note that there won't be any details about the transaction being signed, so please only use a trusted source (for example, an internal swap transaction) for this flow, even if just using a test wallet.
- [ ] Enabling the feature flag, ensure that data (e.g. SIWE) and typed data (e.g. EIP-2612 permits) can be signed with the keyring---as above, no details will be presented, so prefer to use a trusted source.
- [ ] Enabling the feature flag, ensure read-only wallets display a note about read-only wallets not being able to sign in the above scenarios.


mhluongo:
What are you waiting for to undraft here?

Shadowfiend:
I haven't had a chance to test this since I factored these changes out of the big changeset 😅

Shadowfiend:
Ok now I blocked this on #1935, then it'll be good to go.

Shadowfiend:
Okay we ready.

jagodarybacka (commented):
With flag turned on:
✅ internal swap on Ethereum mainnet
✅ SIWE (mainnet)
✅ signing data (mainnet)
❌ internal swap on Polygon

Shadowfiend:
Ooh that's a good question. Let's not for now since we're feature flagged and that's off my current test path. Definitely need to take a look though.

ui/components/Signing/Signer/SignerBaseFrame.tsx
```diff
@@ -0,0 +1,39 @@
...
+  children: ReactElement
+}
+
+export default function SignerBaseFrame({
+  children,
+  signingActionLabel,
+  onConfirm,
+  onReject,
+}: SignerBaseFrameProps): ReactElement {
+  const [isOnDelayToSign /* , setIsOnDelayToSign */] = useState(false)
```
mhluongo, line 17: What's the deal with this comment? Will `isOnDelayToSign` ever be `true`?

mhluongo (approved):
Moving this along! :steam_locomotive:

ui/components/Signing/Signer/SignerBaseFrame.tsx
Shadowfiend: Yes. This is the feature we currently have where the sign button is disabled for a random number of seconds to avoid clickjacking-style attacks.

PR #1918, opened 2022-08-10
Allow to switch default wallet from dapp connection popup

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/1916

### What
* Allow user to switch off default wallet toggle from within dapp
connection popup.
* Fix issue with switching off default tally without other wallets installed

image


ui/components/Shared/SharedTooltip.tsx
```diff
@@ -15,8 +17,17 @@ interface Props {
   }) => ReactElement
 }
 
-function getHorizontalPosition(width: number) {
-  return `right: -${width / 2 + 4}px;`
+function getHorizontalPosition(horizontal: HorizontalPosition, width: number) {
+  switch (horizontal) {
```
greg-nagy, line 21: nostalgic change :)

ui/pages/DAppConnect/SwitchWallet.tsx
```diff
@@ -0,0 +1,51 @@
...
+import SharedButton from "../../components/Shared/SharedButton"
+import SharedTooltip from "../../components/Shared/SharedTooltip"
+import { useBackgroundSelector } from "../../hooks"
+
+export default function SwitchWallet({
+  switchWallet,
+}: {
+  switchWallet: () => void
+}): ReactElement {
+  const isDefaultWallet = useBackgroundSelector(selectDefaultWallet)
```
greg-nagy, line 12: non blocker, planting seeds here: in angular land the concept of smart and dumb components got really prominent. The diffentiation between the 2 was that dumb components did not need to know anything about the context they live in. Purely communicating through porps and events (event handler function props in react)

Smart components were more like and orchestrator for the dumb components. Their html code is usually short, but they collect the data from many sources and pass them down to the dumb components.

greg-nagy (approved):
code looks good and works well ✨

ui/pages/DAppConnect/SwitchWallet.tsx
jagodarybacka: I know this idea, it was popular in React too some time ago, but as hooks are a standard right now I see a lot less people advocating to use it. We can discuss it later if you want

PR #1920, opened 2022-08-11
Default to main route if wallet is stuck on unknown route

jagodarybacka:
### What

We have changed routes names before - if a user hasn't used the wallet for a while and it updates to a newer version then it can be stuck on a location that has been removed. Let's default to the main route if the wallet is in an unknown location.


greg-nagy (approved):
nice bug hunt and fix!

PR #1921, opened 2022-08-11
Revert treating sign message input as hex to fix SIWE

jagodarybacka:
### Why

Let's go back to the previous solution for sign message to unblock SIWE that is expecting input to be parsed as ASCII.

---

**This is not a final solution.**


greg-nagy (approved):
😢 but makes sense 🙏

Ty

PR #1926, opened 2022-08-12
:grin: Optimistic Fungibility: Support NFTs on L2s

mhluongo:
Introduce the [SimpleHash API](https://app.simplehash.com/) as a supplement to Alchemy to get a wallet's NFTs on Optimism and Arbitrum.

# To Test

- [ ] Flip `SUPPORT_OPTIMISM` and `SUPPORT_ARBITRUM` to `true`
- [ ] Load an account you're familiar with, and confirm the correct NFTs show up on Ethereum and Polygon.
- [ ] Load a read-only account with NFTs on Optimism and Arbitrum. Here's [one I found](https://arbiscan.io/address/0xdeb7c10565b37a6f76a9fcf174bc865fbde65a97).
- [ ] Scroll through the entire rendered list of NFTs on both chains, confirming that there aren't any odd runtime errors and that stuff looks right.


background/lib/alchemy.ts
```diff
@@ -289,3 +289,50 @@ export function transactionFromAlchemyWebsocketTransaction(
...
+      network.name === "Polygon" ? "polygon-mainnet.g" : "eth-mainnet"
+    }.alchemyapi.io/nft/v2/${process.env.ALCHEMY_KEY}/getNFTs/`
+  )
+  requestUrl.searchParams.set("owner", address)
+  requestUrl.searchParams.set("filters[]", "SPAM")
+
+  // TODO validate data with ajv
+  const result = await (await fetch(requestUrl.toString())).json()
+  return result.ownedNfts.filter(
+    // TODO this is a misleading typeguard for unvalidated JSON
```
jagodarybacka, line 335: it is not meant to be a type guard, this filters nfts that shouldn't be visible in the UI

background/lib/nfts.ts
```diff
@@ -0,0 +1,88 @@
+import { AddressOnNetwork } from "../accounts"
+import { EVMNetwork } from "../networks"
+import { NETWORK_BY_CHAIN_ID } from "../constants"
+import { getNFTs as alchemyGetNFTs, AlchemyNFTItem } from "./alchemy"
+import { getNFTsByOwners, SimpleHashNFTModel } from "./simple-hash"
```
jagodarybacka, line 5: What do you think about keeping names the same?
```suggestion
import { getNFTs as simpleHashGetNFTs, SimpleHashNFTModel } from "./simple-hash"
```

jagodarybacka (commented):
Code looks good, I will test it later. 

I would leave Polygon and Ethereum on Alchemy for now as we have tested it quite a lot and we don't want to risk bugs because of new data provider I guess 😅

background/lib/alchemy.ts
mhluongo: We aren't validating that the data is the same shape as `AlchemyNFTItem` though? Not against the actual filter at all, I just want to add validation rather than assume Alchemy gives us the correctly shaped JSON

background/lib/nfts.ts
mhluongo: I like it, though it makes me think the function signatures should be more similar as well.

background/lib/alchemy.ts
mhluongo: Not a great comment either, I'll reword it

mhluongo:
> I would leave Polygon and Ethereum on Alchemy for now as we have tested it quite a lot and we don't want to risk bugs because of new data provider I guess sweat_smile

Definitely — I don't think we should be switching providers without more confidence that this, I just wanted a backup so we aren't stuck without NFT support on our next two chains. While I'm optimistic, it's quite possible it falls over in production due to traffic, or the API churns frequently, etc

background/lib/alchemy.ts
```diff
@@ -289,3 +289,50 @@ export function transactionFromAlchemyWebsocketTransaction(
...
+  const requestUrl = new URL(
+    `https://${
+      network.name === "Polygon" ? "polygon-mainnet.g" : "eth-mainnet"
+    }.alchemyapi.io/nft/v2/${process.env.ALCHEMY_KEY}/getNFTs/`
+  )
+  requestUrl.searchParams.set("owner", address)
+  requestUrl.searchParams.set("filters[]", "SPAM")
+
+  // TODO validate data with ajv
+  const result = await (await fetch(requestUrl.toString())).json()
```
greg-nagy, line 333: I know this is just a refactor, but please add exception handling to the network request + json parsing. Both of them are really error-prone and bit us in the past.

mhluongo: Is it blocking for this PR? I haven't been finding edge cases for Alchemy, just focusing on the new provider. Happy to follow up with a full AJV validation + error handling for both providers, but it's going to bloat this quite a bit.

mhluongo:
Addressed both your concerns @jagodarybacka! Added a follow-on issue and assignment myself for the error handling @greg-nagy https://github.com/tallycash/extension/issues/1927

background/lib/nfts.ts
greg-nagy: 🤩

.env.defaults
```diff
@@ -1,6 +1,7 @@
 DEBUG=false
 ALCHEMY_KEY="oV1Rtjh61hGa97X2MTqMY9kEUcpxP-6K"
 BLOCKNATIVE_API_KEY="f60816ff-da02-463f-87a6-67a09c6d53fa"
+SIMPLE_HASH_API_KEY="tally_sk_428718ba-abc9-453a-af95-fd07d046f115_cp3n5shchhcf05xk"
```
greg-nagy, line 4: Is this a publishable key?

greg-nagy: Note to others: we have access for the simplehash service through `doggos@`

greg-nagy (approved):
Woooo ... \o/

CleanShot 2022-08-12 at 22 26 19

Originally I intended to block this PR on the missing error handling in the alchemy fetch but realised that we handle exceptions in the `fetchThenUpdateNFTsByNetwork` so my original blocker status has softened. Also #1927 is there, so I think we are good here.

I think we should merge asset lists down the road, or use fallbacks instead of `this network then this api`, but that's out of the scope for this PR.

Good to have you back in the codebase OD!

background/lib/alchemy.ts
greg-nagy: softened my stance on this by the the time I have fully finished the PR review

background/lib/nfts.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+function alchemyNFTtoNFT(original: AlchemyNFTItem): NFT {
+  const { contract, chainID } = original
+  return {
+    contract,
+    name: original.title,
+    tokenID: original.id.tokenId,
+    media: original.media
+      .filter((m) => m.gateway)
```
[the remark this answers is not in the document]
mhluongo: I have no idea, just refactoring Alchemy code. All I've done is add SimpleHash and the indirection layer so far!

.env.defaults
mhluongo: Yes it is! I have a prod key to set up in CI

PR #1928, opened 2022-08-12
Re-run Prettier

0xDaedalus:
Looks like we have some files that are not formatted correctly - this fixes that

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,80 +4,80 @@ on:
   pull_request: {}
   push:
     branches:
-    - main
+      - main
     tags:
-    - v*
+      - v*
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
-    - name: Read .nvmrc
-      run: echo "::set-output name=NVMRC::$(cat ./.nvmrc)"
-      id: nvm
-    - name: Use Node + Yarn
-      uses: actions/setup-node@v2
-      with:
-        node-version: "${{ steps.nvm.outputs.NVMRC }}"
-        cache: "yarn"
-    - run: yarn install --frozen-lockfile
-    - name: Dev build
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -4,80 +4,80 @@ on:
   pull_request: {}
   push:
     branches:
-    - main
+      - main
```
greg-nagy, line 7: It had happen to me that prettier broke yml files. I don't think this is the case, and even if it is we can figure out after merge.

greg-nagy (changes requested):
Please don't format the generated validators. They are added them to the project for convinience but they are generated files and we use only their ts version.

mhluongo:
Meta: when you do stuff like this, consider [adding the commit hash to the git blame ignore file](https://akrabat.com/ignoring-revisions-with-git-blame/)

Shadowfiend:
Meta meta: if you run into something like this, consider checking why `yarn lint` didn't catch it. In this case, I'm pretty sure you ran `prettier` directly instead of `eslint .`, which is how we run prettier here. The eslint config has the ignores that apply to the project, including the `validate` directory. None of our lint calls hit HTML at the moment—if we want to do that we should add a script for it and make sure the workflows run it.

The YAML formatting should be uses the `**/*` glob which probably doesn't hit hidden directories like `.github`—may be worth explicitly adding `.github/**/*` or perhaps even seeing about `.*/**/*` or so.

`Signing/index.tsx` is definitely from my signing work, and I'm not so clear on why it didn't get hit. Local checking indicates `eslint` isn't doing much with that file 🤔

0xDaedalus:
> Signing/index.tsx is definitely from my signing work, and I'm not so clear on why it didn't get hit. Local checking indicates eslint isn't doing much with that file 🤔

Looks like its because we've disabled eslint for that file - turns out its just a whitespace issue.  I'll go ahead and close this PR since the other files affected are ignored by eslint - and we can worry about formatting Signing/index.tsx once we enable linting for that file

PR #1930, opened 2022-08-12
🔬 Overview upgrade

### LINKED ISSUE #1932: Overview page - implement new design
- [ ]  total balance header has updated design
- [ ] accounts list
- [ ] collapsible accounts list if > 3 elements
- [ ] networks chart

image

### LINKED ISSUE #1933: Overview page - display data in Networks chart
Show real data about accounts and networks:
- Networks chart: sum of all addresses' balances on the given chain 
image

### LINKED ISSUE #1934: Overview page - display data in Accounts list
Show real data about accounts and networks
- Accounts table: each address should only show up once as its own row. Each row's balance should be total aggregate balance across all chains
image

jagodarybacka:
Resolves #1934
Resolves #1933
Resolves #1932

### What

Create new overview page with accounts and network stats.

image

### Testing

Add different kind of account, regular, read only, Ledger, with high and low balance on Polygon and Ethereum and check if list and chart is displaying correct data.


mhluongo:
Screenshot from 2022-08-17 17-07-13

The data looks right, but the page feels really crowded... not sure if that's the design.

The NFT section not scrolling with the page makes it really difficult to use... that works out on the wallet page because it still gets 2/3rd of the screen real estate, but with 1/3 I can't easily browse NFTs on the overview page.

Also seems like we should order accounts by their total value, largest to smallest. Eg in the above, it feels backward. If I had 40 addresses the whole screen would be the smallest portfolios.

mhluongo:
Here's what happens when I add `vitalik.eth` to the above

Screenshot from 2022-08-17 17-12-51

jagodarybacka:
@mhluongo 
> The data looks right, but the page feels really crowded... not sure if that's the design.

Design was made for 2 networks, 4 networks look crowded indeed, will ask @VladUXUI what to do with it, now I'll try to give the legend more space. Maybe we can also hide networks with 0%? As they are not even on the chart

> The NFT section not scrolling with the page 

Yeah, NFTs scrolling definitely needs some love, not sure if this is a blocker

> order accounts by their total value

good idea!

> when I add vitalik.eth

Isn't it expected? Vitalik.eth account is in the collapsed part (if you are asking where it is), you can toggle it with "view all". It is also very large account so  the rest of accounts' balances are <1% of all accounts in the wallet

mhluongo:
> Isn't it expected? Vitalik.eth account is in the collapsed part (if you are asking where it is), you can toggle it with "view all". It is also very large account so the rest of accounts' balances are <1% of all accounts in the wallet

Oh, I just meant... it's so important it should be up top! Just the same point about descending sort order.

> not sure if this is a blocker

It's almost bad enough that I'd rather see the NFT tab not there... but yeah, not my call, and doing that in another PR of course makes sense.

jagodarybacka:
Scrolling NFTs list is fixed, images are lazy loading thanks to @greg-nagy 💖

0xDaedalus (approved):
👑 👸 📈

PR #1935, opened 2022-08-12
Siwe While You Sign: Consistently fix all prefixed data signing

### LINKED ISSUE #1573: Handle string message for personal_sign
### Discord Discussion Link

https://discord.com/channels/808358975287722045/982241852042387517/982257225965928468

Chrome

Use `personal_sign` with a human-readable string.

### What did not work?

Different result compared to other wallets.

Other wallets detect when the input is not a hex string and convert to hex (and assume UTF-8 encoding)

One example: https://github.com/cardstack/cardwallet/blob/1c334864dc7fc919cb7bf7a5dc68aaf9c97edb54/cardstack/src/parsers/signing-requests.ts#L216

Another: https://github.com/MetaMask/metamask-extension/blob/43bd885be7e78f24c7edae594d88721b22a02fc9/app/scripts/lib/personal-message-manager.js#L332

web3.js also does the same: https://github.com/ChainSafe/web3.js/blob/63e73bb3f8d4720edcc31441c3b613758ad96cec/packages/web3-core-helpers/src/formatters.js#L227

Personally, I feel that bad input should be rejected as a best practice in programming, but I also didn't see clear 'official' JSON RPC documentation for `personal_sign` indicating this must be hex.
The closest I see is for `eth_sign` which shows hex in examples: https://eth.wiki/json-rpc/API#example-21

Should we give in and perform the same handling, or is there some documentation to enforce the data representation?

### Version

Community Edition v0.13.7

### Relevant log output

_No response_

### LINKED ISSUE #1763: Swaps on Sushiswap with Sushiguard are failing
### **Build Info/PR/Commit:**

https://github.com/tallycash/extension/tree/polygon-qa-3
### **Reproduction Steps:**

1. Be on Ethereum chain
2. Swap

### **Expected Result:**
Successful

### **Actual Result:**

Odd characters in Signature, failed. 

https://www.awesomescreenshot.com/video/9857632?key=cad66cc8ea1712cf292a368a0b3c1a1b

### LINKED ISSUE #1915: SIWE not working
### Discord Discussion Link

na

Chrome

Sign in with Ethereum on login.xyz & mintkudos.xyz

### What did not work?

Sign in with Ethereum doesn't work. 

Dev feedback from mintkudos side: _"The 401 error is caused by the fact that the signature verification isn't working for the Tally wallet specifically. We use ethers.utils.verifyMessage to verify the signature against the login message, and the verification is invalid for signatures generated by the Tally wallet it seems.

Essentially we want the user to sign a message to login and then we verify in the backend to make sure that it was signed by the particular account. For some reason the Tally wallet shows the message to be signed in hex? Not sure if this is the reason but it's never been a problem with any other wallet providers that we've tested with!"_

With login.xyz can't connect at all. I've tested an older wallet version, 0.13.13, and there it still works. Screenshot of the sign message there below.

image

### Version

Community Edition v0.14.3

### Relevant log output

_No response_

Shadowfiend:
This is a second attempt at the fair shake #1826 gave the problem.

With some tremendous sleuthing by @beemeeupnow in #1573, we determined
an original sin in the extension was always treating the data passed to
`SigningService.signData` and callees as if it were a UTF-8 message string,
rather than hex bytes. This worked fine as long as the assumption was
true, but any attempts to sign actual data would fail. An incomplete
analysis led to #1826, which changed the signing pipeline to *always* pass
hex data to `SigningService.signData`. This decoupled the signing information
displayed to the user from the actual bytes being signed (allowing for
semantic enrichment), but in the process made it so *all* signatures failed,
as the underlying calls signed the hex string as if it were a regular message,
rather than the bytes represented by the hex string.

#1921 reverted the above change so that at least calls that sent regular
messages, even if hex encoded, would be signed correctly.

This PR fixes all problematic scenarios by:
- Depending on `@tallyho/hd-keyring` 0.4.0, which added the `signMessageBytes`
  call on keyrings to allow directly signing bytes instead of a message string.
- Updating `signData` and its callees to only accept and deal with hex strings.
- Reverting the previous revert, thus restoring the behavior where the `signData`
  call always receives hex from the internal Ethereum provider.

The final frontier was dealing with an issue where a Sign In With Ethereum request
was issued using Ethers, the `personal_sign` data came in as a hex string
representing the UTF-8 Sign In With Ethereum string. When the `parseSigningData`
function passed this hex string to the Siwe library, the Siwe library did not
recognize it as a valid Sign In With Ethereum string, and the extension did not
provide the enriched Sign In With Ethereum view of the data.

To fix this, the `parseSigningData` function heuristically normalizes hex data
to its UTF-8 string equivalent in cases where:
- The data is longer than 32 bytes (excluding most hashes).
- The data can be interpreted as UTF-8, then re-serialized to a UTF-8 byte hex string, without changing.

Fixes #1915, fixes #1573.

This PR notably does _not_ handle #1763 . That issue is related to supporting unprefixed signatures, direct signatures of bytes, which this PR does not implement. Instead this PR fixes all versions of signatures that include the Ethereum message signing prefix (`\x19Ethereum Signed Message\n`), and fixes displaying UTF-8 data that is sent for signing as hex bytes.

## Testing

The easiest way to test is to test data signing. A couple of samples:
- [ ] Go to https://login.xyz/, Click the very top Get Started button, and hit Sign In With Ethereum. You should see a signature confirmation from the extension with the top header “Sign in with Ethereum” and the rich Sign In With Ethereum data. Signing should complete the sign-in in the dApp.
- [ ] Go to https://mintkudos.xyz/. Switch your extension to Polygon. Click “Send Kudos”. You should get a prompt to sign a message from Kudos. Signing should take you to https://mintkudos.xyz/create .


background/services/keyring/index.ts
```diff
@@ -17,8 +18,9 @@ import {
   EIP191Data,
   EIP712TypedData,
   UNIXTime,
-} from "../../types"
```
greg-nagy, line 20: (Just to save the log lookup) This probably should stay here.

Shadowfiend: Uggghhh. Should be fixed.

background/utils/signing.ts
```diff
@@ -87,7 +88,32 @@ export const parseSigningData: (signingData: string) => {
...
+      // Below, if the signing data is not a valid UTF-8 string, we move on
+      // with an undefined possibleMessageString.
+      // eslint-disable-next-line no-empty
+    } catch (err) {}
+
+    // If the hex was parsable as UTF-8 and re-converting to bytes in a hex
+    // string produces the identical output, accept it as a valid string and
+    // set the interpreted data to the UTF-8 string.
+    if (
+      possibleMessageString !== undefined &&
```
greg-nagy, line 109: Why not put this simply in the try block?

background/utils/signing.ts
```diff
@@ -87,7 +88,32 @@ export const parseSigningData: (signingData: string) => {
...
-  const data = checkEIP4361(signingData)
+  let normalizedData = signingData
+
+  // Attempt to normalize hex signing data to a UTF-8 string message. If the
+  // signing data is <= 32 bytes long, assume it's a hash or other short data
+  // that need not be normalized to a regular UTF-8 string.
+  if (signingData.startsWith("0x") && signingData.length > 66) {
+    let possibleMessageString: string | undefined
+    try {
+      possibleMessageString = toUtf8String(signingData)
```
greg-nagy, line 99: If you pass in a `false` or any value as a 2nd parameter to the `toUtf8String()`, it's possible to suppress the throwing.  

https://github.com/ethers-io/ethers.js/blob/master/packages/strings/src.ts/utf8.ts#L290
https://github.com/ethers-io/ethers.js/blob/608864fc3f00390e1260048a157af00378a98e41/packages/strings/src.ts/utf8.ts#L105

greg-nagy (approved):
works well, fixes the manifesto dapp problems, awesome debugging and in depth analysis

kudos ... you know ... because we can (bad joke eel™️)

background/utils/signing.ts
Shadowfiend: The default error func throws, right? https://github.com/ethers-io/ethers.js/blob/608864fc3f00390e1260048a157af00378a98e41/packages/strings/src.ts/utf8.ts#L56

We can suppress by passing our own, but it's not clear we want that. We use the thrown error to note that the conversion failed. If `onError` could trigger returning a marker value we might go that way, but the correct behavior when there's an error for this function is to immediately stop trying to convert (via the internal `throw`) and not replace `possibleMessageString` (via the empty `catch`).

background/utils/signing.ts
Shadowfiend: I generally try to avoid using try/catch for flow control; in this case it's impossible to avoid, but I wanted to keep the use contained to the area where it's strictly required. As soon as whole blocks go into a `try` you increase the risk of catching exceptions you did not realize were in the block and continuing as if they were exceptions you expected. It's rough.

PR #1938, opened 2022-08-12
Fix NFT slideup bug introduced by Alchemy refactor

mhluongo:
The lack of a `chainID` was breaking the NFT detail view 😅

Screenshot from 2022-08-12 15-24-35

# To Test

- [ ] Load an account with NFTs on mainnet or Polygon
- [ ] Confirm that clicking on an NFT for a detail view doesn't lead to an error page


0xDaedalus (approved):
🚢

PR #1950, opened 2022-08-15
Remove erroneous Trezor derivation path

### LINKED ISSUE #1943: Using Ledger with Trezor Seed Phrase / Derivation Path does not recognize Accounts 2+
### Discord Discussion Link

_No response_

Chrome

I have a Ledger using a Trezor Recovery Seed Phrase (I'm guessing Ledger and Trezor use a different derivation path)

When connecting that Ledger to Tally, it recognizes the first Account, but the second+ are all different to the ones I use with Trezor, or even as they are recognized in the Ledger Live App.

This happens using the Ledger Derivation Path:
image

As I ask to change the derivation path to "Trezor m/44'/61'/0'/0), it sends me back to the "Check your Ledger" screen
image
image

I'm stuck in this loop, unable to access the second Trezor generated address

### What did not work?

Accessing the Trezor derivation path using a Ledger

This happens in Version: 0.14.3_72cdb16

### Version

Community Edition v0.14.3

### Relevant log output

_No response_

mhluongo:
The path we were using works with MEW's derivation notation, but was
erroring on a Ledger because... *drumroll*... it's for ETC.

https://blog.trezor.io/trezor-integration-with-myetherwallet-3e217a652e08
https://www.reddit.com/r/TREZOR/comments/81qffg/comment/dv4exam/

Closes #1943

### DIFF
--- a/ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
+++ b/ui/components/Onboarding/OnboardingDerivationPathSelect.tsx
@@ -18,15 +18,11 @@ const initialDerivationPaths: Option[] = [
   },
   {
     value: "m/44'/60'/0'/0",
-    label: "Ethereum",
+    label: "BIP 44 (Trezor, MetaMask)",
   },
   {
     value: "m/44'/1'/0'/0",
     label: "Ethereum Testnet",
-  },
-  {
-    value: "m/44'/61'/0'/0",
-    label: "Trezor",
   },
   {
     value: "m/44'/60'/0'",


Shadowfiend:
😬

jagodarybacka:
@mhluongo I'm not sure what are testing steps here, do we need both Ledger and Trezor to check this PR? Not sure who has Trezor in the team, I don't :(

mhluongo:
>  I'm not sure what are testing steps here

Great question.

How @ramaruro found this in #1943 was by loading a recovery phrase originally generated on Trezor onto a Ledger, then connecting that Ledger to the extension.

If no one has a Trezor handy, we could either...
* ship this to a user that has one and confirm that it works, or...
* generate a recovery phrase using another wallet with the documented Trezor derivation path, load that into a Ledger, then confirm this path works when connecting it to the extension

As this is the first time this has come up, I think as long as there are no regressions, we can probably release this + follow up with users after.

jagodarybacka (approved):
I think it won't break anything so let's ship it

PR #1955, opened 2022-08-15
Implement Optimism TX Signing on Ledger

0xDaedalus:
### To Test
- [ ] Attempt to swap Optimism assets on Uniswap while using a ledger.

### DIFF
--- a/background/services/chain/utils.ts
+++ b/background/services/chain/utils.ts
@@ -94,13 +94,12 @@ export function ethersTransactionRequestFromEIP1559TransactionRequest(
 export function ethersTransactionRequestFromLegacyTransactionRequest(
   transaction: LegacyEVMTransactionRequest
 ): EthersTransactionRequest {
-  const { to, input, from, type, nonce, gasPrice, value, chainID, gasLimit } =
+  const { to, input, type, nonce, gasPrice, value, chainID, gasLimit } =
     transaction
 
   return {
     to,
     data: input ?? undefined,
-    from,
     type: type ?? undefined,
     nonce,
     gasPrice,
--- a/background/services/ledger/index.ts
+++ b/background/services/ledger/index.ts
@@ -384,14 +384,15 @@ export default class LedgerService extends BaseService<Events> {
         }
 
         if (
-          typeof tx.maxPriorityFeePerGas === "undefined" ||
-          typeof tx.maxFeePerGas === "undefined" ||
-          tx.type !== 2
+          tx.type !== 0 &&
+          tx.type !== 1 &&
+          tx.type !== 2 &&
+          tx.type !== null
         ) {
-          throw new Error("Can only sign EIP-1559 conforming transactions")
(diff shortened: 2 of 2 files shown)


mhluongo (approved):
Code looks good, and Uniswap works on both mainnet and Optimism using a Ledger!

background/services/ledger/index.ts
```diff
@@ -384,34 +384,36 @@ export default class LedgerService extends BaseService<Events> {
...
-          typeof tx.maxPriorityFeePerGas === "undefined" ||
-          typeof tx.maxFeePerGas === "undefined" ||
-          tx.type !== 2
+          tx.type !== 0 &&
+          tx.type !== 1 &&
+          tx.type !== 2 &&
+          tx.type !== null
         ) {
-          throw new Error("Can only sign EIP-1559 conforming transactions")
+          throw Error(`Unknown transaction type ${tx.type}`)
```
Shadowfiend, line 392: I don't... believe this is a correct usage of `Error`. It's meant to be used through an instantiating expression afaik.

background/services/ledger/index.ts
```diff
@@ -384,34 +384,36 @@ export default class LedgerService extends BaseService<Events> {
...
           r: tx.r,
           s: tx.s,
           v: tx.v,
-
           blockHash: null,
           blockHeight: null,
           asset: transactionRequest.network.baseAsset,
           network: transactionRequest.network,
-        }
+        } as const // narrow types for compatiblity with our internal ones
```
Shadowfiend, line 416: Why do this instead of declare the type, which will error at the declaration site instead of the return expression and doesn't require the extra comment to boot?

background/services/chain/utils.ts
```diff
@@ -94,13 +94,12 @@ export function ethersTransactionRequestFromEIP1559TransactionRequest(
...
   transaction: LegacyEVMTransactionRequest
 ): EthersTransactionRequest {
-  const { to, input, from, type, nonce, gasPrice, value, chainID, gasLimit } =
+  const { to, input, type, nonce, gasPrice, value, chainID, gasLimit } =
     transaction
 
   return {
     to,
     data: input ?? undefined,
-    from,
```
Shadowfiend, line 103: How come we choose to intentionally lose data here?

Shadowfiend (commented):
Couple of questions despite the merge.

background/services/ledger/index.ts
0xDaedalus: This was chosen to be consistent with https://github.com/tallycash/extension/blob/main/background/services/chain/utils.ts#L293

background/services/chain/utils.ts
0xDaedalus: `ethers.utils.serializeTransaction` errors if given a non EIP-1559 transaction with a `from` field.

background/services/ledger/index.ts
0xDaedalus: 👍 ty

PR #1956, opened 2022-08-15
Standardize coingeckoPlatformID

0xDaedalus:
remove coingeckoPlatformID from `CoinGeckoAsset` and add it to `Network` instead.


mhluongo (approved):
:+1:

PR #1960, opened 2022-08-16
Ensure that thrown errors are instantiated with "new"

0xDaedalus:
Cleanup some errors that were being thrown without the `new` keyword.


mhluongo (approved):
:muscle:

Issue #1895, opened 2022-08-07
Green screen when wanting to SIWE while extension is on Polygon


Naxsun:
### Discord Discussion Link

na

### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

Trying to connect to Zilly.social

### What did not work?

Extension can connect while it was on Polygon but then attempting to SIWE gives a green screen. When switching to Ethereum it then works as in I can sign the message.

The option 'Sign in with Ethereum' remains displayed on the dapp though.

[logs_v0_14_2__2022-08-07T08_42_15+02_00.txt](https://github.com/tallycash/extension/files/9276309/logs_v0_14_2__2022-08-07T08_42_15%2B02_00.txt)

Screen recording & logs:

https://user-images.githubusercontent.com/71384922/183278686-459064e0-f8e2-4011-9635-c18c37420d7a.mov

 

### Version

Community Edition v0.14.2

beemeeupnow:
We also have a similar issue going the other direction.

For example, https://mintkudos.xyz expects Polygon. If you have Ethereum selected in the extension, you will end up with a blank green screen instead of the signature approval page.

PR #1961, opened 2022-08-16
Orthogonal Networks: Don't notify dApps when extension account changes

### LINKED ISSUE #1846: Don't change network for dApps when the network is changed for the extension
* https://matrix.to/#/!WZRftnsYNnOtzOGuZy:thesis.co/$AO0U9C2IkgGRwR4m0Q8bX4K5B_LstdurVxbJi9cSR2k?via=thesis.co&via=matrix.org 
* https://github.com/tallycash/extension/blob/c934f88f075b0ca71de889923d02db193549902f/background/services/provider-bridge/index.ts#L268 

⚠️ My original assessment was wrong! For the real reason see https://github.com/tallycash/extension/pull/1963#issuecomment-1217990405 ⚠️

mhluongo:
Switching networks in the extension shouldn't impact different dApp connections.

# To Test

- [x] Visit app.uniswap.org and connect to Tally Ho
- [x] Change Tally Ho to Polygon, and make sure Uniswap hasn't switched networks
- [x] Change Tally Ho back to Ethereum, then change Uniswap to Polygon. Ensure the extension is still set to mainnet

Closes #1846

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1165,10 +1165,6 @@ export default class Main extends BaseService<never> {
         addressNetwork
       )
       this.store.dispatch(setReferrerStats(referrerStats))
-
-      this.providerBridgeService.notifyContentScriptsAboutAddressChange(
-        addressNetwork.address
-      )
     })
 
     uiSliceEmitter.on(


greg-nagy:
@0xDaedalus @mhluongo have you tested this PR? I am running a test on main after this is merged, and the network is still changing for me. 

The other thing is that this codepath is used only for notifying the dapps about the account change. See the parameters. The network part has been reworked [to use the network that's set for the dApp](https://github.com/tallycash/extension/blob/fdc7daafe442bbd0ee5e4ac00ccaf203c5363a81/background/services/provider-bridge/index.ts#L273)

I know I have suggested this as the source of the problem in the element thread, but I was mistaken.

mhluongo:
Hmmm, I did spend a fair amount of time testing on Uniswap. This patch + refreshing the extension and hard-refreshing the app.uniswap.org got the behavior I believe we want.

Can you give steps to reproduce the issue you're seeing @greg-nagy?

mhluongo:
The underlying belief for me here is that we don't *want* the dApp network **or** account to change when the extension network or account changes. The idea being that dApps have a "switch network" button, and to change account you can disconnect and then reconnect with a new account.

0xDaedalus:
Excellent find - I think in some ways this is a product question here of "should we support dapps without a switch network button".  They seem to be in the minority as crypto moves more and more towards a multi-network paradigm but it sure would be unfortunate if there was a viral dapp or airdrop that we were unable to support because of this issue.

Shadowfiend:
I think the actual product question is “should we tie our extension's primary network to the dApp's network bidirectionally to deal with dApps without a switch network button”. We can support dApps that don't have a switch network button in a way other than updating the main network the extension displays, for example with a network switcher in the lightning menu. Not saying that's the right thing, just saying there are ways to solve the dApp-without-a-network-switcher problem without ALSO tying the dApp's network to the extension UI's network.

Presumably if the dApp's network is fully attached to the extension UI's network, the extension should also display the network for the dApp in the current foreground tab and window, which might change from tab to tab or window to window. There should never be a scenario where the extension UI's network doesn't match the dApp's expected network (the source of #1895, presumably?).

greg-nagy:
I went ahead and implemented a fix for the original problem. @Shadowfiend 's thoughts are correct and deep, but I think we can also get an answer to this question by trying out this fix and see if we have bug reports. 

The reason I did this is that I felt that dapps without network swithcer are rare, but we will see.

PR #1963, opened 2022-08-16
Revert "Orthogonal Networks: Don't notify dApps when extension account changes"

greg-nagy:
Reverts tallycash/extension#1961

https://github.com/tallycash/extension/pull/1961#issuecomment-1216858204

> So the situation after this PR: network is still changing for the dapps if the extension changes network but we have lost the change account functionality.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1165,6 +1165,10 @@ export default class Main extends BaseService<never> {
         addressNetwork
       )
       this.store.dispatch(setReferrerStats(referrerStats))
+
+      this.providerBridgeService.notifyContentScriptsAboutAddressChange(
+        addressNetwork.address
+      )
     })
 
     uiSliceEmitter.on(


0xDaedalus (approved):

mhluongo:
We need to slow down on these PRs, this process doesn't make sense.

@greg-nagy how are we supposed to test this, and what's the desired behavior? What went wrong last time, in reproducible steps?

greg-nagy:
@mhluongo Mae culpa, I thought the linked comment was descriptive enough.

What we want: When a user changes an account in the extension it should trigger account change in the dapps as well. (Until we have a proper multi account dapp support with UI)

What we don't want: When a user change a network in the extension it should NOT trigger network change in the dapps. Because we have decoupled the network for the dapps from each other and from the extension. 

Current behavior before the original PR and after the revert PR is merged: 
- account is changed in the extension > account is changed in the dapp 
- network is changed in the extension > network is changed in the dapp 

CleanShot 2022-08-17 at 01 27 02

Behavior with #1961 
- account is changed in the extension > account is _NOT_ changed  in the dapp
- network is changed in the extension > network _IS_ changed in the dapp 

CleanShot 2022-08-17 at 01 32 11

About the quick revert: I am sorry, this is not how it should have happened. But we need to cut a release for the sign fixes to get into the chrome store in time for the petition dapp, that's why I made the call and asked @0xDaedalus for a quick review.

mhluongo:
Got it, I didn't realize we were waiting on new UI for the account change behavior! Can't wait to break this modality, it's really annoying when using multiple dApps.

mhluongo:
Re: the original testing, here's a gif of the functionality working as (I) expected on the `orthogonal-networks` branch. This wasn't a testing snafu, IMO, it was a product understanding misalignment.

uniswap-network-switcher-test

greg-nagy:
> Re: the original testing, here's a gif of the functionality working as (I) expected on the orthogonal-networks branch. This wasn't a testing snafu, IMO, it was a product understanding misalignment.

Wow, I was scratching my head for a while because of this. Your recording works as expected. The good news is that it helped to pinpoint the source of the problem. The difference between my and your recording is that you used the network selector on uniswap first. 

Every dApp uses the default network until its built-in network selector is used for the first time. The default network is changed by the network selector in the extension. 

[`getActiveOrDefaultNetwork` ](https://github.com/tallycash/extension/blob/1f89fe78f3cbf71e2e258a1f418eb3d0a3d02ecc/background/services/internal-ethereum-provider/index.ts#L297)  <  [`notifyContentScriptsAboutAddressChange`](https://github.com/tallycash/extension/blob/1f89fe78f3cbf71e2e258a1f418eb3d0a3d02ecc/background/services/provider-bridge/index.ts#L273)  < [`newSelectedAccount` ui listener](https://github.com/tallycash/extension/blob/1f89fe78f3cbf71e2e258a1f418eb3d0a3d02ecc/background/main.ts#L1155-L1172) < [`setNewSelectedAccount`](https://github.com/tallycash/extension/blob/1f89fe78f3cbf71e2e258a1f418eb3d0a3d02ecc/background/redux-slices/ui.ts#L150) < [`setSelectedNetwork`](https://github.com/tallycash/extension/blob/1f89fe78f3cbf71e2e258a1f418eb3d0a3d02ecc/background/redux-slices/ui.ts#L156-L169)

The possible solutions I can think of 
1) modifying `setSelectedNetwork` 
  - introduce a new action for it which listener doesn't modify the contentscripts
  - add a parameter for the payload and skip contentscript notification based on it 
2) filter in the listener and only notify the content scripts if it was an account change
3) create a new entry for every dapp on first interaction in the db and decouple it from the default network after the first call

I think the proper fix would be 3. the quickest would be 1. (2. would be my least preferred)

> Got it, I didn't realize we were waiting on new UI for the account change behavior! Can't wait to break this modality, it's really annoying when using multiple dApps.

Yeah, I really would like to see this to be finished properly too, but don't want to push our product priorities.

PR #1965, opened 2022-08-16
Remove some redundant optimism block estimate calculations

0xDaedalus:
Calculate the optimism block estimate once and reuse it rather than calculating three times.


jagodarybacka (approved):
✅

greg-nagy (commented):
Optimising a hack ... this is how products evolve sustainably :)

PR #1969, opened 2022-08-17
▦ Orthogonal Networks Season 2: Don't change network for the dapps when the network for the extension changes

### LINKED ISSUE #1846: Don't change network for dApps when the network is changed for the extension

greg-nagy:
save network for dapps on first encounter

Switching networks in the extension
shouldn't impact different dApp connections.

Without this the network is changed for dapps
with the network selection in the extension
until dapp's networkswitcher is used for the
first time.

## Test

- [x] Visit app.uniswap.org and connect to Tally Ho
- [x] Change Tally Ho to Polygon, and make sure Uniswap hasn't switched networks
- [x] Change Tally Ho back to Ethereum, then change Uniswap to Polygon. Ensure the extension is still set to mainnet
- [x] Change network in uniswap. Ensure the network in the extension is unaffected.
- [x] Change accounts in the extension. Ensure that the account is changed for uniswap.

Closes #1846

This PR builds on the findings of #1961 and #1963


greg-nagy:
cc @mr-michael 

These thoughts and concerns are also valid here: https://github.com/tallycash/extension/pull/1961#issuecomment-1216982723

TL;DR: This change will remove the extension dapp network <> dapp network connection. This is a good thing for those dapps with their own network selector. For those that don't have that, it will remove the possibility of changing the network.

Is this something that we are ok with?

mhluongo:
Yay!!!

mhluongo:
orthogonal-networks-test

mhluongo (approved):
Simple, and works like a charm. Approving but leaving unmerged to make sure the team is aligned

Shadowfiend:
(Despite my comment trying to lay out the options linked above, I'm 100% in favor of this approach.)

0xDaedalus:
I think this is happening because we persist the dapp connections in indexdb and don't flush them if an account is removed -  probably something we should do tbh.

0xDaedalus:
FYI I think merging this would prevent users from changing networks on sushiswap.

greg-nagy:
> I have this behavior, if you connect and disconnect and then try to connect again but you start on a different network - uniswap is changing it's network, it is happening even after wallet reinstall so maybe it is on Uniswap side?

It's because we store the currently active network in extension as a default for the dapp almost immediately on dapp load. After this, the only way to change the network for the dapp is through its own network switcher. But the network switcher won't have any effect until permission is granted for the dapp.  

Fresh install > Import wallet 
  - set wallet to polygon > load app.uniswap > grant permission > uniswap will be on polygon
  - wallet is on ethereum > load app.uniswap > change network on uniswap to polygon > change network in wallet to polygon > connect wallet/grant permission > uniswap is on ethereum

To fix this we either need to allow `wallet_switchEthereumChain` without permission or delay the settings of the network for a dapp to a later point in the flow. Maybe after permission is granted. Until then the dapp uses the active network in the wallet. 

OR 

We need to revert this change and only notify content script on address change (and not on network change)

jagodarybacka:
> would prevent users from changing networks on sushiswap

@0xDaedalus but even on current main sushiswap is only working on Ethereum

greg-nagy:
this PR should have been closed long ago ... because it stores info from every visited website in indexeddb

also #2314 implemented the proper fix

PR #1972, opened 2022-08-17
Only warn about inconsistent EIP-1559 gas prices on EIP-1559 gas networks

0xDaedalus:
This should clean up the logs a fair bit

### DIFF
--- a/background/lib/gas.ts
+++ b/background/lib/gas.ts
@@ -6,7 +6,11 @@ import Blocknative, {
   BlocknativeNetworkIds,
 } from "../third-party-data/blocknative"
 import { BlockPrices, EVMNetwork } from "../networks"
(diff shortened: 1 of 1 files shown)


background/lib/gas.ts
```diff
@@ -150,10 +154,16 @@ export default async function getBlockPrices(
     }
   }
 
-  if (feeData.maxPriorityFeePerGas === null || feeData.maxFeePerGas === null) {
+  EIP_1559_COMPLIANT_CHAIN_IDS.has(network.chainID)
```
jagodarybacka, line 157: hmmm, what is this line doing here?

0xDaedalus: Oof 😳

jagodarybacka (approved):
✅

PR #1973, opened 2022-08-17
Optimism - Live Transaction Fee Updates

0xDaedalus:
Currently we only get a transactions estimated L1 Rollup fee when the transaction is [first added](https://github.com/tallycash/extension/blob/optimism-live-gas-updates/background/services/chain/index.ts#L383).  This means that if a user leaves the `Sign Transaction` window open for a while without interacting with it the fee preview displayed may become out-of-sync with the real fee.  This PR attempts to alleviate that.

### To Test
- [ ] Open up the sign transaction screen on Optimism (For example from a Uniswap swap)
- [ ] Make a note of what the $ amount of the fee is
- [ ] Wait 2 minutes (or less if you've set the GAS_PRICE_POLLING_FREQUENCY env var to lower than 120)
- [ ] See the $ amount of the fee change

Note: If gas prices are stable on Ethereum this may look like its not working despite in fact updating.


background/main.ts
```diff
@@ -690,11 +696,33 @@ export default class Main extends BaseService<never> {
...
-      this.store.dispatch(
-        estimatedFeesPerGas({ estimatedFeesPerGas: blockPrices, network })
-      )
-    })
+    this.chainService.emitter.on(
+      "blockPrices",
+      async ({ blockPrices, network }) => {
+        if (network.chainID === OPTIMISM.chainID) {
+          const { transactionRequest: currentTransactionRequest } =
+            this.store.getState().transactionConstruction
```
0xDaedalus, line 704: I really do not like this but am not sure of a better way to figure out if a transactionRequest currently exists and if its on Optimism - any thoughts here?

mhluongo: Maybe you don't need to know! Can't you just dispatch the action, and handle this in the reducer?

mhluongo: In other words: just push the information periodically from the background script. Don't discriminate on whether we need it right this second, just like any other gas information... we need it constantly.

One day we'll probably want to have a way to slow stuff like that down, but it's way easier than adding a full request/response cycle or having the background script awkwardly "look into" the redux store like this

mhluongo (changes requested):
I think this is the wrong approach... I could be missing a resource constraint though

background/main.ts
0xDaedalus: The issue is that in order to retrieve the information from the background script we need access to the transaction's `input` so we need to maintain that state somewhere and retrieve.  Right now I believe that the only place we maintain it is in redux - so its a choice between maintaining and mirroring tx state in a service - or doing this awkward pulling - both are not great tradeoffs tbh.

mhluongo:
@Shadowfiend ideas here? This solution is ikrsome... maybe we need some sort of tx request subscription from the frontend?

background/main.ts
```diff
@@ -690,11 +695,33 @@ export default class Main extends BaseService<never> {
...
+          if (
+            currentTransactionRequest?.network.chainID === OPTIMISM.chainID &&
+            !isEIP1559EnrichedTransactionSignatureRequest(
```
jagodarybacka, line 706: Do I get it right that this won't be true when they will ship bedrock release? https://help.optimism.io/hc/en-us/articles/8309198945179-What-will-change-in-bedrock-

jagodarybacka: so will we have to fix it in the future or will it fall back to EIP1559 compatible code and will work out of the box 😂 ?

0xDaedalus: This particular line of code will need to be fixed in the future (we'll just need to remove it) - the rest of the EIP-1559 stuff _should_ more or less work out of the box but we will definitely need to do some testing to make sure that thats the case.

Shadowfiend: Not entirely clear on *why* we exclude EIP1559 here? Right now it can't be optimism AND EIP1559, right? Or is that me misunderstanding?

background/main.ts
```diff
@@ -690,11 +695,33 @@ export default class Main extends BaseService<never> {
...
+              currentTransactionRequest
+            )
+          ) {
+            // If there is a currently pending transaction request on Optimism,
+            // we need to update its L1 rollup fee as well as the current estimated fees per gas
+            const estimatedRollupFee =
+              await this.chainService.estimateL1RollupFee(
+                currentTransactionRequest.network,
+                currentTransactionRequest.input
+              )
```
Shadowfiend, line 716: It feels like we don't have to exit our current structure completely to handle this. In particular, we can change to:

```typescript
  estimatedFeesPerGas: {
    [ETHEREUM.chainID]: EstimatedGradedFeesPerGas | undefined
    [OPTIMISM.chainID]: EstimatedRollupFees | undefined
  }
```

And go from there in terms of structure.

In this scenario, the `estimatedFeesPerGas` action can become an async thunk that handles the L1 rollup fee lookup via a standard contract interaction through the internal provider.

background/main.ts
0xDaedalus: Good point - I think I had added it to satisfy the compiler at one point but looks like its no longer needed.

mhluongo (dismissed):

mhluongo:
Good with this as long as we're coming back to cover the tech debt later. Onward!

PR #1974, opened 2022-08-18
Fix error on NFT preview slideup with no image

jagodarybacka:
### Testing

1. Find NFT with no preview image
2. Click on the NFT

Bug: Unexpected Error
Expected: Opens slideup with selected NFT

image


greg-nagy:
@jagodarybacka just to be on the safe side: when we fetch the nfts from alchemy [we drop every preview that's not coming from the gateway](https://github.com/tallycash/extension/blob/a2d25eb784b0ce2836f532b025058141e74cfb6c/background/lib/nfts.ts#L26). I think even in those cases we can have the `raw` property returned to us and we might be able to use that. 

The fix in itself is ok, but I just would like to verify if we can do anything to improve the experience.

greg-nagy (approved):
let's merge this great usage of stability opportunity

PR #1981, opened 2022-08-18
Fix derivation path label

jagodarybacka:
Fix the label that was too long 
image


0xDaedalus (approved):
Not sure why CI isn't running but LGTM - maybe close & reopen if its still stuck when you get on.

jagodarybacka:
Merging as this was approved before

PR #1984, opened 2022-08-18
Optimism - Show correct gwei estimates to user in gas fee preview button

0xDaedalus:
image

### To Test
- [ ] Attempt to sign a tx on Optimism - it should give a preview of both $ price and gwei that are reasonable.

### DIFF
--- a/background/networks.ts
+++ b/background/networks.ts
@@ -3,7 +3,9 @@ import { Slip44CoinType } from "./constants/coin-types"
 import { HexString, UNIXTime } from "./types"
 import type { CoinGeckoAsset } from "./assets"
 import type {
+  EnrichedEIP1559TransactionRequest,
   EnrichedEIP1559TransactionSignatureRequest,
+  EnrichedEVMTransactionRequest,
   EnrichedEVMTransactionSignatureRequest,
   PartialTransactionRequestWithFrom,
 } from "./services/enrichment"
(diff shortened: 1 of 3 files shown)


background/networks.ts
```diff
@@ -151,6 +153,7 @@ export type LegacyEVMTransactionRequest = Pick<
   chainID: LegacyEVMTransaction["network"]["chainID"]
   gasLimit: bigint
   estimatedRollupFee: bigint
+  estimatedRollupGwei: bigint
```
0xDaedalus, line 156: There is some intentional tech debt here in that we are adding both `estimatedRollupFee` and `estimatedRollupGwei` as mandatory properties of LegacyEVMTransactionRequests.

This is not strictly true - since there are networks that implement LegacyEVMTransactions which do not roll up to Ethereum.  Once we choose to support one of those networks - we'll probably need to split this type into something like `LegacyEVMTransactionRequest` and `LegacyEvmRollupTransactionRequest`.

ui/components/NetworkFees/FeeSettingsText.tsx
```diff
@@ -60,6 +69,43 @@ const getFeeDollarValue = (
...
+    transactionData &&
+    !isEIP1559EnrichedTransactionRequest(transactionData) &&
+    EVM_ROLLUP_CHAIN_IDS.has(network.chainID)
+  ) {
+    estimatedSpendPerGas += transactionData.estimatedRollupGwei
+  }
+
+  const estimatedGweiAmount =
+    typeof estimatedFeesPerGas !== "undefined" &&
+    typeof selectedFeeType !== "undefined"
```
jagodarybacka, line 102: selectedFeeType can be undefined I guess, it will work fine

jagodarybacka (approved):
Works fine, $ value seems ok.

This screen needs some love but it can be done later I guess
image

PR #1985, opened 2022-08-18
Add unit tests to ensure transactions are being populated via the correct method.

0xDaedalus:


background/services/chain/tests/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+          from: "0x0d18b6e68ec588149f2fc20b76ff70b1cfb28882",
+          network: ETHEREUM,
+          nonce: 1,
+          maxPriorityFeePerGas: 1n,
+          maxFeePerGas: 2n,
+        }
+
+      const stub = sandbox.stub(
+        chainService,
+        "populatePartialEIP1559TransactionRequest" as any // `as any` casting here since we are stubbing a private method
```
jagodarybacka, line 35: please turn off the linter rule for this line to avoid warning

0xDaedalus: I think I found a better solution than `any` - let me know what you think

hyphenized: looks like a valid approach 🤔

mhluongo:
Ser a description would be :ok_hand:

mhluongo:
These aren't unit tests quite unit tests — could we also test inputs / outputs for baseline accuracy? The callstack stuff is nice, but we could fail signing pretty hard (returning all 0 signature, etc) and this wouldn't tell us anything

mhluongo:
It's fine IMO if you want to kick that to another PR, but calling out that the commit description...

> Add unit tests that verify that partial transactions are correctly enriched.

Isn't accurate to the code in this PR... as I read it, we aren't verifying correctness.

0xDaedalus:
Fair point - will amend the commit - the intention of these tests was to test the logical branching which is a unit test of the `populatePartialTransactionRequest` function imo.  I think the actual population would be a unit test of the called functions.

mhluongo (approved):
Tests gud :rock:

PR #1989, opened 2022-08-19
Handle NFTs list without react-window

jagodarybacka:
### What
Use lazy image loading instead of list virtualisation library to align scrolling behaviour between tabs

### Testing
 
Add many accounts with a lot of NFTs and check if NFTs lists are working as before without significant performance problems

### DIFF
--- a/ui/components/NFTs/NFTsImage.tsx
+++ b/ui/components/NFTs/NFTsImage.tsx
@@ -40,7 +40,7 @@ export default function NFTsImage({
         height={height}
         borderRadius={8}
       >
-        <img alt={alt} src={imageSrc} />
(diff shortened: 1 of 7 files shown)


mhluongo:
Excited about this!

Shadowfiend:
I'm going to flag--we will probably want a virtualization library for some if not all of this stuff at some point, as it will help us with performance of redux, scrolling, etc for any place that has potentially many items (activity list being the major offender). So let's be thinking in the background about how we might solve some of the issues we ran into with react-window in the mid-range future.

mhluongo (approved):
Great interim solution, thanks for knocking this!

PR #1990, opened 2022-08-19
remove loading skeleton

greg-nagy:


jagodarybacka (approved):
Looking great, now images are loading as we wanted 💖

PR #1991, opened 2022-08-19
Add RSK as "coming soon" 

EvelinaBuiciag:
We are working on tally's support of RSK. Works with SUPPORT_RSK on and off.

image

### DIFF
--- a/background/services/preferences/defaults.ts
+++ b/background/services/preferences/defaults.ts
@@ -23,7 +23,7 @@ const defaultPreferences: Preferences = {
     ],
   },
   currency: USD,
-  defaultWallet: true,
+  defaultWallet: false,
   selectedAccount: {
     address: "",
     network: ETHEREUM,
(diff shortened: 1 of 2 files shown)


mhluongo:
@EvelinaBuiciag do you have a timeline on integration completion? We're thinking this can be merged + released after Arbitrum and Optimism go out, and we can find a way to co-market with RSK and maybe have some folks on a community call — WDYT?

Y'all have been awesome contributors, we're really impressed with the work and consistency!

alepc253:
hi @mhluongo, thanks for your kind words! we are currently working on the integration but we need [these issues](https://github.com/tallycash/extension/issues/1315#issuecomment-1214436623) solved to unlock it and that is why the timeline is on both teams.
Regarding the alternative to Alchemy (re link), we know that this is something that you want to think about carefully in terms of architecture but our idea here has always been to define a fallback schema where calls are resolved by using plain json-rpc requests if Alchemy is not available for the network.
About co-market, of course, more than happy to make it happen and the same about presenting again on a community call.

PR #1992, opened 2022-08-19
Fix optimism transaction enrichment

0xDaedalus:
Despite the [ethers js docs](https://docs.ethers.io/v5/api/providers/types/) stating that receipt.effectiveGasPrice "will simply be equal to the transaction gasPrice" on chains that do not support EIP-1559 - it seems that this is not yet the case with Optimism.
    
This commit fixes a bug where transaction enrichment was failing due to effectiveGasPrice being undefined and calling .toBigInt on it - but is not a perfect solution because transaction.gasPrice does not necessarily take into account L1 rollup fees.

### To Test
- [ ] Submit a transaction on Optimism (e.g. uniswap swap).  You should not see `Uncaught TypeError: Cannot read properties of undefined (reading 'toBigInt')` in the console after submitting


Shadowfiend:
Thoughts on adding a note to the comments on the legacy transaction type to clarify both that they're the base type for optimism and to the `gasPrice` specifically to flag the issue you noted above?

Wondering if in the longer term it would make sense to also add a new field once we can take L1 fees into account, either to represent the L2-only fee or to represent the combined fee, so both pieces of information are available to the system 🤔

0xDaedalus:
Excellent idea - the general comment was added in https://github.com/tallycash/extension/pull/1995 and I've just pushed an update to this PR to codify the `gasPrice` issue.

I do think that longer term we'll need to split prices and fees into l1 and l2 - probably as part of the Arbitrum work.

mhluongo (approved):
:ship:

PR #1901, opened 2022-08-08
Warn before sending funds to a known ERC-20 contract 

hyphenized:
Adds a warning when user inputs a contract address as the destination

warning

Refs #261

PR #1993, opened 2022-08-20
⚠Warn before sending funds to a known ERC-20 contract

hyphenized:
Displays a warning when user inputs a contract address as the destination in the Send Tab

warning

Refs #261
Duplicate of #1901 

## To Test

- [X] Input a contract address like `0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48` (USDC), after loading a warning should appear and the send button should remain disabled.
- [X] Input a normal address, no warning should appear and the send button should work normally again.
- [X] Changing the destination address back to a contract address that previously triggered the warning should not send new requests to the provider until the Send tab is closed.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -77,6 +77,7 @@ import {
   rejectTransactionSignature,
   transactionSigned,
   clearCustomGas,
+  setETHAddressType,
 } from "./redux-slices/transaction-construction"
 import { selectDefaultNetworkFeeSettings } from "./redux-slices/selectors/transactionConstructionSelectors"
 import { allAliases } from "./redux-slices/utils"
@@ -627,6 +628,16 @@ export default class Main extends BaseService<never> {
             })
           )
         }
+      }
+    )
+
+    transactionConstructionSliceEmitter.on(
+      "lookupETHAddressType",
+      async (address: string) => {
+        const isContractAddress =
+          await this.enrichmentService.checkIsEthereumContractAddress(address)
+
+        this.store.dispatch(setETHAddressType([address, isContractAddress]))
       }
     )
 
--- a/background/redux-slices/transaction-construction.ts
+++ b/background/redux-slices/transaction-construction.ts
@@ -58,6 +58,7 @@ export type TransactionConstruction = {
   signedTransaction?: SignedTransaction
   broadcastOnSign?: boolean
   transactionLikelyFails?: boolean
+  ETHAddressLookupCache: Record<string, boolean>
   estimatedFeesPerGas: { [chainID: string]: EstimatedFeesPerGas | undefined }
   customFeesPerGas?: EstimatedFeesPerGas["custom"]
   lastGasEstimatesRefreshed: number
@@ -86,6 +87,7 @@ export const initialState: TransactionConstruction = {
   estimatedFeesPerGas: {},
   customFeesPerGas: defaultCustomGas,
   lastGasEstimatesRefreshed: Date.now(),
+  ETHAddressLookupCache: {},
 }
 
 export type Events = {
@@ -93,6 +95,7 @@ export type Events = {
   requestSignature: SignOperation<TransactionRequest>
   signatureRejected: never
   broadcastSignedTransaction: SignedTransaction
+  lookupETHAddressType: string
 }
 
 export type GasOption = {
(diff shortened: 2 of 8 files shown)


mhluongo:
Tested, and it works! Finally going to give this the deep review it deserves.

Screenshot from 2022-08-20 15-00-57

ui/_locales/en/messages.json
```diff
@@ -29,7 +29,8 @@
     "assetAmount": "Asset / Amount",
     "sendTo": "Send To:",
     "sendButton": "Continue",
-    "receiveAddress": "Receive address"
+    "receiveAddress": "Receive address",
+    "sendToContractWarning": "This is a token address, sending assets will result in loss of funds."
```
mhluongo, line 33: Excellent that this is i18n'able. Note that currently we're checking for a contract address, though, not a token address!

background/main.ts
```diff
@@ -630,6 +631,16 @@ export default class Main extends BaseService<never> {
       }
     )
 
+    transactionConstructionSliceEmitter.on(
+      "lookupETHAddressType",
+      async (address: string) => {
```
mhluongo, line 636: Prefer the `AddressOnNetwork` type so this works across chains! A contract on what chain likely won't be on another

background/main.ts
```diff
@@ -630,6 +631,16 @@ export default class Main extends BaseService<never> {
...
+        const isContractAddress =
+          await this.enrichmentService.checkIsEthereumContractAddress(address)
```
mhluongo, line 638: Instead of *just* checking if the address is a contract, I think you should request an entire annotation here. We have a bunch of other guards we want to add as well.

That said, it'd also make a lot of sense to save this check for one more screen, and rely on the emitted enriched transaction request that's already happening. That way you don't need this extra wiring

background/redux-slices/transaction-construction.ts
```diff
@@ -58,6 +58,7 @@ export type TransactionConstruction = {
   signedTransaction?: SignedTransaction
   broadcastOnSign?: boolean
   transactionLikelyFails?: boolean
+  ETHAddressLookupCache: Record<string, boolean>
```
mhluongo, line 61: Style note: we lowercase any acronyms at the beginning of non-class non-type identifiers, eg variables and map keys. Eg `ethAddressLookupCache`

ui/pages/Send.tsx
```diff
@@ -148,6 +154,35 @@ export default function Send(): ReactElement {
...
+  const isLoadingAddressDetails =
+    currentAccount.network.family === "EVM" &&
+    !!destinationAddress &&
+    typeof contractLookupCache[destinationAddress] === "undefined"
+
+  useEffect(() => {
+    if (
+      destinationAddress &&
+      currentAccount.network.family === "EVM" &&
+      !contractLookupCache[destinationAddress]
```
mhluongo, line 168: Instead of growing the Redux state size, we can rely on the `EnrichmentService` to cache this.

Okay, okay, it's not caching today! But that would be a better place to worry about that.

Rationale: the latency to pull from the background script is low, it's not like hitting an API endpoint. On the other hand, as Redux size grows, the frontend / background script synchronization overhead quickly grows to slow down renders / spike CPU time.

mhluongo (changes requested):
Excellent effort so far! I included some thoughts on how you can do this that's more in line with the service model.

Outside the approach / architectural improvement here, the biggest blocker is that that this PR prevents anyone from sending tokens to contract-based multisigs like Gnosis Safe. I think moving this interaction a screen deeper and making the warning dismissable is the way to go here.

background/main.ts
hyphenized: You're definitely right, thanks!

ui/pages/Send.tsx
hyphenized: Tyvm for the explanation!! I initially wanted to keep the cache in local state but found it a bit difficult to interact with the enrichment service without using the redux store.

background/main.ts
hyphenized: You're right! That'd be much better and simpler as well. I've moved the warning to the next screen, thanks for the feedback!

mhluongo (approved):
Excellent first contribution @hyphenized!

Issue #492, opened 2021-11-30
Use 4byte to look up all unknown transaction function signatures by selector


mhluongo:
Secure signing means users need to know the implication of what they're signing... otherwise the signature doesn't represent user consent. And for most users, a hex string ain't that.

We can use [4byte](https://4byte.directory) to look up function selectors to get additional metadata, making many otherwise inscrutable contract calls somewhat bearable. We can apply this in two places in the wallet to great success

- [ ] Show function signatures before a user signs a transaction
- [ ] Show function signatures in activities when we don't have a better visualization. More context on that problem space [here](https://github.com/tallycash/tally-extension/pull/400#discussion_r756425046)

cc @henryboldi as I know you want to see more tx introspection / activity metadata.

tallysam:
This service no longer exists.

Shadowfiend:
I believe it does fwiw, it's just at www.4byte.directory . The referencing commits seem to use the right format.

PR #1998, opened 2022-08-22
4byte: Enrich transaction requests and activities with ABIs

mhluongo:
This PR covers the background script components of #492. Each transaction and transaction signature request is enriched by looking up its function signature in the [4byte.directory](https://4byte.directory) database. These signatures can be used to (eventually) replace our bland "Contract interaction" interface with something contract-specific.

Here's the data from the last few contract interactions in one of my wallets.

Screenshot from 2022-08-21 23-10-34

While I was at it, I found and fixed a major enrichment bug that replaced contract names for generic contract interactions with the name of the sender.

# To Test

- [ ] Interact with a contract from an address that resolves to an ENS name. Make sure the activity shows the contract info, not the senders.
- [ ] Load a read-only wallet with an interesting history — maybe LP'ing, registering an ENS name, or voting on-chain. Watch the network tab, and you should see `4byte.directory` API calls
- [ ] Confirm the calls aren't for the same hex selector — none should appear twice thanks to caching.
- [ ] Check the Chrome background script console, and confirm that selectors are cached under "IndexedDB > tally/enrichment"
- [ ] Finally, confirm the Redux store has a `functionSignature` for many (though not all) activities with `type === "contract-interactions"`


background/lib/four-byte.ts
```diff
@@ -0,0 +1,62 @@
...
+        results.count !== 0 &&
+        "results" in results &&
+        results.results.length > 0
+      ) {
+        const {
+          id,
+          created_at: createdAt,
+          text_signature: functionSignature,
+          hex_signature: functionSelector,
+        } = results.results[0] as {
```
0xDaedalus, line 41: This line gives me a lot of pause although I'm not sure of a better way to approach the problem than using a service like 4byte - so I'd like to open up a discussion to see if anyone has any ideas.

1).  Its pretty trivial to intentionally collide function selectors - just requires a bit of GPU time.  I would _really_ like to avoid the situation where our wallet shows a contract interaction as "safe" but it is in fact malicious which this change opens us up to.

2).  The converse scenario is also a negative experience - showing the contract interaction as something unrecognized when it is in fact a trusted function on a trusted contract has the potential to cause a lot of user confusion.

If we just take the `n`th result every time we hit 4byte - both of these scenarios are possible.

e.g. as an example - [if we look up the ABI for the ERC-20 `approve` function](https://www.4byte.directory/api/v1/signatures/?hex_signature=0x095ea7b3) - the `0th` result is `sign_szabo_bytecode(bytes16,uint128)`.

mhluongo: Here's the call [sorted properly](https://www.4byte.directory/api/v1/signatures/?hex_signature=0x095ea7b3&ordering=created_at)

mhluongo: Give that another look @0xDaedalus! This is still treating function selectors as unique IDs in the database (boo), but using the oldest one seems like a decent patch for us to start using these while I think through the grand solution outlined above.

background/lib/four-byte.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Calculate a function selector from an ABI fragment.
+ *
+ * @param functionSignature an ABI fragment used to get the function selector,
+ *        in the form `"function(uint256,uint256)"`
+ */
+export function calculateFunctionSelector(functionSignature: string): string {
+  const abi = [`function ${functionSignature}`]
+  const iface = new Interface(abi)
+  return iface.getSighash(Object.keys(iface.functions)[0])
+}
```
Shadowfiend, line 23:
```suggestion
export function calculateFunctionSelector(functionSignature: string): string {
  return FunctionFragment.from(`function ${functionSignature}`).format("sighash")
}
```
Believe this will do the trick, and may honestly invalidate the need for the wrapper function.

background/lib/four-byte.ts
Shadowfiend: One way to potentially help mitigate here is just to try and parse the call data with the ABI. At least then we know the ABI structure is correct for the data, even if you can structure a collision to absorb additional call data in a seemingly innocuous way.

Shadowfiend (commented):
A few quick (not necessarily blocking) thoughts:
- `types.ts` is an anti pattern IMO. The refactoring that's happened recently *should* allow for types to be declared alongside the relevant code instead of in a segregated dumping ground file. 
- `lib` has kind of become a dumping ground of unrelated stuff as well. I suggest putting the 4byte stuff under the enrichment directory (potentially in a subdir 🤷) unless/until we start consuming it directly from elsewhere. 
- Do we have a sense of how often this could fail due to rate limits or otherwise using an API that may not be meant for frequent hits? (See the TOS at the bottom of [the site](https://www.4byte.directory/), and keep in mind we'll run this for every historical transaction over time on potentially many wallets.)

Now the meaty one. IMO @0xdaedalus's point about collisions means this whole code path should be disconnected from enrichment until we've implemented the additional bits you've described @mhluongo. Initially I was going to say it depends on how we want to implement it in the UI (is it a potentially inaccurate aid or a security feature?), but until we know that I feel like it's subject to knowledge rot that will eventually make it seem like a good security feature when it is currently arguably more of a vector. 

I'll leave it there, ultimate call can be someone else's, but (as I've said before) I find this naive approach poor at best.

mhluongo:
> Now the meaty one. IMO @0xDaedalus's point about collisions means this whole code path should be disconnected from enrichment until we've implemented the additional bits you've described @mhluongo. Initially I was going to say it depends on how we want to implement it in the UI (is it a potentially inaccurate aid or a security feature?), but until we know that I feel like it's subject to knowledge rot that will eventually make it seem like a good security feature when it is currently arguably more of a vector.

Did you see my follow-up @Shadowfiend? It wouldn't just have to be a function selector collision... it would also have to be the first time the selector has been seen by the service, ever.

If the concern is newly deployed contracts, we can set a date after which we don't accept 4byte submissions by themselves. Say, today. That way, we can get useful human-readable metadata in the wallet while the bigger metadata push above is implemented... without opening up a vector for manipulation.

Shadowfiend:
The issue is creating a malicious function whose sighash collides with a matching 4byte signature, not vice versa, no?

background/utils/view-model-transformer.ts
```diff
@@ -45,7 +45,14 @@ export const transactionPropertiesForUI = [
   "annotation.spender.annotation.balance",
   "annotation.spender.annotation.nameOnNetwork",
   "annotation.spender.annotation.nameOnNetwork.name",
-  "annotation.contractName",
+  "annotation.contractInfo",
+  "annotation.contractInfo.nameOnNetwork",
+  "annotation.contractInfo.nameOnNetwork.name",
+  "annotation.contractInfo.nonce",
```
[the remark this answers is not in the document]
Shadowfiend: > can

Thus "effectively". They can, but is it useful to know for the system? I hear you re pick list, just weird to add something that has no direct applicability to the UI to a list that is meant to reduce needless data shuttling 😬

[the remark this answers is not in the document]
mhluongo: The alternative is a misleading type though! If you think we should distinguish address annotations from contract annotations, we can introduce a new type?

mhluongo:
All of that is a UI concern though. This PR doesn't include an interface that suggests this information should be trusted as correct — it gives us the information necessary to approach feature parity with MetaMask, which is already vulnerable to that same attack.

You two are suggesting we shouldn't store spoofable information, rather that that we should very carefully design an interface to explain the risk. We aren't holding to that threat model throughout the rest of the extension though, or we wouldn't be relying on eg trusted RPCs.

Happy to rename variables here to make the intent more clear, or include more documentation on collisions, or both...? Also happy to add a feature flag here, and happy to add a cutoff date that solves 40% of the collision issue. But this isn't a UI PR, and solving all of these concerns is going to be way more work than a single PR's worth of review.

Thoughts?

background/utils/view-model-transformer.ts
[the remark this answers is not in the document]
Shadowfiend: We have a type, it just isn't reflective I believe. ActivityItem.

Shadowfiend:
I had the mind flash of a variable rename the other day but forgot it before suggesting. I'd be fine with a rename that made it clear this wasn't a known good value.

0xDaedalus:
Very pro variable rename 👍

mhluongo:
Moving this back to draft and giving it a rethink.

Shadowfiend:
Closing this for now; we'll be doing a rework to pull in transaction simulations soon, and this may factor into that work as a fallback or similar.

PR #2009, opened 2022-08-23
Display the network corresponding to the current transaction when signing.

0xDaedalus:
This PR fixes an issue where the Sign Transaction screen was displaying the wallet's currently selected network on the Sign Transaction screen in cases where the transaction being signed was for a network that differed from the currently selected network.

image

image

### To Test
- [ ] Select `Ethereum` in the wallet
- [ ] Attempt to sign a transaction on Optimism
- [ ] The network shown in the upper-left should be Optimism


jagodarybacka (changes requested):
There is a problem because SIWE and other non-tx signing are not coming with network info, IMO it's not looking good. If network doesn't matter for them maybe there is no point in showing network? 
image

0xDaedalus:
Hmmmmnnnn good catch I wasn't aware we had transactions coming in without a network.  I think the answer here is to ensure that transactions always have a network (corresponding to the dapp connection the tx came in on) - I'll work on this.

jagodarybacka (approved):
LGTM, both txs and messages ✅

PR #2015, opened 2022-08-24
2011: create the change password feature

### LINKED ISSUE #2011: Add the ability for users to change their Tally Ho password.
Requirements:
- [ ] Add an [ALLOW_CHANGE_PASSWORD feature flag ](https://github.com/tallycash/extension/blob/optimism-gas-price/background/features.ts)and gate the following functionality behind it.
- [ ] Add unit tests for any added methods in keyring service.
- [ ] Add functionality to [keyring service](https://github.com/tallycash/extension/blob/optimism-gas-price/background/services/keyring/index.ts) that allows a caller to encrypt unlocked vaults with a newly designated passphrase.
- [ ] Add a "Change Password" option to [`Settings.tsx`](https://github.com/tallycash/extension/blob/optimism-gas-price/ui/pages/Settings.tsx) with formatting similar to the `Bug Report` option.
- [ ] Add a dedicated page for changing a users password.  This page should:
  - [ ] require users to re-enter their password to access it even if the extension is unlocked (probably via the typical lock screen).
  - [ ] Use the [`PasswordStrengthBar`](https://github.com/tallycash/extension/blob/optimism-gas-price/ui/components/Password/PasswordStrengthBar.tsx)
  - [ ] Require the user to enter their new password twice and confirm that the entries match
- [ ] Remove the [copy about not being able to change passwords](https://github.com/tallycash/extension/blob/optimism-gas-price/ui/components/Keyring/KeyringSetPassword.tsx#L74:L84) from the initial password creation flow.

Notes:
- It might be a good idea to modify and reuse the existing [`KeyringSetPassword`](https://github.com/tallycash/extension/blob/optimism-gas-price/ui/components/Keyring/KeyringSetPassword.tsx) component for the Change Password screen.
- Adding unit / integration tests are heavily encouraged while developing feature as it is security-sensitive and not likely to change much once implemented.
- The users password is never (and should never be) persisted.  Password verification happens [here](https://github.com/tallycash/extension/blob/optimism-gas-price/background/services/keyring/index.ts#L187:L196) and will fail if a given password attempting to decrypt a vault does not match the password used to encrypt it.

rgdalessandro:
Closes #2011

NOTE:
This PR should only be merged AFTER this other PR it is dependent on --> https://github.com/tallycash/extension/pull/2128

**Changes:**
- create new feature flag --> `ALLOW_CHANGE_PASSWORD`
- `ui` changes in `routes.ts`, `Settings.tsx` and `KeyringSetPassword.tsx` to create the change password flow
- `background` changes in `main.ts` to listen for the `changePassword ` emitter
- `background` changes to the `keyrings.ts` redux slice to notify the `ui` whether `didPasswordChangeSucceed`
- A few tiny improvements to `.gitignore`, `README.md`, `storage.ts`, `SharedInput.tsx`, `signing-hooks.ts` and `Popup.tsx`

**Before:**
Screen Shot 2022-08-24 at 4 48 52 PM

**After:**
Screen Shot 2022-08-24 at 4 54 53 PM

Screen Shot 2022-08-24 at 4 55 21 PM

**To test:**
- [ ] Set `ALLOW_CHANGE_PASSWORD = true` by adding the variable to your environment or editing the `features.ts` file.
- [ ] Navigate to the Settings page and click the new link for "Change password" to see the Change Password page.
- [ ] Attempt to change the password by providing a random string as the "Current Password" and observe validation errors
- [ ] Attempt to change the password by providing the correct "Current Password" and observe navigation back to Settings
- [ ] Force the unlock page to load (choose your favorite method) and verify the old password no longer unlocks
- [ ] Type the new password into the unlock page and observe the unlock is successful


.gitignore
```diff
@@ -8,3 +8,4 @@ app/extension-scripts/api/playground/
 *.rest
 .DS_Store
 .vscode
+.idea
```
rgdalessandro, line 11: add `.idea` folder to the `.gitignore` of users of JetBrains IDE's (such as WebStorm or IntelliJ)

background/redux-slices/keyrings.ts
```diff
@@ -154,3 +154,13 @@ export const createPassword = createBackgroundAsyncThunk(
...
+
+// TODO: define changePassword thunk and emitter
+export const changePassword = (
+  previousPassword: string,
+  newPassword: string
+): void => {
+  // eslint-disable-next-line no-empty
+  if (previousPassword && newPassword) {
+  }
+}
```
rgdalessandro, line 166: This is only a mock of the changePassword thunk which will be fleshed out in a subsequent commit.

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -63,69 +91,98 @@ export default function KeyringSetPassword(): ReactElement {
     }
   }
 
+  const dispatchChangePassword = (): void => {
+    if (validatePreviousPassword() && validatePassword()) {
+      changePassword(previousPassword, password)
```
rgdalessandro, line 96: This will change to `dispatch(changePassword(previousPassword, password))` once `changePassword` has been completed.

ui/routes/routes.tsx
```diff
@@ -36,7 +36,14 @@ interface PageList {
 
 const pageList: PageList[] = [
   {
-    path: "/keyring/set-password",
+    path: "/keyring/set-password/initial-password",
```
0xDaedalus, line 39: I think it might be cleaner to have these as `/keyring/set-password` and `/keyring/change-password`

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -63,69 +91,98 @@ export default function KeyringSetPassword(): ReactElement {
...
         onSubmit={(event) => {
           event.preventDefault()
           dispatchCreatePassword()
         }}
       >
+        {isChangePassword && (
+          <div className="input_wrap input_wrap_padding_bottom">
+            <SharedInput
+              type="password"
+              label="Previous Password"
```
0xDaedalus, line 124: Probably want `Current Password` here.

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -14,26 +18,40 @@ import SharedButton from "../Shared/SharedButton"
...
+  const [previousPasswordErrorMessage, setPreviousPasswordErrorMessage] =
+    useState("")
   const [password, setPassword] = useState("")
   const [passwordErrorMessage, setPasswordErrorMessage] = useState("")
   const [passwordConfirmation, setPasswordConfirmation] = useState("")
   const history = useHistory()
+  const location = useLocation()
+  const isInitialPassword = location.pathname.includes("initial-password")
+  const isChangePassword =
+    ALLOW_CHANGE_PASSWORD && location.pathname.includes("change-password")
```
0xDaedalus, line 35: I don't think we need the feature flag here since there will be no way for users to get to this screen from settings if the ALLOW_CHANGE_PASSWORD flag is not on.

background/redux-slices/keyrings.ts
mhluongo: Feel free to leave comments like this in commit messages, most of us read them as part of review!

mhluongo:
Nit: No need to include "WIP" anywhere when a PR is in draft!

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -14,26 +18,40 @@ import SharedButton from "../Shared/SharedButton"
...
 export default function KeyringSetPassword(): ReactElement {
+  const [previousPassword, setPreviousPassword] = useState("")
+  const [previousPasswordErrorMessage, setPreviousPasswordErrorMessage] =
+    useState("")
   const [password, setPassword] = useState("")
   const [passwordErrorMessage, setPasswordErrorMessage] = useState("")
   const [passwordConfirmation, setPasswordConfirmation] = useState("")
   const history = useHistory()
+  const location = useLocation()
+  const isInitialPassword = location.pathname.includes("initial-password")
```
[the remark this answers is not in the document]
rgdalessandro: Thank you for reviewing this PR. I do agree that simplifying the `&&` is better.

I went ahead and took care of the simplification in lines 107-120 of this file. I felt the HOC might increase complexity unnecessarily. Of course we may need to discuss this piece further.

ui/components/Keyring/KeyringSetPassword.tsx
rgdalessandro: Good call, I have removed it from this file.

background/redux-slices/keyrings.ts
```diff
@@ -154,3 +155,11 @@ export const createPassword = createBackgroundAsyncThunk(
     await emitter.emit("createPassword", password)
   }
 )
+
+// FIXME: pair program this method
+export const checkPassword = createBackgroundAsyncThunk(
```
0xDaedalus, line 160: I think we probably want this to be `changePassword`.  We can handle checking if the password is correct or not inside of a method we add to the keyring service.

background/services/keyring/index.ts
```diff
@@ -144,6 +144,43 @@ export default class KeyringService extends BaseService<Events> {
...
+      const saltedKey = await deriveSymmetricKeyFromPassword(
+        password,
+        currentEncryptedVault.salt
+      )
+      try {
+        await decryptVault<SerializedKeyringData>(
+          currentEncryptedVault,
+          saltedKey
+        )
+        this.#cachedKey = saltedKey
```
0xDaedalus, line 172: I don't think you'll want to overwrite the cached key when verifying a password

background/services/keyring/index.ts
```diff
@@ -144,6 +144,43 @@ export default class KeyringService extends BaseService<Events> {
...
+   * Verify if the provided password is valid.
+   * Do not actually lock or unlock the vault.
+   *
+   * @param password A user-chosen string used to encrypt keyring vaults.
+   *        Unlocking will fail if an existing vault is found, and this password
+   *        can't decrypt it.
+   * @returns true if the password is able to unlock the vault,
+   *          and false otherwise.
+   */
+  async verifyPassword(password: string): Promise<boolean> {
```
0xDaedalus, line 157: my spidey sense is telling me that we might want this to be a `private` method thats called from a sibling public`changePassword` method

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -63,69 +89,102 @@ export default function KeyringSetPassword(): ReactElement {
     }
   }
 
+  // FIXME: pair program this method
+  const dispatchChangePassword = async (): Promise<void> => {
+    if (validateCurrentPassword() && validatePassword()) {
+      checkPassword(currentPassword)
+      // TODO: how do we actually know if the current password is valid?
```
0xDaedalus, line 96: Take a look at https://github.com/tallycash/extension/blob/main/background/redux-slices/accounts.ts#L355:L364 for some inspiration

0xDaedalus (commented):
Excellent progress - really starting to take shape!

background/redux-slices/keyrings.ts
rgdalessandro: As suggested, I have renamed the action to `changePassword`

background/services/keyring/index.ts
rgdalessandro: As suggested, I have removed the line of code `this.#cachedKey = saltedKey`

background/services/keyring/index.ts
rgdalessandro: As suggested, I have created a new class method in the keyrings service class called `changePassword` that now references `verifyPassword`. I have also made `verifiedPassword` a private class method.

ui/components/Keyring/KeyringSetPassword.tsx
rgdalessandro: I have taken a look at the link and still remain confused.

I think my main confusion here is that I see examples and understand how `background` can listen for emitter events generated by `background`, but I cannot find examples and do not understand how `ui` can listen for emitter events generated from `background`.

dispatchChangePassword() is part of `ui` but the action that generates the emitter event is in `background`. This is the problem I need to solve.

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -14,26 +18,58 @@ import SharedButton from "../Shared/SharedButton"
 import SharedInput from "../Shared/SharedInput"
 import titleStyle from "../Onboarding/titleStyle"
 import SharedBackButton from "../Shared/SharedBackButton"
-import SharedBanner from "../Shared/SharedBanner"
 import SharedToggleButton from "../Shared/SharedToggleButton"
 import PasswordStrengthBar from "../Password/PasswordStrengthBar"
 
 export default function KeyringSetPassword(): ReactElement {
```
jagodarybacka, line 24: I'm wondering if this is even a good idea to include setting and changing password within one component. A lot of additions to this file and overall logic is getting more complicated. I would rather have two components, `KeyringSetPassword` and `KeyringChangePassword` even if it means some lines will be the same in both files. This is not a change request yet, would love to hear your opinion and @0xDaedalus

background/services/keyring/index.ts
```diff
@@ -144,6 +144,72 @@ export default class KeyringService extends BaseService<Events> {
...
+   * Verify if the provided password is valid.
+   * Do not actually lock or unlock the vault.
+   *
+   * @param password A user-chosen string used to encrypt keyring vaults.
+   *        Unlocking will fail if an existing vault is found, and this password
+   *        can't decrypt it.
+   * @returns true if the password is able to unlock the vault,
+   *          and false otherwise.
+   */
+  private static async verifyPassword(password: string): Promise<boolean> {
```
jagodarybacka, line 157: Do we need to make this method static? What is the reason? Can we make this change later if needed because right now we are only using `verifyPassword` inside its own class so it seems to me like it doesn't have to be static

jagodarybacka (commented):
Hey, nice to see your code 💖 Some comments 👇

ui/pages/Settings.tsx
```diff
@@ -21,15 +22,20 @@ import { getLanguageIndex, getAvalableLanguages } from "../_locales"
 import { getLanguage, setLanguage } from "../_locales/i18n"
 import SettingButton from "./Settings/SettingButton"
 
+interface ISettingRow extends ReactElement {
```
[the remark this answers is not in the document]
rgdalessandro: Hello @jagodarybacka ! Thank you for reviewing this PR. You are correct, this interface is no longer needed -- nice catch!

A week ago when I started this task, it seemed to be necessary as my IDE was giving me TS errors without it. However, after several rounds of refactoring and also pulling in latest changes, it definitely seems it is not longer necessary. Thank you!

background/main.ts
```diff
@@ -853,6 +854,18 @@ export default class Main extends BaseService<never> {
...
+    keyringSliceEmitter.on("changePassword", async (stringifiedPasswords) => {
+      const didPasswordChangeSucceed = await this.keyringService.changePassword(
+        stringifiedPasswords
+      )
+
+      if (didPasswordChangeSucceed) {
+        this.store.dispatch(setDidPasswordChangeSucceed(true))
+      } else {
+        this.store.dispatch(setDidPasswordChangeSucceed(false))
+      }
```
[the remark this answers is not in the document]
rgdalessandro: Thank you, I have made the suggested simplification.

background/services/keyring/index.ts
rgdalessandro: My linter asked me to make this method static because there is an ESlint rule called `class-methods-use-this` which requires class methods to either use the `this` keyword somewhere in the body of the method definition or otherwise be `static`.

I seemed to have 2 options to satisfy the linter... either make the method static or add a `// eslint-disable-next-line class-methods-use-this` line. I opted to go making the method static. Totally open to discussion on this in case I should have used the `eslint-disable-next-line` approach or if there is a good way to use the `this` keyword within the body of the method definition.

ui/components/Keyring/KeyringSetPassword.tsx
0xDaedalus: I'm inclined to agree here.  Initially I had thought we could get away with something like a `KeyringContainer` HOC but with how the useEffect is implemented I think there's a chance that could be a whole different sort of complicated.  With how absolutely application-critical the `KeyringSetPassword` component is I think in this case its worth it to be a bit less DRY and repeat some code.

0xDaedalus:
Hey @rgdalessandro - if you can please avoid force-pushing once a PR is out of draft - makes it quite difficult for those of us who review commit-by-commit

0xDaedalus (changes requested):

background/services/keyring/index.ts
```diff
@@ -144,6 +144,70 @@ export default class KeyringService extends BaseService<Events> {
...
+   *          and false otherwise.
+   */
+  async changePassword(stringifiedPasswords: string): Promise<boolean> {
+    const { currentPassword, newPassword } = JSON.parse(stringifiedPasswords)
+    try {
+      const isCurrentPasswordValid = await this.verifyPassword(currentPassword)
+
+      if (isCurrentPasswordValid) {
+        this.#cachedKey = await deriveSymmetricKeyFromPassword(newPassword)
+        await this.persistKeyrings()
```
Shadowfiend, line 200: We will need to clear older versions of the vault out when there is a password change, or reencrypt them against the new password, otherwise the old password continues to be capable of unlocking key material. We should augment https://github.com/tallycash/extension/blob/491d9fb86e6dc26dba41b0b36ef23d11747d3c30/rfc/rfc-1-keyring-design.adoc with this information/requirement as well.

Shadowfiend:
Want to flag that it's going to be bad form in general to do security-critical background script changes (as with the keyring service) in the same PR as UI stuff. It makes it harder to review, for code that is particularly crucial to review carefully.

To that end, and I realize this is going to require a force-push in spite of @0xDaedalus's request 😅, can we please:
- Move the commits that update `background/services/keyring` to a separate PR (it can be targeted at `main` or at this branch, whichever is easiest).
- Keep this PR focused on the UI pieces.

I realize this is a bit disruptive to the workflow of actually building out the future, but I think it will be helpful to split the conversations and the review along this security boundary.

ui/components/Keyring/KeyringSetPassword.tsx
rgdalessandro: As suggested, I have split this into two separate components.

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -117,87 +104,23 @@ export default function KeyringSetPassword(): ReactElement {
             value={defaultWallet}
           />
         </div>
-        <SharedButton
-          type="primary"
-          size="large"
-          onClick={dispatchCreatePassword}
```
rgdalessandro, line 123: This `onClick` was removed because I found that `dispatchCreatePassword` was incorrectly firing twice. Once from the `onClick` and once from the `onSubmit` of the Form. By removing this `onClick` prop, then `dispatchCreatePassword` is now firing only once, as intended.

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -14,26 +18,58 @@ import SharedButton from "../Shared/SharedButton"
...
   const dispatch = useBackgroundDispatch()
 
   useEffect(() => {
-    if (areKeyringsUnlocked) {
+    const initialPasswordProvided = isInitialPassword && areKeyringsUnlocked
+    const changePasswordSuccess = isChangePassword && didPasswordChangeSucceed
+    const changePasswordFailed =
+      isChangePassword && didPasswordChangeSucceed === false
+
+    if (initialPasswordProvided || changePasswordSuccess) {
```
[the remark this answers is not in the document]
rgdalessandro: As suggested I have implemented `dispatch(setSnackbarMessage("Password successfully changed"))` on `changePasswordSuccess`

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -63,69 +113,96 @@ export default function KeyringSetPassword(): ReactElement {
     }
   }
 
+  const dispatchChangePassword = async (): Promise<void> => {
+    if (validateCurrentPassword() && validatePassword()) {
+      dispatch(
+        changePassword(
+          JSON.stringify({ currentPassword, newPassword: password })
```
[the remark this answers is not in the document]
rgdalessandro: I found the solution. I had to create a type `ChangePasswordPayload` and assign it to the `changePassword` property of  type `Event` in `keyrings.ts` to resolve this. I am no longer using any `JSON.stringified` values, only a simple JS object of type `ChangePasswordPayload`

background/services/keyring/index.ts
```diff
@@ -144,6 +144,70 @@ export default class KeyringService extends BaseService<Events> {
...
+   * Create a new vault encrypted with a new user-chosen password.
+   * Do not actually lock or unlock any vaults.
+   *
+   * @param stringifiedPasswords A JSON-stringified object with this shape:
+   * { currentPassword: string, newPassword: string }
+   *
+   * @returns true if the password was successfully changed,
+   *          and false otherwise.
+   */
+  async changePassword(stringifiedPasswords: string): Promise<boolean> {
```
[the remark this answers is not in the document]
rgdalessandro: As requested, I have eliminated all uses of `JSON.stringify` and `JSON.parse`

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -63,69 +113,96 @@ export default function KeyringSetPassword(): ReactElement {
...
+  let buttonText = "Begin the hunt"
+  if (isChangePassword) {
+    backButtonPath = "/settings"
+    headerText = "Let's change your password"
+    onSubmit = dispatchChangePassword
+    inputLabel = "New Password"
+    buttonShowLoadingOnClick =
+      !currentPasswordErrorMessage && !passwordErrorMessage
+    buttonText = "Change password"
+  }
```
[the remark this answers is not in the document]
rgdalessandro: As suggested, I have split up the UI into two separate components and have used i18n throughout.

ui/components/Keyring/KeyringSetPassword.tsx
```diff
@@ -63,69 +113,96 @@ export default function KeyringSetPassword(): ReactElement {
...
-          Set Tally Ho as default wallet
-          <SharedToggleButton
-            onChange={(toggleValue) => {
-              dispatch(setNewDefaultWalletValue(toggleValue))
-            }}
-            value={defaultWallet}
-          />
+        {isInitialPassword && (
+          <div className="set_as_default_ask">
+            Set Tally Ho as default wallet
```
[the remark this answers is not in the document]
rgdalessandro: Done

ui/components/Keyring/KeyringChangePassword.tsx
```diff
@@ -0,0 +1,209 @@
...
+          font-weight: 500;
+        }
+        .restore {
+          display: none;
+          position: fixed;
+          bottom: 26px;
+        }
+        .button_wrap {
+          padding-top: 10px;
+        }
```
rgdalessandro, line 205: I tried to abstract these styles to a file called `styles.ts` and then referencing here them simply as `${styles}`.

However, I was unable to get `styled-jsx` to cooperate with this approach. After several attempts I decided to simply inline all these same styles in both `KeyringChangePassword.tsx` and `KeyringSetPassword.tsx`. It's not keeping the code quite as DRY but at least this approach cooperated with `styled-jsx` and the styles were applied correctly.

background/services/keyring/index.ts
rgdalessandro: As requested, `keepOnlyOneEncryptedVault()` was created to clear out all older versions of vaults and only keep the latest one after a password change.

background/services/keyring/storage.ts
```diff
@@ -86,3 +86,25 @@ export async function writeLatestEncryptedVault(
...
+ * removing all other vaults from extension storage.
+ *
+ * @param encryptedVault - an encrypted keyring vault
+ */
+export async function keepOnlyOneEncryptedVault(
+  encryptedVault: EncryptedVault
+): Promise<void> {
+  await browser.storage.local.set({
+    tallyVaults: {
+      version: 1,
```
rgdalessandro, line 101: I guessed at `version: 1` being the correct value here. Please correct me if I am wrong.

rgdalessandro:
> Want to flag that it's going to be bad form in general to do security-critical background script changes (as with the keyring service) in the same PR as UI stuff. It makes it harder to review, for code that is particularly crucial to review carefully.
> 
> To that end, and I realize this is going to require a force-push in spite of @0xDaedalus's request 😅, can we please:
> 
> * Move the commits that update `background/services/keyring` to a separate PR (it can be targeted at `main` or at this branch, whichever is easiest).
> * Keep this PR focused on the UI pieces.
> 
> I realize this is a bit disruptive to the workflow of actually building out the feature*, but I think it will be helpful to split the conversations and the review along this security boundary.

As requested, this PR has been split up into two PS's:
1) changes to the keyring service --> https://github.com/tallycash/extension/pull/2128

2) all other changes, including redux changes and `ui` changes

PR #2017, opened 2022-08-24
Fix sending ETH on Optimism.

### LINKED ISSUE #2014: Can't send assets on Optimism
### What

Trying to send assets is redirecting user instead of setting up transaction

https://user-images.githubusercontent.com/20949277/186460581-6138f957-4385-42ec-b5e1-40b294641694.mov

0xDaedalus:
Closes #2014 

This PR allows users to send ETH to other addresses through the extension.

Things Learned:
- [ETH is an ERC-20 token on Optimism](https://optimistic.etherscan.io/token/0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000)
- We were doing rollup fee estimation incorrectly - which was causing us to overestimate gas fees and prevented transfer ETH transactions from going through once we correctly identified it as a baseAsset.

### To Test
- [ ] Attempt to send ETH to an address on Optimism.


background/services/chain/index.ts
```diff
@@ -823,11 +829,12 @@ export default class ChainService extends BaseService<Events> {
 
   async estimateL1RollupFee(
     network: EVMNetwork,
-    input: string | null | undefined
+    transaction: UnsignedTransaction
```
jagodarybacka, line 832: is there any place that is using this function and doesn't need to run `unsignedTransactionFromEVMTransaction(transactionRequest)` before passing it as this argument? 
If there isn't - maybe it's worth moving this `unsignedTransactionFromEVMTransaction` transformation to this function?

background/constants/currencies.ts
```diff
@@ -25,6 +25,20 @@ export const ETH: FungibleAsset & CoinGeckoAsset & NetworkBaseAsset = {
   },
 }
 
+export const OPTIMISTIC_ETH: FungibleAsset & CoinGeckoAsset & NetworkBaseAsset =
+  {
```
jagodarybacka, line 29: base asset needs homeNetwork field
https://github.com/tallycash/extension/blob/send-eth-optimism/background/services/indexing/db.ts#L67
image

jagodarybacka (changes requested):

background/constants/currencies.ts
0xDaedalus: Hmmmmnn I was hoping to avoid changing the `NetworkBaseAsset` type any further - let me see if I find an alternative way to just adding it

background/services/indexing/db.ts
```diff
@@ -61,7 +61,7 @@ export type CachedTokenList = TokenListCitation & {
  */
 function assetID(asset: AnyAsset): string {
   const id = `${asset.symbol}`
-  if (!("contractAddress" in asset)) {
+  if (!("contractAddress" in asset) || "coinType" in asset) {
```
0xDaedalus, line 64: @Shadowfiend curious for your take on this

jagodarybacka (approved):
Sending ETH on OP is fixed 👑

PR #2031, opened 2022-08-26
Let's load all the tx history as 1st step

greg-nagy:
and we fall back to a limited time window if it doesn't work

Note: We have concerns about how this wil behave
with accounts with huge tx history.

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -907,6 +907,23 @@ export default class ChainService extends BaseService<Events> {
     const blockHeight =
       (await this.getBlockHeight(addressNetwork.network)) -
       BLOCKS_TO_SKIP_FOR_TRANSACTION_HISTORY
+
+    // Let's try to go all in on tx history as a first step.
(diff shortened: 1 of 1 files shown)


background/services/chain/index.ts
```diff
@@ -907,6 +907,23 @@ export default class ChainService extends BaseService<Events> {
     const blockHeight =
       (await this.getBlockHeight(addressNetwork.network)) -
       BLOCKS_TO_SKIP_FOR_TRANSACTION_HISTORY
+
+    // Let's try to go all in on tx history as a first step.
+    // If it works we are happy. If it doesn't we fall back on the limited time window.
+    try {
+      return await this.loadAssetTransfers(
```
0xDaedalus, line 914: I believe we'll need to change the underlying method here to order transactions `desc` since we're limited to 1,000 transactions

mhluongo:
I'm going to offer up an alternative approach here I think you guys will like → should lead to a similar API call reduction without throwing the baby out with the bathwater.

Issue #2042, opened 2022-08-27
Newly generated wallets don't need more than a quick asset transfer check


mhluongo:
If we know the extension has generated a local keyring, we also know that any address from that keyring doesn't need a historical asset transfer lookup.

It's an annoying one to track in our current codebase, as `ChainService` doesn't have `KeyringService` as a dependency... but it will mean no history checks for *most* newly onboarded users, saving on resources.

H/t @Shadowfiend for this insight.

PR #2047, opened 2022-08-27
📖 Conservative View of History: Rely on a single `getAssetTransfers` to pull wallet history in the happy case 

mhluongo:
Remove the sliding window transfer lookup construct in `ChainService`, instead immediately looking up history when an account is tracked.

While the block height concept isn't absolutely removed for now, in practice in the happy path we should have a single call for a new account's transfer history.

Explicitly out of scope for this PR, though I'd like to address some of these issues in follow-ups
* Intelligent retries/backoffs for the original history pull. The current behavior is a single dumb hourly retry, but doesn't change parameters.
* Removing the dumb hammer-retry logic getting more recent asset transfers.
* Even though I call this a "single call" for history lookups, it's actually 2 — one for incoming and another for outgoing transactions. I'd like to get it down to 1 for new wallets. #2042
* ... and, one day, pulling history based on nonce and chain bisection. Alchemy misses a ton of juicy stuff today, and my hope for this wallet is still to replace EtherScan for the vast majority of use cases... and that means a more consistent history than this or the previous behavior can give us.

# To test

- [x] From a fresh install, load a read-only account with history
- [x] Ensure that over the next 5-10 minutes, the wallet activity list is populated.
- [x] Make sure there aren't any unexpected errors in the background page console

Closes #2031, alternative to #2031

### DIFF
--- a/background/lib/alchemy.ts
+++ b/background/lib/alchemy.ts
@@ -39,13 +39,17 @@ export async function getAssetTransfers(
   provider: AlchemyProvider | AlchemyWebSocketProvider,
   addressOnNetwork: AddressOnNetwork,
   fromBlock: number,
-  toBlock?: number
+  toBlock?: number,
+  order: "asc" | "desc" = "desc",
+  maxCount = 1000
 ): Promise<AssetTransfer[]> {
   const { address: account, network } = addressOnNetwork
 
   const params = {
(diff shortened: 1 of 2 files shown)


0xDaedalus (approved):
Fantastic.  

Even better than the solution I initially had in mind and excellent follow-on steps ✅

PR #2049, opened 2022-08-27
Introduce an initial CODEOWNERS file

mhluongo:
As the code base continues to mature and more users rely on Tally Ho, we also need to scale our security practices.

As a small step, this PR adds myself and @Shadowfiend as required to sign off on all keyring-related code changes, as well as @0xDaedalus to sign off on dependency graph changes.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Any changes to keyring code deserve extra scrutiny
+/background/services/keyring/* @shadowfiend @mhluongo
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,9 @@
-# Any changes to keyring code deserve extra scrutiny
+# Any changes to keyring code deserve extra scrutiny to prevent key
+# exfiltration and general "roll your own crypto" mistakes. Newer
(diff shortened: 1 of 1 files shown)


greg-nagy:
💔

Shadowfiend:
> 💔

Adding responsibility to folks when they're on vacation isn't cool! 😜

------

Separately, should we make these teams rather than individuals? It'll let us manage the team membership as things evolve without having to do PRs—of course you could argue doing PRs is precisely the best way to do that, which I'm open to as well hehe. The other advantage of using teams is if we want a given team to own more than one thing across, for example, repos (e.g. setting CODEOWNERS for all of tallycash/hd-keyring to the same as the keyring service here).

mhluongo:
I suspect we should add way more here, but for now I wanted to go minimal.

If y'all want to add a "keyring security" team and a "dependency security" team, happy to revise!

@greg-nagy you're more than welcome to have more work, lol. But maybe not while you're out 😝

mhluongo:
I do like that teams mean avoiding per-addition discussions, but it also means users and auditors can't easily audit based on the commit history.

0xDaedalus (approved):
✅

Issue #1967, opened 2022-08-16
Enormous amount of background requests to UNS name resolution


0xDaedalus:
<img width="1022" alt="image" src="https://user-images.githubusercontent.com/94649004/184991411-4be0e2f1-f1c0-4c2e-a232-ec6802b91e40.png">

~200k of these messages in half an hour of monitoring.

I'm guessing we're still looking up names somewhere we shouldn't be - probably when we get a batch of new blocks.

Shadowfiend:
`0x2105dc7…` -> is this abbreviation done by the console or are we getting real weird in what we're sending here?

0xDaedalus:
just a console abbreviation - full URL is available in the `networks` tab

hyphenized:
I'm looking into this, it seems we're requesting enrichment for accounts related transactions that are coming from our request to get asset transfers through Alchemy's WebSocket connection. So I think @0xDaedalus is correct 👀. I'm working on a fix

Shadowfiend:
That is actually correct behavior--but we should probably queue enrichment for older transactions while fast-pathing signature requests.

0xDaedalus:
Closed by https://github.com/tallycash/extension/pull/2068

PR #2054, opened 2022-08-29
Reduce number of UNS address lookups

hyphenized:
This PR adds caching to the UNS resolver in an effort to reduce the amount of calls made to the Unstoppable Domains API while syncing blocks.

Refs: 
#1967

- [ ] Watch network requests in the background script, should no longer send requests in burst for the same address

### DIFF
--- a/background/services/name/resolvers/uns.ts
+++ b/background/services/name/resolvers/uns.ts
@@ -1,9 +1,12 @@
 import { AddressOnNetwork, NameOnNetwork } from "../../../accounts"
-import { ETHEREUM, POLYGON } from "../../../constants"
+import { ETHEREUM, POLYGON, SECOND } from "../../../constants"
 import logger from "../../../lib/logger"
 import { isDefined } from "../../../lib/utils/type-guards"
 import { sameNetwork } from "../../../networks"
 import { NameResolver } from "../name-resolver"
+
+// Time until response is stale
+const RESPONSE_TTL = 5 * SECOND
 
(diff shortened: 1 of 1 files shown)


mhluongo:
Full disclosure, I'm not super familiar with `NameService` so kicking this review to @Shadowfiend. But! Do you know why the generic TTL stuff in `background/services/name/index.ts` didn't cover this @hyphenized?

hyphenized:
@mhluongo I think it's because the UNS lookup methods don't need `chainId` and that's what's being used in the cache strategy at the `NameService`, so it's working but it's making a request in multiple chains for the same address

jagodarybacka (commented):
Can we normalize addresses in the cache so addresses with different upper/lowercase letters are pointing to the same cache entry? 
image

jagodarybacka (approved):
It's a big improvement already so I think address normalization can be added later. I would love to merge it but I would like to hear @0xDaedalus thoughts first :)

0xDaedalus (approved):
Code looks fantastic - and I can see requests getting de-duplicated.  📈

0xDaedalus (approved):

Shadowfiend:
Btw, I do think having the resolvers cache locally for semantic differences like this is 👌 

I’ll add the flag that we could probably comfortably extend the TTL in this cache to be closer to minutes than seconds. Eventually we probably want to let the resolver communicate back some info to the caller about TTL from the resolver as well. Alternatively, we could push the entire concept of caching onto the resolvers with a couple of helpers to make it easy to do the basics.

Issue #2039, opened 2022-08-27
Replace `eth.link` token URLs with `eth.limo`


mhluongo:
We need to

- [ ] Change the default preferences
- [ ] Migrate old preferences so anyone that currently has the wallet has the new links

jagodarybacka:
Closed by #2056

PR #2056, opened 2022-08-29
Replace .eth.link token lists urls to .eth.limo

hyphenized:
Updates default and stored token lists URL preferences, the later through a migration

Refs: 
#2039


jagodarybacka (approved):
LGTM ✨

Issue #2007, opened 2022-08-23
Add visual functionality to Analytics settings page.


0xDaedalus:
Depends on #2006 

Add onclick confirmation screens for turning off analytics and deleting private data (You will probably want to use [`SharedSlideUpMenu`](https://github.com/tallycash/extension/blob/optimism-gas-price/ui/components/Shared/SharedSlideUpMenu.tsx#L1:L1) for these guys)

Display [`Snackbar`](https://github.com/tallycash/extension/blob/optimism-gas-price/ui/components/Snackbar/Snackbar.tsx#L1:L1) notifications once a user has turned off analytics or deleted private data.

Add and maintain the state of the user's analytics preferences to the [UI reducer](https://github.com/tallycash/extension/blob/optimism-gas-price/background/redux-slices/ui.ts#L24:L24)

Figma: https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=12140%3A213853

PR #2058, opened 2022-08-29
Add analytics settings page

### LINKED ISSUE #2006: Add analytics settings page
Depends on #2005 

This page should be visible when clicking the `Analytics` button on the settings menu.

It's not necessary to have the `Delete my data` button do anything to close this issue as that functionality will be tackled down the line.

image

kkosiorowska:
Closes #2006 

This PR displays the analytics page after clicking the button. PR includes only a visual layer, functionalities will be added in #2007 .

**Changes**
- add analytics settings page
- set back button and title is in the same line without "back" test for Analytics and Export Log page

**UI**

Analytics Settings page:
Screenshot 2022-08-30 at 09 55 43

Back buton and title in the same line for Export Logs page:
Screenshot 2022-08-29 at 11 52 50

**To Test** 
- [ ]   Set up SUPPORT_ANALYTICS flag to true next click the analytics button on the settings page. The analytics page should open.
- [ ] Check that back button and title is in the same line without "back" text for Analytics and Export Log page


ui/_locales/en/messages.json
```diff
@@ -81,6 +81,18 @@
       "ariaLabel": "Open bug report"
     },
     "analyticsSetUp": {
+      "title": "Analytics",
+      "toggleTitle": "Analytics are on",
+      "toggleDesc": "Analytics help us understand your needs and empowers us to build a better wallet for you!",
+      "recordTitle": "What we will record",
+      "recordItem1": "Number of swaps / Sends",
```
jagodarybacka, line 88:
```suggestion
      "recordItem1": "Number of swaps / sends",
```

ui/_locales/en/messages.json
```diff
@@ -81,6 +81,18 @@
...
+      "recordItem2": "Number of accounts",
+      "recordItem3": "Gas paied overall",
```
jagodarybacka, line 90:
```suggestion
      "recordItem3": "Gas paid overall",
```

ui/_locales/en/messages.json
```diff
@@ -81,6 +81,18 @@
...
+      "noRecordTitle": "We will never record",
+      "noRecordItem1": "See/save your seed phrase",
```
jagodarybacka, line 92:
```suggestion
      "noRecordItem1": "See / save your seed phrase",
```

ui/_locales/en/messages.json
```diff
@@ -81,6 +81,18 @@
...
+      "noRecordItem2": "Make your address and transaction",
```
jagodarybacka, line 93: This sentence doesn't make sense to me, why not "Your address and transactions you make" or something similar cc @VladUXUI

jagodarybacka, line 93: I think it's better to have more "descriptive" names for these strings without indexes

jagodarybacka (changes requested):

ui/public/index.css
```diff
@@ -148,3 +148,10 @@ label {
   opacity: 60%;
   background-color: var(--green-120);
 }
+
+.text {
```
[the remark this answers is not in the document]
kkosiorowska: I moved this class to setting scope.

ui/pages/Settings.tsx
```diff
@@ -245,6 +245,12 @@ export default function Settings(): ReactElement {
             margin: 0 auto;
             padding: 16px 0px;
           }
+          .text {
```
jagodarybacka, line 248: seems like this class is in a wrong file?

kkosiorowska: I moved this class to the Settings file because this class is also used by SettingsExportLogs

jagodarybacka: Are you sure it is still working? `styled-jsx` is not applying styles from parent component to child components by default (sometimes it is but it's a "bug") if I remember correctly

ui/_locales/en/messages.json
0xDaedalus: I think we can get rid of the verbs here. so something like:

`Your seed phrase` 
and 
`Your address and transactions`

ui/pages/Settings.tsx
kkosiorowska: Good point. I'll move the styles to the global scope, but rename the class so that it doesn't conflict with the others.

ui/_locales/en/messages.json
kkosiorowska: One point.  What is the final decision and who should decide about this texts?

VladUXUI: Yeah, that was a slippage in text, don't think anybody really checked it ?:)
Let's go with what @0xDaedalus proposed, as that seems the shortest and ties in to the "title"

ui/_locales/en/messages.json
VladUXUI: Here let's also write Swaps insted of swaps

ui/_locales/en/messages.json
kkosiorowska: Vlad approved the texts below:
Screenshot 2022-08-30 at 09 55 43

ui/components/Shared/SharedBackButton.tsx
```diff
@@ -5,9 +5,11 @@ import classNames from "classnames"
 export default function SharedBackButton({
   path,
   onClick,
+  withoutBackText,
 }: {
   path?: string
   onClick?: () => void
+  withoutBackText?: boolean
```
Shadowfiend, line 12: Wondering if in future work we should consider just using `children`, and making existing uses explicitly do `<SharedBackButton...>Back</SharedBackButton>`. It seems overall clearer to me, but could see arguments in either direction.

Anyway, obviously not blocking :)

kkosiorowska: I think this is a good point. Maybe, we should create another task for this which should describe how this component should looks like for pages because this task was only about the analytics settings page.

jagodarybacka (approved):
Looks good 💖

Issue #2043, opened 2022-08-27
Add TTL-based caching for UNS lookups


mhluongo:
![Screenshot from 2022-08-27 13-48-37](https://user-images.githubusercontent.com/427505/187042081-53c23528-4e4c-41f4-8588-54ed45018be8.png)

We're hammering UNS unnecessarily (and likely other name resolvers as well). `NameService` was meant to cache name and address resolutions — let's do it.

mhluongo:
Funny, we've actually already implemented this... go us! But it looks like our expiry is 10 seconds by default :thinking:

mhluongo:
Closed by #2054

PR #2062, opened 2022-08-29
Social Norms: Only look up names for normalized addresses

mhluongo:
All EVM addresses that hit `NameService` should be normalized to prevent cache misses.

# To test

- [ ] Add a UNS name as a read-only account. There should be no behavior change.
- [ ] Load up an account that's sent or received assets from a UNS name. The wallet activity list should correctly show the UNS name.
- [ ] In the console network tab, ensure there are no back-to-back UNS lookups of the same address within 5 seconds.

Refs #2043, builds on #2054


0xDaedalus (approved):
👨‍🍳

PR #2066, opened 2022-08-29
Avoid adding transaction hashes for known asset transfers to the queue of transactions to retrieve.

0xDaedalus:
Now that we've moved to an approach of querying a given accounts most recent 1k asset transfers every 90 seconds - we need to ensure that we are not retrieving the same transactions unnecessarily.

Prior to this change - the "transactionsToRetrieve" queue would fill up every 90 seconds even after some transactions were correctly processed via "handleQueuedTransactionAlarm".  The performance hit from that was especially large on accounts with a large asset transfer history due to how we handle determining if a given transaction hash is already in the queue inside of "queueTransactionHashToRetrieve".

### To Test
- [ ] drop the following log [inside of the chain service constructor](https://github.com/tallycash/extension/blob/unnecessary-tx-enrichment/background/services/chain/index.ts#L229)
```
setInterval(() => console.log("Transactions in queue: ", this.transactionsToRetrieve.length), 10_000)
```
- [ ] Load a new account into the wallet and open the logs for the background page.
- [ ] You should see the size of this.transactionsToRetrieve decrease by 5 every minute.

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -1049,14 +1049,22 @@ export default class ChainService extends BaseService<Events> {
 
     const firstSeen = Date.now()
 
-    /// send all found tx hashes into a queue to retrieve + cache
-    assetTransfers.forEach((a) =>
-      this.queueTransactionHashToRetrieve(
-        addressOnNetwork.network,
-        a.txHash,
-        firstSeen
+    const savedTransactionCacheKeys = new Set(
+      (await this.db.getAllSavedTransactions()).map(
+        (tx) => `${tx.hash}-${tx.network.chainID}`
       )
     )
+    /// send all found tx hashes into a queue to retrieve + cache
+    assetTransfers.forEach((a) => {
+      const cacheKey = `${a.txHash}-${addressOnNetwork.network.chainID}`
+      if (!savedTransactionCacheKeys.has(cacheKey)) {
+        this.queueTransactionHashToRetrieve(
+          addressOnNetwork.network,
+          a.txHash,
+          firstSeen
+        )
+      }
+    })
   }
 
   /**
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -16,7 +16,6 @@ import {
   EIP1559TransactionRequest,
   EVMNetwork,
   BlockPrices,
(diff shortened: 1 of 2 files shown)


background/services/chain/index.ts
```diff
@@ -1049,14 +1049,22 @@ export default class ChainService extends BaseService<Events> {
...
     const firstSeen = Date.now()
 
-    /// send all found tx hashes into a queue to retrieve + cache
-    assetTransfers.forEach((a) =>
-      this.queueTransactionHashToRetrieve(
-        addressOnNetwork.network,
-        a.txHash,
-        firstSeen
+    const savedTransactionCacheKeys = new Set(
+      (await this.db.getAllSavedTransactions()).map(
```
mhluongo, line 1053: Does this not get expensive? It seems like we could either...
* do a query against the db for all new tx hashes / networks to test for uniqueness, or
* only pull the tx hashes from the db rather than the entire objects, using eg `db.chainTransactions.keys()`

I'm for merging this regradless as it solves an immediate problem, but it shouldn't take long to save ourselves from a future footgun

mhluongo (approved):
Great catch! This PR works, and is good to merge... though I'm hoping you'll consider a slight improvement on the filter mechanism :)

background/services/chain/index.ts
mhluongo: On my personal wallet only against mainnet, this is 870 items or so... and we can expect that to increase as we add more and more networks. `1000 * numNetworks` is a likely upper limit we'll see in the short term

0xDaedalus: I would hope that there is virtually no difference between retrieving 10 transactions and retrieving 10,000 (since there is no `where` in our query) but I confess I did not measure this.

We do need to pull at least the transaction and the network to mimic our current way of ensuring cross-network transaction uniqueness.

That being said - there are 16 ^ 64 possible transaction hashes which means that the chance of a collision happening is even smaller than when using `uuid's`.  Should we drop the `sameNetwork` requirement in both cases?

mhluongo: 👌

mhluongo (approved):
:rocket:

PR #2067, opened 2022-08-29
Fix ERC-20 balance issue on Polygon

0xDaedalus:
### To Test
- [ ] Load the test wallet into the extension (0x208e94d5661a73360d9387d3ca169e5c130090cd)
- [ ] Select the Polygon network.
- [ ] Open the logs for a few minutes.
- [ ] There should be no warnings containing `Alchemy token balance response didn't validate, did the API change?` in the logs.


mhluongo (approved):
:ship:

PR #2068, opened 2022-08-29
Drastically reduce unnecessary enrichment of ERC-20 Transfers.

### LINKED ISSUE #2063: Prevent "thundering herd" UNS lookups by the `EnrichmentService`
Screenshot from 2022-08-29 14-00-16

This is a mess :) and appears to be breaking all sorts of things. I believe fixing this without losing functionality is going to require an `EnrichmentService` refactor into phased enrichment.

0xDaedalus:
Closes #2063 

This PR addresses an issue where transactions producing multiple ERC-20 `Transfer` events were significantly impacting the performance of the extension.

In extreme cases - transactions such as [this one](https://etherscan.io/tx/0x3c712768212cc75eec5329c6d08ed4a4edc679ee5c4553c6433c7b16259816fb) containing thousands of ERC-20 `Transfer's` would cause the extension to attempt to enrich each address involved in the transfer - slowing the extension to a standstill.  This PR aims to keep the desired behavior of enriching addresses involved in ERC-20 `Transfer` events while only enriching the events that are relevant to the accounts loaded into the extension.

### To Test
- [x] load up `mhluongo.eth` into the extension
- [x] open up the background logs
- [ ] go eat a delicious sandwich
- [x] Check the console in 10-15 minutes - you should not see an enormous amount of `429` errors related to querying unstoppabledomains.  Some errors may still be present because of network conditions and whatnot - but nothing like what we see in https://github.com/tallycash/extension/issues/2063
- [x] ERC-20 transactions should properly enrich with ENS/UNS names in wallet activity tab.

### DIFF
--- a/background/services/enrichment/transactions.ts
+++ b/background/services/enrichment/transactions.ts
@@ -58,16 +58,9 @@ async function annotationsFromLogs(
   )
   const relevantAddresses = [
     ...new Set(
-      getDistinctRecipentAddressesFromERC20Logs(relevantTransferLogs)
-        .concat(
-          tokenTransferLogs.flatMap<string>(
-            ({ senderAddress, recipientAddress }) => [
-              senderAddress,
-              recipientAddress,
-            ]
-          )
-        )
-        .map(normalizeEVMAddress)
+      getDistinctRecipentAddressesFromERC20Logs(relevantTransferLogs).map(
+        normalizeEVMAddress
+      )
     ),
   ]
 
--- a/background/services/enrichment/transactions.ts
+++ b/background/services/enrichment/transactions.ts
@@ -85,42 +85,58 @@ async function annotationsFromLogs(
       .filter(([, annotation]) => isDefined(annotation))
   )
 
-  const subannotations = tokenTransferLogs.flatMap<TransactionAnnotation>(
-    ({ contractAddress, amount, senderAddress, recipientAddress }) => {
-      // See if the address matches a fungible asset.
-      const matchingFungibleAsset = assets.find(
-        (asset): asset is SmartContractFungibleAsset =>
-          isSmartContractFungibleAsset(asset) &&
-          sameEVMAddress(asset.contractAddress, contractAddress)
+  const subannotations = (
+    await Promise.all(
+      tokenTransferLogs.map(
+        async ({
+          contractAddress,
+          amount,
+          senderAddress,
+          recipientAddress,
+        }) => {
+          // See if the address matches a fungible asset.
+          const matchingFungibleAsset = assets.find(
+            (asset): asset is SmartContractFungibleAsset =>
+              isSmartContractFungibleAsset(asset) &&
+              sameEVMAddress(asset.contractAddress, contractAddress)
+          )
+
+          if (!matchingFungibleAsset) {
+            return undefined
+          }
+
+          // Try to find a resolved annotation for the recipient and sender and otherwise fetch them
+          const recipient =
+            annotationsByAddress[normalizeEVMAddress(recipientAddress)] ??
+            (await enrichAddressOnNetwork(chainService, nameService, {
+              address: recipientAddress,
+              network,
+            }))
+          const sender =
+            annotationsByAddress[normalizeEVMAddress(senderAddress)] ??
+            (await enrichAddressOnNetwork(chainService, nameService, {
+              address: senderAddress,
+              network,
+            }))
+
+          return {
+            type: "asset-transfer" as const,
+            assetAmount: enrichAssetAmountWithDecimalValues(
+              {
+                asset: matchingFungibleAsset,
+                amount,
+              },
+              desiredDecimals
+            ),
+            sender,
+            recipient,
+            timestamp: resolvedTime,
+            blockTimestamp: block?.timestamp,
+          }
+        }
       )
-
-      if (!matchingFungibleAsset) {
-        return []
-      }
-
-      // Try to find a resolved annotation for the recipient and sender
-      const recipient =
-        annotationsByAddress[normalizeEVMAddress(recipientAddress)]
-      const sender = annotationsByAddress[normalizeEVMAddress(senderAddress)]
-
-      return [
-        {
-          type: "asset-transfer",
-          assetAmount: enrichAssetAmountWithDecimalValues(
-            {
-              asset: matchingFungibleAsset,
-              amount,
-            },
-            desiredDecimals
-          ),
-          sender,
-          recipient,
-          timestamp: resolvedTime,
-          blockTimestamp: block?.timestamp,
-        },
-      ]
-    }
-  )
+    )
+  ).filter(isDefined)
 
   return subannotations
 }


mhluongo:
💃🕺

mhluongo (approved):
Looks good. This behavior deserves more documentation, but it's not blocking. Wish I had a sandwich.

background/services/enrichment/transactions.ts
```diff
@@ -92,42 +85,58 @@ async function annotationsFromLogs(
...
-      ]
-    }
-  )
+          if (!matchingFungibleAsset) {
+            return undefined
+          }
+
+          // Try to find a resolved annotation for the recipient and sender and otherwise fetch them
+          const recipient =
+            annotationsByAddress[normalizeEVMAddress(recipientAddress)] ??
```
Shadowfiend, line 110: Just want to flag that `annotationsByAddress` feels misnamed <_< I would call this `addressEnrichmentsByAddress` or just `resolvedAddressEnrichments` or so.

background/services/enrichment/transactions.ts
```diff
@@ -92,42 +85,58 @@ async function annotationsFromLogs(
...
-        (asset): asset is SmartContractFungibleAsset =>
-          isSmartContractFungibleAsset(asset) &&
-          sameEVMAddress(asset.contractAddress, contractAddress)
-      )
-
-      if (!matchingFungibleAsset) {
-        return []
-      }
+  const subannotations = (
+    await Promise.all(
```
Shadowfiend, line 89: This will kill the entire enrichment if any one address resolution fails 🙊

Shadowfiend:
I am… Quite puzzled as to how this changes behavior. Before:

```typescript
  const relevantAddresses = [
    ...new Set(
      getDistinctRecipentAddressesFromERC20Logs(relevantTransferLogs)
        .concat(
          tokenTransferLogs.flatMap<string>(
            ({ senderAddress, recipientAddress }) => [
              senderAddress,
              recipientAddress,
            ]
          )
        )
        .map(normalizeEVMAddress)
    ),
  ]
```

“Look up all distinct recipient addresses from the relevant logs, then add all sender and recipient addresses and normalize.” This then feeds into the `annotationsByAddress` resolver that looks up the names for all of those sender/recipient addresses.

```typescript
      const recipient =
         annotationsByAddress[normalizeEVMAddress(recipientAddress)]
      const sender = annotationsByAddress[normalizeEVMAddress(senderAddress)]
```

“Look up the recipient and sender in the pre-resolved list.”

After:

```typescript
  const relevantAddresses = [
    ...new Set(
      getDistinctRecipentAddressesFromERC20Logs(relevantTransferLogs).map(
        normalizeEVMAddress
      )
    ),
  ]
```

“Look up all distinct recipient addresses from the relevant logs and normalize.” Then feed `annotationsByAddress`.

```typescript
          const recipient =
             annotationsByAddress[normalizeEVMAddress(recipientAddress)] ??
             (await enrichAddressOnNetwork(chainService, nameService, {
               address: recipientAddress,
               network,
             }))
           const sender =
             annotationsByAddress[normalizeEVMAddress(senderAddress)] ??
             (await enrichAddressOnNetwork(chainService, nameService, {
               address: senderAddress,
               network,
             }))
```

“Look up the recipient and sender in the pre-resolved list; if not found, resolve.”

These should both look up the exact same addresses, except that the irrelevant addresses may result in multiple hits (because they're not pre-resolved and therefore caching relies on a timeout instead of a different process) and it will probably halve the throughput (because we await the recipient and then do the sender, so if both are always distinct we'll do one at a time instead of two at a time).

All of this to say: I think the diagnosis of the issue was correct, but the fix was actually semi-accidental. Instead of looking up all mentioned addresses at the same time (via `N` different lookups inside a big `Promise.allSettled`), we look up all *relevant* addresses (via `M` lookups inside a `Promise.allSettled`), then do the remainder half and half (via `(N - M) / 2` lookups inside the following `Promise.all` resolving all addresses that weren't deemed relevant and pre-resolved). This is enough to mitigate the 429s, but we're actually still doing an enormous number of hits.

I think it's worth digging into https://github.com/tallycash/extension/commit/d48430ff8e0c727a3cccc5412dc500d5dfdefee1, trying to see if @mhluongo can provide us a transaction that was manifesting the specific issue he tried to address, and then see if we can enrich mostly *just* relevant addresses up front (i.e., addresses where one of our accounts was sender or recipient). There doesn't seem to be a clear reason why failing to enrich an irrelevant address in an ERC20 transfer log should result in `Contract creation` as a transaction recipient, except that we might have been `throw`ing somewhere we shouldn't have and short-circuiting an enrichment that way.

mhluongo:
Wow, I completely forgot I wrote any of the related code here. I feel like this needs line-by-line docs here.

Upside, this is really simple to reproduce. I'll try to find an example tx in the next day or two.

background/services/enrichment/transactions.ts
0xDaedalus: Very good catch - this should be a blocker for cutting the next release

0xDaedalus:
> but we're actually still doing an enormous number of hits.

I don't think this is the case - we're capped out at 2 lookups per relevant `Transfer` log aren't we?  

Looking at the logs after having the extension open for 30 mins I see 567 uns lookups for 452 transactions in the chain service db which seems entirely reasonable.

background/services/enrichment/transactions.ts
0xDaedalus: Addressed here: https://github.com/tallycash/extension/pull/2069

Shadowfiend:
Then it feels like something else was going on IMO, and it's not clear how this mitigated it---possibly the 429s were resulting in some sort of cascade or retry or other outcome. The point of this function is batch handling all transfer logs for a given transaction, and a single transfer log can only ever have one sender and one recipient---the issue you flagged was transactions with many transfers, and I'm not seeing how this diff would actually change the behavior beyond spreading the queries out temporally. The iteration here (`tokenTransferLogs.map`) goes over all transfer logs--the only place we check relevance is for the address pre-resolution, and what we seem to be tweaking here is how many of the addresses we preresolve. 

Another possibility is that this approach 429s once and early, and the `Promise.all` usage makes it so we block further attempts sooner. 

Yet another possibility is that I'm missing something obvious here, of course. I have been staring at it trying to understand for a while though.

0xDaedalus:
Not obvious at all - [this is the line responsible](https://github.com/tallycash/extension/blob/dont-block-log-enrichment/background/services/enrichment/transactions.ts#L101:L103) and it probably deserves its own line of documentation.

Basically we don't track the spam ERC-20's - so we filter them out prior to attempting to enrich their senders & recipients.

Shadowfiend:
Ahhh, this was already in place, but we were pre-resolving so it didn't actually effect the requests. Got it.

Ok, so with all this in mind, I think the flow of this code would be much clearer and require fewer comments if we went:
- Extract transfer logs.
- Resolve matching fungible asset and filter out anything without a matching asset. This would replace the current approach for `relevantTransferLogs`, or supplement it.
- Pull distinct addresses and resolve.
- Assemble the final subannotations.

Dealing with two separate lists of token logs (`tokenTransferLogs` and `relevantTransferLogs`) at different stages is what produces confusion right now I think.

PR #2069, opened 2022-08-30
Support partial subannotation population in the case where one annotation fails.

0xDaedalus:
This fixes a potential problem where failing to enrich one subannotation from a `Transfer` log prevents all subannotations from populating.


jagodarybacka (approved):
LGTM 👑

PR #2072, opened 2022-08-30
Beyond Community Edition: Remove "Community Edition" prefix from bug reports

mhluongo:
We haven't been CE since v0.14.0 😀

# To test

- [x] Run the version bump script and confirm that the new version it adds in `.github/ISSUE_TEMPLATE/BUG.yml` doesn't include the label "Community Edition"


Shadowfiend (approved):
Worth it(tm).

PR #2073, opened 2022-08-30
More Release Owners: Separate release management from dependency audits

mhluongo:
This is closer to my original intent with the `CODEOWNERS` config — keyring code should be hard to change, dependencies should be hard to change, non-dependency releases should be a bit looser.

We should agree on a process for adding people to the @tallycash/extension-dependency-auditors team. The idea here is that anyone who joins it needs to agree that they won't merge without auditing every dependency change... a pretty serious lift / responsibility.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,5 +5,7 @@
 /background/services/keyring/* @shadowfiend @mhluongo
 # Any changes to dependencies deserve extra scrutiny to help prevent supply
 # chain attacks
-package.json @0xDaedalus @shadowfiend @mhluongo
-yarn.lock @0xDaedalus @shadowfiend @mhluongo
+package.json @greg-nagy @0xDaedalus @shadowfiend @mhluongo
+yarn.lock @tallycash/extension-dependency-auditors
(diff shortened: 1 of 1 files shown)


jagodarybacka:
I think we should define what this dependency audit will look like first

mhluongo:
> I think we should define what this dependency audit will look like first

Basic idea: if you sign off on a `yarn.lock`-impacting PR, you're stating that you do not believe the dependency upgrade includes malicious code.

Agreeing on how many minutes / hours / days we spend on that as the project progresses is out-of-scope of this PR, IMO... I think we need to have a discussion with the team. It should likely be a function of

* How many MAU do we have?
* How many dollars is the code protecting?
* How much do we trust the underlying dependencies to do their own diligence? Eg I have pretty high confidence in `ethers` / @ricmoo to look out for supply chain attacks
* How many other mitigations are we deploying? We aren't yet using something like [`lavamoat`]](https://github.com/LavaMoat/LavaMoat) and while I don't think @kumavis 's work here would be a get-out-of-jail-free card, it'd certainly give us some extra confidence.

0xDaedalus (approved):

PR #2083, opened 2022-08-30
Avoid looking up historical transfers for newly created accounts.

### LINKED ISSUE #2042: Newly generated wallets don't need more than a quick asset transfer check

0xDaedalus:
Closes #2042 

This change should free up some alchemy bandwidth if and when we get an influx of new users similar to last week.

### To Test
- [ ] Load up a fresh extension.
- [ ] Open devtools - go to network tab - press cmd + r - click on one of the websocket connections to alchemy - go to the `Messages` tab.
- [ ] Create a new account.
- [ ] You should not see any `alchemy_getAssetTransfers` messages go out over the connection when creating the account.
- [ ] ~90 seconds after creating the account - you should see an `alchemy_getAssetTransfer` message go out over the connection.

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -64,6 +64,7 @@ import {
   OPTIMISM_GAS_ORACLE_ABI,
   OPTIMISM_GAS_ORACLE_ADDRESS,
 } from "./utils/optimismGasPriceOracle"
+import KeyringService from "../keyring"
 
 // We can't use destructuring because webpack has to replace all instances of
 // `process.env` variables in the bundled output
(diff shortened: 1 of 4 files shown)


background/services/keyring/index.ts
```diff
@@ -348,6 +348,18 @@ export default class KeyringService extends BaseService<Events> {
     return newKeyring.id
   }
 
+  async getKeyringSourceForAddress(
```
mhluongo, line 351: Docs plz

background/services/chain/index.ts
```diff
@@ -693,24 +699,30 @@ export default class ChainService extends BaseService<Events> {
...
-    this.loadHistoricAssetTransfers(addressNetwork).catch((e) => {
+    this.getLatestBaseAccountBalance(addressNetwork).catch((e) => {
       logger.error(
-        "chainService/addAccountToTrack: Error loading historic asset transfers",
+        "chainService/addAccountToTrack: Error getting latestBaseAccountBalance",
         e
       )
     })
+    if (
+      (await this.keyringService.getKeyringSourceForAddress(
```
mhluongo, line 715: Now that we have `KeyringService` as a dependency, it'd be nice to use `KeyringService.emitter.on("...")` rather than calling out to it like this... perhaps a good future refactor

mhluongo (commented):
Left a couple comments while I test!

mhluongo (approved):
Looks good, works well. Going to give this a little bit to merge in case @Shadowfiend wants to weigh in

Issue #2085, opened 2022-08-30
Consider a prompt for users to set Tally Ho as the default wallet


mhluongo:
> Been thinking about this a bit more, and i think we run the risk of going to hard on the other side, if people don't know how to disable Tally as default, they won't know how to enable it as default. So i propose we add this bar for when Tally is disabled.
> It's safe to assume that if it's disabled people don't actively use it for transactions or maybe don't even do too much in it. So it shouldn't be too much of a user friction to have the bar there.
> 
> If enabled, the bar should disappear in 10 seconds or so. Giving user a bit of time to disable back
>
> <img width="2108" alt="image" src="https://user-images.githubusercontent.com/80846886/187387635-d8897302-cebc-4bca-ac0d-17a6e29db25c.png">
>
> What do you all think?

_Originally posted by @VladUXUI in https://github.com/tallycash/extension/issues/2026#issuecomment-1231330412_

### Figma link 
https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=12524%3A214935

### Requirements

- [ ] Add banner with Toggle button
- [ ] Change content of banner on enable
- [ ] Hide banner after 5 seconds of enabling
- [ ] If disable toggle is clicked back off in those 5 seconds, leave banner there
- [ ] Banner should be always visible if TallyHo is not set as default

VladUXUI:
Given this feature, and the fact that is off by default, i wonder if we should remove the toggle from onboarding entirely?
Would give users less cognitive load and make it cleaner

VladUXUI:
Adding a gif to this flow for clarity
![CleanShot 2022-09-30 at 10 52 07](https://user-images.githubusercontent.com/80846886/193220514-ee4b925d-45eb-46f9-8859-59c2121df129.gif)

mr-michael:
@jagodarybacka quick note that we just updated figma to include a toast

Issue #2025, opened 2022-08-26
Incremental onboarding improvements


mhluongo:
We are considering a major onboarding flow revamp. First, though, we'd like to explore an incremental design that uses the same flow, but makes use of a tab rather than a restrictive extension popup window.

The goals here are to
* stem immediate user pain
* prevent onboarding from being accidentally aborted by extension popup behavior
* prevent the need for "re-onboarding"
* minimize bugs related to an extension that hasn't yet been onboard, like the famous "It looks like you don't have any accounts, that's sad for a wallet!" screen :)
* use a feature flag for further user evaluation and a product decision

# Tasks

- [x] Adapt the current onboarding design to a responsive tab @VladUXUI 
- [x] #2026
- [x] Implement the new tab onboarding design behind a feature flag
- [x] Ensure the new tab onboarding design isn't susceptible to Halborn's [Demonic vulnerability](https://halborn.com/disclosures/demonic-vulnerability/), which recently impacted MetaMask, Coinbase Wallet, & Phantom.

VladUXUI:
https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=12428%3A213365

Took a quick stab at a few screens. see how they would look on full screen without changing design/content too much.
hard to resist not changing a bunch of things, but it should be decent for our first iteration. If Hailey decided later that we should prioritise this, w'll do a bigger (not huge) redesign

The one screen that is not in the extension flow is the congrats screen.
This is a bit of an edit from our Ledger congrats screen, but i think it's important for this flow
![image](https://user-images.githubusercontent.com/80846886/186850808-a0ed57c0-942a-4a89-8fcc-cf6b9ea57898.png)

mhluongo:
For those shortcuts, we should be able to detect the OS and only show the relevant one

mhluongo:
@VladUXUI what do you think of this tweak to the last page?

![Add-Wallet-Ledger-Step4](https://user-images.githubusercontent.com/427505/188214749-785d9c37-db74-43a7-ab01-83af57dfd266.png)

VladUXUI:
> @VladUXUI what do you think of this tweak to the last page?
> 
> ![Add-Wallet-Ledger-Step4](https://user-images.githubusercontent.com/427505/188214749-785d9c37-db74-43a7-ab01-83af57dfd266.png)

100% Looks awesome and easier to scan

beemeeupnow:
Is this complete?

PR #2086, opened 2022-08-30
👛 You Must Have a Wallet to Have a Default Wallet: Don't set Tally Ho as the default wallet before onboarding

### LINKED ISSUE #2026: Disable the "default wallet" setting by default for all wallets
It's a huge user complaint and a clear source of churn.

The behavior we want

- [ ] Prior onboarding, the wallet shouldn't act as if "default wallet" is set
- [ ] In the onboarding flow, the setting should default to "off"

mhluongo:
Turn off the "set Tally Ho as default wallet" preference default. This avoids the error where users try to use the wallet before onboarding, and are greeted with a frustrated green screen message.

We have product alignment for this new behavior to avoid churn / uninstalls. There's an ongoing discussion about where to go from here in #2085 that I suggest all who are interested check out.

# To test

- [ ] Download & install [MetaMask](https://metamask.io) 
- [ ] Complete MetaMask onboarding
- [ ] Install Tally Ho, but don't onboard
- [ ] Visit [CowSwap](https://cowswap.exchange/#/swap?chain=mainnet) and connect to MetaMask. Tally Ho shouldn't pop up
- [ ] Onboard with Tally Ho, importing or creating a new wallet. Flip the "use as default wallet" setting to "on"
- [ ] Visit [CowSwap](https://cowswap.exchange/#/swap?chain=mainnet) and connect. "Injected" should be shown instead of MetaMask

Closes #2026

### DIFF
--- a/background/services/preferences/defaults.ts
+++ b/background/services/preferences/defaults.ts
@@ -23,7 +23,7 @@ const defaultPreferences: Preferences = {
     ],
   },
   currency: USD,
-  defaultWallet: true,
+  defaultWallet: false,
   selectedAccount: {
     address: "",
     network: ETHEREUM,
--- a/background/services/preferences/tests/index.integration.test.ts
+++ b/background/services/preferences/tests/index.integration.test.ts
@@ -28,10 +28,10 @@ describe("Preference Service Integration", () => {
 
   describe("setDefaultWalletValue", () => {
     it("should correctly toggle defaultWallet in indexedDB", async () => {
-      // Should default to true
-      expect(await preferenceService.getDefaultWallet()).toEqual(true)
-      await preferenceService.setDefaultWalletValue(false)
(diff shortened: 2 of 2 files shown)


jagodarybacka (changes requested):
Testing steps are not aligned with our app behaviour - do we want to turn "default" to ON after onboarding? Right now it is still OFF so MM is the default wallet

mhluongo:
> Testing steps are not aligned with our app behaviour - do we want to turn "default" to ON after onboarding or are we leaving it to the user to switch it manually?

My mistake! Adjusting the testing steps.

Per #2025, we want it to be "off" in the onboarding flow, and rely on users switching it on based on what they want.

jagodarybacka (approved):
LGTM 🌸

PR #2087, opened 2022-08-30
Further reduce calls to alchemy - introduce Active Networks.

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/2040
Closes https://github.com/tallycash/extension/issues/2041

This guy will need some thorough testing - the best way I've found to do it is by opening up dev tools and monitoring an individual websocket connection for messages.

### To Test
- [ ] Importing a new wallet into the extension should not produce any messages on the alchemy polygon ws connection
- [ ] Importing a ledger wallet into the extension should not produce any messages on the alchemy polygon ws connection
- [ ] Adding a read-only address to the extension should not produce any messages on the alchemy polygon ws connection
- [ ] Switching from Ethereum to Polygon inside the wallet for the first time should load balances for the account you have selected
- [ ] Switching from Ethereum to Polygon inside the wallet for the first time should add every added account to polygon
- [ ] Reloading a wallet that has only used the Ethereum chain should not produce any messages on the alchemy polygon ws connection
- [ ] Reloading a wallet that has used both Polygon and Ethereum should produce messages on the alchemy polygon ws connection
- [ ] Switching to Polygon for the first time via a dapp should work, produce messages on the alchemy polygon ws connection, and correctly load account information into the dapp.

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -25,6 +25,7 @@ import {
   OPTIMISM,
   EVM_ROLLUP_CHAIN_IDS,
   GOERLI,
+  NETWORK_BY_CHAIN_ID,
 } from "../../constants"
 import {
   SUPPORT_ARBITRUM,
@@ -61,6 +62,7 @@ import {
   OPTIMISM_GAS_ORACLE_ABI,
   OPTIMISM_GAS_ORACLE_ADDRESS,
 } from "./utils/optimismGasPriceOracle"
+import { IndexableTypeArray } from "dexie"
 
 // How many queued transactions should be retrieved on every tx alarm, per
 // network. To get frequency, divide by the alarm period. 5 tx / 5 minutes →
@@ -166,7 +168,7 @@ export default class ChainService extends BaseService<Events> {
 
   supportedNetworks: EVMNetwork[]
 
-  activeNetworks: EVMNetwork[]
+  private activeNetworks: EVMNetwork[]
 
   assetData: AssetDataHelper
 
@@ -229,11 +231,11 @@ export default class ChainService extends BaseService<Events> {
       ...(SUPPORT_OPTIMISM ? [OPTIMISM] : []),
     ]
 
-    this.activeNetworks = [ETHEREUM]
+    this.activeNetworks = []
 
     this.providers = {
       evm: Object.fromEntries(
-        this.activeNetworks.map((network) => [
+        this.supportedNetworks.map((network) => [
           network.chainID,
           makeSerialFallbackProvider(network),
         ])
@@ -251,11 +253,12 @@ export default class ChainService extends BaseService<Events> {
     await super.internalStartService()
 
     const accounts = await this.getAccountsToTrack()
+    const activeNetworks = await this.getActiveNetworks()
 
     // get the latest blocks and subscribe for all active networks
     // TODO revisit whether we actually want to subscribe to new heads
     // if a user isn't tracking a relevant addressOnNetwork
-    this.activeNetworks.forEach(async (network) => {
+    activeNetworks.forEach(async (network) => {
       const provider = this.providerForNetwork(network)
       if (provider) {
         Promise.allSettled([
@@ -281,7 +284,7 @@ export default class ChainService extends BaseService<Events> {
           // Schedule any stored unconfirmed transactions for
           // retrieval---either to confirm they no longer exist, or to
           // read/monitor their status.
-          this.activeNetworks.map((network) =>
+          (await this.activeNetworks).map((network) =>
             this.db
               .getNetworkPendingTransactions(network)
               .then((pendingTransactions) => {
@@ -314,6 +317,32 @@ export default class ChainService extends BaseService<Events> {
     return USE_MAINNET_FORK
       ? this.providers.evm[ETHEREUM.chainID]
       : this.providers.evm[network.chainID]
+  }
+
+  async getActiveNetworks(): Promise<EVMNetwork[]> {
+    if (this.activeNetworks.length > 0) {
+      return this.activeNetworks
+    }
+
+    // Since activeNetworks will be an empty array at extension load (or reload time)
+    // we need a durable way to track which networks an extension is tracking.
+    // The below code should only be called once per extension reload for extensions
+    // with active accounts
+    const chainIdsToTrack = new Set(await this.getChainIDsToTrack())
+    if (chainIdsToTrack.size > 0) {
+      const activeNetworks: EVMNetwork[] = []
+      chainIdsToTrack.forEach((chainID) => {
+        const network = NETWORK_BY_CHAIN_ID[chainID]
+        if (network) {
+          activeNetworks.push(network)
+        }
+      })
+      this.activeNetworks = activeNetworks
+      return activeNetworks
+    }
+
+    // Default to supporting Ethereum so ENS resolution works during onboarding
+    return [ETHEREUM]
   }
 
   /**
@@ -649,6 +678,10 @@ export default class ChainService extends BaseService<Events> {
 
   async getAccountsToTrack(): Promise<AddressOnNetwork[]> {
     return this.db.getAccountsToTrack()
+  }
+
+  async getChainIDsToTrack(): Promise<string[]> {
+    return this.db.getChainIDsToTrack()
   }
 
   async removeAccountToTrack(address: string): Promise<void> {
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -318,6 +318,10 @@ export default class ChainService extends BaseService<Events> {
       : this.providers.evm[network.chainID]
   }
 
+  /**
+   * Pulls the list of active networks from memory or indexedDB.
+   * Defaults to ethereum in the case that neither exist.
+   */
   async getActiveNetworks(): Promise<EVMNetwork[]> {
     if (this.activeNetworks.length > 0) {
       return this.activeNetworks
@@ -341,7 +345,8 @@ export default class ChainService extends BaseService<Events> {
     }
 
     // Default to supporting Ethereum so ENS resolution works during onboarding
-    return [ETHEREUM]
+    this.activateNetwork(ETHEREUM)
+    return this.activeNetworks
   }
 
   /**
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -62,7 +62,6 @@ import {
   OPTIMISM_GAS_ORACLE_ABI,
   OPTIMISM_GAS_ORACLE_ADDRESS,
 } from "./utils/optimismGasPriceOracle"
-import { IndexableTypeArray } from "dexie"
 
 // How many queued transactions should be retrieved on every tx alarm, per
 // network. To get frequency, divide by the alarm period. 5 tx / 5 minutes →
@@ -349,11 +348,7 @@ export default class ChainService extends BaseService<Events> {
    * Adds a supported network to list of active networks.
    */
   async activateNetwork(network: EVMNetwork): Promise<void> {
-    if (this.providers.evm[network.chainID]) {
-      logger.error(`${network.name} is already active.`)
-      return
-    }
-    this.providers.evm[network.chainID] = makeSerialFallbackProvider(network)
+    this.activeNetworks.push(network)
   }
 
   /**
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -265,20 +265,9 @@ export default class ChainService extends BaseService<Events> {
     // TODO revisit whether we actually want to subscribe to new heads
     // if a user isn't tracking a relevant addressOnNetwork
     activeNetworks.forEach(async (network) => {
-      const provider = this.providerForNetwork(network)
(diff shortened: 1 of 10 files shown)


background/services/chain/index.ts
```diff
@@ -340,6 +318,42 @@ export default class ChainService extends BaseService<Events> {
...
+
+    // Default to supporting Ethereum so ENS resolution works during onboarding
+    this.activateNetwork(ETHEREUM)
+    return this.activeNetworks
+  }
+
+  /**
+   * Adds a supported network to list of active networks.
+   */
+  async activateNetwork(network: EVMNetwork): Promise<void> {
```
mhluongo, line 353: Maybe worth checking if the network is supported

background/services/chain/index.ts
```diff
@@ -664,6 +678,10 @@ export default class ChainService extends BaseService<Events> {
     return this.db.getAccountsToTrack()
   }
 
+  async getChainIDsToTrack(): Promise<string[]> {
```
mhluongo, line 681: Feels like `ChainService` should only be exposing `EVMNetwork` s rather than chain IDs

background/services/chain/index.ts
```diff
@@ -1264,6 +1282,25 @@ export default class ChainService extends BaseService<Events> {
     )
   }
 
+  /**
+   * Get the latest block for a network and save it to the db.
```
mhluongo, line 1286: Maybe `fetchLatestBlockForNetwork()`? I'd expect "get" to check the cache first rather than force-pull.

background/services/chain/index.ts
```diff
@@ -340,6 +318,42 @@ export default class ChainService extends BaseService<Events> {
...
+    // Default to supporting Ethereum so ENS resolution works during onboarding
+    this.activateNetwork(ETHEREUM)
+    return this.activeNetworks
+  }
+
+  /**
+   * Adds a supported network to list of active networks.
+   */
+  async activateNetwork(network: EVMNetwork): Promise<void> {
+    this.activeNetworks.push(network)
```
mhluongo, line 354: This should probably emit an event for other services to respond to

background/services/chain/index.ts
mhluongo: I'm sure we're not following this everywhere, but reaffirming that we want `ChainService` to be subscribed to network info then deliver it to people who ask, rather than just be a new interface for providers without any caching / resource management

mhluongo (commented):
Draft thoughts for a draft PR

background/services/indexing/index.ts
```diff
@@ -581,30 +582,27 @@ export default class IndexingService extends BaseService<Events> {
 
     // get the prices of all assets to track and save them
     const assetsToTrack = await this.db.getAssetsToTrack()
+    const activeNetworks = await this.chainService.getActiveNetworks()
 
     // Filter all assets based on supported networks
     const activeAssetsToTrack = assetsToTrack.filter(
-      (asset) =>
+      async (asset) =>
```
hyphenized, line 589: The async here invalidates the filter?

0xDaedalus: Good catch - ty 👑

background/services/chain/index.ts
0xDaedalus: Definitely agreed - and you're 100% right about not following this everywhere because this function is just moving existing code from https://github.com/tallycash/extension/pull/2087/files#diff-91b61b0f030f6297dc4a42d91331e187c77d1ca989cb2f277a62c4aacb4245b2L276 into its own method.  I think refactoring it is out of the scope of this PR.

jagodarybacka (approved):
No issues found, let's go 🚀

background/services/chain/index.ts
```diff
@@ -346,6 +321,89 @@ export default class ChainService extends BaseService<Events> {
...
+      })
+      return this.activeNetworks
+    }
+
+    // Default to supporting Ethereum so ENS resolution works during onboarding
+    this.activateNetworkOrThrow(ETHEREUM.chainID)
+    return this.activeNetworks
+  }
+
+  async subscribeToNetworkEvents(network: EVMNetwork): Promise<void> {
```
Shadowfiend, line 350: This should probably be private and pushed lower in the class, right? Seems like we want external callers to always go through `activateNetworkOrThrow`.

background/services/chain/index.ts
```diff
@@ -346,6 +321,89 @@ export default class ChainService extends BaseService<Events> {
...
+    const networksToTrack = await this.getNetworksToTrack()
+    if (networksToTrack.length > 0) {
+      networksToTrack.forEach((network) => {
+        this.activateNetworkOrThrow(network.chainID)
+      })
+      return this.activeNetworks
+    }
+
+    // Default to supporting Ethereum so ENS resolution works during onboarding
+    this.activateNetworkOrThrow(ETHEREUM.chainID)
```
Shadowfiend, line 346: Would it make sense to move this logic to `getNetworksToTrack`? The idea being that we are saying “if we are tracking no networks explicitly, track Ethereum mainnet”.

That would make the purpose of this method a little tighter.

background/services/chain/index.ts
```diff
@@ -346,6 +321,89 @@ export default class ChainService extends BaseService<Events> {
...
+    }
+
+    // Since activeNetworks will be an empty array at extension load (or reload time)
+    // we need a durable way to track which networks an extension is tracking.
+    // The below code should only be called once per extension reload for extensions
+    // with active accounts
+    const networksToTrack = await this.getNetworksToTrack()
+    if (networksToTrack.length > 0) {
+      networksToTrack.forEach((network) => {
+        this.activateNetworkOrThrow(network.chainID)
```
Shadowfiend, line 340: Both here and in the final method `return`, if we don't `await` the `activateNetworkOrThrow` calls, I don't think the return value will reflect all the active networks (since `this.activeNetworks` is updated by `activateNetworkOrThrow` async, likely the next tick). Is that the behavior we're wanting?

If not, we may need to do the whole `Promise.all` or `allSettled` + `map` dance here. On the bright side, at that point we can just return the result of the `map`/`Promise` combo, since `activateNetworkOrThrow` returns the activated network when available.

background/services/chain/index.ts
```diff
@@ -346,6 +321,89 @@ export default class ChainService extends BaseService<Events> {
...
+    const networkToActivate = this.supportedNetworks.find(
+      (ntwrk) => toHexChainID(ntwrk.chainID) === toHexChainID(chainID)
+    )
+    if (!networkToActivate) {
+      throw new Error(`Network with chainID ${chainID} is not supported`)
+    }
+
+    this.activeNetworks.push(networkToActivate)
+
+    const existingSubscription = this.subscribedNetworks.find(
```
Shadowfiend, line 386: Do we expect that we would have an existing subscription to a network that we didn't activate?

background/services/chain/index.ts
```diff
@@ -346,6 +321,89 @@ export default class ChainService extends BaseService<Events> {
...
+      (networkSubscription) =>
+        networkSubscription.network.chainID === networkToActivate.chainID
+    )
+
+    if (!existingSubscription) {
+      this.subscribeToNetworkEvents(networkToActivate)
+      const addressesToTrack = new Set([
+        ...(await this.getAccountsToTrack()).map((account) => account.address),
+      ])
```
Shadowfiend, line 395:
```suggestion
      const addressesToTrack = new Set(
        await this.getAccountsToTrack()).map((account) => account.address),
      )
```

I expect this would work without the extra array unwrap/rewrap.

background/services/chain/index.ts
```diff
@@ -355,10 +413,10 @@ export default class ChainService extends BaseService<Events> {
 
     if (!provider) {
       logger.error(
-        "Request received for operation on unsupported network",
+        "Request received for operation on an inactive network",
```
Shadowfiend, line 416: Should we error in this case or attempt to activate the network?

background/services/chain/serial-fallback-provider.ts
```diff
@@ -660,3 +663,17 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
+  return new SerialFallbackProvider(
+    network,
+    () =>
+      new AlchemyWebSocketProvider(
+        getNetwork(Number(network.chainID)),
+        ALCHEMY_KEY
+      ),
+    () => new AlchemyProvider(getNetwork(Number(network.chainID)), ALCHEMY_KEY)
+  )
+}
```
Shadowfiend, line 679: This function is only used in one place and has very little to do with SerialFallbackProvider itself—it's mostly making decisions on the providers to use (which may vary by network in the future independent of the fallback provider itself) … Overall I'm not really convinced it's worth lifting it to an exported function, fwiw.

background/services/internal-ethereum-provider/index.ts
```diff
@@ -339,12 +341,31 @@ export default class InternalEthereumProviderService extends BaseService<Events>
...
+    chainID: string
+  ): Promise<EVMNetwork | undefined> {
+    const activeNetworks = await this.chainService.getActiveNetworks()
+    const activeNetwork = activeNetworks.find(
+      (network) => toHexChainID(network.chainID) === toHexChainID(chainID)
     )
-    return network
+    if (activeNetwork) {
+      return activeNetwork
+    }
```
Shadowfiend, line 361: This is already handled by `activateNetworkOrThrow`, right?

Shadowfiend (commented):
Left some mostly non-blocking thoughts.

I do want to draw attention to [the behavior of the return from `getActiveNetworks`](https://github.com/tallycash/extension/pull/2087#discussion_r961788850) though—I think the behavior there may differ from expectations in a significant way.

background/services/chain/index.ts
0xDaedalus: Fantastic improvement - thanks. ❤️

background/services/chain/index.ts
0xDaedalus: This is a good catch - and should be changed to your suggestion because whats going on is opaque - but the return value _will_ actually reflect all active networks since adding a network to the activeNetworks list is synchronous.

The asynchronous part is subscribing to the network events - so we do for a brief moment have an "activeNetwork" which is not subscribed to anything.

The reason we use a `forEach` here is so that if 1 of N networks is unable to connect (lets say due to alchemy issues) we can still connect to the other N-1 networks.  I do think that a Promise.allSettled is with an `await` is better here as that makes this more explicit.

background/services/chain/serial-fallback-provider.ts
0xDaedalus: You're right - this is setup for a follow-on PR which incorporates connecting to providers on-demand rather than connecting to all supported networks at app startup.

background/services/internal-ethereum-provider/index.ts
0xDaedalus: I'd like to separate out getting an active network and activating one (even though we have the default of returning an already active network inside of `getActiveNetworkOrThrow`) even though its not as DRY.  My thinking is that returning an already active network from `getActiveNetworkByChainId` should be expected behavior but returning an already active network from `activateNetworkOrThrow` should logger.warn.

background/services/chain/index.ts
0xDaedalus: No we don't - this is just me being extra careful.

background/services/chain/index.ts
0xDaedalus: I'd like to match the preexisting behavior for the time being

background/services/chain/index.ts
Shadowfiend: > the return value will actually reflect all active networks since adding a network to the activeNetworks list is synchronous

Interesting, this is me misunderstanding a subtlelty of the `async` execution model. I assumed the whole function call was wrapped in a promise, but in fact the call only goes async on the first `await`, all previous code does in fact execute synchronously for the caller.

I'm actually quite comfortable keeping this as-is with that in mind, tbh. The other way is more explicit, but it also introduces a lot of promise noise. It may be worth adding a small note that the `activeNetworks` will be updated synchronously in the called function.

background/services/chain/index.ts
Shadowfiend: The only reason I bring this up is that we're in many ways *not* matching the existing behavior. Prior to this PR, receiving a call for an unsupported network is the only thing that would throw—but in this case, we could receive a call for a supported but deactivated network, and we would throw for that as well. In that sense it's safer to activate the network, since that reduces the number of possible calls that would have succeeded before but would fail now.

That said, it seems unlikely we'd find ourselves in this situation—except perhaps via dApp connections?

background/services/chain/serial-fallback-provider.ts
Shadowfiend: I'll withhold judgement until then in that case :)

background/services/internal-ethereum-provider/index.ts
Shadowfiend: Mmm I see. Perhaps it would make sense to move `getActiveNetworkByChainId` up to `ChainService` and call that from here? Can't quite figure out why this feels so itchy to me though, so I'm fine keeping it as-is.

PR #2089, opened 2022-08-31
Make default account names more random

### LINKED ISSUE #2075: Make default accounts names and avatars more random
Let's try to avoid cases where multiple accounts added in a row have the same default account names and avatars as this is not looking UX friendly.
image

jagodarybacka:
Resolves #2075

### What
Avoid adding two accounts with the same default name and avatar in the row.
### Testing
Add multiple accounts and check if no two addresses in a row have the same name and avatar

### DIFF
--- a/background/redux-slices/accounts.ts
+++ b/background/redux-slices/accounts.ts
@@ -120,13 +120,15 @@ function newAccountData(
             (existingAccountsCount % availableDefaultNames.length)
         )
     )
-  const defaultAccountName =
-    sameAccountOnDifferentChain?.defaultName ??
-    availableDefaultNames[defaultNameIndex]
 
-  // Move used default names to the start so they can be skipped above.
-  availableDefaultNames.splice(defaultNameIndex, 1)
-  availableDefaultNames.unshift(defaultAccountName)
+  let defaultAccountName = sameAccountOnDifferentChain?.defaultName
+
+  if (typeof defaultAccountName === "undefined") {
+    defaultAccountName = availableDefaultNames[defaultNameIndex]
+    // Move used default names to the start so they can be skipped above.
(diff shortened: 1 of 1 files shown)


hyphenized:
We need more dog names 😆

kkosiorowska:
It looks really good 😃  but after several attempts managed to reproduce the issue again.

Screenshot 2022-09-01 at 09 17 57

Also, I am wondering about a few things. 🤔
1. The second account added can have the same avatar. Are we able to do something about it ?
Screenshot 2022-09-01 at 08 49 47

2. We have 8 names available by default. The list of new accounts is presented below.  The second element is the same as the last but we have one more name available - Foz. I wonder if we are able to use all available names first ?
(1. Sport 2. Lola 3. Matilda 4. Topa 5. Phoenix 6. Sirius 7. Atos 8. Lola) 
Screenshot 2022-09-01 at 09 04 36

jagodarybacka:
@kkosiorowska 
> 1. The second account added can have the same avatar.

Have you removed any accounts in the meantime, between adding first account and adding second one with husky pfp? 

>  wonder if we are able to use all available names first ?

RIght now we are trying to make these pfp/names "random", not sure if we want to just make a randomized list and use them one by one.

kkosiorowska:
@jagodarybacka 

> Have you removed any accounts in the meantime, between adding first account and adding second one with husky pfp?

I haven't removed anything in the meantime. I was just wondering if it is possible and after several attempts, it happened.

kkosiorowska:
@jagodarybacka 
Anyway for my perspective it looks good and resolved the most important issue. 💖

kkosiorowska (approved):
For my perspective it looks good and resolved the most important issue. 💖

PR #2090, opened 2022-08-31
Add all new accounts to all networks

### LINKED ISSUE #1864: Read-only address disappears when switching to Polygon
### Discord Discussion Link

na

Chrome

Switch between read-only accounts on mainnet and Polygon

### What did not work?

One read only address is can only be selected as an address when the extension is on mainnet. When switching to Polygon it's no longer on the list. This was also the first read only address that was added. The balance of this address is still included in the overview page though.

The other two read only addresses that were added after that, are in the list both when the extension is on mainnet and on Polygon.

Side note. All of these 3 addresses were added through an ENS or UD. These domain names are loaded when the extension is on mainnet but not when it is on Polygon.

### Version

Community Edition v0.14.1

### Relevant log output

_No response_

jagodarybacka:
Resolves #1864

### What
- Add all types of accounts to all supported networks at once.
- Refactor add account methods to avoid code duplication.
### Testing
- add multiple accounts of all types - import, read only, Ledger - to the wallet and make sure they are visible on the accounts list no matter what chain is selected

image

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -471,7 +471,7 @@ export default class Main extends BaseService<never> {
     this.connectPopupMonitor()
   }
 
-  async addAccount(addressNetwork: AddressOnNetwork): Promise<void> {
+  async addAccountToTrack(addressNetwork: AddressOnNetwork): Promise<void> {
     await this.chainService.addAccountToTrack(addressNetwork)
   }
 
@@ -485,6 +485,19 @@ export default class Main extends BaseService<never> {
       network,
       name,
     })
+  }
+
+  async addAccountToAllNetworks(address: string): Promise<void> {
+    await Promise.allSettled(
+      this.chainService.supportedNetworks.map(async (network) => {
+        const addressNetwork = {
+          address,
+          network,
+        }
+        await this.chainService.addAccountToTrack(addressNetwork)
+        this.store.dispatch(loadAccount(addressNetwork))
+      })
+    )
   }
 
   async removeAccount(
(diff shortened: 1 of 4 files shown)


background/main.ts
```diff
@@ -487,6 +487,19 @@ export default class Main extends BaseService<never> {
...
   }
 
+  async addAccountToAllNetworks(address: string): Promise<void> {
+    await Promise.allSettled(
+      this.chainService.supportedNetworks.map(async (network) => {
+        const addressNetwork = {
+          address,
+          network,
+        }
+        await this.chainService.addAccountToTrack(addressNetwork)
```
mhluongo, line 497: Left and right hands are fighting here @jagodarybacka. This is the opposite of what we want to do for performance.

mhluongo: Instead, I recommend we add a layer of indirection in Redux... where we have all `addressNetworks` to select from, but only those chosen get added to `accountsToTrack` in the background script

mhluongo (changes requested):
I think you and @0xDaedalus need to get aligned here

background/main.ts
jagodarybacka: I'm aware that adding accounts to track to all supported chains at once means more performance problems. But what this PR is changing is only adding read-only accounts the same way we add imported and Ledger accounts. I agree we need a better way to optimize tracking accounts but I think having all types of accounts behave the same way will be a good first step. Then we can improve adding accounts to track.

0xDaedalus: I'm inclined to agree with Matt here.  While it would be nice to have all accounts behaving the same way what we want is a fix that doesn't impede our scaling and that we can extend to imported and ledger accounts in a subsequent PR.

jagodarybacka:
I'm merging my work from this PR with @0xDaedalus PR - https://github.com/tallycash/extension/pull/2100

PR #2092, opened 2022-09-01
Fix missing activities for read only accounts

hyphenized:
This fixes an issue with imported accounts, where activities never load since the UI slice's selected account stores a dirty address that never gets matched to it's normalized version in the activities slice.

## To test
1. Import this address `0x00976385980F6a03606C29EB97627691d7e3CAA6` or any non normalized address 
2. Activities should show up now


jagodarybacka (approved):
Thank you for fixing it 💖

PR #2096, opened 2022-09-01
Remove dApp permissions with account

### LINKED ISSUE #1978: Removing an account from the wallet should remove its associated dapp permissions from indexDB
_Originally posted by @jagodarybacka in https://github.com/tallycash/extension/issues/1969#issuecomment-1219357589_

jagodarybacka:
Resolves #1978

### What
When user removes address we should remove its dApp permissions too from db and redux store

### Testing
1. Add a couple of accounts to the wallet
2. Open background script devtools so you can observe how dAppPermissions database is changing 
image

3. Connect to any dapp - confirm in the UI it is connected on both dapp and wallet side, refresh db and check new connection is saved there
4. Remove address that is connected to the dapp
5. Confirm it is correctly disconnected from dapp and wallet, UI and db
6. Add that address again - it shouldn't be connected to that dapp
7. Confirm dapps connections feature is working as before in general (connect, switch accounts, disconnect, sign tx)

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -85,6 +85,7 @@ import {
   requestPermission,
   emitter as providerBridgeSliceEmitter,
   initializePermissions,
+  revokePermissionsForAddress,
 } from "./redux-slices/dapp"
 import logger from "./lib/logger"
 import {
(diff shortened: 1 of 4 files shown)


0xDaedalus (approved):
Looks good and works great.  Left one non-blocking comment and will wait a bit to merge in case you want to address it.

0xDaedalus (approved):
🎉

PR #2098, opened 2022-09-01
Populate ActivityItem `spender` when able to.

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/2091

### To Test
- Load a new wallet on version `0.14.7`
- Approve a token you have not previously approved before
- Upgrade to version to this branch
- Open the Activity Page.
- [ ] Activity page should not crash.

### DIFF
--- a/background/redux-slices/migrations/to-16.ts
+++ b/background/redux-slices/migrations/to-16.ts
@@ -1,0 +1,114 @@
+import { NETWORK_BY_CHAIN_ID } from "../../constants"
+import { EVMNetwork } from "../../networks"
+
+type OldState = {
+  activities: {
+    [address: string]: {
+      [chainID: string]: {
+        entities: {
+          [txHash: string]: {
+            hash: string
+            annotation: {
+              type: string
+              spenderName: string
+              spenderAddress: string
+              spender?: {
+                address: string
+                network: EVMNetwork
+                annotation: {
+                  nameOnNetwork: {
+                    name: string
+                    network: EVMNetwork
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+  [otherSlice: string]: unknown
+}
+
+type NewState = {
+  activities: {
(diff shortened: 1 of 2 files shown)


background/redux-slices/migrations/to-16.ts
```diff
@@ -0,0 +1,114 @@
...
+              annotation &&
+              annotation.type === "asset-approval" &&
+              (annotation.spenderAddress || annotation.spenderName)
+            ) {
+              const spender = {
+                address: annotation.spenderAddress,
+                network: NETWORK_BY_CHAIN_ID[chainID],
+                annotation: {
+                  nameOnNetwork: {
+                    name: annotation.spenderName,
```
mhluongo, line 88: If there's no `spenderName` I don't think there should be a `nameOnNetwork`... versus eg

```typescript
annotation: {
    nameOnNetwork: {
        name: undefined,
        network: ETHEREUM
    }
}
```

which I believe this code will do now?

background/redux-slices/migrations/to-16.ts
```diff
@@ -0,0 +1,114 @@
...
+                    network: NETWORK_BY_CHAIN_ID[chainID],
+                  },
+                },
+              }
+
+              const { spenderName, spenderAddress, ...newAnnotation } =
```
mhluongo, line 94: Maybe consider renaming `newAnnotation`? It's not the new annotation, after all. `oldAnnotationProps` perhaps?

mhluongo (commented):
Couple questions before I test!

background/redux-slices/migrations/to-16.ts
0xDaedalus: right you are ✅

jagodarybacka (approved):
-> `0.14.7` 
-> Approve SUSHI on Polygon 
-> `main` -> activities broken 
-> this branch -> activities fixed

LGTM 🚢

PR #2100, opened 2022-09-01
Introduce active networks with read only accounts

jagodarybacka:
Ref #1864

### What
Handle read only accounts the same way as imported and Ledger accounts - add them to active networks

image


jagodarybacka:
Accidentally fixed by https://github.com/tallycash/extension/pull/2087

PR #2102, opened 2022-09-01
Queue transaction retrievals

hyphenized:
This PR adds a small delay( 2.5 secs ) between each transaction retrieval, this should give the API breathing room to process all the necessary calls involved in the process for each transaction, such as:

- Getting transaction details

- Enriching transaction(Highest cost) x2
> - Getting transaction count (nonce)
> - Getting code
> - Getting balance for enriched address
> - Lookup of names for enriched address

And finally:  
- Getting block height
- Getting account balance (conditional)

Additionally this also considerably reduces the possibility of multiple users spamming the API at the same time and getting rate limited


background/services/chain/index.ts
```diff
@@ -1121,7 +1122,12 @@ export default class ChainService extends BaseService<Events> {
         // If more transactions can be retrieved in this alarm, bump the count,
         // retrieve the transaction, and drop from the updated queue.
         fetchedByNetwork[network.chainID] += 1
-        this.retrieveTransaction(network, hash, firstSeen)
+        queue = queue.then(() =>
```
jagodarybacka, line 1125: I think a comment here explaining why this is needed would be nice 😄

0xDaedalus: Awesome direction - but I don't think this is _quite_ what we want here.

In the current implementation - if we try to retrieve 5 transactions - and the first transaction fails retrieve - the other 4 still get retrieved.

In this implementation - I believe that if the first transaction fails to retrieve then the following 4 will fail as well.

hyphenized: @jagodarybacka Good idea, thanks! I'll add one.
@0xDaedalus 😨 Good catch! This relies too much on `retrieveTransaction` to not throw any errors. Thanks, I'll change it.

0xDaedalus (approved):
Tested and looks good to me - I'd love to further expand on this with something like:

```
type Options = {
  concurrency: number,
  wait: number,
}

async resolvePromises(promises: Array<Promise<any>>, options: Options) {
 ....
}
```
and use that throughout the codebase - but thats definitely out of the scope of this PR.

0xDaedalus (approved):

hyphenized:
> Tested and looks good to me - I'd love to further expand on this with something like:
> 
> ```
> type Options = {
>   concurrency: number,
>   wait: number,
> }
> 
> async resolvePromises(promises: Array<Promise<any>>, options: Options) {
>  ....
> }
> ```
> 
> and use that throughout the codebase - but thats definitely out of the scope of this PR.


Ty, I agree! As a side note I just came across this library `https://github.com/caolan/async` which has some interesting async utilities 👀

PR #2105, opened 2022-09-02
WIP: Make broader user of makeFetchWithTimeout

### LINKED ISSUE #2036: Make broader user of `makeFetchWithTimeout`.
We introduced [makeFetchWithTimeout](https://github.com/tallycash/extension/blob/main/background/services/name/resolvers/uns.ts#L8:L8) as part of a hotfix in `v0.14.6` to alleviate issues resulting from long http response times when querying UNS servers.  

We should extend and standardize this functionality to all `fetch` requests we make in the app with a default timeout (open to suggestions on what it should be) that is also configurable in case we need a shorter or longer timeout for a particular http endpoint.

kkosiorowska:
Closes #2036 

This PR makes broader user of `makeFetchWithTimeout`.

**Changes**

- make timeout value configurable
- set default timeout value
- change all fetch functions to `fetchWithTimeout`

**To Test**
- [ ] check that all fetch functions are working properly

### DIFF
--- a/background/lib/alchemy.ts
+++ b/background/lib/alchemy.ts
@@ -18,6 +18,7 @@ import {
   isValidAlchemyTokenMetadataResponse,
 } from "./validate"
 import { AddressOnNetwork } from "../accounts"
+import fetchWithTimeout from "../utils/fetching"
 
 /**
  * Use Alchemy's getAssetTransfers call to get historical transfers for an
@@ -338,7 +339,7 @@ export async function getNFTs({
   requestUrl.searchParams.set("pageSize", "100")
 
   // TODO validate data with ajv
-  const result = await (await fetch(requestUrl.toString())).json()
+  const result = await (await fetchWithTimeout()(requestUrl.toString())).json()
   return result.ownedNfts
     .filter((nft: AlchemyNFTItem) => typeof nft.error === "undefined")
     .map((nft: AlchemyNFTItem) => ({ ...nft, chainID: network.chainID }))
--- a/background/services/name/resolvers/uns.ts
+++ b/background/services/name/resolvers/uns.ts
@@ -1,31 +1,12 @@
 import { AddressOnNetwork, NameOnNetwork } from "../../../accounts"
 import { ETHEREUM, POLYGON, SECOND } from "../../../constants"
-import logger from "../../../lib/logger"
 import { isDefined } from "../../../lib/utils/type-guards"
 import { sameNetwork } from "../../../networks"
+import fetchWithTimeout from "../../../utils/fetching"
 import { NameResolver } from "../name-resolver"
 
 // Time until response is stale
 const RESPONSE_TTL = 5 * SECOND
-
-const makeFetchWithTimeout = (timeoutMs: number) => {
-  return async function fetchWithTimeout(
-    requestInfo: RequestInfo,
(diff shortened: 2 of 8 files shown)


background/lib/alchemy.ts
```diff
@@ -338,7 +339,7 @@ export async function getNFTs({
   requestUrl.searchParams.set("pageSize", "100")
 
   // TODO validate data with ajv
-  const result = await (await fetch(requestUrl.toString())).json()
+  const result = await (await fetchWithTimeout()(requestUrl.toString())).json()
```
0xDaedalus, line 342: Very close to what we want here but we want `fetchWithTimeout` to have the same API as `fetch` - e.g. `await fetchWithTimeout(requestUrl.toString())`

background/utils/fetching.ts
```diff
@@ -0,0 +1,27 @@
...
+      ...options,
+      signal: controller.signal,
+    })
+    clearTimeout(id)
+    return response
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, prettier/prettier
+const fetchWithTimeout = (timeoutMs = DEFAULT_TIMEOUT) => makeFetchWithTimeout(timeoutMs)
```
0xDaedalus, line 25: A couple of notes

a) I think we can `export const fetchWithTimeout` here`
b) I assume you want this to be the default `fetchWithTimeout` that's used throughout the application.  In that case I'd probably just have this export a [fetchWithTimeout](https://github.com/tallycash/extension/pull/2105/files#diff-39ae85fe22d82d57964144c1d2510a9ed8cff40b55dc9865ae8430640da8865bR6) that with the default timeout.

background/utils/fetching.ts
```diff
@@ -0,0 +1,27 @@
+import { logger } from "ethers"
+
+const DEFAULT_TIMEOUT = 3_000
```
0xDaedalus, line 3: lets make this 10s

background/services/name/resolvers/uns.ts
```diff
@@ -80,15 +61,14 @@ const cacheAsyncResults = <F extends (...args: any[]) => Promise<any>>(
     return result
   }) as F
 }
-const fetchWithTimeout = makeFetchWithTimeout(3_000)
```
0xDaedalus, line 83: With the default being changed to 10s - lets revert this guy so we still have 3s timeouts when hitting UNS.

0xDaedalus (changes requested):

mhluongo:
Calling out that we might want to consider using `fetchJson` from `ethers` for many of these instead — it includes built-in exponential back-offs and timeouts, IIRC.

mhluongo:
Confirmed reading the code... [it does what we need](https://github.com/ethers-io/ethers.js/blob/ec1b9583039a14a0e0fa15d0a2a6082a2f41cf5b/packages/web/src.ts/index.ts#L362). It's semi-private younger sibling `_fetchData` can cover any non-JSON as well.

0xDaedalus:
Good catch Matt and I am inclined to agree.   I think what we have here is still a good incremental step forward so I'll make a follow-on issue.

0xDaedalus (approved):
💯

PR #2110, opened 2022-09-02
Enable Tally users to use Matcha.xyz

0xDaedalus:
matcha.xyz does not play nicely with Tally - so lets add it to our list of websites on which to impersonate metamask.

### To Test
- [ ] Navigate to matcha.xyz.
- [ ] With "Impersonate Metamask" set to off - you should not be able to connect.
- [ ] With "Impersonate Metamask" set to on - you should be able to connect by clicking on the Metamask option.


jagodarybacka (approved):
Works fine ✅ 

I don't like that this button is not doing anything but it's on their side 😬
image

PR #2111, opened 2022-09-03
Display current network's base asset earlier in assets list

### LINKED ISSUE #2084: MATIC (ERC-20) is showing up before ETH on the Ethereum wallet page
Screenshot from 2022-08-30 12-34-12

Seems like we're considering the MATIC ERC-20 a base asset when we shouldn't be.

hyphenized:
Closes #2084 

image

### To Test
- [x] Load an account with both ETH and MATIC (e.g. `0xb23da923a507761cd1561b32cc6a0cc47b345fbc`)
- [x] When using the Ethereum network ETH should be displayed at the top of the list, the same goes for MATIC when the current network is Polygon

### DIFF
--- a/background/redux-slices/selectors/accountsSelectors.ts
+++ b/background/redux-slices/selectors/accountsSelectors.ts
@@ -138,6 +138,18 @@ const computeCombinedAssetAmountsData = (
         return 1
       }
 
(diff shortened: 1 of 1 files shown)


mhluongo (approved):
:ship:

PR #2115, opened 2022-09-03
More Chain Service Tests

0xDaedalus:
### Highlights:
- No need to import `import "fake-indexeddb/auto"` at the top of every service test anymore
- Removed unused `setAccountsToTrack` method from chain service.
- Added ew factories to make testing transaction-related flows easier
- Added Examples of how to integration test functionality that persist to indexedDB (addAccountToTrack tests)
- Extracted pending transaction handler into its own method & added integration test.


background/services/chain/tests/index.integration.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  TransactionRequestWithNonce,
+} from "../../../networks"
+import {
+  createAnyEVMTransaction,
+  createChainService,
+  createLegacyTransactionRequest,
+} from "../../../tests/factories"
+import SerialFallbackProvider from "../serial-fallback-provider"
+
+type ChainServiceExternalized = Omit<ChainService, ""> & {
```
mhluongo, line 16: Why are we `Omit` ing an empty string? :thinking:

background/tests/factories.ts
```diff
@@ -47,3 +49,50 @@ export const createSigningService = async (
     overrides.chainService ?? createChainService()
   )
 }
+
+export const createLegacyTransactionRequest = (
```
mhluongo, line 53: Where did you get this test data? A comment would be nice, unsure if it's "special" at all

background/services/chain/tests/index.integration.test.ts
```diff
@@ -0,0 +1,96 @@
...
+describe("Chain Service", () => {
+  const sandbox = sinon.createSandbox()
+  let chainService: ChainService
+
+  beforeEach(async () => {
+    sandbox.restore()
+    chainService = await createChainService()
+    await chainService.startService()
+  })
+  it("handlePendingTransactions should update nonce tracking, subscribe to transaction confirmations, and persist the transaction to indexedDB", async () => {
```
mhluongo, line 35: I'd love to land regression tests for our nonce issues in here after this PR :point_right: :point_left: :blush: cc @jagodarybacka

background/services/chain/tests/index.integration.test.ts
```diff
@@ -0,0 +1,96 @@
...
+      validOptimismEVMTransaction
+    )
+    // Handling a pending transaction should update the last seem EVM transaction nonce
+    expect(
+      chainServiceExternalized.evmChainLastSeenNoncesByNormalizedAddress[
+        chainID
+      ][validOptimismEVMTransaction.from]
+    ).toBe(validOptimismEVMTransaction.nonce)
+
+    // provider.once should be called inside of subscribeToTransactionConfirmation
```
mhluongo, line 84: Ooh, seeing an opportunity to do some performance regression tests here too :hugs:

background/services/chain/tests/db.integration.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    it("should correctly add accounts to track to indexedDB", async () => {
+      expect(await db.getAccountsToTrack()).toHaveLength(0)
+      await db.addAccountToTrack(account1)
+      expect(await db.getAccountsToTrack()).toHaveLength(1)
+      await db.addAccountToTrack(account2)
+      const accountsToTrack = await db.getAccountsToTrack()
+      expect(accountsToTrack).toEqual([account1, account2])
+    })
+
+    it("should not add the same account twice.", async () => {
```
mhluongo, line 34: This doesn't really do what it says on the time... maybe do it with a few test accounts, and make sure you're checking the set for uniqueness, rather than length?

mhluongo (approved):
Really excited about this! I'm looking for one more :+1: to merge though, how about it @jagodarybacka @hyphenized @Shadowfiend?

background/services/chain/tests/index.integration.test.ts
0xDaedalus: Because we want our `Externalized` type to inherit from `ChainService` but overwrite its private methods.  Without an Omit this type of inheritance is correctly disallowed.  

tl;dr this is to trick the type system

jagodarybacka (approved):
Looks fine ✅ Do you want to address any non blocking comments here @0xDaedalus? If not I will merge it later

PR #2118, opened 2022-09-04
Manage connected dapps

### LINKED ISSUE #1290: Menu - Connected websites
Connected websites is a page in the main menu that should allow users to disconnect for any dApp they have previously connected too.

UI Requirements:
 - An item in main menu
 - Show a list of current connected websites of the currently selected account
 - Ability to independently disconnect from each website
 - Show a snackbar after a disconnect to confirm website was disconnected
 
[ Figma link](https://www.figma.com/file/EiXpBpNz2HcoeFJftgC5oc/?node-id=634%3A2366)
 
image

hyphenized:
Closes #1290

Allows users to manage their connected dApps through settings, and hopefully sleep better at night.

Needs an empty state message 😶 

image

## To test
- [ ] Set SUPPORT_MANAGE_DAPPS in your environment to true 
- [ ] Go to https://app.uniswap.com, connect your wallet
- [ ] Go to settings, click on connected websites, Uniswap should appear in the list
- [ ] Click on the disconnect button, Uniswap's `Connect Wallet` button should appear again


mhluongo:
@VladUXUI thoughts on the empty state?

Shadowfiend:
Feels like navigating to the page on an empty state shouldn't be possible?

hyphenized:
> Feels like navigating to the page on an empty state shouldn't be possible?

That's true, I thought about hiding the option if there were no connected dapps but that might affect the UX

VladUXUI:
image

Awesome work here @hyphenized!
Let's use this for our empty state. Agreed that since this is a settings menu, it should not disappear.
This is the same empty state we use for NFT and other places, so sources should be in the code already

ui/pages/Settings/ConnectedWebsitesListItem.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <div className="container">
+      <div className="connected-website-item">
+        <img className="logo" src={icon} width={32} alt={title} />
+        <div className="connected-website-details">
+          <span className="title">{title}</span>
+          <span className="host">{host}</span>
+        </div>
+        <SharedIcon
+          color="var(--error)"
+          width={32}
```
VladUXUI, line 44: If this is the disconnect button, let's use 24px. All of our icons are either 24 or 16px

jagodarybacka (commented):
Feature looks awesome 💖 Let's implement Vlad's design for empty page.

---
Also I've found 🐛 :
- connect address to uniswap and cowswap, have both dapps open
- with cowswap tab active click disconnect on uniswap item

Result: cowswap dapp is disconnected too. It's not a big problem because it is not loosing connection - after reload it is fine. But would be nice to take a look or add new issue if this will stay unresolved.

https://user-images.githubusercontent.com/20949277/188434423-243606a4-734c-444b-b9f8-31ce99e70e28.mov

ui/pages/Settings/ConnectedWebsitesListItem.tsx
hyphenized: Sure can! I've fixed the size and changed the default color to `--green-40` and `error` on hover

hyphenized:
Thanks for the comments everyone!

I could not reproduce the bug on chrome/chromium, feels like I missed something 🤔. I also tried clicking away from the pop-up before the success toast shows up but no success

sneakybug

jagodarybacka (approved):
I can't reproduce the problem with uni/cow swap today 🤔 

Anyway, this looks great ✨ 🚀

PR #2123, opened 2022-09-05
Update the redux state version to 17

kkosiorowska:
This PR updates the redux state version to 17 for the analytics page.


jagodarybacka (approved):
✅

PR #2128, opened 2022-09-05
2011: create keyring methods and tests to enable password change feature

rgdalessandro:
This PR contains changes to the `keyring service` that are necessary to enable the change password feature.

(See issue#2011 --> https://github.com/tallycash/extension/issues/2011)

1. the `verifyPassword` method is created in `services/keyring/index.ts` to verify if a provided password is valid

2. the `changePassword` method is created in `services/keyring/index.ts` to create a new vault encrypted with a new user-chosen password

3. the `persistKeyrings` method is modified in `services/keyring/index.ts` to take the `deleteOlderVaults` optional parameter when clearing out all older vaults is desired (like when changing the password)

4. the `keepOnlyOneEncryptedVault` method is created in `services/keyring/storage.ts` to write one encrypted vault to extension storage, while removing all other vaults from extension storage

5. a test was added in `keyring-integration.test.ts`


background/services/keyring/storage.ts
```diff
@@ -86,3 +86,25 @@ export async function writeLatestEncryptedVault(
...
+ * removing all other vaults from extension storage.
+ *
+ * @param encryptedVault - an encrypted keyring vault
+ */
+export async function keepOnlyOneEncryptedVault(
+  encryptedVault: EncryptedVault
+): Promise<void> {
+  await browser.storage.local.set({
+    tallyVaults: {
+      version: 1,
```
rgdalessandro, line 101: I took a guess that `version: 1` was the right value to use here. Please correct me if I am wrong.

background/services/keyring/index.ts
```diff
@@ -144,6 +153,72 @@ export default class KeyringService extends BaseService<Events> {
...
+        password,
+        currentEncryptedVault.salt
+      )
+      try {
+        await decryptVault<SerializedKeyringData>(
+          currentEncryptedVault,
+          saltedKey
+        )
+      } catch (err) {
+        // if we weren't able to load the vault, don't unlock
```
mhluongo, line 182: This comment isn't accurate

background/services/keyring/index.ts
```diff
@@ -144,6 +153,72 @@ export default class KeyringService extends BaseService<Events> {
...
+   * Do not actually lock or unlock the vault.
+   *
+   * @param password A user-chosen string used to encrypt keyring vaults.
+   *        Unlocking will fail if an existing vault is found, and this password
+   *        can't decrypt it.
+   * @returns true if the password is able to unlock the vault,
+   *          and false otherwise.
+   */
+  // eslint-disable-next-line class-methods-use-this
+  private async verifyPassword(password: string): Promise<boolean> {
```
mhluongo, line 167: I think this is a better fit for `./encryption.ts` rather than bringing all these details into `index.ts`

background/services/keyring/index.ts
```diff
@@ -144,6 +153,72 @@ export default class KeyringService extends BaseService<Events> {
...
+    }
+
+    return false
+  }
+
+  /**
+   * Create a new vault encrypted with a new user-chosen password.
+   * Do not actually lock or unlock any vaults.
+   *
+   * @param changePasswordPayload An object with this shape:
```
mhluongo, line 196: The types cover this comment already

background/services/keyring/index.ts
```diff
@@ -144,6 +153,72 @@ export default class KeyringService extends BaseService<Events> {
...
+   * { currentPassword: string, newPassword: string }
+   *
+   * @returns true if the password was successfully changed,
```
mhluongo, line 199: Should this perhaps throw? Not a strong opinion, just curious

background/services/keyring/storage.ts
mhluongo: Yep, it's actually enforced by the type right now!

background/services/keyring/storage.ts
```diff
@@ -86,3 +86,25 @@ export async function writeLatestEncryptedVault(
...
   }
 }
+
+/**
+ * Write one encryptedVault to extension storage, while
+ * removing all other vaults from extension storage.
+ *
+ * @param encryptedVault - an encrypted keyring vault
+ */
+export async function keepOnlyOneEncryptedVault(
```
mhluongo, line 96: This name is odd. Maybe prefer something like `overwriteEncryptedVaults`?

Aside from the naming... the plan here was to keep old vaults unless they are being rekeyed. Instead of deleting all of them indiscriminately, we should only delete the latest whose password was changed.

background/tests/keyring-integration.test.ts
```diff
@@ -284,6 +284,26 @@ describe("KeyringService when initialized", () => {
     expect(service.locked()).toEqual(false)
     expect(await service.unlock(testPassword)).toEqual(true)
   })
+
+  it("successfully changes the vault password when provided with the correct current password", async () => {
```
mhluongo, line 288: I'd love to see a test of the actual storage behavior — it looks like this only tests the derived in-memory key, but doesn't show that eg the old key material was deleted.

mhluongo (changes requested):
Appreciate your breaking the PR out, this is way easier to review!

Big blocker here is the unnecessary vault deletion — we want to preserve old unrelated vaults and only delete the most recent one if its password has been changed. We plan to rely on the existing behavior to allow wallets to be "logged out" of without accidentally killing people's keys

0xDaedalus:
> Big blocker here is the unnecessary vault deletion — we want to preserve old unrelated vaults and only delete the most recent one if its password has been changed.

Are you saying here that we want to unencrypt all old vaults and reencrypt them with the new password?

mhluongo:
> Are you saying here that we want to unencrypt all old vaults and reencrypt them with the new password?

So there's what we should do now, and what the plan is for the future.

**What I think we should do now**: re-encrypt the latest vault. In normal operation, there should only be 1 vault anyway. Don't delete unrelated vaults that will only be there in extraordinary cases.

**What I think we should do in the future**: Add a "forget password" function that let's people start a new vault. Keep old vaults around. Eventually, expose an interface in the settings page or elsewhere to manage those vaults.

Goal is to minimize accidental deletion of key material... except when we're sure it should be deleted, as in the case of key rotation.

background/services/keyring/index.ts
rgdalessandro: This comment has been updated to be more accurate and relevant.

background/services/keyring/index.ts
rgdalessandro: As suggested, I attempted to move this method out of the `KeyringService` class in `index.ts` and place it within the `encryption.ts` file. However, this led to linting errors because now the `encryptions.ts` file would need this additional import statement `import { getEncryptedVaults } from "./storage"`, but also the `storage.ts` already has this import statement currently `import { EncryptedVault } from "./encryption"` 

This led to the following linting error:
`ESLint: Dependency cycle detected.(import/no-cycle)`

Rather than trying to suppress this error. I kept `verifyPassword()` in its current file. Please let me know if this is acceptable or if I need to definitely find a way to move the logic into `encryption.ts` while also keeping the linter happy by finding a work around to the import cycle.

background/services/keyring/index.ts
rgdalessandro: I have updated the JSDoc description to be more qualitative and useful instead of something resembling a type definition.

background/services/keyring/index.ts
rgdalessandro: You bring up a good point, and I thought about it some more. I am still leaning towards the `return false` instead of throwing an error.

My thinking is that if something unexpected happens during execution, then throwing an error may be the more intuitive way to go. However, a user providing an incorrect password doesn't seem completely unexpected to me, as it is part of the normal flow of a password change to allow the user to keep re-attempting the password change if they initially fat-finger or forget their password. When an incorrect password is provided, then I believe the `changePassword` method is doing its job by reporting back that the password change was not successful. We later will use this to kick off a redux action so that the UI can reflect the appropriate validation error in the form.

I am not too strongly tied to this opinion, and open to changing this to throwing an error if that general consensus.

background/services/keyring/storage.ts
rgdalessandro: As suggested, I have renamed my new method to `overwriteLatestEncryptedVault()`. I have also renamed an existing method to match, `appendLatestEncryptedVault()`. I believe these names help convey to developers exactly what we want to do and how these methods differ.

I have also updated the logic so that only the `currentLatest` vault is overwritten, and all others are kept as they were.

0xDaedalus:
I think there might be competing ideas between what you suggest above and the implementation suggested in  https://github.com/tallycash/extension/pull/2015#discussion_r958684628

background/tests/keyring-integration.test.ts
rgdalessandro: I tried to improve testing with some checks around service.lock() and service.unlock()

I am not yet sure how I will test the localStorage before/after a password change.

mhluongo:
It's the same idea, just restricted a bit. Best phrased as "only re-encrypt the most recent vault". In our current implementation, if the vault has the same password... it should be the most current vault.

jagodarybacka:
This PR is stale and not aligned with the current roadmap - let's close it

Issue #2127, opened 2022-09-05
Gnosis Safe creation transaction fails to broadcast


brandoncurtis:
### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

Deploy a new Gnosis Safe using the gnosis-safe.io UI and Ledger Nano X.

### What did not work?

The Safe creation transaction repeatedly fails to broadcast.

Fails whether or not I change the fee rate from the default.

Based on the displayed fee on the Ledger and the errors in the logs, I believe this is a problem in fee calculation resulting in an insufficient fee.

### Version

v0.15.0

### Relevant log output

```shell
SerialFallbackProvider.send
Skipping fallback for unidentified error,{"code":-32000,"response":"{\"jsonrpc\":\"2.0\",\"id\":17345,\"error\":{\"code\":-32000,\"message\":\"err: max fee per gas less than block base fee: address [REDACTED] maxFeePerGas: 8000000000 baseFee: 16715458487 (supplied gas 272084)\"}}"},for provider,{"_isProvider":true,"_events":[],"_emitted":{"block":-2},"formatter":{"formats":{"transaction":{},"transactionRequest":{},"receiptLog":{},"receipt":{},"block":{},"blockWithTransactions":{},"filter":{},"filterLog":{}}},"anyNetwork":false,"_network":{"name":"homestead","chainId":1,"ensAddress":"[REDACTED]
    at SerialFallbackProvider.send (chrome-extension://eajafomhmkipbjmfmhebemolkcicgfmd/background.js:58321:24) 

SerialFallbackProvider.send
Skipping fallback for unidentified error,{"code":-32000,"response":"{\"jsonrpc\":\"2.0\",\"id\":17362,\"error\":{\"code\":-32000,\"message\":\"err: max fee per gas less than block base fee: address [REDACTED] maxFeePerGas: 8000000000 baseFee: 17221112807 (supplied gas 272084)\"}}"},for provider,{"_isProvider":true,"_events":[],"_emitted":{"block":-2},"formatter":{"formats":{"transaction":{},"transactionRequest":{},"receiptLog":{},"receipt":{},"block":{},"blockWithTransactions":{},"filter":{},"filterLog":{}}},"anyNetwork":false,"_network":{"name":"homestead","chainId":1,"ensAddress":"[REDACTED]
    at SerialFallbackProvider.send (chrome-extension://eajafomhmkipbjmfmhebemolkcicgfmd/background.js:58321:24) 

[...]

Issue #2106, opened 2022-09-02
Can not set-up transactions on Polygon


Naxsun:
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/1015282104562626623

### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

Set-up transactions

### What did not work?

Transactions remain pending, don't show up on Polygonscan. The user disabled the extension and re-enabled it, re-started the browser in between, but the pending transactions remained.

_"ChainService.retrieveTransaction
Error retrieving transaction [REDACTED]
    at ChainService.retrieveTransaction (chrome-extension://eajafomhmkipbjmfmhebemolkcicgfmd/background.js:59334:24) "_

Logs: [logs_v0_15_0__2022-09-02T18_36_2403_00.txt](https://github.com/tallycash/extension/files/9479454/logs_v0_15_0__2022-09-02T18_36_2403_00.txt)

![image](https://user-images.githubusercontent.com/71384922/188193136-a40d4d6d-4570-4582-a9f6-0ebb83fd01d9.png)

### Version

v0.15.0

Naxsun:
Same issue:
https://discord.com/channels/808358975287722045/888500174685614090/1015375728067494028

Naxsun:
User LLLA: https://discord.com/channels/808358975287722045/888500174685614090/1016096095497097307

Naxsun:
User YoZero: https://discord.com/channels/808358975287722045/888500174685614090/1016248533650780170

Naxsun:
Successfully reproduced, logs:
[logs_v0_15_0__2022-09-04T21_55_51+02_00.txt](https://github.com/tallycash/extension/files/9487688/logs_v0_15_0__2022-09-04T21_55_51%2B02_00.txt)

VladUXUI:
Issue seems to be from Gas being set too low

Naxsun:
Was able to reproduce this. The pending transactions can be a result of a skipped nonce. A skipped nonce can be created by a dropped transaction.

To reproduce:
- Make an in-wallet swap on Polygon, set the custom gas low. E.g. Miner 1 and Max Base: 1
--> Dropped transaction
- Set-up another transaction, send or swap, but with ok / suggested gas settings
- The new transaction will be pending, because the dropped transaction did update the nonce

[...]

PR #2130, opened 2022-09-06
🛢 Transaction nonce and gas fixes

jagodarybacka:
This PR is touching multiple issues but they are tightly connected with each other

### What

**Nonce issue**
Ref #2127, #2106 - not closing for now
- Nonce should be released if the transaction failed to broadcast.
This will not fix the problem for users that already have pending or dropped transactions and nonce is in a corrupted state but it will prevent the same issue in the future.

**Gas settings issue**
Ref https://github.com/tallycash/extension/issues/2133 - 90% solved, I see the problem with custom settings - they are losing value, gas limit is ignored.
- We had a problem that the gas settings that the user was changing were not reflected in the transaction. That was because `isEIP1559EnrichedTransactionSignatureRequest(partialRequest)` was expecting `partialRequest` to have `          maxFeePerGas` and `maxPriorityFeePerGas` already set which was never true as this was the function that was supposed to set these values.

**Estimated gwei value issue**
Resolves https://github.com/tallycash/extension/issues/2112 
- Fixed the problem with incorrect gas price calculation - now Optimism and L1 gas calculation logic are not mixed up

### Testing
- setup transaction on any network (all networks should be tested)
- check Estimated network fee button value, compare with current gas prices on given network - they should be similar
- open gas settings panel, change selected option and save
- Estimated network fee should change according to selected gas setting
- set custom, very low gas fee in settings
- sign tx - it should fail to broadcast
- setup another tx, change gas settings and remember what you selected
- sign tx, check it on the scanner website - priority fee and other gas values should be set to the values you selected
- check tx nonce and previous tx nonce - they should be `N` and `N - 1`


0xDaedalus:
Can you add testing steps please

mhluongo:
# Potential testing steps

- [ ] Switch to an account on Polygon
- [ ] Attempt to swap on QuickSwap. Set the gas price to anything custom
- [ ] The tx should fail to broadcast. Note the nonce
- [ ] Attempt again to swap on QuickSwap without changing the gas price.
- [ ] It should go through, with the same nonce as the failed tx above

LMK if that's right @jagodarybacka

0xDaedalus:
Hmmmnn

> The tx should fail to broadcast. Note the nonce

This is not happening for me - quickswap works fine even with gasprice at 0.01 gwei.

0xDaedalus:
https://github.com/tallycash/extension/issues/2112 could be interfering here

jagodarybacka:
Yup, these are correct testing steps but testing is blocked because of - https://github.com/tallycash/extension/issues/2133 At least for me

mhluongo:
Can we force a tx to fail with a code change, or maybe some other way like trying to spend more of an ERC-20 than we have?

jagodarybacka:
I will just solve the gass issue first

0xDaedalus:
Gas estimates are not accurate for Optimism:
image

jagodarybacka:
@0xDaedalus hmmm, I can't reproduce such high Optimism gas price
Ethereum:
image
Optimism:
image
Polygon:
image
They all seems fine 🤔

PR #2138, opened 2022-09-06
Add new galaxy url to metamask impersonation list

0xDaedalus:
### To Test
- [x] Have Metamask Installed alongside Tally
- [x] Set Tally to Default
- [x] Visit https://galxe.com/
- [x] tally connect prompt should pop up instead of metamask connect prompt


mhluongo (approved):
:muscle:

PR #2139, opened 2022-09-06
Update the amount of gwei estimated for an optimism transaction rollup in real time.

### LINKED ISSUE #2000: Gwei gas prices are not updating live on Optimism
They are initially populated but do not update when price updates

0xDaedalus:
Closes #2000 

### To Test
- [x] set the `periodInMinutes` for the [`blockPrices` alarm](https://github.com/tallycash/extension/blob/main/background/services/chain/index.ts#L219) to `0.25`
- [x] Set up a swap transaction from uniswap 
- [x] Navigate to the final sign screen and wait
- [x] You should see the gwei cost updating every ~15s or so.


mhluongo (approved):
:muscle:

PR #2140, opened 2022-09-06
Enrichment Interaction: Annotate "to" address on transactions, not "from"

### LINKED ISSUE #2131: "Interacting With" address is wrong when confirming in-app swap
image

mhluongo:
Pretty sure wallet users know enough about their own wallet 😀

# To Test

- [ ] Try to swap from a wallet with an ENS or UNS name on mainnet
- [ ] In the confirmation page, the "Interacting with" field should be "Ox Router", rather than the wallet's ENS or UNS name

Closes #2131

### DIFF
--- a/background/services/enrichment/transactions.ts
+++ b/background/services/enrichment/transactions.ts
@@ -225,16 +225,11 @@ export default async function resolveTransactionAnnotation(
       // If the tx has no data, it's either a simple ETH send, or it's relying
       // on a contract that's `payable` to execute code
 
-      const [recipient, sender] = await Promise.all([
-        enrichAddressOnNetwork(chainService, nameService, {
-          address: transaction.to,
-          network,
-        }),
-        enrichAddressOnNetwork(chainService, nameService, {
-          address: transaction.from,
-          network,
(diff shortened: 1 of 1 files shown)


background/services/enrichment/transactions.ts
```diff
@@ -209,149 +205,164 @@ export default async function resolveTransactionAnnotation(
...
+      typeof transaction.input === "undefined"
     ) {
-      const [sender, recipient] = await Promise.all([
+      // If the tx has no data, it's either a simple ETH send, or it's relying
+      // on a contract that's `payable` to execute code
+
+      const [recipient, sender] = await Promise.all([
         enrichAddressOnNetwork(chainService, nameService, {
-          address: erc20Tx.args.from ?? transaction.from,
+          address: transaction.to,
```
hyphenized, line 230: Can we reuse the enriched address result from line 214 here? 👀

mhluongo: Looks like it!

0xDaedalus (approved):
👑

PR #2144, opened 2022-09-07
Consistent ENS names across networks.

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/1873

### To Test
- [x] Add a read-only address via ENS name with an ENS avatar (e.g. vitalik.eth)
- [x] The ENS name and avatar should consistently be shown in the top of the extension when switching between Ethereum / Polygon / Optimism

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -26,6 +26,7 @@ import {
   OPTIMISM,
   EVM_ROLLUP_CHAIN_IDS,
   GOERLI,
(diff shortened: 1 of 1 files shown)


mhluongo (approved):
Tested, simple, looks good for now. I do hate that the name "flashes" in the frontend when you first switch networks... that makes me think we're going about this the wrong way a bit... but this is an immediate improvement.

PR #2146, opened 2022-09-07
Correctly parse wrapped asset Withdrawals & Deposits.

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/2121
Closes https://github.com/tallycash/extension/issues/1871

### To Test
- [x] Swap ETH to WETH or ETH to WETH
- [x] WETH balance should update when ETH balance does.
- [x] Do the same with MATIC and WMATIC


mhluongo (approved):
:ok_hand:

PR #2148, opened 2022-09-07
Disable GOERLI tracking by default.

0xDaedalus:
Now that load testing is complete we no longer need to track GOERLI by default.  More context here: https://github.com/tallycash/extension/pull/2108

### To Test
- [x] Checkout `v0.15.1`
- [x] Add a wallet via seed phrase
- [x] dev tools => Application => indexedDB => accountsToTrack - you should see addresses being tracked with chain ID 5.
- [x] Upgrade to this branch
- [x] dev tools => Application => indexedDB => accountsToTrack - you should _not_ see addresses being tracked with chain ID 5.
- [x] Your address should still be being tracked on Ethereum (chain ID 1)
- [x] Switching to Polygon / Optimism should add your address to `accountsToTrack` on that network.


mhluongo:
I can't seem to get Goerli tracking after checking out `v0.15.1`, are you sure that's the right release?

0xDaedalus:
Yes - although after double checking this it looks like adding a read-only address does not track on Goerli.  You'll have to import a seed phrase I think

mhluongo:
Per our live discussion, for whatever reason the `v0.15.1` tag doesn't show the expected Goerli behavior... but the prod extension and `main` do. Worth following up to fix our release process / make tags more useful.

mhluongo (approved):
Tests are green, let's do it!

PR #2154, opened 2022-09-08
Display uniswap contract name when signing optimism transactions on uniswap

0xDaedalus:
image

### To Test
- [ ] Setup an optimism swap on uniswap.
- [ ] The `Send to` field should show `🦄 Uniswap`


mhluongo (approved):
:unicorn:

PR #2159, opened 2022-09-09
🕸 Optimism gas explainer design changes 

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/2119

### What

Redesigned Optimism gas slide up
image

### Testing

- [ ] setup tx on Optimism, open gas settings - it should show new design
- [ ] setup tx on Eth and Polygon, open gas settings - it should show regular gas settings panel


jagodarybacka:
@VladUXUI do we want to change something in this design?

ui/components/NetworkFees/NetworkSettingsSelectOptimism.tsx
```diff
@@ -15,56 +15,107 @@ export default function NetworkSettingsOptimism(): ReactElement {
...
+        <div className="fee_chart_sign">=</div>
+
+        <div className="fees_chart_item">
+          <div className="fees_icon icon_gas" />
+          <span>Estimated Gas</span>
+        </div>
+      </div>
+
+      <div className="simple_text">
+        The estimated gas cost for Optimism transactions includes Optimism fee +
```
0xDaedalus, line 41: includes `an` Optimism fee + `an` Ethereum roll-up fee (`the` fee....`

0xDaedalus (approved):
Just a small non-blocking copy edit suggestion.

Also want to point out that the size of the new images is different but both are labled as `@2x`.  I'm not sure if the `@2x` suffix is supposed to indicate images of a uniform size but wanted to mention it just in case cc @VladUXUI

jagodarybacka:
> different but both are labled as @2x

oh that's my bad, I was renaming them because they exported with wrong names, will fix it

0xDaedalus (approved):
image

PR #2165, opened 2022-09-09
Fix OP <> ETH swaps on Optimism

0xDaedalus:
### To Test
- [x] Attempt to swap OP for ETH in the wallet
- [x] You should not see "Price Impact Too High" - and you should be able to swap
- [x] Swaps should work both ways


jagodarybacka (approved):
Works 👑 

But I see some problems on this screen:
- gwei value is missing
- exchange route could have better styles
Do we have issues for them?
image

PR #2171, opened 2022-09-09
Tabs vs Popups: Prefer tabbed over popup-based onboarding

mhluongo:
Remove the secondary popup-based onboarding, pushing users through the tabbed flow instead.

# To Test

- [x] With `SUPPORT_TABBED_ONBOARD=false`, install the extension and onboard in the popup. Everything should work as expected.
- [x] With `SUPPORT_TABBED_ONBOARD=true`, reinstall the extension.
- [x] Close the onboarding tab.
- [x] Click the extension command button. Instead of a popup, you should see the onboarding tab open.

Refs #2025


jagodarybacka (commented):
We are getting 2 onboarding pages when user is using `alt + T`:

https://user-images.githubusercontent.com/20949277/189638618-406db901-9abc-4750-a2d0-09f16a681f90.mov

Shadowfiend:
Continuing from [this discussion](https://github.com/tallycash/extension/pull/2088#discussion_r963612805), part of the issue is that web pages can cancel an event bubble and block the browser shortcut (a feature I hate and have an extension to block lol), whereas I don't think we can do that with the extension shortcut. Definitely in unexplored territory and uncertain about this though.

mhluongo:
Ah, the is unrelated to a shortcut collision. The idea of this PR is to always show the onboarding tab *instead of* the extension popup if onboarding hasn't finished.

So the missing logic is "if there's already an onboarding tab, switch focus to that".

mhluongo:
Confirmed, the shortcut key doesn't debounce :cry:

mhluongo:
Mk, going to hack this first... if there are still issues then I'm going to move the tabbing code to before the React rendering (better spot for it anyway, but requires more state logic)

mhluongo:
Mk, I fixed this "the React way" but it's still not where I'd like it. Lmk what you think here @jagodarybacka and I'll experiment to see if I can avoid rendering at all

mhluongo:
Pushed an alternative approach I'm happier with. This is ready for another look @jagodarybacka @Shadowfiend

jagodarybacka (approved):
Looks complicated but works great ✨ 

I will give @Shadowfiend a chance to take a look before merging

ui/index.ts
```diff
@@ -19,10 +19,40 @@ export async function attachUiToRootElement(
...
+  const store = await newProxyStore()
+
+  if (SUPPORT_TABBED_ONBOARDING) {
+    const state = store.getState()
+    if (Object.keys(state.account?.accountsData?.evm).length === 0) {
+      // we're onboarding! look for an onboarding tab, or open a new one,
+      // rather than rendering the popup
+      const baseURL = browser.runtime.getURL("tab.html")
+      const tabs = (await browser.tabs.query({ url: baseURL })).filter(
+        (tab) => tab?.url && tab.url.includes("onboarding")
```
mhluongo, line 44: Notably, this won't work for the Ledger onboarding flow... in that case we'll get a new onboarding tab. I'll be unifying those in a future PR.

mhluongo:
Ping @0xDaedalus in case you want to swoop, @Shadowfiend is out for a bit

ui/index.ts
0xDaedalus: Should we make this more specific - tally-onboarding for example?  Or is the `baseURL` enough of a filter where we don't need to worry about conflicts.

0xDaedalus (approved):
✅

ui/index.ts
mhluongo: Base URL already confirms it's specific to the extension :)

PR #2173, opened 2022-09-09
Successfully Navigating Onboarding: Rewire onboarding for more consistent back behavior and page styling

mhluongo:
Use nested routes and a navigation component for more consistent back-button behavior in tabbed onboarding.

Out of scope for this PR
- [ ] [Forking the Ledger onboarding](https://github.com/tallycash/extension/issues/2172). The state is complex enough that it deserves its own PR.
- [ ] [Breaking the "DAO intro" into separate routable pages.](https://github.com/tallycash/extension/issues/2227)

# To Test
- [ ] With `SUPPORT_TABBED_ONBOARDING=false`, confirm onboarding works as expected.
- [ ] With `SUPPORT_TABBED_ONBOARDING=true`, test the new wallet, view-only, and import seed flows. Each page should have a consistent header & back button.

Refs #2025


ui/pages/Onboarding/Tabbed/ImportSeed.tsx
```diff
@@ -232,15 +224,14 @@ export default function OnboardingImportSeed(props: Props): ReactElement {
         .bottom {
           justify-content: space-between;
           flex-direction: column;
+          /* margin-top: ${HIDE_IMPORT_DERIVATION_PATH ? "35px" : "24px"}; */
```
jagodarybacka, line 227: Do we want to keep it as a comment or remove?

jagodarybacka (changes requested):
I love these short file names 💖 Also, much cleaner implementation than before IMO

---

- [ ] Back button doesn't make sense on the first page:

https://user-images.githubusercontent.com/20949277/189641278-384d861f-2967-4e51-9611-3426ffd28dff.mov

- [ ] Back button is not working here:
image

- [ ] Just want to confirm - back button on this screen should go back to the same seed phrase?

image

- [x] `Let's secure...` should be centered
image

- [x] There is no margin between text area and button
image

- [ ] Importing with seedprase - when password is set and we are on the import seed screen - back button is not working there
- [ ] By clicking back button we are able to add multiple accounts with seed phrase or read only without opening extension even once - is that what we want?

mhluongo:
Excellent QA, thanks @jagodarybacka. Working on the styling stuff in particular.

I could use advice from @VladUXUI on some of these navigation cases... eg I think the back button should allow us to reset a password... but not sure we have the background support needed for it yet. What should it do in the meantime?

> Just want to confirm - back button on this screen should go back to the same seed phrase?

Here, I believe the intent is to generate a new recovery phrase. If it's creating multiple accounts I need to figure that out.

> Back button doesn't make sense on the first page:

Agreed! Going to hit it in a future PR along with this :point_down: 

> Breaking the "DAO intro" into separate routable pages.

Shadowfiend:
As far as I've seen it would be a departure from typical react practices. I absolutely hate it, but consistency to the community's style is a reasonably important part of using a framework (and language). If we dig and find that isn't the case there should be nothing stopping us. If we find it is... I don't consider this my call at this point, so I'm completely neutral 😉

jagodarybacka:
>departure from typical react practices

@Shadowfiend can you give us some sources? 
Every React project I worked with before had short file names that were not including folder names etc. It seems to me that rules around naming things are more like a "what is popular in this moment" thing and shorter names are more readable IMO.

henryboldi:
The inspiration for the file naming convention here came from Dan Abramov, which I believe sparked the convention becoming more popular:

> Everything is in a single repository, and each JS filename is unique and can be imported absolutely from any other file in the source.
>
>This means filenames are relatively verbose, e.g. require('AdsManagerPrivacyDialogButton'), but the benefit is it lets people navigate projects they aren’t intimately familiar with, and that IDE “quick open” popup always lands me in the right place.
>
>You might think this is a terrible idea, but it actually works great for us. I joined Facebook a year ago, and it was weird because nobody does it in the open source, but I found it really easy to work with.
>
> The main reason it works so well is because Facebook invests tons into internal tooling, including cross-project search, IDE integrations, component explorer, test runners, code review tooling, etc. It probably would be a bad idea unless you can build tooling that really benefits from this choice.
>
>For example, cross-project grep works great because there are never any conflicting filenames. I can search for a require() of any component and find all matches with zero false positives. The IDE can insert imports automatically without guessing which folder they are coming from. The internal jsfiddle-like tool lets you use any components without writing imports. You never need to change paths when you move files.

Admittedly, it's a better fit for large projects with more tooling. And "quick open" has likely improved in the past 5 years since this comment was made.

jagodarybacka:
Thank you @henryboldi! 
I think that at this point it is just a manner of personal taste, I was in a project where only folder names mattered, every file was named like `index.js` `model.js` `view.js` etc. and there were no problems because of it.

mhluongo:
Let's double-down on non-repetitive naming and save ourselves the RSI :)

mhluongo:
Addressed all feedback and added issues for the back-button behavior that's a little too tricky for this humble PR

https://github.com/tallycash/extension/issues/2229
https://github.com/tallycash/extension/issues/2230

This is ready for another look @jagodarybacka @Shadowfiend @0xDaedalus

0xDaedalus (approved):
lezgo

PR #2178, opened 2022-09-12
⏱ Pending transaction fixes

jagodarybacka:
### Dropped tx in activities
Resolves https://github.com/tallycash/extension/issues/1817

Handle blockheight in activities screen the same way as the activity is labeled - `pending`, `dropped`, `failed`. We can add migration so it would fix activities for existing extensions but it is not needed IMO - but open to suggestions 😄 

### Pending Sign button
Resolves https://github.com/tallycash/extension/issues/2051

Fix the problem with pending/loading sign button and transactions that can't be signed.

#### Problem
After user clicks "Sign" button `handleConfirm()`function is called. Each `handleConfirm()` had its rules when the transaction was not yet ready to be signed. In that case, function had no effect but Sign button was already in a "loading" state and as users were waiting for it to finish the action window was often losing focus - and disabling the button.  

#### Solution
Let's add `canConfirm` prop and let's block the Sign button if the transaction is not ready to be signed - for example, it takes a while after gas settings are changed.

### Testing

1. Setup transaction with custom, too low fee settings
2. Sign and check activities - block height should be "dropped"
3. Setup transaction with correct fee settings
4. Click sign, transaction should be successful
5. Check other signing types - Tally pledge, login.xyz, mintkudos...

### DIFF
--- a/background/redux-slices/utils/activity-utils.ts
+++ b/background/redux-slices/utils/activity-utils.ts
@@ -3,6 +3,11 @@ import { ConfirmedEVMTransaction } from "../../networks"
 import { EnrichedEVMTransaction } from "../../services/enrichment"
 import { HexString } from "../../types"
 
+enum TxStatus {
+  FAIL = 0,
+  SUCCESS = 1,
+}
+
 type FieldAdapter = {
   readableName: string
   transformer: (tx: EnrichedEVMTransaction) => string
@@ -73,12 +78,19 @@ function blockHeightTransformer(
   blockHeight: number | null,
   status: number | undefined
 ): string {
-  if (blockHeight !== null && status !== undefined && status !== 1)
+  if (
+    blockHeight !== null &&
+    status !== undefined &&
+    status !== TxStatus.SUCCESS
+  ) {
     return "(failed)"
-
-  if (blockHeight !== null) return blockHeight.toString()
-
-  if (blockHeight === null && status === 0) return "(dropped)"
+  }
+  if (blockHeight !== null) {
+    return blockHeight.toString()
+  }
+  if (blockHeight === null && status === TxStatus.FAIL) {
+    return "(dropped)"
+  }
 
   return "(pending)"
 }
--- a/background/redux-slices/utils/activity-utils.ts
+++ b/background/redux-slices/utils/activity-utils.ts
@@ -69,8 +69,18 @@ function gweiTransformer(value: bigint | null | undefined): string {
   return `${weiToGwei(value) || "0"} Gwei`
 }
 
-function blockHeightTransformer(blockHeight: number | null): string {
-  return blockHeight === null ? "(pending)" : blockHeight.toString()
+function blockHeightTransformer(
+  blockHeight: number | null,
+  status: number | undefined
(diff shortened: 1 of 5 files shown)


background/redux-slices/utils/activity-utils.ts
```diff
@@ -69,8 +69,18 @@ function gweiTransformer(value: bigint | null | undefined): string {
   return `${weiToGwei(value) || "0"} Gwei`
 }
 
-function blockHeightTransformer(blockHeight: number | null): string {
-  return blockHeight === null ? "(pending)" : blockHeight.toString()
+function blockHeightTransformer(
+  blockHeight: number | null,
+  status: number | undefined
+): string {
+  if (blockHeight !== null && status !== undefined && status !== 1)
```
0xDaedalus, line 76: I believe that this function will always return `"(failed)"` since the return statement is on a different line than the `if` statement.  A fantastic reason to always use curly braces with `if` statements.  

Also - once that is fixed - I think it would be helpful to add a comment here about why we care about `1` and what it means. Alternatively if its fairly simple we can assign `1` a descriptive variable name and use it inside the `if`.

ui/pages/SignTransaction.tsx
```diff
@@ -62,16 +62,17 @@ export default function SignTransaction(): ReactElement {
 
   if (isLocked) return <></>
 
+  const canConfirm =
```
0xDaedalus, line 65: These canConfirm's really feel like they should be derived inside of `SignTransactionContainer` but I get that it might be quite difficult to do that with the way our current component heirarchy is setup.  Maybe something we can do after the signing flow refactor lands

0xDaedalus (changes requested):

background/redux-slices/utils/activity-utils.ts
jagodarybacka: > return statement is on a different line

I'm pretty sure it is [not the case ](https://stackoverflow.com/questions/2125066/is-it-a-bad-practice-to-use-an-if-statement-without-curly-braces)- if `if` has no curly braces then first line after if is considered its block, second line is not. 
Anyway, I can't do it in a single line because linter is complaining so I will add curly braces everywhere.

> why we care about 1

This is copied from the activities component but yup, will improve it.

ui/pages/SignTransaction.tsx
jagodarybacka: That was my initial idea but as each signing type has it's own confirmation rules it felt better to keep it near `handleConfirm()`, also I wanted to make as little changes here as possible to avoid adding more problems :joy:

background/redux-slices/utils/activity-utils.ts
```diff
@@ -3,6 +3,11 @@ import { ConfirmedEVMTransaction } from "../../networks"
 import { EnrichedEVMTransaction } from "../../services/enrichment"
 import { HexString } from "../../types"
 
+enum TxStatus {
```
0xDaedalus, line 6: ❤️

0xDaedalus (approved):
👑

PR #2179, opened 2022-09-12
Dam the Waterfall - Fix Blocknative Cascading Reconnect

### LINKED ISSUE #1445: Repeated Websocket connection attempts to blocknative, http requests to alchemy.
Version: 0.13.3

After opening up and using Tally Ho in a chrome profile that I hadn't used for a couple of weeks (which has the Tally Ho extension installed via chrome store)  - the application felt very sluggish and the following was present in the background page logs:

https://user-images.githubusercontent.com/94649004/168485903-0338c0bb-0548-453d-93ca-7122bdb79a24.mov

https://user-images.githubusercontent.com/94649004/168486090-bf547cdb-3d5b-4294-8665-d99404aebaee.mov

0xDaedalus:
Closes #1445

By calling Blocknative.connect inside of `getBlockPrices` - we were opening a websocket connection each time we polled for block prices - which then stayed open for an indefinite amount of time.  Once a critical mass of websocket connections were opened - they would begin to disconnect and then automatically reconnect - causing a loop that used up _significant_ resources.  Switching to a singleton mitigates this issue and still lets us allow on the sdk's reconnect logic in case of connectivity issues.

### To Test
- [ ] Drop a console.log [inside of the blockPrices event handler](https://github.com/tallycash/extension/blob/main/background/main.ts#L701:L701)
- [ ] Drop another log [inside of getBlockPrices](https://github.com/tallycash/extension/blob/blocknative-reconnect/background/lib/gas.ts#L98:L98)
- [ ] Set the [blockPrices handler](https://github.com/tallycash/extension/blob/blocknative-reconnect/background/services/chain/index.ts#L226:L226) period to `0.1`
- [ ] Load up the extension - you should see logs that indicate blockPrices are correctly being retrieved.
- [ ] Disconnect from wifi - you should see errors in the console that indicate blockPrices are not correctly being retrieved
- [ ] Turn wifi back on - you should once again be getting block prices.

### DIFF
--- a/background/lib/gas.ts
+++ b/background/lib/gas.ts
@@ -17,6 +17,8 @@ import { gweiToWei } from "./utils"
 // `process.env` variables in the bundled output
 const BLOCKNATIVE_API_KEY = process.env.BLOCKNATIVE_API_KEY // eslint-disable-line prefer-destructuring
 
+let blocknative: Blocknative
+
 type PolygonFeeDetails = {
   maxPriorityFee: number // gwei
   maxFee: number // gwei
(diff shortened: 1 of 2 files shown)


jagodarybacka (approved):
Looks good 👑

PR #2181, opened 2022-09-12
Greatly increase available swap assets on L2's

0xDaedalus:
We were deduplicating assets that shared a name but belonged to different networks - now we don't.

### To Test
- [ ] Open the swap screen on Optimism
  - [ ] There should be more than 10 buy assets to choose from
  - [ ] ETH should not be duplicated in the buy asset list
- [ ]  Open the swap screen on Polygon
  - [ ] There should be more assets available to swap than on `main`
  - [ ] MATIC should not be duplicated in the buy asset list


mhluongo:
Muh USDC!!!

mhluongo (approved):
Excellent :raised_hands:

PR #2185, opened 2022-09-13
Correctly display token decimals on approval activities

### LINKED ISSUE #1786: Correctly display token decimals on approval activities
### Discord Discussion Link

_No response_

Chrome

Approve exact KEEP tokens amount and swap

### What did not work?

I was swapping and approving the exact amount of KEEP tokens, and noticed that the amount is reflected in not convenient manner

Снимок экрана 2022-07-11 в 15 46 12

Also when I swap I would like the wallet automatically bring me to the "activity" page after I sign the tx, so that I could see what is happening and not to switch to 'activity' manually.

And I would like this suggestion to be implemented, it would be very convenient for executing swaps: https://github.com/tallycash/extension/issues/1407

### Version

Community Edition v0.13.13

### Relevant log output

_No response_

kkosiorowska:
Closes #1786 

This PR correctly displays token decimals for activity item.

**UI**

Before
178268160-ef6a2ad2-ff4e-402d-87dd-abb9d27eab01

After
Screenshot 2022-09-13 at 16 07 56

Screenshot 2022-09-14 at 13 03 36

**To Test**
- [ ] go to the activity on wallet and check how token amounts are displayed for the activity item
- [ ] check if tooltip with whole value is visible


mhluongo:
Confirming the thousands separator is still localized correctly? Only asking because I see " " in the before and "," in the after. Both are appropriate, they should just be consistent to a user's locale.

jagodarybacka (changes requested):
Would be nice to add title attribute to the value as we are cutting it and user can't see the whole value

kkosiorowska:
> Confirming the thousands separator is still localized correctly? Only asking because I see " " in the before and "," in the after. Both are appropriate, they should just be consistent to a user's locale.

@mhluongo 
Yes, the thousands separator is still localized correctly. It's just a difference of local user settings.

jagodarybacka (approved):
Looks good 👍

PR #2186, opened 2022-09-13
Remove CUSTOM_GAS_SELECT flag

jagodarybacka:
### What
Remove CUSTOM_GAS_SELECT feature flag logic as it is `true` for some time now


0xDaedalus (approved):
✅

jagodarybacka:
Disabled automerge because checks were stuck, now they are running correctly

PR #2188, opened 2022-09-14
Show manage dApps page

### LINKED ISSUE #2263: Some dapp connections push disconnect button offscreen
image

jagodarybacka:
Resolves #2263
### What

This PR:
- removes feature flag and makes manage dApp permissions subpage public
- fix revoking app permissions - the same set of networks should be used during granting and revoking permissions
- wait for the database to remove permissions before UI will display success snackbar

### Testing
- [x] connect to several dapps
- [x] go to settings -> connected website
- [x] disconnect some dapps, confirm they are really disconnected on all networks
- [x] reload extension, disconnected apps should be still disconnected

image


background/main.ts
```diff
@@ -1141,9 +1141,7 @@ export default class Main extends BaseService<never> {
       "denyOrRevokePermission",
       async (permission) => {
         await Promise.all(
-          // We use supportedNetworks here because we currently grant
```
jagodarybacka, line 1144: This comment was not really true because during granting permissions we were using hardcoded list - see line 1131

background/main.ts
```diff
@@ -1141,9 +1141,7 @@ export default class Main extends BaseService<never> {
...
-          // dapp permission for all supported networks when approving a dapp.
-          this.chainService.supportedNetworks.map(async (network) => {
+          [ETHEREUM, POLYGON, OPTIMISM, GOERLI].map(async (network) => {
```
greg-nagy, line 1144: non blocker, but this approach makes me a bit nervous. Really want to finish the per dapp connection feature

greg-nagy (approved):
code looks good, works well, want to ship it, 
but not sure if we want to ship now so leaving an approve for now 

edit: waiting for product confirmation, then will merge

background/main.ts
jagodarybacka: yup, this is quick fix for this bug (revoking permission was not removing goerli in previous implementation)

0xDaedalus: I think what we want here is to keep `supportedNetworks` and add a db migration to remove GOERLI permissions through a migration - similar to https://github.com/tallycash/extension/pull/2148

0xDaedalus: Ah I see we're also not using `supportedNetworks` when granting permissions - i guess we can push this off to another issue

greg-nagy:
product decision: #2214 is a blocker before we can release this

mhluongo (approved):
:rocket:

PR #2189, opened 2022-09-14
❌🗿 - Replace `newHeads` subscription with polling for latest blocks.

0xDaedalus:
This change gives a number of improvements - including:
- Progress towards Dealchemification
- Significant reduction in background browser CPU usage
- Significant reduction in Alchemy CUP credit usage

Prior to this change - we were writing to disk and emitting an event that modified redux state at least once every ~2 seconds (Polygon block time).  This commit standardizes the rate at which we fetch latest blocks to once every 15 seconds, and moves us from a subscription-based method of fetching new blocks to a poll-based method.

Anecdotally it seems that this change reduces background CPU usage by ~90% - which makes sense since we drastically reduce disk I/O as a result of this change.

### To Test
- [x] Load up the extension
- [x] You should see new blocks populating in Dev Tools => Application => IndexedDB => chain servicedb => blocks approximately every 15 seconds (you will have to refresh the db to see changes)
- [x] Initiate a uniswap swap
- [x] When on the sign page - leave it open - after ~15 seconds you should see the gwei estimate update.  (This may take longer if there is low volatility in gas prices)


background/services/chain/utils/index.ts
```diff
@@ -69,7 +69,8 @@ export function blockFromWebsocketBlock(
     hash: gethResult.hash,
     blockHeight: BigNumber.from(gethResult.number).toNumber(),
     parentHash: gethResult.parentHash,
-    difficulty: BigInt(gethResult.difficulty),
+    // PoS networks will not have block difficulty.
+    difficulty: gethResult.difficulty ? BigInt(gethResult.difficulty) : 0n,
```
mhluongo, line 73: Exciting times

background/services/chain/index.ts
```diff
@@ -1027,6 +1035,34 @@ export default class ChainService extends BaseService<Events> {
     )
   }
 
+  /*
+   * Fetch, persist, and emit the latest block on a given network.
+   */
+  private async pollLatestBlock(
```
mhluongo, line 1041: This implementation can miss blocks, right?

mhluongo: It's not clear to me whether or not the other could (probably), but it's a lot more obvious this one can... eg in Polygon I'd expect around 7 blocks every alarm, and this will only catch one of them.

mhluongo (changes requested):
I see some immediate issues with missed blocks, but haven't fully grokked the implications. I think our block recording behavior deserves a written rationale tbh

background/services/chain/index.ts
0xDaedalus: Can you expand on why we care about missed blocks?

As far as I can tell the only thing we do with the blocks we get in real time outside of persisting them is update the `blockHeight` and `baseFeePerGas` in redux - which seems like quite an insignificant benefit considering how performance intensive fetching and persisting blocks is.

0xDaedalus: I guess another thought here is that we are already necessarily missing hundreds if not thousands of blocks per day - whenever a user does not have their browser running

mhluongo: Had a sync conversation — some quick notes.

* The original idea with `newHeads` was to use our block tracking to do event log inspection to discover asset transfers, as well as provider-agnostic gas price tracking.
* We aren't using it for that today, and it's expensive.
* I'm good to merge this with some extra testing, and @0xDaedalus will write up next steps with the competing goals of performance and going provider-agnostic.

mhluongo: cc @Shadowfiend to get read in when you're back online

0xDaedalus:
Some next steps:
- In order to stop relying on `alchemy_getAssetTransfers`, we will eventually need to actually poll-for and index each block.  This will likely result in too much disk I/O as we support more chains with shorter block times (BSC comes to mind) so we will probably need to rethink some of the architecture around this.  Batching writes to disk could go a long way here.
- We can potentially use the `emit` of blocks to get away from our `blockPrices` alarm and dependency on blockNative
- Right now we rely on blocks we have in the db for transaction enrichment - with a fallback to querying a network in case a block does not exist in the db.  This should be unaffected by this PR due to the fallback.

greg-nagy:
A bit of context for how severe the problem was before this PR. 

After upgrading from 0.15.1 > 0.16.0 Nax's cpu was almost constantly running around ~350.
After this PR the problem is probably gone. 

(It's still possible that the source of the problem was historical data but given that the install was 2 weeks old I think we are safe to assume that this was the root cause) 

v0 16 0 - cpu perf issue

Shadowfiend:
> In order to stop relying on alchemy_getAssetTransfers, we will eventually need to actually poll-for and index each block. This will likely result in too much disk I/O as we support more chains with shorter block times (BSC comes to mind) so we will probably need to rethink some of the architecture around this. Batching writes to disk could go a long way here.

This isn't strictly true, btw. We don't have to index each block, we have to look at it. If it doesn't have data relevant to the current accounts, we could ignore it, with the caveat that when adding future accounts we might have to recheck elided blocks.

PR #2195, opened 2022-09-15
Fix password preview

### LINKED ISSUE #2187: Allow the user to preview the password
**Problem**

A user should be able to view the password. Clicking the eye icon does not show the password on password screen from onboarding and unlock Wallet screen.

**Solution**

Clicking the eye icon should allow you to preview your password.

Password screen from onboarding

Screenshot 2022-09-12 at 13 12 12

Unlock Wallet

Screenshot 2022-09-12 at 13 08 30

hyphenized:
Makes the preview password button work

_eyecanseenow

Closes #2187

## To Test:
- [ ] Unlock your wallet, try clicking on the 👁️ . You should be able to see your password (and everyone else too!)
- [ ] Type a very long password, text/mask should not appear behind the toggler icon
- [ ] Check other inputs aren't horribly broken

### DIFF
--- a/ui/components/Shared/SharedInput.tsx
+++ b/ui/components/Shared/SharedInput.tsx
@@ -1,5 +1,12 @@
-import React, { ChangeEvent, ReactElement, useEffect, useRef } from "react"
+import React, {
+  ChangeEvent,
+  ReactElement,
+  useEffect,
+  useReducer,
+  useRef,
+} from "react"
 import classNames from "classnames"
+import { useTranslation } from "react-i18next"
 import { useParsedValidation, useRunOnFirstRender } from "../../hooks"
 import { PropsWithIcon } from "./types"
 
@@ -50,6 +57,7 @@ export function SharedTypedInput<T = string>(
     iconSmall,
   } = props
   const inputRef = useRef<HTMLInputElement | null>(null)
+  const { t } = useTranslation("translation", { keyPrefix: "shared" })
 
   useEffect(() => {
     if (autoFocus) inputRef.current?.focus()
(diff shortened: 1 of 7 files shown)


ui/components/Shared/SharedInput.tsx
```diff
@@ -68,39 +76,67 @@ export function SharedTypedInput<T = string>(
...
+          className={classNames({
+            error: !isEmpty && (errorMessage ?? parserError !== undefined),
+            small: isSmall,
+            password: type === "password",
+          })}
+          step={step}
+          ref={inputRef}
+          maxLength={maxLength}
+        />
+        {(iconMedium || iconSmall) &&
```
0xDaedalus, line 117: I wonder if we want to condense this into a `hasIcon` variable in case we add an `iconLarge` in the future.

0xDaedalus (commented):
Absolutely love this - two small nits but also a general question.

I wonder if this PR gives the `SharedInput` component functionality that is a bit too specific for a `Shared` component.  Maybe it could make sense to go a bit more general here and do a bit of renaming to reflect it - e.g. `showPassword` => `showInput`, `hidePasswordHint` => `hideInput`, etc...  So instead of password-specific functionality inside of a SharedInput component - we have Input functionality that can be extended to password fields.

cc @greg-nagy @hyphenized @jagodarybacka curious for your thoughts here

hyphenized:
> Absolutely love this - two small nits but also a general question.
> 
> I wonder if this PR gives the `SharedInput` component functionality that is a bit too specific for a `Shared` component. Maybe it could make sense to go a bit more general here and do a bit of renaming to reflect it - e.g. `showPassword` => `showInput`, `hidePasswordHint` => `hideInput`, etc... So instead of password-specific functionality inside of a SharedInput component - we have Input functionality that can be extended to password fields.
> ...

Thanks @0xDaedalus I agree, It feels a bit dirty. I thought it was justified because there are inputs of type `password`  and this generic input was supposed to handle that case. But, it might be better to move this password specific logic one level above into a `PasswordInput` component that just reuses the `SharedInput` altogether... 🤔  

I'll push this later approach in a separate commit, I think the result should be much cleaner. Nevertheless, I'm up for ideas/suggestions 😃

hyphenized:
@mhluongo any chance you can re-open this to edit the base branch to main? 👀

0xDaedalus:
Hmmmmn Github is not being nice here - looks like I can't reopen it since the branch has been deleted and I can't edit the target branch.  Hyphenized are you able to change the base branch to `main` on your end?

hyphenized:
Not at all... 😕  guess I'll open another PR, thanks for trying !

PR #2207, opened 2022-09-15
Fix fallback logic for unsupported `alchemy_filteredNewFullPendingTransactions`

0xDaedalus:
### To Test
- [x] Do an in-wallet swap on optimism
- [x] The pending transaction should show up in the activity tab


mhluongo (approved):
:muscle:

Issue #2204, opened 2022-09-15
Optimism - can't deposit on beefy


jagodarybacka:
https://app.beefy.finance/
- connect wallet
- try to deposit
- allow asset spend window is showing
- gas price is not showing in $
- tx is failing with `Error: invalid transaction: intrinsic gas too low`

https://user-images.githubusercontent.com/20949277/190394093-8decd94b-53c4-4ec2-8de8-414ccb93fa3e.mov

PR #2208, opened 2022-09-15
Populate `from` on LegacyTransactions when estimating gas limit

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/2204
Closes https://github.com/tallycash/extension/issues/2201

This one was tricky - we need a `from` field when estimating gas limits for a given legacy tx - but `from` is also disallowed when serializing legacy transactions.  Luckily we correctly separate our concerns with `ethersTransactionRequestFromLegacyTransactionRequest` and `ethersTransactionFromSignedTransaction` so the fix was simple to implement - although difficult to pinpoint.

### To Test
- [ ] Follow the steps to recreate #2204 - you should be able to approve and deposit.

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -483,6 +483,7 @@ export default class ChainService extends BaseService<Events> {
         transactionRequest
       )
     } catch (error) {
+      logger.error("Error estimating gas limit: ", error)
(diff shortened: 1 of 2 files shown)


greg-nagy (approved):
let's merge! this is the theme of the day anyways 

CleanShot 2022-09-15 at 19 23 14

PR #2215, opened 2022-09-15
Handle `null` transaction.value from dapp transaction requests

0xDaedalus:
Closes https://github.com/tallycash/extension/issues/2212

image

### To Test
- [ ] Attempt to swap  on Velodrome


greg-nagy (approved):
let's 🐿

PR #2219, opened 2022-09-15
Fix "0x Router" not displaying in Activity List for Tally Swaps

0xDaedalus:
### To Test
- [ ] Make an in-wallet swap
- [ ] Activity list should display `To: 0x Router` for the corresponding transaction

image

### DIFF
--- a/background/utils/view-model-transformer.ts
+++ b/background/utils/view-model-transformer.ts
@@ -24,6 +24,7 @@ export const transactionPropertiesForUI = [
   "annotation.blockTimestamp",
   "annotation.type",
   "annotation.source",
+  "annotation.contractInfo.annotation.nameOnNetwork.name",
   "annotation.recipient.address",
   "annotation.recipient.network",
   "annotation.recipient.annotation",


greg-nagy (approved):
@0xDaedalus this fixes the issue for upcoming transactions but without a migration won't work for activities that are already in the store 

I am approving this to avoid back-and-forth because I think it's good enough as is, but want to let you make the call whether you want to fix it for previous txs or not

0xDaedalus:
I don't think we want a migration - this has been the status quo for a few releases now and should be seen strictly as an improvement.

Furthermore, our ui code has guarded against `contractInfo` not existing the whole time this bug was present - so we know that the extension works even if it doesn't.

One takeaway from this bugfix is that our types on the UI don't actually match what is in redux (We really have a `Partial<EnrichedEVMTransaction> rather than an `EVMTransaction` as part of `ActivityItem`) so I think aligning those types is a next step here.

greg-nagy:
> One takeaway from this bugfix is that our types on the UI don't actually match what is in redux (We really have a Partial<EnrichedEVMTransaction> rather than an EVMTransactionas part ofActivityItem`) so I think aligning those types is a next step here.

We dispatch the `activityEncountered` from both the `ChainService` and `EnrichmentService`. Because of this the type was not correct — more precisely the activities always had periods when they weren't the correct type.  I would introduce a new type for what's in the UI.

PR #2220, opened 2022-09-15
Improve i18n typing

hyphenized:
This PR just adds type autocompletion for language keys and fixes a few cases where the language server wasn't able to infer correctly which key was being used.

fuzzysearchisnice

It also fixes the type for `dispatch` which is asynchronous due to `webext-redux`


ui/components/Shared/types.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type AsyncifyFn<K> = K extends (...args: any[]) => any
+  ? (...args: Parameters<K>) => Promise<ReturnType<K>>
+  : never
```
Shadowfiend, line 4: Do we use this utility anywhere else? If not I would leave it alongside the one place it's used.

ui/pages/Settings/SettingsAnalytics.tsx
```diff
@@ -62,24 +59,22 @@ export default function SettingsAnalytics(): ReactElement {
...
-              {t(`${prefix}.recordItem${value}`)}
+              {t(`recordItem${value}`)}
             </li>
           ))}
         </ul>
-        <h2 className="title_error">{t(`${prefix}.noRecordTitle`)}</h2>
+        <h2 className="title_error">{t(`noRecordTitle`)}</h2>
         <ul className="list">
-          {["Seed", "Transactions"].map((value) => (
+          {(["Seed", "Transactions"] as const).map((value) => (
```
Shadowfiend, line 75: Both this and the previous arrays look like they could probably be pulled to named constants for clarity (though I realize this preceded these changes).

ui/pages/Settings/SettingsAnalytics.tsx
```diff
@@ -104,20 +99,26 @@ export default function SettingsAnalytics(): ReactElement {
...
       </section>
       <AnalyticsSlideUpMenu
         isOpen={showAnalyticsMenu}
         onCancel={() => setShowAnalyticsMenu(false)}
         onSubmit={() => handleCollectAnalyticsSubmit()}
-        prefix={`${prefix}.analyticsOffSlideUpMenu`}
+        title={t("analyticsOffSlideUpMenu.title")}
+        description={t("analyticsOffSlideUpMenu.desc")}
+        submitLabel={t("analyticsOffSlideUpMenu.submitBtn")}
+        snackBarMsg={t("analyticsOffSlideUpMenu.snackbar")}
```
Shadowfiend, line 112: Why this change? The previous approach seemed better parametrized, and the commit message doesn't seem to explain the decision.

Shadowfiend (commented):
Love the improved typing 💪 Left a few non-blocking comments and questions.

ui/pages/Settings/SettingsAnalytics.tsx
hyphenized: The previous approach relied on typing the prefix prop inside `AnalyticsSlideUpMenu` to be aware of which translations it was going to take, which made me feel it was better to refactor this component to be more functional and just take the text it was going to use as props. But, now that you mention it there might be nothing wrong with that 🤔

ui/pages/Settings/SettingsAnalytics.tsx
hyphenized: Refactored!

jagodarybacka (approved):
LGTM, feels great to have some help with these strings 😄

PR #2221, opened 2022-09-16
Remove account activities on account removal

### LINKED ISSUE #2194: Remove Account does not remove activities from storage
### Discord Discussion Link

_No response_

Chrome

Simple reproduction:
* Add read-only account
* View the Activities section for the account and make sure they load
* Remove the read-only account

### What did not work?

The activities remain as orphaned data in redux

### Version

v0.16.0

### Relevant log output

_No response_

hyphenized:
Removes account activities on account removal and adds a migration to check and remove orphaned accounts activities

Closes #2194

## To test
- [ ] Add an account, wait until activities load, then remove it
- [ ] Check the redux store, the removed account should no longer be in the activities state


0xDaedalus (commented):
Awesome PR - I've tested the redux functionality and everything works as expected ✅ 

We do, however, need to go one step deeper in order to merge this PR.  In this current state - removing an address results in the activity of that address not being available in the UI should the user change their mind and re-add the address at a later time.

This behavior is unfortunate - and I think stems from the fact that we currently treat redux state as a quasi-persistant db.  We definitely want to move away from this going forward - but until we do it unfortunately means that in some cases (like this one) we need to modify db state as well as modifying redux state to achieve the change we want.

At a quick glance I think we'll need to clear the corresponding account's saved transactions as well as saved asset transfer lookups which are checked against in [chainService's loadAssetTransfer function](https://github.com/tallycash/extension/blob/main/background/services/chain/index.ts#L1151:L1183) to fully remove orphaned activities.

Very happy to work with you on this guy through PR comments - I know that this unfortunately increases the scope of this bugfix quite a bit but I think it's still doable in one PR.

One problem I think we'll run into is that its currently possible for a transaction to be associated with more than 1 activity - e.g. if `0xa` sends  ETH to `0xb` and both are loaded into the wallet - we'll have both a `send` and `receive` activity in the store but only 1 transaction in the db (I think).

cc @mhluongo @Shadowfiend curious for your thoughts - maybe there's a simple way here that I'm not seeing?

hyphenized:
Thanks for the guidance 0xDaedalus (and pointing out where to look!)😄. I'll set up something to clear orphaned data from the 	db first and see if I can find anything else that might cause issues along the way

mhluongo:
I'm not sure we want to remove stuff from IndexedDB here — if we do, and the user re-adds an account, we have to re-fetch everything... maybe I'm missing your rationale here though @0xDaedalus.

I've been treating IndexedDB — at least for `ChainService` and `IndexingService` — as a primarily append-only database. That won't work forever from a performance perspective, but not sure we should make the decision to shift from that lightly, and I think we have a while now that we've dialed down block polling.

jagodarybacka:
Hey, I'm working on totally new activities slice here so I think this will be no longer needed when new slice will land https://github.com/tallycash/extension/pull/2311

hyphenized:
> Hey, I'm working on totally new activities slice here so I think this will be no longer needed when new slice will land #2311

Awesome, thanks for letting me know 😄 !

PR #2222, opened 2022-09-16
Block reject button when transaction is not ready

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/2164
Resolves https://github.com/tallycash/extension/issues/2202
### What
Same as `Sign` button `Reject` button should be blocked when the transaction is not ready to be signed.

This is because sometimes there are pending dispatched actions when the user is rejecting transaction and there is no easy way to cancel them. When they are finally dispatched then the transaction should be already rejected but it is being repopulated with data. Because of this signing screen becomes visible again. 

### Testing
- [ ] setup transaction (send, swap, siwe, with dapp etc)
- [ ] try rejecting transaction
- [ ] try signing transaction
- [ ] try above after gas fee change

Expected: Signing screen shouldn't be visible again after user rejects, after sign there should be no endlessly spinning Sign button

https://user-images.githubusercontent.com/20949277/190655998-46378df8-9cf7-41d1-b1af-d97dec78dbcc.mov

### DIFF
--- a/ui/components/Shared/SharedButton.tsx
+++ b/ui/components/Shared/SharedButton.tsx
@@ -205,6 +205,7 @@ export default function SharedButton(
             line-height: 24px;
             text-align: center;
             padding: 0 17px;
+            transition: background-color 0.2s, color 0.2s;
           }
           .button:hover {
             background-color: var(--gold-80);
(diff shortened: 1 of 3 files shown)


greg-nagy (approved):
(^_^) 

there is still some flickering when the UI shows up and it's strange that the reject button can  become active before the sign button, but these are so small micro improvements that I don't think we should worry about them

PR #2228, opened 2022-09-16
Fix OP Transactions not working when using Ledger

0xDaedalus:
I think that our friction around this functionality could be an indicator that we are making too broad use of ethersTransactionFromTransactionRequest and it might need to be broken down into two similar but distinct functions.  However, a change of that size is out of the scope of this bugfix.

### To Test
- [x] Attempt to send assets / swap / interact with a dapp on Optimism using a ledger.
- [x] Attempt to send assets / swap / interact with a dapp on Polygon / Ethereum using a ledger.

### DIFF
--- a/background/services/ledger/index.ts
+++ b/background/services/ledger/index.ts
@@ -13,6 +13,7 @@ import {
   getAddress as ethersGetAddress,
 } from "ethers/lib/utils"
 import {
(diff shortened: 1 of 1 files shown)


mhluongo (approved):
LGTM :+1:

PR #2234, opened 2022-09-18
Reduce ram/cpu usage for eslint

hyphenized:
After testing the extension on Firefox, I found my eslint server crashing constantly due to being out of memory. A couple of tests afterwards revealed the problem to be in the exclude directives used by `.tsconfig-eslint.json`. 

As shown in the screenshot below, both my RAM usage and CPU time went down by a lot (~70%/~50%) when linting after adding output directories to the excluded paths in `.tsconfig-eslint.json`.

#justjavascriptthings

Left side has same settings as PR
Middle does not exclude `validate` and `local-chain` folders
Right side only excludes `dist/firefox`


mhluongo (approved):
🙇‍♂️

PR #2238, opened 2022-09-19
Chain Service DB Tests

0xDaedalus:
This PR is intended to be a partial implementation as well as description of tests for the chain service db.

Some things to note:
- The implemented tests are not _necessarily_ indicative of the types of tests we want core contributors to be writing - but rather are meant to serve as an example for contributors interested in writing tests.
- The `.todo` tests are meant as guidance specifically for outside contributors.
- [This commit](https://github.com/tallycash/extension/commit/1d097b554c5c62d8286ba935309a45b96b1a2406) improves the typing of a function  currently that may potentially return undefined but is not described as doing so and has it return `null` in the case where no blocks are found.


jagodarybacka (approved):
Looks good, I'm a little bit afraid that these test files will be growing fast and it will become hard to read them at some point but let's not worry about it now 😄 

I'll let @greg-nagy take a look too

background/services/chain/tests/db.integration.test.ts
```diff
@@ -36,7 +35,151 @@ describe("Chain Database ", () => {
...
+      expect((await db.table("balances").toArray()).length).toEqual(0)
+      const accountBalance = createAccountBalance()
+      await db.addBalance(accountBalance)
+      const accountBalances = await db.table("balances").toArray()
+      expect(accountBalances.length).toEqual(1)
+      expect(accountBalances[0].address).toEqual(accountBalance.address)
+    })
+  })
+  describe("addBlock", () => {
+    it.todo("should correctly persist blocks to indexedDB") // Implementation should be similar to addBalance
```
greg-nagy, line 52: ❤️ `.todo`

background/services/chain/tests/db.integration.test.ts
```diff
@@ -36,7 +35,151 @@ describe("Chain Database ", () => {
...
+    it("should return chainIds corresponding to the networks of accounts being tracked", async () => {
+      await db.addAccountToTrack(account1)
+      expect(await db.getChainIDsToTrack()).toEqual(new Set(ETHEREUM.chainID))
+      await db.addAccountToTrack(account2)
+      expect(await db.getChainIDsToTrack()).toEqual(
+        new Set([ETHEREUM.chainID, POLYGON.chainID])
+      )
+    })
+    it("should disallow duplicate chain ids", async () => {
+      expect((await db.getChainIDsToTrack()) instanceof Set).toEqual(true)
```
greg-nagy, line 87: Creative cheating ✨

greg-nagy (approved):
The code and tests are good and straightforward. 

I am thinking about the added value of testing mostly crud operations on a mocked system. 
I think this is a good next step, but not sure if this is the type of tests are the most useful for us on the long run.

PR #2247, opened 2022-09-19
Remove unused HiddenDevPanel component

0xDaedalus:
This component is defined but not used anywhere in the application.


mhluongo (approved):
:hocho: my fav kind of PR

hyphenized:
Maybe we could have a shadowy super doggos panel 🤔

greg-nagy:
I wanted to use that for my not-so-secret runtime enable feature flag idea :( 

Also curious about what could live in the `shadowy super doggos panel` because I just fell in love with this name :)

hyphenized:
Ehhh it's a vague idea, but for example, recently I kept adding the chain service as a global to the window object so I could force the wallet to sync transactions. Though adding to the hidden dev panel some kind of feature to do these common operations did cross my mind, I think in hindsight it would be faster to run these commands in the background page using the console 😄

greg-nagy:
I fell in love with this concept :)

PR #2252, opened 2022-09-19
Akashi-Kaikyo (Pearl) Bridge: Don't warn when sending ETH or ERC-20s to known contracts

### LINKED ISSUE #2250: Warning when depositing funds into OP bridge
Currently we display the following warning when users attempt to use the official OP bridge https://app.optimism.io/bridge (because the bridge is custodial).

We need to fix this issue this week. Here are two options to do that:

1. Name the OP teleportr contract: "🔴 Optimism Teleportr" so this warning doesn't display
2. Add this contract to an allowlist  so this warning doesn't display

Option #1 is preferable (because later on we can add Kraken, Coinbase, and other ~trusted CEX deposit addresses), but if we need to (aka we can't have this complete this week otherwise), option #2 is okay for now.

Screen Shot 2022-09-19 at 4 34 21 PM

Related: https://github.com/tallycash/extension/issues/2117

mhluongo:
> The Akashi Kaikyo Bridge ([Japanese](https://en.wikipedia.org/wiki/Japanese_language): 明石海峡大橋, [Hepburn](https://en.wikipedia.org/wiki/Hepburn_romanization): Akashi Kaikyō Ōhashi) is a [suspension bridge](https://en.wikipedia.org/wiki/Suspension_bridge) which links the city of [Kobe](https://en.wikipedia.org/wiki/Kobe) on the Japanese island of [Honshu](https://en.wikipedia.org/wiki/Honshu) to Iwaya on [Awaji Island](https://en.wikipedia.org/wiki/Awaji_Island). It is part of the [Kobe-Awaji-Naruto Expressway](https://en.wikipedia.org/wiki/Kobe-Awaji-Naruto_Expressway), and crosses the busy and turbulent [Akashi Strait](https://en.wikipedia.org/wiki/Akashi_Strait) (Akashi Kaikyō in Japanese). It was completed in 1998,[[1]](https://en.wikipedia.org/wiki/Akashi_Kaikyo_Bridge#cite_note-structurae-1) and has the [second longest central span of any suspension bridge](https://en.wikipedia.org/wiki/List_of_longest_suspension_bridge_spans) in the world,[[3]](https://en.wikipedia.org/wiki/Akashi_Kaikyo_Bridge#cite_note-hsbe-3) at 1,991 metres (6,532 ft), only behind the [1915 Çanakkale Bridge](https://en.wikipedia.org/wiki/1915_%C3%87anakkale_Bridge) that was opened in March 2022.
— [Wikipedia](https://en.wikipedia.org/wiki/Akashi_Kaikyo_Bridge)

Add the Optimism bridge contracts to known contracts, and avoid misleading warnings.

In the future, this should be revised by...
- [ ] Having a specific warning for custodial bridges like Optimism Teleport
- [ ] Distinguishing between contracts that are "safe" to send funds without data (multisigs, exchanges) and those that aren't (Uniswap).

# To Test

- [ ] Attempt to [bridge ETH to Optimism](https://app.optimism.io/bridge/deposit) using the "Warp Speed" setting. The contract name should be "Optimism Teleportr", and there should be no warning about sending assets to the contract.
- [ ] Attempt to [bridge ETH to Optimism](https://app.optimism.io/bridge/deposit) using the "Standard Speed" setting. The contract name should be "Optimism Gateway" , and there should be no warning about sending assets to the contract.
- [ ] Attempt to send funds to an ENS address. The ENS address should show up correctly in the signature request and the wallet activity list.

Closes #2250


background/services/enrichment/transactions.ts
```diff
@@ -241,7 +241,16 @@ export default async function resolveTransactionAnnotation(
       if (typeof transaction.value !== "undefined") {
         // Warn if we're sending ETH to a contract. This is normal if you're
         // funding a multisig or exchange, but it's good to double check
-        if (recipient.annotation.hasCode) {
+        // If the annotation is a built-in contract or in the address book,
+        // skip the warning.
+        if (
+          recipient.annotation.hasCode &&
```
0xDaedalus, line 247: This seems like it could be a good candidate for a test - if you don't want to stub this monster function - the logic inside of the `if` statement could be factored out into its own function and that function could be unit tested.

mhluongo: Thanks, I hate it. Coming right up

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -126,7 +126,7 @@ export default function SignTransactionSpendAssetInfoProvider({
             Approve{" "}
             <SharedAddress
               address={spender.address}
-              name={spender.annotation.nameOnNetwork?.name}
+              name={spender.annotation.nameRecord?.resolved.nameOnNetwork.name}
```
0xDaedalus, line 129: I know that this conforms to our existing type - but I think the nesting here (and in the annotation object in general) could be confusing to new contributors.

Maybe this is evidence of us starting to treat `annotation` too much as a catch-all and losing some idiomaticity in the process?  I'd love for this line to be closer to `spender.annotation.name` and I'm sure there are other annotations where we can reduce some of the nesting complexity as well.

Anyway - something to think about during our next refactor.

0xDaedalus (approved):
👑 🚢

mhluongo:
Aw I didn't even get my final test in! I'll take it though :grin:

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
mhluongo: Tbh the type is a little verbose already in the service, so it'll get cut down.

`spender.annotation.name` isn't enough IMO — we should be showing where names where resolved from in the UI at some point, otherwise there are all sorts of odd phishing schemes that can arise down the road. But something like

`spender.annotation.nameRecord` that gives us...

```
{
    name: "mhluongo.eth",
    network: ETHEREUM,
    system: "ens",
    expiredAt: 123,
    from: {
       address: ...,
       network: ...,
   }
}
```

Without the `...resolved.nameOnNetwork` would be :ok_hand:

background/services/name/index.ts
```diff
@@ -25,17 +25,15 @@ import { RESOLVE_RNS_NAMES } from "../../features"
 
 export { NameResolverSystem }
 
-type ResolvedAddressRecord = {
-  from: {
-    name: DomainName
-  }
+export type ResolvedAddressRecord = {
+  from: NameOnNetwork
```
Shadowfiend, line 29: This is a complete departure from how `ResolvedNameRecord` works below lol.

PR #2253, opened 2022-09-19
Reflective Ledgend: Add Ledger signing flow to new signing UI

Shadowfiend:
The Ledger signing UI is a bit closer to the original designs,
displaying an image corresponding to the current ledger connection
state throughout the signing flow. Once `Sign` is clicked, the
signing UI also displays something that is as close as possible to
what the Ledger will display for easier verification, instead of
another Tally-specific summary of the signing details.

## Testing

- [x] All of this should be behind a feature flag. Ensure signing is unchanged with the feature flag disabled. There is a slight touch to `SignTransaction` to simplify `canConfirm` types due to some of the tweaks here, testing would be useful though the logic outcome should be roughly identical.
- [x] Enabling the feature flag, ensure that Ledger transactions can be signed---note that there won't be any details about the transaction being signed, so please only use a trusted source (for example, an internal swap transaction) for this flow, even if just using a test wallet.
  - [x] Ensure that, after clicking `Sign`, Tally displays the exact same fields and values as the Ledger.
- [x] Enabling the feature flag, ensure that data (e.g. SIWE) and typed data (e.g. EIP-2612 permits) can be signed with a Ledger---as above, no details will be presented, so prefer to use a trusted source.
  - [x] Ensure that, after clicking `Sign Data` in both cases, Tally displays the exact same fields and values as the Ledger.[^1]
- [x] Enabling the feature flag, ensure the Ledger displays that it is connected on the signing screen when the Ledger is connected, unlocked, and has blind data signing enabled. If any of those is not true, ensure that the Ledger displays as disconnected (and the Check Ledger button replaces the Sign button).

[^1]: Note that you will need the latest Ledger firmware for non-typed data messages to display the same value. Older firmwares displayed a message hash as opposed to message data, and Tally Ho is set up to display message data. We could do some hoop-jumping to special-case older firmwares correctly, but for now are explicitly choosing not to.


Shadowfiend:
All righty folks I believe this one is ready to go!

mhluongo:
Without the flag

Screenshot from 2022-09-20 12-21-49
Screenshot from 2022-09-20 12-22-01

and with it

Screenshot from 2022-09-20 12-19-10

Screenshot from 2022-09-20 12-20-31

Screenshot from 2022-09-20 12-20-45

In that last screen, I'd love to see the proper address component being used (linked and maybe ellided, etc) but otherwise it looks good.

ui/components/Signing/Signer/SignerLedger/SignerLedgerConnectionStatus.tsx
```diff
@@ -0,0 +1,61 @@
...
+  ledgerState,
+  mustEnableArbitraryDataSigning,
+}: SignerLedgerConnectionStatusProps): ReactElement {
+  switch (ledgerState.state) {
+    case "available":
+      if (mustEnableArbitraryDataSigning) {
+        return (
+          <img
+            src="/images/connect_ledger_indicator_disconnected.svg"
+            alt="Ledger is connected but cannot sign"
```
mhluongo, line 20: I'd like to see us lean in to i18n here → trying to move to only merging PRs that do it properly

mhluongo (changes requested):
Really want to see the i18n land. Otherwise appears good to go!

Shadowfiend:
> In that last screen, I'd love to see the proper address component being used (linked and maybe ellided, etc) but otherwise it looks good.

We can add linking, but reusing that component is probably the wrong move. The goal of this screen is to be as close to the Ledger display as possible rather than facilitate understanding, and that will likely deviate from the goals of the shared address component that's meant to make addresses easier for the user to understand. Open to a different take though.

mhluongo:
> The goal of this screen is to be as close to the Ledger display as possible rather than facilitate understanding, and that will likely deviate from the goals of the shared address component that's meant to make addresses easier for the user to understand.


I agree with your rationale on wanting to show the full thing. Maybe the `SharedAddress` just needs a "no shortening" prop? I suspect this will come up elsewhere as well.

Shadowfiend:
I think the fundamental goals are different enough that it would need to be a different component, even if still a shared one. We've already seen how mixed goals can lead to extremely messy components with the SharedButton leviathan.

Shadowfiend:
I've unchecked all test checkboxes as the i18n requires rechecking everything :/ I also fixed a bug where SIWE messages in particular were being presented as hex on the Ledger signing confirmation screen and as normal UTF-8 text on the Ledger itself. This is fixed by carrying the original UTF-8 message alongside the parsed SIWE data when dealing with signing data, which is then used to feed the correct display for the Ledger signing confirmation.

The full Ledger flow should now be i18ned, with two keys moved from `signTransaction` to `ledger` in the messages JSON.

mhluongo (approved):
Very excited seeing this move along!

PR #2254, opened 2022-09-19
Force sqlite3 5.1.1 which supports Apple Silicon

Shadowfiend:
@redux-devtools/cli requires sqlite3, which has not had darwin-arm64 builds until the latest version 5.1.1. Because the version constraint on @redux-devtools/cli is loose enough to be fulfilled by sqlite 5.1.1, we add a direct dependency to 5.1.1 to force the resolution to pull it in, and let yarn do the rest.

Fixes issues with `yarn install` on Apple Silicon.


Shadowfiend:
Confirmed this locally, but happy to get some confirmation from e.g. @kkosiorowska who I know recently had issues here.

greg-nagy (approved):
For me `brew install openssl@1.1` solved the issue so can't verify if this fix actually solves the problem. 

But otoh this is a good change anyway, so let's 🚢

If we don't hear about this problem ever again we will be happy :)

PR #2260, opened 2022-09-20
App permission details

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/2214
### What

Fix the problem with missing dapps favicons and titles.
Fix the problem with overflowing text on connect to dapp page.

### How

To fetch favicon and title properly we have to use `browser.tabs.get(id)`, we can't rely on `browser.runtime.onConnect` because it is not returning website details consistently. If we use `browser.tabs.get` without `tabs` extension permission it won't return any details that we care about ([see docs page](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/Tab), both title and favicon are `undefined` without `tabs` permission).

### Test

- [ ] allow to use extension in incognito mode (problems with title and favicon are happening mostly when you visit given app for the first time, at least for me)
- [ ] connect wallet to multiple dapps, check if dapps connection screen contains correct icon and title (not the URL as a title), check if text is not overflowing

image

### DIFF
--- a/manifest/manifest.json
+++ b/manifest/manifest.json
@@ -32,7 +32,7 @@
     "default_title": "Tally Ho",
     "default_popup": "popup.html"
   },
-  "permissions": ["alarms", "storage", "unlimitedStorage", "activeTab"],
+  "permissions": ["alarms", "storage", "unlimitedStorage", "activeTab", "tabs"],
   "background": {
     "persistent": true,
     "scripts": ["background.js", "background-ui.js"]
--- a/ui/pages/DAppConnect/RequestingDApp.tsx
+++ b/ui/pages/DAppConnect/RequestingDApp.tsx
@@ -1,5 +1,7 @@
 import React, { ReactElement } from "react"
 
+const TITLE_MAX_LENGTH = 55
+
 export default function RequestingDAppBlock(props: {
   title: string
   url: string
@@ -10,8 +12,14 @@ export default function RequestingDAppBlock(props: {
     <div className="request_wrap">
       <div className="dapp_favicon" />
       <div className="info">
-        <div className="dapp_title">{title}</div>
-        <div className="dapp_url">{url}</div>
+        <div className="dapp_title" title={title}>
+          {title.length > TITLE_MAX_LENGTH
+            ? `${title.slice(TITLE_MAX_LENGTH)}...`
+            : title}
+        </div>
+        <div className="dapp_url" title={url}>
+          {url}
+        </div>
       </div>
       <style jsx>{`
         .request_wrap {
@@ -33,13 +41,18 @@ export default function RequestingDAppBlock(props: {
           color: #fff;
           font-size: 16px;
           font-weight: 500;
+          overflow-wrap: anywhere;
         }
         .dapp_url {
           color: var(--green-40);
           font-size: 16px;
+          white-space: nowrap;
+          overflow: hidden;
+          text-overflow: ellipsis;
         }
         .info {
           margin-left: 16px;
+          width: calc(100% - 48px - 16px);
         }
       `}</style>
     </div>
--- a/background/services/provider-bridge/index.ts
+++ b/background/services/provider-bridge/index.ts
@@ -120,16 +120,12 @@ export default class ProviderBridgeService extends BaseService<Events> {
     }
 
     const { origin } = new URL(url)
+
     const completeTab =
       typeof tab !== "undefined" && typeof tab.id !== "undefined"
-        ? {
-            ...tab,
-            // Firefox sometimes requires an extra query to get favicons,
-            // unclear why but may be related to
-            // https://bugzilla.mozilla.org/show_bug.cgi?id=1417721 .
-            ...(await browser.tabs.get(tab.id)),
-          }
+        ? await browser.tabs.get(tab?.id)
         : tab
+
     const faviconUrl = completeTab?.favIconUrl ?? ""
     const title = completeTab?.title ?? ""
 


0xDaedalus (commented):
Tested and looks great - I would however love to get started testing this file as part of this PR.  

What do you think about adding a `index.unit.ts` file, writing a test for:

```
    if (typeof url === "undefined") {
      return
    }
 ```
 
 And adding a few `todo` tests based on some of the other functionality in `onMessageListener`?

File: manifest/manifest.json [hunk 1]
```diff
@@ -32,7 +32,7 @@
     "default_title": "Tally Ho",
     "default_popup": "popup.html"
   },
-  "permissions": ["alarms", "storage", "unlimitedStorage", "activeTab"],
+  "permissions": ["alarms", "storage", "unlimitedStorage", "activeTab", "tabs"],
```
greg-nagy, line 35: This will trigger a user modal on install and disable the extension before the user accepts it on auto-update. 

https://developer.chrome.com/docs/extensions/mv3/permission_warnings/#update_permissions

```
If the user manually updates an extension that now includes the [tabs](https://developer.chrome.com/docs/extensions/reference/tabs/) permission, they will get a warning on the management page.

If the extension is updated automatically it will be disabled until the user agrees to the new permissions.
```

But we have some other approaches that we can try:
- there is the `favicon` permission which might solve our headaches 
- we can create some kind of polling / double check mechanism that relies on the `activeTab` permission. 
- we can also extend the content script/`TallyWindowProvider` to actively check for the favicon and send it through the port communication

If we can't solve the problem without `tabs` permission, we should accept the flaky favicons IMO, but this will be a product decision.

ui/pages/DAppConnect/RequestingDApp.tsx
```diff
@@ -10,8 +12,14 @@ export default function RequestingDAppBlock(props: {
     <div className="request_wrap">
       <div className="dapp_favicon" />
       <div className="info">
-        <div className="dapp_title">{title}</div>
-        <div className="dapp_url">{url}</div>
+        <div className="dapp_title" title={title}>
+          {title.length > TITLE_MAX_LENGTH
+            ? `${title.slice(TITLE_MAX_LENGTH)}...`
```
greg-nagy, line 17: Could this be solved with ellipsis?

hyphenized:
Nice work! Looks like the reason why the `activeTab` permission is not enough to get the `faviconUrl` is that the user has to explicitly invoke the extension. No clue on why the port event is not returning the `faviconUrl` with the tab's data though, that's definitely got to be a bug... 

Paging @mhluongo and @Shadowfiend in case they have something to say about the updated permissions 👀

greg-nagy (changes requested):
Please try to find a solution for this without adding the `tabs` permission. If it's a no go let's ask for product decision, because that permission is scarry.

greg-nagy:
> Looks like the reason why the activeTab permission is not enough to get the faviconUrl is that the user has to explicitly invoke the extension. No clue on why the port event is not returning the faviconUrl with the tab's data though, that's definitely got to be a bug...

I think not getting the favicons is actually the correct behavior and we are relying on the info available at `onConnect` time because there the security restriction is not yet enforced properly. (If memory serves this info is stripped from the port object later on)

mhluongo:
We should be able to get the favicon info with our content script powers on the page, right? Maybe that's the angle.

manifest/manifest.json
hyphenized: I think the 3rd option is the way to go here, correct me if I'm wrong but the favicon permission does not seem mature enough yet and the `activeTab` approach won't work unless the user interacts with the wallet popup before connecting to the dApp... It might be easier to grab the icon manually from the dom with something like `link[rel*=icon]` or try to fallback into `${origin}/favicon.ico`/`favicon.png`. Although with the latter, there's still a risk of missing the image

greg-nagy: Hm, did some digging. 

- Yes, the `activeTab` only works on user action so it's a no go 
- `favicon` is for manifest v3 which we are not 
- because we are on manifest v2 we can use the `chrome://favicon/http://<domainpath>` as a fallback in the UI for a quick fix 
  - it has it's limitations [as pointed out in this stackoverflow article](https://stackoverflow.com/questions/10665321/reliably-getting-favicons-in-chrome-extensions-chrome-favicon/73415941#73415941) but those wont affect us if we apply this as a fallback in the UI
  - this will need to be updated when we migrate to manifest v3
- the most robust solution seems to be the 3., the content script approach

jagodarybacka: Ty @greg-nagy for the research, I couldn't find a perfect solution, will try with content script

ui/pages/DAppConnect/RequestingDApp.tsx
jagodarybacka: I want to keep 2 lines of text here. Will try with [line-clamp](https://caniuse.com/css-line-clamp) and see if we can have CSS only solution

provider-bridge/index.ts
```diff
@@ -17,6 +17,19 @@ export function connectProviderBridge(): void {
       event.source === window && // we want to recieve msgs only from the in-page script
       event.data.target === PROVIDER_BRIDGE_TARGET
     ) {
+      // if dapp wants to connect let's grab its details
+      if (event.data.request.method === "eth_requestAccounts") {
```
greg-nagy, line 21: This is an interesting spot to implement this, but I think this is the best for this. 
We ruse the `TallyWindowProvider` elsewhere and in that context this change wouldn't have made sense. 

Nice ✨

ui/public/index.css
```diff
@@ -160,4 +160,19 @@ label {
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
-}
\ No newline at end of file
+}
+
+.ellipsis_multiline {
```
greg-nagy, line 165: 🤩

greg-nagy (approved):
Love this fix 🚀

0xDaedalus:
Please resist the urge to merge PRs without tests 😅 - especially when tests have been requested

PR #2262, opened 2022-09-20
📑 Tabbed onboarding QA build

jagodarybacka:


greg-nagy:
We need to tweak the tabbed onboarding a bit before we can start to QA it

PR #2266, opened 2022-09-20
Don't show "Tally Ho as default" page onboarding

### LINKED ISSUE #2257: Hide 'Tally Ho set as default' page from onboarding
Since 'tis no longer set as default

Screen Shot 2022-09-19 at 11 07 04 PM

mhluongo:
We no longer flip the option on by default.

Closes #2257

# To Test

- [ ] Onboard with `SUPPORT_TABBED_ONBOARD=false`. You shouldn't see an intro page about Tally being set as default
- [ ] Onboard with `SUPPORT_TABBED_ONBOARD=true`. You shouldn't see an intro page about Tally being set as default


greg-nagy (approved):
I think it's sad that we have turned this to default off :(

but the PR and the code is good 

let's 🚢

PR #2267, opened 2022-09-21
Refactor window-provider

hyphenized:
Addresses this TODO:	[window-provider/index.ts:185](https://github.com/tallycash/extension/blob/main/window-provider/index.ts#L185)

## To Test

- [ ] Check that everything works exactly the same


greg-nagy (changes requested):
hey @hyphenized 

The changes are looking good here! This is an approve on my part. 

I know it's not fair, but this PR is overlapping somewhat with #2460 and that's a priority, so I will ask you to update this PR when that's merged. I am blocking this PR until then.

hyphenized:
No worries, I'll ping you after I update it. Thanks for the review!

hyphenized:
Looks like this wasn't affected by the other PR changes @greg-nagy

greg-nagy (approved):
let's 🚢

PR #2270, opened 2022-09-21
Localize Shared, SignData, and SignTransaction components

0xDaedalus:
### To Test
- [ ] Click through the application 
- [ ] Try signing different types of transactions
- [ ] Try signing transactions with ledger
- [ ] Things should look the same.

### DIFF
--- a/ui/_locales/en/messages.json
+++ b/ui/_locales/en/messages.json
@@ -69,6 +69,26 @@
       "step1": "Accept or Reject transaction on device",
       "step2": "Refresh page"
     },
+    "notConnected": {
+      "title": "Connect to Ledger",
+      "helpMessage": "Take the following steps:",
+      "step1": "Plug in Ledger",
+      "step2": "Enter PIN to unlock",
+      "step3": "Open Ethereum app"
+    },
+    "multipleLedgersConnected": {
+      "title": "Multiple Ledgers are connected",
+      "helpMessage": "Take the following steps:",
+      "step1": "Remove all Ledgers but one",
+      "step2": "Enter PIN to unlock",
+      "step3": "Open Ethereum app"
+    },
+    "wrongLedger": {
+      "title": "Wrong Ledger",
+      "helpMessage": "Looks like you are using the wrong Ledger.",
+      "connectLedger": "Connect the Ledger containing this account:",
+      "refresh": "Refresh the page"
+    },
     "derivationPaths": {
       "ledgerLive": "Ledger Live",
       "bip44": "BIP 44 (Trezor, 🦊)",
@@ -230,7 +250,8 @@
       "connectedStatus": "Connected"
     },
     "selectToken": "Select token",
-    "backButtonText": "Back"
+    "backButtonText": "Back",
+    "saveBtn": "Save"
   },
   "signing": {
     "signatureRequired": "Your signature is required",
@@ -256,8 +277,27 @@
     "noSigning": "Read-only accounts cannot sign",
     "sendTo": "Send to",
     "contractCreation": "Contract creation",
-    "interactingWith": "Interacting with"
-  },
+    "interactingWith": "Interacting with",
+    "unknownNetwork": "Unknown Network",
+    "detailPanelName": "Details",
+    "rawDataPanelName": "Raw data",
+    "rawDataCopyMsg": "Raw data copied to clipboard",
+    "copyRawData":  "Copy hex",
+    "spendAmount": "Spend Amount",
+    "help": "I need help",
+    "refresh": "Refresh",
+    "spendAsset": {
+      "approve": "Approve",
+      "infinite": "Infinite",
+      "spend": "Spend",
+      "tokens": "tokens",
+      "spendLimit": "Spend limit",
+      "tooltip1": "Spend limit is the amount of funds from a particular asset, that you allow a contract to spend.",
+      "tooltip2": "Infinite tx has the drawback that if the contract is mailicous, it can steal all your funds.",
+      "changeLimit": "Change limit",
+      "invalidAmount": "Invalid amount"
+    }
+    },
   "swap": {
     "title": "Swap Assets",
     "from": "Swap from:",
--- a/ui/_locales/en/messages.json
+++ b/ui/_locales/en/messages.json
@@ -56,6 +56,19 @@
     }
   },
   "ledger": {
+    "activation": {
+      "title": "Activate blind signing",
+      "helpMessage": "Take the following steps:",
+      "step1": "Open Ethereum app on Ledger",
+      "step2": "Navigate to Settings",
+      "step3": "Enable blind signing"
+    },
+    "busy": {
+      "title": "Ledger is busy",
+      "helpMessage": "Looks like your Ledger has another signature in progress.",
+      "step1": "Accept or Reject transaction on device",
+      "step2": "Refresh page"
+    },
     "derivationPaths": {
       "ledgerLive": "Ledger Live",
       "bip44": "BIP 44 (Trezor, 🦊)",
@@ -70,6 +83,7 @@
     "toBeDetermined": "TBD",
     "settingsTitle": "Network fees (Gwei)",
     "gasLimit": "Gas limit",
+    "estimatedNetworkFee": "Estimated network fee",
     "insufficientBaseAsset": "Not enough {{symbol}} for network fees",
     "totalMax": "Total Max",
     "saveSettings": "Save settings",
@@ -233,6 +247,16 @@
       "chainID": "Chain ID",
       "expiration": "Expiration"
     }
+  },
+  "signTransaction": {
+    "awaitingHardwareSignature": "Awaiting hardware wallet signature",
+    "cannotRejectWarning": "Tx can only be Rejected from Ledger",
+    "reject": "Reject",
+    "checkLedger": "Check Ledger",
+    "noSigning": "Read-only accounts cannot sign",
+    "sendTo": "Send to",
+    "contractCreation": "Contract creation",
+    "interactingWith": "Interacting with"
   },
   "swap": {
     "title": "Swap Assets",
--- a/ui/_locales/en/messages.json
+++ b/ui/_locales/en/messages.json
@@ -214,7 +214,9 @@
     "close": "Close",
     "accountItemSummary": {
       "connectedStatus": "Connected"
-    }
+    },
+    "selectToken": "Select token",
+    "backButtonText": "Back"
   },
   "swap": {
     "title": "Swap Assets",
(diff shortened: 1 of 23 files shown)


ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -109,7 +115,7 @@ export default function SignTransactionSpendAssetInfoProvider({
   return (
     <SignTransactionBaseInfoProvider
       title="Approve asset spend"
```
hyphenized, line 117: 👀

hyphenized: Looks good, missed a label here `ui/components/SignData/SignTypedDataInfo.tsx:49`

ui/components/SignTransaction/SignTransactionContainer.tsx
```diff
@@ -125,15 +127,15 @@ export default function SignTransactionContainer({
...
+        {isWaitingForHardware ? t("awaitingHardwareSignature") : title}
       </h1>
       <div className="primary_info_card standard_width">
         {isWaitingForHardware ? reviewPanel : detailPanel}
       </div>
       {isWaitingForHardware ? (
         <div className="cannot_reject_warning">
           <span className="block_icon" />
-          Tx can only be Rejected from Ledger
+          {t("cannotRejectWarning")}
```
Shadowfiend, line 138: Just a heads up that I'm planning on moving this and `checkLedger` to the `ledger` prefix in #2253 . I'll take care of updating this component as well.

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
0xDaedalus: Eagle eyes 👀 😄 

The `title` for the `SignTransactionBaseProvider` requires a slightly more complicated refactoring than just replacing an English string with `t("...")` [due to type constraints ](https://github.com/tallycash/extension/pull/2270/files/4e2c9b9ad01e11d8ead4f8cd7402d1da983182eb#diff-6c2bf2337eac96235069635135596124726645749b7fbfd39c883e70e57b9703R5)- so I've left that for a follow-up PR.

ui/components/SignTransaction/SignTransactionContainer.tsx
Shadowfiend: (This is now done in that PR.)

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
Shadowfiend: Tbqh I would vote not to do this in the current flow at all. Having this represented in a type at all is… Not a great solution. The new signing flow leaves this string entirely up to the component, as 0xch4d intended.

That said, if we really want to handle this in the current flow I can probably take it on.

Shadowfiend:
> Try signing different types of transactions

I assume this is trying to sign all types of data right—typed data, messages, and transactions?

Going through now.

ui/_locales/en/messages.json
```diff
@@ -214,8 +248,56 @@
...
+    "help": "I need help",
+    "refresh": "Refresh",
+    "spendAsset": {
+      "approve": "Approve",
+      "infinite": "Infinite",
+      "spend": "Spend",
+      "tokens": "tokens",
+      "spendLimit": "Spend limit",
+      "tooltip1": "Spend limit is the amount of funds from a particular asset, that you allow a contract to spend.",
+      "tooltip2": "Infinite tx has the drawback that if the contract is mailicous, it can steal all your funds.",
```
Shadowfiend, line 296:
```suggestion
      "tooltip2": "Infinite tx has the drawback that if the contract is malicous, it can steal all your funds.",
```

```suggestion
      "tooltip2": "Infinite tx has the drawback that if the contract is malicous, it can steal all your funds.",
```

While we're in here...

ui/_locales/en/messages.json
```diff
@@ -214,8 +248,56 @@
...
+    "spendAmount": "Spend Amount",
+    "help": "I need help",
+    "refresh": "Refresh",
+    "spendAsset": {
+      "approve": "Approve",
+      "infinite": "Infinite",
+      "spend": "Spend",
+      "tokens": "tokens",
+      "spendLimit": "Spend limit",
+      "tooltip1": "Spend limit is the amount of funds from a particular asset, that you allow a contract to spend.",
```
Shadowfiend, line 295:
```suggestion
      "tooltip1": "Spend limit is the amount of funds from a particular asset that you allow a contract to spend.",
```

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -135,25 +141,19 @@ export default function SignTransactionSpendAssetInfoProvider({
               customStyles="margin: 10px 0 0;"
               height={32}
             >
-              Spend{" "}
+              {t("spend")}{" "}
               {asset.symbol ?? (
                 <SharedAddress address={transactionDetails.to ?? ""} />
               )}{" "}
-              tokens
+              {t("tokens")}
```
Shadowfiend, line 148: This probably needs to be reworked as a formatting thing. There's no guarantee that “<verb> <asset> tokens” is going to be the right order for this in another language—instead, we need the i18n text to be one key that looks something like “Spend {{asset-symbol}} tokens”, and need to use the [react-i18next `Trans` component](https://react.i18next.com/latest/trans-component#alternative-usage-which-lists-the-components-v11.6.0) to let us slot in the `SharedAddress` component or simple text for `{{asset-symbol}}`.

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -123,7 +129,7 @@ export default function SignTransactionSpendAssetInfoProvider({
             </div>
           </div>
           <span className="site">
-            Approve{" "}
+            {t("approve")}{" "}
```
Shadowfiend, line 132: As below, this needs to be something like “Approve {{spending-contract}}” or so.

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -48,7 +54,7 @@ export default function SignTransactionSpendAssetInfoProvider({
       })
 
   const approvalLimitDisplayValue = `${
-    approvalLimitString ?? "Infinite"
+    approvalLimitString ?? t("infinite")
```
Shadowfiend, line 57: As below, just swapping out the word “infinite” here probably isn't enough to i18n. We'll need to update the container text so we're doing e.g. `{asset-amount} {asset-symbol}` or so, and use [i18next's plurals functionality](https://www.i18next.com/translation-function/plurals). Similarly, we may need a separate key for infinite asset, `Infinite {asset-symbol}`, as some languages may/probably will require a different wording when presenting infinity vs a definite number. Not sure if the plurals stuff supports this directly.

ui/components/SignTransaction/SignTransactionTransferInfoProvider.tsx
```diff
@@ -52,7 +54,7 @@ export default function SignTransactionTransferInfoProvider({
       infoBlock={
         <div className="sign_block">
           <div className="container">
-            <div className="label">Send to</div>
+            <div className="label">{t("sendTo")}</div>
```
Shadowfiend, line 57: We may need to suggest translations can use something like “Recipient” here. I wonder if we name the key like that, and set the text to “Send to” for English, where it reads naturally. In other languages “Send to X” might not make sense, and in the UI layout it may make sense to treat it more like a label rather than part of a phrase.

ui/components/SignData/EIP4361Info.tsx
```diff
@@ -17,30 +18,34 @@ const CHAIN_NAMES: (chain: number) => string = (chain) => {
...
   return (
     <>
       <div className="domain">{signingData.domain}</div>
       <div className="divider spaced" />
       <div className="subtext">
-        Wants you to sign in with your
+        {t("subtext1")}
         <br />
-        Ethereum account:
+        {t("subtext2")}
```
Shadowfiend, line 29: Phew, this is rough af. The manual break feels like an issue, but also what happens if the translation is too long here? We should talk to Vlad a little about this one, as it's one of the few places we use longer phrases that could run into space problems.

Shadowfiend (approved):
Ok, went through and checked the signing flows and the Ledger flows for message data, SIWE, typed data (permit only), and transactions, as well as the Ledger states for disconnected, multiple connected, wrong connected, and normal flow.

Few thoughts:
- We should probably allow for translating the typed data field keys so we can hit common keys with translations. We should be able to use the `default` functionality on `t` to set a default `capitalize(key)` when the key doesn't have an i18ned key.
- I left a few notes in places where we're translating phrases by translating fragments of the phrases, including places where numbers are involved. I don't think this is going to work, we're going to need to do some interpolation.

I don't think any of this is blocking since it's strictly an improvement, so I'm giving this the green light.

ui/_locales/en/messages.json
0xDaedalus: addressed in https://github.com/tallycash/extension/pull/2289

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
0xDaedalus: Yeah - I am conscious of these edge cases but I'd like to tackle them in follow up translation polish PRs and just kind of get the bulk of the english string => json work out of the way first.  This stuff is so mind-numbing that its _quite_ difficult to context switch to real programming in the middle of one of these PRs.

Shadowfiend: I'll get you in the new flow 😉

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
Shadowfiend: The layout of things here means we don't actually need to worry about this 😮‍💨

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
Shadowfiend: Handling this one in #2274 .

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
Shadowfiend: Handling this one in #2274.

PR #2274, opened 2022-09-21
Auto Detailing: Add signing data to the new signing flow

Shadowfiend:
The diff is big here but most of the components are pulled from our existing signing summaries (for spend assets, etc), with a different wrapping API that matches the new signing flow. The only real significant addition here is `useSwitchablePanels`, which is a top-level hook that makes using `SharedSwitchablePanel` a little less annoying.

## Testing

- [x] All of this should be behind a feature flag. Ensure signing is unchanged with the feature flag disabled.
  - [x] Check asset transfers are still labeled correctly in the UI on the old flow. An i18n key was changed.
  - [x] Ensure dismissing a “sending funds to a contract is dangerous” warning is still dismissable. Sending to 0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe should show this warning. The state management was modified for this flow.
  - [x] Ensure gas management and raw data panels still work as expected. A very small change was made here.
- [x] Enabling the feature flag, ensure that each type of transaction has its summary data displayed correctly.
  - [x] ETH transfer (can be from the extension).
  - [x] ERC20 transfer (can be from the extension).
  - [x] Approval transaction (the Tally test wallet does not have SUSHI approved, so an easy way is to try to swap SUSHI).
  - [x] Normal contract interaction (a swap with an approved asset will work here).
- [x] For all of the above, make sure fee changes are applied correctly and, if there are other internal form elements (e.g., the approval transaction amount), that these are also applied correctly.


Shadowfiend:
Ok this fellow is nearly ready but I'm going to apply the i18n from #2270 with some improvements to the new flow.

Shadowfiend:
This gentleman should be ready to go.

mhluongo (approved):
:postal_horn: :postal_horn: :postal_horn:

PR #2284, opened 2022-09-22
Do not annotate contract deployments

hyphenized:
Found this while profiling, `transaction.to` can be `null` for contract deployments according to Alchemy `getTransaction` docs

nullisanobject

Also noticed we were enriching the transaction twice as a consequence of saving it twice. So fixed that as well.

## To test
- [ ] Load `junk junk....` account
- [ ] Wait until it retrieves this transaction, or do it manually `0x7fdc19bc064de9b71a389177415fdb609a8bf92325d0130d05a26a4d486d2e14`
- [ ] No error should pop up


background/services/chain/index.ts
```diff
@@ -1268,10 +1269,7 @@ export default class ChainService extends BaseService<Events> {
         network
       ).getTransaction(hash)
 
-      const transaction = transactionFromEthersTransaction(result, network)
-
-      // TODO make this provider type specific
-      await this.saveTransaction(transaction, "alchemy")
```
Shadowfiend, line 1274: The gap between this and the below step may not be short, e.g. if we're retrieving a pending transaction (done on startup). We could also have the transaction receipt fail to retrieve (for connectivity reasons or others), and lose the transaction entirely in this flow, or considerably delay presenting it in the UI.

Overall unless we're seeing performance impact, I would recommend against taking this approach.

background/services/enrichment/transactions.ts
```diff
@@ -207,7 +207,7 @@ export default async function resolveTransactionAnnotation(
 
   // If the tx has a recipient, its a contract interaction or another tx type
   // rather than a deployment.
-  if (typeof transaction.to !== "undefined") {
+  if (typeof transaction.to !== "undefined" && transaction.to !== null) {
```
Shadowfiend, line 210: I think we should take one of two approaches here:
- We update the type of `EVMTransaction.to` to express that `null` is a possible value. This will probably break stuff.
- We update the conversion to `EVMTransaction` to convert `null` to `undefined`.

I favor doing the second one, in `transactionFromEthersTransaction`. Unfortunately although Ethers gives us types, Ethers isn't built with `strictNullChecks` enabled, so `to?: string` allows a `null` value, which is why this slipped through.

background/services/chain/index.ts
hyphenized: Thanks, you're right, taking into account the delay it's better to show up right away what's available

background/services/enrichment/transactions.ts
hyphenized: I've added a conversion to `transactionFromEthersTransaction`, thanks for the guidance 👍

jagodarybacka (approved):
I think it's good to go :)

PR #2285, opened 2022-09-23
Debounce middleware call to persist the store

hyphenized:
Added a debounce to the redux cache middleware since `JSON.stringifying` the entire store after every dispatch is a bit expensive. Still need to test what could happen if the browser crashes before saving triggers.

## Results:

About 3.8x~ less time spent calling `encodeJSON`

### With debounce:
debounce

### Without: 
nodebounce

Note: While this improves performance, for some reason CPU still reaches 100% every so often when inspecting the background page on brave.

## To test:
- [ ] Reinstall the extension
- [ ] Start profiling the extension, add an account, time spent calling `encodeJSON` should decrease considerably


mhluongo:
Don't be afraid to split separate stuff like this into even smaller independent PRs — it saves us from needing to do performance testing per commit (for example), and it saves you from a review nit around eg the manifest build slowing down the merge for an important performance improvement.

hyphenized:
That makes sense, I'll split this up in 3 PRs. Thanks Matt!

greg-nagy:
I won't have time to finish the testing of this PR today, but will test tomorrow

hyphenized:
> I won't have time to finish the testing of this PR today, but will test tomorrow
Let me know if there's anything I can do to help!

greg-nagy (approved):
Tested and the fix works really well, ty ❤️

We will still have cpu issues in the future, but this is a huge improvement!

PR #2286, opened 2022-09-23
🏆 Achievements tab

jagodarybacka:
### What

Create an achievements tab on the Portfolio page to show Galxe OATs/NFTs there

image

### Testing
- [x] add account with no Galxe NFts - achievements tab should show empty state
- [x] add account with some Galxe NFTs - they should not show on Portfolio's NFTs tab, they should be on Achievements tab. They should be visible on Wallet's NFTs tab.
- [ ] get new Galxe OAT and confirm it is added to the list


jagodarybacka:
@VladUXUI @mr-michael I've implemented that banner already but if we don't want this banner right now we have to change text below "empty bowl". Or we can leave that banner and we don't have to add the one on the Wallet page.
image

background/lib/nfts.ts
```diff
@@ -14,6 +14,11 @@ export type NFT = {
     url: string
   }[]
   contract: { address: string }
+  isOAT: boolean
+}
+
+function isOAT(url: string | null | undefined) {
```
mhluongo, line 20: This will scoop of non-OAT Galxe NFTs... can we change the name? Maybe `isGalxeAchievement`?

ui/_locales/en/messages.json
```diff
@@ -149,9 +149,12 @@
     }
   },
   "nfts": {
-    "empty": "Looks like you don&apos;t have any NFTs yet? Get some and view them here!",
+    "empty": "Looks like you don't have any NFTs yet? Get some and view them here!",
     "noTitle": "No title"
   },
+  "oats": {
+    "empty": "Nothing to see here :(\nBut you can start with the quest below"
```
mhluongo, line 156: Let's avoid the frowny face  in the copy here? I imagine this wasn't your call, but typically we should use positive language outside of error cases

ui/components/NFTs/OATsEmpty.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { ReactElement } from "react"
+import { useTranslation } from "react-i18next"
+import OATsBanner from "./OATsBanner"
+
+export default function OATsEmpty(): ReactElement {
```
mhluongo, line 5: Same, this isn't just an OATs feature

mhluongo: I'd suggest we call these "Achievements" in the general case, throughout the UI codebase... and "Galxe NFTs" or "Galxe achievements" when being that specific matters

mhluongo (commented):
Impressed with how quickly this came together. Left some quick thoughts.

jagodarybacka:
New banner design was created, converting to draft for a moment

mhluongo (approved):
LGTM :ok_hand:

PR #2287, opened 2022-09-23
⛓ Add arbitrum support

greg0x:
# What

Add Arbitrum support

Kudos to @0xDaedalus for the awesome preparation ❤️

# Test
- [ ] remove and install extension, import a seed phrase
- [ ] try switch to arbitrum and 
  - [ ] see if the balance loads
  - [ ] see if the activity loads
- [ ] try sending tokens on arbitrum between wallets
- [ ] try swapping on arbitrum 
- [ ] try connecting to dapps on arbitrum and use them

### DIFF
--- a/background/services/chain/tests/index.unit.test.ts
+++ b/background/services/chain/tests/index.unit.test.ts
@@ -100,7 +100,7 @@ describe("Chain Service", () => {
         chainService as unknown as ChainServiceExternalized
       ).lastUserActivityOnNetwork[ETHEREUM.chainID]
       chainService.markNetworkActivity(ETHEREUM.chainID)
-      expect(lastUserActivity).toBeLessThan(
+      expect(lastUserActivity).toBeLessThanOrEqual(
         (chainService as unknown as ChainServiceExternalized)
           .lastUserActivityOnNetwork[ETHEREUM.chainID]
       )


ui/components/NetworkFees/NetworkSettingsChooser.tsx
```diff
@@ -46,7 +46,7 @@ export default function NetworkSettingsChooser({
             onSave={saveNetworkSettings}
           />
         ) : (
-          <NetworkSettingsOptimism />
+          <NetworkSettingsOptimismAndArbitrum />
```
0xDaedalus, line 49: Arbitrum's gas estimation implementation differs from Optimism's - we should have everything we need in the `NetworkSettingsSelectLegacy` file to allow users to adjust their gas prices on Arbitrum and won't want to display the disabled gas screen we have for Optimism.

background/networks.ts
```diff
@@ -165,14 +160,34 @@ export type LegacyEVMTransactionRequest = Pick<
...
+
 /**
  * An EIP1559 EVM transaction, whose type is set to `1` or `2` per EIP1559 and
  * whose EIP1559-related fields are required, while `gasPrice` (pre-EIP1559) is
  * mandated to be `null`.
  */
 export type EIP1559Transaction = EVMTransaction & {
   gasPrice: null
-  type: 1 | 2
+  type: KnownTxTypes
```
0xDaedalus, line 190: Maybe I have a knowledge gap here but my understanding is that Arbitrum does not support EIP-1559 (despite having an EIP-1559-style gas auction mechanism built in).

0xDaedalus: Ok - it looks like Arbitrum "mostly" implements EIP-1559 but [ignores tips](https://developer.arbitrum.io/arbos/gas#tips-in-l2).

I think this means we can implement gas controls very similar to how we do on Ethereum and Polygon.

ui/components/NetworkFees/NetworkSettingsChooser.tsx
greg-nagy: https://developer.arbitrum.io/arbos/gas
https://medium.com/offchainlabs/understanding-arbitrum-2-dimensional-fees-fd1d582596c9

I might have a knowledge gap here. Based on what I read in those articles the `eth_estimateGas` returns the correct gas amount, and we can't change the tip. 

```
The sequencer prioritizes transactions on a first-come first-served basis. Because tips do not make sense in this model, they are ignored. Arbitrum users always just pay the basefee regardless of the tip they choose.
```

Doublechecking the code path though.

background/networks.ts
greg-nagy: My understanding is that arbitrum is using EIP-2718 which is backward compatible with EIP-1559. So from our point of view, all we need to do is accept the extended tx types if we don't want to use the rest. Might not be the best solution but renaming everything seemed like an overshoot for something that we don't want to use at all.

Happy to be convinced otherwise!

Started a [discussion here](https://matrix.to/#/!VGkUlrCkVskNDeSMmM:thesis.co/$fC9Znn-NnHiw7WvYpCfDaxit3kxNhqL9nXg5nQd2qnE?via=thesis.co&via=matrix.org)

greg-nagy: > I think this means we can implement gas controls very similar to how we do on Ethereum and Polygon.

Why should we do this? Changing the gas fee won't affect how quickly a tx is mined based on what I understand, and we get an estimation from the network that's sufficient for the tx to go through.

What am I missing?

0xDaedalus: Changing the gas fee won't change how quickly a tx is mined but it can affect if a transaction is accepted at all or not.

During periods of high gas volatility (1st week of arbitrum odyssey being a prime example) we would certainly have an increase in dropped transactions due our gas price estimates being on a 15 second interval and arbitrum block time being so short.

Incidentally the same is true for Optimism - it just so happens that we don't need a whole new design to implement gas controls on Arbitrum (since `eth_estimateGas` returns the _total_ gas needed for a transaction to be executed - as opposed to Optimism's implementaiton which returns just the l1 portion).

ui/components/NetworkFees/NetworkSettingsChooser.tsx
greg-nagy: > we should have everything we need in the NetworkSettingsSelectLegacy file to allow users to adjust their gas prices on Arbitrum and won't want to display the disabled gas screen we have for Optimism.

I follow your lead on this, but man, this feels nasty :)

0xDaedalus: Lets CC @VladUXUI on this to get his thoughts.  

The question here is do we want to:
a) Present users with the same screen we have on Optimism about gas controls
b) Allow users to edit their gas settings via the legacy network settings select.

a) means that in times of high volatility users might get frustrated at failing transactions
b) means that we might be using an older design that is out-of-date with Vlads vision

@greg-nagy would you mind posting a screenshot of the legacy gas settings?

ui/components/NetworkFees/NetworkSettingsChooser.tsx
```diff
@@ -34,21 +35,37 @@ export default function NetworkSettingsChooser({
...
-  return (
-    <>
-      <div className="wrapper">
-        {transactionDetails &&
-        EIP_1559_COMPLIANT_CHAIN_IDS.has(transactionDetails.network.chainID) ? (
-          <NetworkSettingsSelect
+  function networkSettingsSelectorFinder() {
+    if (transactionDetails) {
+      if (transactionDetails.network.name === "Optimism") {
+        return <NetworkSettingsOptimismAndArbitrum />
```
jagodarybacka, line 41: If you prepared this component to handle both OP and Arbitrum then why Arbitrum is using `NetworkSettingsSelectLegacy`?

greg-nagy: There is a convo above about which approach we should take, and I haven't reverted these changes yet.

ui/components/NetworkFees/NetworkSettingsChooser.tsx
greg-nagy: @VladUXUI 

Here are the screenshots. What are your thoughts?

a) 
CleanShot 2022-09-30 at 11 15 10

b) 
CleanShot 2022-09-30 at 11 14 00

VladUXUI: I would say legacy although i understand it's more prone to bugs.
With Odyssey launching, we don't know how gas is going to go.
Although Nitro should help.
Is there much upside/downside to either? From a dev perpective

0xDaedalus: Nope not much difference at all

0xDaedalus (approved):
Everything here looks quite good.

Created a follow-on issue for the outstanding work - I'm sure we will catch some more stuff in our internal QA push but honestly this feels _really_ solid so as-is.

ui/components/NetworkFees/NetworkSettingsChooser.tsx
VladUXUI: Then let's go with Legacy, although it probably also needs a bit of TLC

PR #2292, opened 2022-09-24
Fix `assetsLoaded` dispatch spam after fetching assets

hyphenized:
Moved the assets event dispatch so that it happens after token lists have loaded, before it was happening 5 x 10 times (each network * 10 default token lists) and the redux action is sort of expensive.

Also removed the  call to `fetchAndCacheTokenLists` from Indexing Service's `internalStartService` seeing that there's a scheduled alarm that runs just before that

Could we change the `periodInMinutes` for handleTokenAlarm?
## To Test
- [ ] Profile the extension, every minute or so time spent calling `assetsLoaded` should decrease about 90%, extension startup should also be faster.


0xDaedalus (approved):
👑

PR #2293, opened 2022-09-24
FIx webpack manifest overrides

hyphenized:
This PR adds ~`["brave", "edge", "firefox", "opera"]` back to the `supportedBrowsers` with a `CI` env check, it also~ sorting for the `manifest.json` files before merging them so that the final output from webpack respects manifest overrides

## To test
- [ ] ~Try to create a production build for brave running `yarn build --config-name brave`~
- [ ] Add an override to `manifest.development.json` it should work as expected


0xDaedalus (commented):
Looks good to me and the effect on CI build times seems to be negligible.  Will wait for @greg-nagy for the final approval here.

greg-nagy:
What was the motivation behind this change? 

We had this setup like this, but I have removed them because I couldn't find any reason not to.
We are not shipping to any other browsers, probably the extension is broken at least partially for most of them. 

I am not against this, I just would like to understand the motivation behind the change.

hyphenized:
> We had this setup like this, but I have removed them because I couldn't find any reason not to. We are not shipping to any other browsers, probably the extension is broken at least partially for most of them.
> 
> I am not against this, I just would like to understand the motivation behind the change.

Ahh my bad, since those browsers are still mentioned in the `README` I thought they were removed because of the memory issue addressed in #1843. Please let me know if the fix for the manifest overrides is good 👀

greg-nagy:
> Ahh my bad, since those browsers are still mentioned in the README I thought they were removed because of the memory issue addressed in https://github.com/tallycash/extension/pull/1843. 

Yes, that was the point they were removed, but the reasoning above was the source of the change. (Any my laziness, because this way I don't need to `--config-name chrome` every time :) )

> Please let me know if the fix for the manifest overrides is good 👀

The overrides part looks good.

greg-nagy (approved):
🤩 lookin' good

PR #2301, opened 2022-09-26
Only poll for gas prices when a network is active

0xDaedalus:
This PR introduces functionality which limits the extension from polling for gas prices when the extension is not active - and lays the groundwork for doing the same with base asset balances, erc20 lookups, and just about any other chain lookup we want to throttle.

## What is going on here.

The central change that this PR introduces is a [central piece of state](https://github.com/tallycash/extension/blob/reduce-the-consumption/background/services/chain/index.ts#L146:L148) inside of chain service which tracks the last time a network was "active".  The current events that trigger a refresh of "active" time are:
-  User has opened or navigated to any page in the extension (currently selected network activity)
-  A dapp has requested to permission to connect to the wallet (dapp's connected network activity)
-  A dapp has requested to switch to a given network (network that is requested to switch to activity)

When any of these events is triggered - one of the following two things happens:
-  If a network was active within [the last 5 minutes](https://github.com/tallycash/extension/blob/reduce-the-consumption/background/services/chain/index.ts#L90:L91) the  "last active time" is refreshed to the current time.
- If a network was not active within the last 5 minutes - blockPrices are immediately requested, and the "last active time" is refreshed to the current time.

Once a network is "active" - gas prices are queried and updated in the background every 15 seconds (per our standard browser alarm pattern) until there has been 5 minutes of no activity on that network.

## What is not going on here

Asset transfer refresh schedule and query frequency is not affected.
Asset balance refresh schedule and query frequency is not affected.
Token price refresh schedule and query frequency is not affected.

## Implementation Notes
- Since this PR does not reduce the frequency of the browser alarms querying for gas prices - we still maintain background-script activity that should prevent the background service worker from being turned off due to inactivity once we switch to manifest v3.
- If we like this pattern - we can extend it to asset transfer and asset balance polling as well - although that will likely require more UX testing that this PR since those queries touch data that we want immediately presentable upon application open.

## To Test
- Change the [`NETWORK_POLLING_TIMEOUT`](https://github.com/tallycash/extension/blob/reduce-the-consumption/background/services/chain/index.ts#L90:L90) to 1 minute.
- Drop some logs inside of [`pollBlockPricesForNetwork`](https://github.com/tallycash/extension/blob/reduce-the-consumption/background/services/chain/index.ts#L1049:L1049) after the first `if` so you can monitor when a network's block prices are being polled.
- Import a wallet and add it to multiple networks
- [ ] After 1 minute of inactivity - you should no longer see logs indicating that block prices are being polled
- [ ] After opening the extension - you should see logs indicating that block prices of the currently selected network are being polled
- [ ] After connecting to a dapp - you should see block prices of the network the dapp is connected to being polled
- [ ] After switching networks in a dapp - you should see block prices of the network you've switched to being polled
- [ ] When signing a transaction - you should see gas prices populating and updating with values similar to what you'd see on https://l2fees.info/

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1082,7 +1082,6 @@ export default class Main extends BaseService<never> {
     )
 
     uiSliceEmitter.on("newSelectedNetwork", (network) => {
-      this.chainService.markNetworkActivity(network.chainID)
       this.internalEthereumProviderService.routeSafeRPCRequest(
         "wallet_switchEthereumChain",
         [{ chainId: network.chainID }],
--- a/background/main.ts
+++ b/background/main.ts
@@ -1075,6 +1075,7 @@ export default class Main extends BaseService<never> {
     )
 
     uiSliceEmitter.on("newSelectedNetwork", (network) => {
+      this.chainService.markNetworkActivity(network.chainID)
       this.internalEthereumProviderService.routeSafeRPCRequest(
         "wallet_switchEthereumChain",
         [{ chainId: network.chainID }],
@@ -1096,6 +1097,14 @@ export default class Main extends BaseService<never> {
       "initializeAllowedPages",
       async (allowedPages: PermissionMap) => {
         this.store.dispatch(initializePermissions(allowedPages))
+      }
+    )
+
+    this.providerBridgeService.emitter.on(
+      "permissionQueriedForChain",
+      async (chainID: string) => {
+        this.chainService.markNetworkActivity(chainID)
+        this.chainService.pollBlockPricesForNetwork(chainID)
       }
     )
 
--- a/background/main.ts
+++ b/background/main.ts
@@ -1110,7 +1110,6 @@ export default class Main extends BaseService<never> {
       "permissionQueriedForChain",
       async (chainID: string) => {
         this.chainService.markNetworkActivity(chainID)
-        this.chainService.pollBlockPricesForNetwork(chainID)
       }
     )
 
--- a/background/main.ts
+++ b/background/main.ts
@@ -733,6 +733,10 @@ export default class Main extends BaseService<never> {
         )
       )
     })
+
+    uiSliceEmitter.on("userActivityEncountered", (addressOnNetwork) => {
+      this.chainService.markNetworkActivity(addressOnNetwork.network.chainID)
+    })
   }
 
   async connectNameService(): Promise<void> {
--- a/background/main.ts
+++ b/background/main.ts
@@ -1034,6 +1034,9 @@ export default class Main extends BaseService<never> {
         resolver: (result: string | PromiseLike<string>) => void
         rejecter: () => void
       }) => {
+        this.chainService.pollBlockPricesForNetwork(
+          payload.account.network.chainID
+        )
         this.store.dispatch(signDataRequest(payload))
 
         const clear = () => {
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -378,6 +378,7 @@ export default class ChainService extends BaseService<Events> {
       logger.warn(
         `${activeNetwork.name} already active - no need to activate it`
       )
+      this.markNetworkActivity(activeNetwork.chainID)
       return activeNetwork
     }
 
@@ -408,6 +409,7 @@ export default class ChainService extends BaseService<Events> {
       })
     }
 
+    this.markNetworkActivity(networkToActivate.chainID)
     return networkToActivate
   }
 
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -1031,6 +1031,10 @@ export default class ChainService extends BaseService<Events> {
     }
   }
 
+  markNetworkActivity(chainID: string): void {
+    this.lastUserActivityOnNetwork[chainID] = Date.now()
+  }
+
   /*
    * Periodically fetch block prices and emit an event whenever new data is received
    * Write block prices to IndexedDB so we have them for later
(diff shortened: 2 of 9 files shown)


ui/pages/Popup.tsx
```diff
@@ -88,6 +90,12 @@ function useConnectPopupMonitor() {
 export function Main(): ReactElement {
   const dispatch = useBackgroundDispatch()
 
+  // Emit an event when the popup page is first loaded.
+  const currentAccount = useBackgroundSelector(selectCurrentAddressNetwork)
+  useEffect(() => {
+    dispatch(userActivityEncountered(currentAccount))
+  })
```
jagodarybacka, line 97: this useEffect will run every time Main is rerendered because it has no dependency array, from the comment above I feel like it is not the effect you want to get
```suggestion
  useEffect(() => {
    dispatch(userActivityEncountered(currentAccount))
  }, [])
```

background/services/chain/index.ts
```diff
@@ -1019,19 +1024,55 @@ export default class ChainService extends BaseService<Events> {
...
+    ) {
+      return
+    }
+
+    const subscription = this.subscribedNetworks.find(
+      ({ network }) => toHexChainID(network.chainID) === toHexChainID(chainID)
+    )
+
+    if (!subscription) {
+      logger.warn(`Can't fetch block prices for unsupported chainID ${chainID}`)
```
jagodarybacka, line 1062: It's not unsupported network, it's not subscribed network - network can be supported but not subscribed to at the moment, right?

background/main.ts
```diff
@@ -1099,6 +1106,14 @@ export default class Main extends BaseService<never> {
       }
     )
 
+    this.providerBridgeService.emitter.on(
+      "permissionQueriedForChain",
+      async (chainID: string) => {
+        this.chainService.markNetworkActivity(chainID)
+        this.chainService.pollBlockPricesForNetwork(chainID)
```
jagodarybacka, line 1113: this can run twice as `markNetworkActivity` will call the same method if the time condition is met

background/services/chain/index.ts
0xDaedalus: Correct - great catch

background/main.ts
0xDaedalus: Yeah you're right - and its not like we need up-to-the-second gas prices when first connecting to a dapp.  Great spot

jagodarybacka (approved):
Works fine 👍

background/services/chain/index.ts
```diff
@@ -218,21 +225,12 @@ export default class ChainService extends BaseService<Events> {
       blockPrices: {
         runAtStart: false,
         schedule: {
-          periodInMinutes:
-            Number(process.env.GAS_PRICE_POLLING_FREQUENCY ?? "120") / 60,
+          periodInMinutes: 0.25, // Every 15 seconds
```
greg-nagy, line 228: Chrome alarms can not be less than a minute.

https://developer.chrome.com/docs/extensions/reference/alarms/#method-create

```
In order to reduce the load on the user's machine, Chrome limits alarms to at most once every 
1 minute but may delay them an arbitrary amount more. That is, setting delayInMinutes or 
periodInMinutes to less than 1 will not be honored and will cause a warning. when can be set 
to less than 1 minute after "now" without warning but won't actually cause the alarm to fire for 
at least 1 minute.
```

We also have a warning in the console about this.

background/services/chain/index.ts
```diff
@@ -218,21 +225,12 @@ export default class ChainService extends BaseService<Events> {
       blockPrices: {
         runAtStart: false,
         schedule: {
-          periodInMinutes:
-            Number(process.env.GAS_PRICE_POLLING_FREQUENCY ?? "120") / 60,
```
greg-nagy, line 222: GAS_PRICE_POLLING_FREQUENCY should be removed from the .env.defaults

background/services/chain/index.ts
```diff
@@ -1012,19 +1017,57 @@ export default class ChainService extends BaseService<Events> {
...
-        this.emitter.emit("blockPrices", { blockPrices, network })
-      })
+      this.subscribedNetworks.map(async ({ network }) =>
+        this.pollBlockPricesForNetwork(network.chainID)
+      )
     )
   }
 
+  async pollBlockPricesForNetwork(chainID: string): Promise<void> {
+    if (
```
greg-nagy, line 1043: ❤️ the double safety here

greg-nagy (commented):
in progress review

background/services/chain/index.ts
jagodarybacka: oh no, I have warnings filtered out in the console 😵‍💫

background/services/provider-bridge/index.ts
```diff
@@ -348,6 +349,9 @@ export default class ProviderBridgeService extends BaseService<Events> {
     origin: string,
     chainID: string
   ): Promise<PermissionRequest | undefined> {
+    // This is analogous to "User opened a dapp on chain X"
+    this.emitter.emit("permissionQueriedForChain", chainID)
```
greg-nagy, line 353: This method will be emitted on every dapp RPC call because the `checkPermission` method is [used in the port communication listener](https://github.com/tallycash/extension/blob/aad9d3408e556d6326f3d3d34df0118a7643ad7a/background/services/provider-bridge/index.ts#L136)

greg-nagy (commented):
This is an awesome PR, wonderful initial small scope change and a could approach to reduce api calls 

The 2 things that worth mentioning: 
- the sub 1 min alert shouldn't work as per chrome documentation
- any connected dapp RPC communication will trigger the `markNetworkActivity`, not just connect and network switch. To test: 
  - drop a console log in the `markNetworkActivity`
  - open uniswap and connect

PR #2307, opened 2022-09-27
Hardcode the 0x Router dApp favicon for swap tx's

### LINKED ISSUE #2244: Hardcode the "0x Router" dApp favicon for swap tx's
Screenshot from 2022-09-19 13-22-09

As this tx was internally initiated, seeing a "missing" dApp favicon is pretty weird.

kkosiorowska:
Closes #2244 

This PR adds the 0x Router dApp favicon for swap tx's.

**Changes**
* add a new icon
* add the internal provider domain name to constants
* add a condition for rendering icon

**UI**

Before
Screenshot 2022-09-27 at 09 46 04

After
Screenshot 2022-09-27 at 09 45 40

**To test**
- [ ] check favicon for swap tx's


kkosiorowska:
@Shadowfiend 

> * Not all traffic on the internal provider is swaps. Do we show 0x when doing an ERC20 transfer as well, for example?

I will change the name of the flag from generic to more clear and specific. I will make sure also that the icon change is only for swap.

> * How spoofable is the domain? Is there a way a dApp could pretend to be 0x by pushing what we think is the internal provider through the pipeline? We can probably ensure there is no such vector by rejecting all requests that have this domain in the provider bridge.

Could you explain it more?  I'm not sure that I understand everything what do you mean.

Shadowfiend:
> Not all traffic on the internal provider is swaps.

This is a bigger issue, brought it up to @VladUXUI internally for further discussion.

> How spoofable is the domain?

This was me misunderstanding how you implemented this, sorry. You adjusted how we dealt with the actual resolved name of the contract, rather than injecting a name in requests from the internal provider itself, which is what I thought I was seeing when I skimmed. It doesn't apply :)

background/constants/index.ts
```diff
@@ -33,6 +33,8 @@ export enum EarnStages {
   Live = "Live",
 }
 
+export const INTERNAL_PROVIDER_DOMAIN_NAME = "0x Router"
```
jagodarybacka, line 36: hmmm I don't really like this const living here alone 😅 ideally I think we could include all internal providers' names in one object like
```suggestion
export const INTERNAL_PROVIDER_DOMAIN_NAME = {
  0X: "0x Router",
  UNISWAP: "🦄 Uniswap"
  // ...
}
```

Shadowfiend: Small terminology point (which is reflected in the new  variable name!)--a provider for the extension specifically refers to the classes we use to connect to on chain data see `SerialFallbackProvider` and the ethers providers).

Strong support for extracting to named constants, btw, though keep in mind the contract info is already in a named constant! Waiting for Vlad's take on how this should work, but my guess is I'm going to suggest a different way of approaching this that does not involve checking string constant equality.

kkosiorowska:
I changed PR to draft state. After the Vlad updates I will back to this.

Shadowfiend:
This work was more or less superseded by #2331 in the new signing flow, since internal swaps now go through a special enrichment and rendering pathway that no longer displays this icon.

PR #2308, opened 2022-09-27
Fix get active networks early returns

hyphenized:
Small fix, added a test too!

## To Test
- [ ] Add the pair of brackets back, test should break


background/services/chain/tests/index.unit.test.ts
```diff
@@ -79,4 +80,48 @@ describe("Chain Service", () => {
...
     })
   })
+
+  describe("getActiveNetworks", () => {
+    let chainService: ChainService
+
+    beforeEach(async () => {
+      sandbox.restore()
+      chainService = await createChainService()
+    })
```
0xDaedalus, line 90: Hmmmn we should really only need this once in this file.

I think what we want to do here is move lines 20-26 into the top-level "Chain Service" describe block.

0xDaedalus (commented):
💯 📈 Fantastic Catch.  

If you wouldn't mind quickly re-arranging the `beforeEach` I'm eager to get this guy merged.

hyphenized:
@0xDaedalus Updated the `markNetworkActivity` test to use sinon fake timers because it was flaky on my end (Maybe because of the Date stub 😖)

greg-nagy (approved):
this is a neat fix, let's 🚢

PR #2312, opened 2022-09-27
LOD-ing: Add data signing details to new signing flow

Shadowfiend:
> [In computer graphics, **level of detail** (**LOD**) refers to the complexity of a 3D model representation.[1][2][3] LOD can be decreased as the model moves away from the viewer or according to other metrics such as object importance, viewpoint-relative speed or position. LOD techniques increase the efficiency of rendering by decreasing the workload on graphics pipeline stages, usually vertex transformations. The reduced visual quality of the model is often unnoticed because of the small effect on object appearance when distant or moving fast.](https://en.wikipedia.org/wiki/Level_of_detail_(computer_graphics))
> 
> [Although most of the time LOD is applied to geometry detail only, the basic concept can be generalized. Recently, LOD techniques also included shader management to keep control of pixel complexity. A form of level of detail management has been applied to texture maps for years, under the name of mipmapping, also providing higher rendering quality.](https://en.wikipedia.org/wiki/Level_of_detail_(computer_graphics))
>
> [It is commonplace to say that "an object has been LOD-ed" when the object is simplified by the underlying LOD-ing algorithm as well as a 3D modeler manually creating LOD models.](https://en.wikipedia.org/wiki/Level_of_detail_(computer_graphics))

This PR adds details for data (typed and not) signing to the new signing flow. Once again we reuse existing components where reasonable, with a few small tweaks to allow for the new structure around them. The framing is set up so that both types of data signing have a common frame, in line with the design's consistency between these, and internal usage of components is improved a bit.

A few other small improvements specific to the new signing flow:
- Addresses are displayed using the `SharedAddress` component.
- The external link icon aligns with the design color-wise, and uses the
  `SharedIcon` component.
- Padding, dividers, and other spacing details have been aligned to the
  existing design, and generally look better.

<details>
  <summary>Screenshots!</summary>

  Sign In With Ethereum in the new signing flow
  Regular message signing in the new signing flow
  Typed Data signing in the new signing flow
  Sign In With Ethereum in the new signing flow with a Ledger account
</details>

## Testing

- [ ] All of this should be behind a feature flag. Ensure signing data is unchanged with the feature flag disabled.
  - [ ] Check Sign In With Ethereum signatures still look visually consistent. Can be checked at https://login.xyz/ under Get Started.
  - [ ] Check regular message signing still looks visually consistent. Can be checked at https://mintkudos.xyz/ on Polygon by hitting Connect Wallet.
  - [ ] Check typed data signing still looks visually consistent. Can be checked at https://app.uniswap.org/#/swap trying to swap from USDC using the Tally test wallet by clicking “Allow the Uniswap Protocol to use your USDC”. Do not confirm this signature so that it remains a viable test path for below.
- [ ] Enabling the feature flag, ensure that each type of data signing has its summary data displayed correctly.
  - [ ] Sign In With Ethereum. Can be checked at https://login.xyz/ under Get Started. Spacing should be tighter (see above).
  - [ ] Check regular message signing still looks visually consistent. Can be checked at https://mintkudos.xyz/ on Polygon by hitting Connect Wallet.
  - [ ] Try all of the above with a Ledger account and ensure the Ledger connection status appears above.


Shadowfiend:
Btw, note the Unknown Network for typed data signing—this seems to be an issue in the current flow as well, and I haven't dug into it yet.

ui/components/SignData/EIP4361Info.tsx
```diff
@@ -15,22 +15,30 @@ const CHAIN_NAMES: (chain: number) => string = (chain) => {
 }
 
 // this overides the type to expect EIP4361Data
-const EIP4361Info: React.FC<{ signingData: EIP4361Data }> = ({
-  signingData,
-}) => {
+const EIP4361Info: React.FC<{
```
0xDaedalus, line 18: Should we stick to the pattern of `(): React Element => ...` here rather than using `React.FC`?

ui/components/Signing/SignatureDetails/DataSignatureDetails/MessageDataSignatureDetails.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export default function MessageDataSignatureDetails({
+  messageRequest,
+}: MessageDataSignatureDetailsProps): ReactElement {
+  switch (messageRequest.messageType) {
+    case SignDataMessageType.EIP4361:
+      return (
+        <DataSignatureDetails
+          requestingSource={(messageRequest.signingData as EIP4361Data).domain}
```
0xDaedalus, line 23: Can we make `SignDataRequest` into a union type of two newly defined types `EIP191SignDataRequest` and `EIP4361SignDataRequest`?  I think that lets us avoid the `as` here which has traditionally been a core source of bugs / regressions for us.

ui/components/Signing/SignatureDetails/DataSignatureDetails/TypedDataSignatureDetails.tsx
```diff
@@ -0,0 +1,153 @@
...
+    return (
+      <div className="value">
+        <SharedAddress address={value} />
+        <SharedIcon
+          icon="new_tab@2x.png"
+          width={16}
+          height={16}
+          color="var(--green-40)"
+          hoverColor="var(--trophy-gold)"
+          ariaLabel="View Address on Etherscan"
```
0xDaedalus, line 21: I think this should be internationalized

ui/components/Signing/SignatureDetails/DataSignatureDetails/TypedDataSignatureDetails.tsx
```diff
@@ -0,0 +1,153 @@
...
+          onClick={() => {
+            window
+              .open(`https://etherscan.io/address/${value}`, "_blank")
```
0xDaedalus, line 24: Does this need to be network aware?  I think the answer is right now is no b/c signing typed data is itself somewhat network agnostic - but maybe we want to respect whatever network the user currently has selected in the wallet?

ui/components/Signing/SignatureDetails/DataSignatureDetails/TypedDataSignatureDetails.tsx
```diff
@@ -0,0 +1,153 @@
...
+              ?.focus()
+          }}
+        />
+        <style jsx>
+          {`
+            div {
+              display: flex;
+              column-gap: 5px;
+              align-items: center;
+
```
0xDaedalus, line 34:
```suggestion
```

ui/components/Signing/SignatureDetails/DataSignatureDetails/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  excludeTitle,
+  children,
+}: DataSignatureDetailsProps): ReactElement {
+  const { t } = useTranslation("translation", { keyPrefix: "signing" })
+
+  return (
+    <div className="primary_info_card standard_width">
+      <div className="sign_block">
+        <div className="container">
+          {excludeTitle ?? false ? (
```
0xDaedalus, line 22: might I suggest adding `excludeTitle = false` on line 13 and getting rid of `?? false` here?

0xDaedalus (changes requested):
Looks good - the only think I'd want to block on is reworking the types a bit to get rid of the `as` typecasting.

ui/components/SignData/EIP4361Info.tsx
Shadowfiend: This is existing code so I'm not touching it for now. 

And yes, it hurts me :)

ui/components/Signing/SignatureDetails/DataSignatureDetails/MessageDataSignatureDetails.tsx
Shadowfiend: This pattern is existing and baked a little deeper than I wanted to go into for unrelated work. Changing it would involve changing the existing flow more than I was comfortable doing in order to move fast on finishing this. 

Deciding not to do this was also painful 😅

Shadowfiend:
> the only think I'd want to block on is reworking the types a bit to get rid of the as typecasting

Mentioned this in direct replies, but it you basically hit on the refactoring I considered big enough and not on the critical path for this feature. If you really want to block, I can fix them, but they're already on the existing code path and I don't think we gain in terms of delivering this feature by reworking the type structure here. 

They are, however, definitely on my hit list 😁

ui/components/Signing/SignatureDetails/DataSignatureDetails/TypedDataSignatureDetails.tsx
Shadowfiend: … Man I strongly disagree here but we've ignored any sort of functional grouping of CSS style parameters across the board, and we don't have stylelint or similar prettier plugin in the mix, so… Fine.

Shadowfiend:
Fixed the other tidbits.

ui/components/Signing/SignatureDetails/DataSignatureDetails/MessageDataSignatureDetails.tsx
0xDaedalus: In that case how would you feel about me adding a commit to this branch that addresses this comment?

I really _really_ do not want to add anymore `n as x` to the codebase.

Alternatively we can create an issue to address this that blocks us flipping the feature flag on until it is resolved

Shadowfiend: Latter option with the blocking issue seems perfect.

0xDaedalus (approved):
👑

PR #2336, opened 2022-09-30
👻 Banish Ghosts: fix `to-16.ts` migration, to fix green screen of death

### LINKED ISSUE #2334: Fix broken migration for brave users
The `to-16.ts` migration fails for some brave users [at line 71](https://github.com/tallycash/extension/blob/9c9339753aaea7c8656d28bffcc05334148daba7/background/redux-slices/migrations/to-16.ts#L71) with the bellow error

CleanShot 2022-09-29 at 12 14 08

The attached activities slice caused this problem
[faulty activities slice.json.zip](https://github.com/tallycash/extension/files/9683369/faulty.activities.slice.json.zip)

To reproduce the problem:
- `npm i -g ts-node`
- extract [test.zip](https://github.com/tallycash/extension/files/9684473/test.zip) into `./background/redux-slices/migrations`
- cd `./background/redux-slices/migrations`
- comment out [these lines](https://github.com/tallycash/extension/blob/7100b1eaa6229ac0afa49cad1ff07a4bf580e783/background/redux-slices/migrations/to-16.ts#L71-L73)
- `ts-node test.ts`
  - ⚠️the error is different with this setup from the one attached and points to the wrong function call (the topmost `forEach` instead of the one with the `.entities`.)

greg-nagy:
# What 

An old migration didn't move the `ids` and `entities` from the address level into the `chainID` level but copied it. Because of this these keys are still in the state and there is is entities field inside the `ids` or `entities` array. This has caused migration issues for some folks on brave. The easy nondestructive fix is simply ignore these properties. Because the slice will be refactored soon it did not worth the risk to remove those keys from state.

# Test

⚠️ We don't have a way to reproduce the green screen issue. We just know the error message/line of code and the state of the environment where it happened. So these tests are only to verify that nothing is broken and generally the migrations
are safe. For more context see #2334 

- [x] clean up the dist folder `rm -rf ./dist/chrome/*`
- [x] checkout `v0.14.0` tag
- [x] `yarn build --config-name chrome`
- [x] Install the chrome folder in brave, import seed
- [x] check activity tab, wait for some activity to load
- [x] checkout this branch 
- [x] `yarn build`
- [x] reload the extension, verify the version number change
- [ ] check that you don't get green screen of death

Closes #2334

### DIFF
--- a/background/redux-slices/migrations/to-16.ts
+++ b/background/redux-slices/migrations/to-16.ts
@@ -68,6 +68,9 @@ export default (prevState: Record<string, unknown>): NewState => {
 
   Object.keys(typedPrevState.activities).forEach((address) => {
     Object.keys(typedPrevState.activities[address]).forEach((chainID) => {
+      if (chainID === "ids" || chainID === "entities") {
+        return
+      }
       Object.keys(typedPrevState.activities[address][chainID].entities).forEach(
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
         (_) => {


0xDaedalus:
Is there a pattern or a test we can introduce to our process of creating redux-migrations to avoid this specific problem in the future?

greg-nagy:
I would like to eliminate the redux cache entirely/almost entirely, which would solve the problem.

I would like to get us to a point where all the services emit an initialize event, and the redux store is populated from that data 
for the indexedDB, we can use the dexie migrations 

What would be extremely valuable is to have a way to test dexie migrations, but I don't see how to do that atm. Do you have a good approach in mind? @0xDaedalus 

(the crux of the problem is that the redux store assumed a shape for some reason that we had no idea that it can and we comforted ourselves by implementing types based on what we thought should be there.)

The only other approach I can think of is if we start using runtime type checks in conjunction with our types heavily.

PR #2338, opened 2022-09-30
Prevent background dapp connections from marking network activity.

0xDaedalus:
In our current implementation - a dapp connected to tally in an active or inactive tab will reset `lastUserNetworkActivity` each time it sends a request to the wallet.

This change updates that logic to only set `lastUserNetworkActivity` when `eth_requestAccounts` is received - which is indicative of a user actively navigating to a dapp.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1117,7 +1117,7 @@ export default class Main extends BaseService<never> {
     )
 
     this.providerBridgeService.emitter.on(
-      "permissionQueriedForChain",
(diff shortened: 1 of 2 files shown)


greg-nagy (approved):
`eth_requestChange` usually happens when a user navigates to a dapp, but not always. 

This change feels to be too restrictive, but I we should start err on the frugality side. 

Let's test this bad boy ✨

PR #2342, opened 2022-09-30
Clean up chain service db tests

0xDaedalus:
### To Test
- [ ] Tests should pass.

### DIFF
--- a/background/services/chain/db.ts
+++ b/background/services/chain/db.ts
@@ -270,8 +270,8 @@ export class ChainDatabase extends Dexie {
       .toArray()
     return lookups.reduce(
       (newestBlock: bigint | null, lookup) =>
-        newestBlock === null || lookup.startBlock > newestBlock
-          ? lookup.startBlock
+        newestBlock === null || lookup.endBlock > newestBlock
+          ? lookup.endBlock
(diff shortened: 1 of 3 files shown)


background/services/chain/db.ts
```diff
@@ -270,8 +270,8 @@ export class ChainDatabase extends Dexie {
       .toArray()
     return lookups.reduce(
       (newestBlock: bigint | null, lookup) =>
-        newestBlock === null || lookup.startBlock > newestBlock
-          ? lookup.startBlock
+        newestBlock === null || lookup.endBlock > newestBlock
```
mhluongo, line 273: Looks like you fixed a bug... should we be manually testing this somehow in PR review?

mhluongo (approved):
Looks good to me, but unclear if we need additional manual QA

background/services/chain/db.ts
0xDaedalus: Unfortunately not possible to test as the bugfix in question is in a piece of dead code

jagodarybacka: if this is a dead code then why we are keeping it here?

PR #2345, opened 2022-09-30
Union Jack: Rework SignDataRequest as MessageSigningRequest discriminated union

### LINKED ISSUE #2332: Refactor `SignDataRequest` type to be a union type of `EIP191SignDataRequest` and `EIP4361SignDataRequest`
_Originally posted by @0xDaedalus in https://github.com/tallycash/extension/pull/2312#discussion_r981507184_

Shadowfiend:
> [The **Union Jack**, or **Union Flag**, is the de facto national flag of the United Kingdom. Although no law has been passed making the Union Jack the official national flag of the United Kingdom, it has effectively become such through precedent. It is sometimes asserted that the term *Union Jack* properly refers only to naval usage, but this assertion was dismissed by the Flag Institute in 2013 following historical investigations.](https://en.wikipedia.org/wiki/Union_Jack)

`SignDataRequest` was an envelope container for (personal, non-typed-data) message signing requests that were further subdivided by type, and different message types (such as SIWE messages) had more specific parsed data types in a `signingData` field. `SignDataMessageType` was an integer enum that was used for message types.

The complicating factor with this approach was that trying to narrow the types in consuming code was impossible, as the types were represented without a way to type-guard the parsed message data. The result was a need to check against `messageType` and the `SignDataMessageType` enum, followed by an ostensibly unguarded `as` cast for the parsed data in `signingData` itself.

The new approach uses a `MessageSigningRequest` envelope type that is specialized by a type parameter, and that type parameter is a discriminated union of possible signing data types called `MessageSigningData`. The envelope type is structured the exact same way as previously, but the typing allows for the `messageType` to be used as a discriminant that guards the `signingData` type as well, removing the need for an `as` cast and requiring instantiating code to adhere to one of the discriminated type/data pairs rather than allowing for instantiation of mixed types.

One other notable change here is moving away from enums, which are a less favored TypeScript feature, to a simple discriminated union type of string constants for the different message types.

## Testing

- [x] Check Sign In With Ethereum signatures still works as expected. Can be checked at https://login.xyz/ under Get Started.
- [x] Check regular message signing still works as expected. Can be checked at https://mintkudos.xyz/ on Polygon by hitting Connect Wallet.

Closes #2332.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -99,7 +99,7 @@ import {
   signDataRequest,
 } from "./redux-slices/signing"
 
-import { SignTypedDataRequest, SignDataRequest } from "./utils/signing"
+import { SignTypedDataRequest, MessageSigningRequest } from "./utils/signing"
 import {
   emitter as earnSliceEmitter,
   setVaultsAsStale,
@@ -1040,7 +1040,7 @@ export default class Main extends BaseService<never> {
         resolver,
         rejecter,
       }: {
-        payload: SignDataRequest
+        payload: MessageSigningRequest
         resolver: (result: string | PromiseLike<string>) => void
         rejecter: () => void
       }) => {
--- a/background/redux-slices/signing.ts
+++ b/background/redux-slices/signing.ts
@@ -1,11 +1,6 @@
 import { createSelector, createSlice } from "@reduxjs/toolkit"
 import Emittery from "emittery"
-import {
-  ExpectedSigningData,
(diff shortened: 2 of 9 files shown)


0xDaedalus (approved):
📈 

Found a separate issue that exists on main while testing.

Ticketed [here](https://github.com/tallycash/extension/issues/2361).

PR #2349, opened 2022-09-30
Trackmania - separate active, tracked, and current networks.

0xDaedalus:
Since we are moving towards a new definition of what an "active" network is as shown in https://github.com/tallycash/extension/pull/2301, https://github.com/tallycash/extension/pull/2318, https://github.com/tallycash/extension/pull/2338 its important we update our nomenclature in the codebase as well.

This PR attempts to split up the idea of an Supported Network / Active Network / Tracked Network / Current Network and codify that semantic splitting through method and variable names.  

The loose definitions are:
- Supported Network: A network that the extension is able to track
- Tracked Network: A network that the extension is tracking or has ever tracked.
- Active Network: A network that the extension is actively tracking / listening to.
- Current Network: The network that a given dapp (or the internal wallet) is connected to.

I suspect once our work to limit JSON-RPC usage the distinction between Supported Network and Tracked Network will disappear (in fact this could be an indicator of the success of that effort) but we are not there quite yet.

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -538,7 +538,7 @@ export default class Main extends BaseService<never> {
       setNewSelectedAccount({
         address: accounts[0].address,
         network:
-          await this.internalEthereumProviderService.getActiveOrDefaultNetwork(
+          await this.internalEthereumProviderService.getCurrentOrDefaultNetworkForOrigin(
             TALLY_INTERNAL_ORIGIN
           ),
       })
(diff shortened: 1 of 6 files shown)


greg-nagy (approved):
Will be interesting to see how this evolves.

The code looks good! The conflicts need to be resolved, than I think it's good to go.

jagodarybacka:
I would love to include the definitions in the codebase, in one place so this is the knowledge that won't be forgotten 😄

jagodarybacka (approved):
ty for adding it to the readme :)

PR #2351, opened 2022-09-30
Know Thy Networks: Properly resolve account totals across networks, fix sign typed data unknown network

Shadowfiend:
This makes a pretty deep plumbing change to allow components to look up the account total for an arbitrary network, rather than only being able to return account totals for the current network.

This removes the dependency the signing header had on redux selectors, to instead display the network for the account whose signature is being requested as passed from its calling component.

## Testing

- [ ] Switch between networks and make sure the account balances displayed in the accounts selector update to follow the network for each account.
- [ ] On a couple of networks, attempt to sign a transaction (any dApp), messages ([SIWE](https://login.xyz/#tryitout) on mainnet and [Kudos](https://mintkudos.xyz/) on Polygon), and signed data ([Uniswap](https://app.uniswap.org/#/swap) USDC swap source, click “Allow the Uniswap protocol to use your USDC”, mainnet only) and make sure that the network at top left is correct.


Shadowfiend:
Rebased here as #2380 landed separately with a more specific fix to the issue; adjusted both the commit message and the PR description to reflect the fact that it's not fixing the original bug but instead just the architectural issue.

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -333,7 +333,70 @@ const getTotalBalance = (
     .reduce((total, assetBalance) => total + assetBalance, 0)
 }
 
-export const selectCurrentNetworkAccountTotalsByCategory = createSelector(
+function getNetworkAccountTotalsByCategory(
```
jagodarybacka, line 336: I think we can came up with a better name as this function returns more than the account total, it returns account details so maybe `getAccountDetailsByCategory`?

Shadowfiend: It actually returns specifically account totals as defined in the type `AccountTotal`. We can absolutely rename that type and all its selectors but… I feel like that's good for a separate PR.

I do think we're using “account details” to mean something else somewhere else. Another approach here would be to adjust how we resolve some of this information—some of it may not need a selector at all, and we may be able to simplify the `AccountTotal` type further.

jagodarybacka (approved):
I think it's ready to go, network switcher works fine, I couldn't find a way to get unexpected network on the Signing screen.

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -333,7 +333,70 @@ const getTotalBalance = (
...
+          address,
+          network,
+          shortenedAddress,
+          accountType,
+          keyringId,
+          accountSigner,
+        }
+      }
+
+      return {
```
greg-nagy, line 371: This level of complexity reminds me of good ol' sql queries. 

Which makes me a bit worried. This kind of complexity ties everything together at runtime. The reason this gives me the chills is that the code space from where a bug/unexpected change can pop up feels wide. What could be a simpler flow to represent this data? 

The output data doesn't seem to be big. Would it make sense to create a derived state and update it on events?

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -333,7 +333,70 @@ const getTotalBalance = (
     .reduce((total, assetBalance) => total + assetBalance, 0)
 }
 
-export const selectCurrentNetworkAccountTotalsByCategory = createSelector(
+function getNetworkAccountTotalsByCategory(
+  accounts: AccountState,
```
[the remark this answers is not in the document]
hyphenized: I agree, also we could skip the 🍛 if we do something like this right?
```ts
function getNetworkAccountTotalsByCategory(
  state: RootState,
  network: EVMNetwork
): CategorizedAccountTotals {
  const accounts = getAccountState(state)
  const assets = getAssetsState(state)
  const accountSignersByAddress = selectAccountSignersByAddress(state)
  const keyringsByAddresses = selectKeyringsByAddresses(state)
  const sourcesByAddress = selectSourcesByAddress(state)
  const mainCurrencySymbol = selectMainCurrencySymbol(state)
```

PR #2359, opened 2022-10-03
Add price impact in-wallet swaps

### LINKED ISSUE #1863: Add price impact & warning for in-wallet swaps
### Discord Discussion Link

na

Add a price impact indicator for in-wallet swaps, and a warning in case it's high.

Example from Matcha:

Screen Shot 2022-07-31 at 08 27 28

kkosiorowska:
Closes #1863 

This PR adds a price impact indicator for in-wallet swaps.

**Changes**

- add a new state for swap about price details
- add prices and price impact to input component
- do a state migration

**UI**

Before
Screenshot 2022-10-03 at 15 30 43

After
Screenshot 2022-10-04 at 14 48 24

Screenshot 2022-10-06 at 08 40 24

Screenshot 2022-10-06 at 13 54 43

Screenshot 2022-10-06 at 17 15 35]

Screenshot 2022-10-06 at 13 37 51

**To test**
- [ ] go to swap and check price impact indicator


VladUXUI:
Looking nice from screens! Will give it a go tomorrow morning

ui/_locales/en/messages.json
```diff
@@ -37,6 +37,10 @@
     "error": {
       "invalidAmount": "Invalid amount",
       "insufficientBalance": "Insufficient balance"
+    },
+    "tooltip": {
```
Shadowfiend, line 41: Naming this `priceImpactTooltip` will probably save us some headaches in the future if we decide to add other tooltips.

Shadowfiend (commented):
Quick architectural note: we should be generating the issue around price impact in redux, probably in the async thunk when we get the swap results, rather than in components. The rule of thumb here is "if we're making decisions about data, try to do it in redux".  The component can then focus on presenting the error if available. 

Amongst other things, this will enable us to easily add the same price impact admonition to the confirmation page.

VladUXUI:
Tried it out and Ui wise it's looking good. There are some UX issues i didn't realise when doing this.

- [ ] We should only show price impact with Red colour if change is bigger than -5%
- [ ] Show price impact in Yellow if change is between 1%-5%
- [ ] Show price impact in Green-40 if change is under 1%
- [ ] For total asset price, if total is less than $0,00, let's show it as <$0.00 (so it looks less like an error)

If these are too much work, i can ticket separately

kkosiorowska:
> 

I agree. Shadowfiend suggestions are more complicated. I will back to you when I finish updates.

kkosiorowska:
I have taken your suggestions into account. 
Please review again : @VladUXUI , @Shadowfiend , @VladUXUI

kkosiorowska:
> Hmm, idk, `+0.05$` is kinda good deal 😂 it should be green right? and it's `+5%` not `+0.05%`. Also this tooltip is misleading I think if the price impact is surplus image

I fixed issue with the price impact calculation. 

@VladUXUI Do you have any suggestion about tooltip ?

VladUXUI:
Yes, let's not show "(%##)" if difference is 0+

kkosiorowska:
@VladUXUI and @jagodarybacka please check it again.

jagodarybacka:
Would be nice to remove the price impact if the input has changed

https://user-images.githubusercontent.com/20949277/194523122-fefa432d-9108-4c0b-835c-3872491f7b51.mov

jagodarybacka (approved):
Code looks nice and clean 👑  I think current behaviour is fine but I would like @VladUXUI to see it first :)

jagodarybacka:
@greg-nagy I think it is fine to merge but do we want to feature flag it for any reason?

background/redux-slices/0x-swap.ts
```diff
@@ -492,3 +502,8 @@ export const selectInProgressApprovalContract = createSelector(
   (state: { swap: SwapState }) => state.swap.inProgressApprovalContract,
   (approvalInProgress) => approvalInProgress
 )
+
+export const selectPriceDetails = createSelector(
```
greg-nagy, line 506: Non-blocker: If you prepare this selector to `priceDetails` being `undefined` you probably can avoid the migration.

Migrations are always scarry and we should be wary when we feel the need for them.

greg-nagy (approved):
this looks neat 🥳

background/redux-slices/0x-swap.ts
kkosiorowska: Sure, thanks for the tip.

PR #2365, opened 2022-10-03
Rework remaining account-related browser alarms to only interact with active accounts.

0xDaedalus:
This PR is the final _large_ step we can take towards reducing our background json-rpc calls (there definitely still quite a few  optimizations we can make going forward).  It introduces a new piece of state to chain service - `lastUserActivityOnAddress` - which is then used in combination with `lastUserActivityOnNetwork` inside of `getAccountsToTrack` to optionally fetch only active accounts from the chain service db.

Here are some key points of the changes this PR introduces
- [Account transfers & balances will immediately be loaded when reactivating an address](https://github.com/tallycash/extension/blob/shut-the-gates/background/services/chain/index.ts#L1053:L1054)
- [Token and base balances will be refreshed for all accounts at application startup](https://github.com/tallycash/extension/blob/shut-the-gates/background/services/indexing/index.ts#L125:L131)
- [Changing networks now mark account activity](https://github.com/tallycash/extension/blob/shut-the-gates/background/services/internal-ethereum-provider/index.ts#L258:L261)
- [Changing accounts via the wallet will mark account activity](https://github.com/tallycash/extension/blob/shut-the-gates/background/main.ts#L1230:L1230)
- [Opening a dapp will mark account activity](https://github.com/tallycash/extension/blob/shut-the-gates/background/main.ts#L1126:L1126)

There is also an important tradeoff here that can be addressed in a followup PR:  After a period of address / network inactivity - users may have to wait up to 1 minute until the next scheduled token balance refresh since there is no straightforward way to call [retrieveTokenBalances](https://github.com/tallycash/extension/blob/shut-the-gates/background/services/indexing/index.ts#L458:L458) from chainService.  Rather than patching a hack on-top of the existing code in this PR this is something we should handle down the line because:

- It will be uncommon for users' balances to change during periods where they are not using the wallet.
- This tradeoff gives us massive gains in terms of how many calls we are making to alchemy.

### To Test
- [x] Add a new account to your wallet
  - [x] Base assets and token balances should populate as normal, activities should populate as normal (after a short delay) 
  - [x] Switch to different networks - same thing should be true
- [x] Close chrome and re-open it - do NOT open the extension
  - [x] Observe dev tools for a few minutes - you should not see background calls to alchemy happening
  - [x] Open the extension - you should see calls going out on the connection of the chain you have selected in the wallet
  - [x] Wait 5 minutes with the extension closed (you may want to set NETWORK_POLLING_TIMEOUT to a smaller number to wait less) and observe background calls to alchemy ceasing after that time.
- [x] Close chrome and re-open it again - do NOT open the extension
  - [x] Navigate to a dapp (e.g. uniswap) - upon connecting you should see queries for base balances and asset transfers on the appropriate network.


background/services/provider-bridge/index.ts
```diff
@@ -128,18 +129,18 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
+    const network =
       await this.internalEthereumProviderService.getCurrentOrDefaultNetworkForOrigin(
         origin
       )
 
     if (event.request.method === "eth_requestAccounts") {
       // This is analogous to "User opened a dapp on chain X"
-      this.emitter.emit("dappOpenedOnChain", chainID)
+      const { address } = await this.preferenceService.getSelectedAccount()
+      this.emitter.emit("dappOpened", { address, network })
```
greg-nagy, line 140: Not every dapp will start the communication flow with `eth_requestAccounts`. 
There are quite a few out there that fires an `eth_accounts` first and if that fails/returns `[]` then it fires `eth_requestAccounts`

greg-nagy (approved):
code looks good, works as described 

let's 🚢

PR #2366, opened 2022-10-03
Poll for gas every 15 seconds while network is active.

0xDaedalus:
Although our development builds properly query gas prices every 15 seconds - [production bundles don't support alarms with periods of under 1 minute](https://developer.chrome.com/docs/extensions/reference/alarms/#method-create).

This PR introduces a change to the `blockPrices` alarm that lets us properly query for gas every 15 seconds both in dev and prod (while still having those queries gated by user network activity).

### To Test
- [ ] `yarn build`
- [ ] Load up the application
- [ ] You should not see the `Alarm period is less than minimum of 1 minutes` message in the console
- [ ] You should see prices being queried every 15 seconds.


background/services/chain/index.ts
```diff
@@ -1041,6 +1045,18 @@ export default class ChainService extends BaseService<Events> {
...
   async pollBlockPrices(): Promise<void> {
+    // Schedule next N polls at even interval
+    for (let i = 1; i < GAS_POLLS_PER_PERIOD; i += 1) {
+      setTimeout(async () => {
+        await Promise.allSettled(
+          this.subscribedNetworks.map(async ({ network }) =>
+            this.pollBlockPricesForNetwork(network.chainID)
+          )
+        )
+      }, (GAS_POLLING_PERIOD / GAS_POLLS_PER_PERIOD) * 1e3 /* convert to ms */ * i)
```
hyphenized, line 1056: Excellent idea! This will run in intervals of 250 ms though, I think we're missing a conversion here 👀

greg-nagy: This has the potential of becoming a feature instead of a bug :) 

With the per address activity tracking this has the potential to become `if (chain has active address) then let's poll for block prices with chain dependent period time`

This would minimize the resource we waste but provide up to date gas prices for transactions.

The absolute best would be to do this only when the user is using gas price info. But that's a far fetched dream for now.

greg-nagy (changes requested):
Awesome idea! Let's fix the timing and it's good to go

(We will probably need tp refactor this `setTimeout` to be user action dependent for manifest v3)

background/services/chain/index.ts
0xDaedalus: Hmmmn yeah I don't think we can do sub 15s polling until we implement your last point - but I also think the dream is not as far fetched as you may think 😄

greg-nagy (approved):

PR #2381, opened 2022-10-05
Fix token list test

jagodarybacka:
Fix broken test that was failing on GH builds. 
This version is not time-dependent as this is using emitted events instead of `wait`.


greg-nagy (approved):
let's fix this annoying blocker! 

ty @jagodarybacka, this is a lifesaver

PR #2382, opened 2022-10-05
😢 But why?: remove nonce if received from dapp

### LINKED ISSUE #2370: Arbitrum: 1 inch approval fails
Tried to swap from usdc to eth but didn't work because Approval kept failing, same with DAI or other assets.
Approval worked on Uniswap.

To test:
https://app.1inch.io/#/42161/unified/swap/USDT/ETH

Error was fail to broadcast

greg-nagy:
1inch sends nonce from mainnet when approving
assets on arbitrum and this causes the broadcast
to fail

we lookup the nonce during the signing process
so it's safe to ignore this parameter from the dapp

Closes #2370

### DIFF
--- a/background/services/internal-ethereum-provider/index.ts
+++ b/background/services/internal-ethereum-provider/index.ts
@@ -213,7 +213,11 @@ export default class InternalEthereumProviderService extends BaseService<Events>
       }
       case "eth_sendTransaction":
         return this.signTransaction(
-          params[0] as JsonRpcTransactionRequest,
+          {
+            ...(params[0] as JsonRpcTransactionRequest),
+            // we populate the nonce during the signing process. If we receive one, it's safer to just ignore it.
+            nonce: undefined,
+          },
           origin
         ).then(async (signed) => {
           await this.chainService.broadcastSignedTransaction(signed)


Shadowfiend:
I'm wary of doing this globally just to fix one dapp. I would almost consider fixing this on the dApp end of the bridge on a dApp by dApp basis, similar to how we handle Metamask impersonation. 

The alternative is if this is how other wallets behave already across the board. If that's the case I think ignoring the dApp-specified nonce is "fine enough"--though as a dApp dev if I tried to set nonces explicitly for the user for some reason I would be pretty furious if the wallet ignored me heh. We should also note that in the PR description.

greg-nagy:
The reason I chose this solution first and foremost is that it felt conceptually off for a dapp to specify nonce for whatever reason. This concern felt clearly to be the responsibility of the wallet and impossible to handle properly from a dapp.  Eg a dApp can't know how long a user is waiting before clicking the sign button, and other nonces can happen in the meantime — a highly unlikely scenario for sure. 

Anyhow MM works like this, haven't tested with other wallets though. 

This extension dev thing would be so much fun without the dapps ... :D

0xDaedalus (approved):
I share @Shadowfiend's wariness - but as it seems this mimics the current behavior of metamask - I guess its good enough

PR #2384, opened 2022-10-05
Posthog / UUID Integration

zuuring:
This pull request adds a new analytics system powered by Posthog which checks if a UUID is stored in browser from Tally website. If analytics is enabled by user, Posthog events will fire.

Any posthog event can be created by simply

`import { posthogEvent } from "./background/lib/posthog"`

and using the function `posthogEvent("Put your event title here");`


background/services/analytics/posthog.ts
```diff
@@ -0,0 +1,83 @@
+import {v4 as uuidv4} from 'uuid';
+import toggleCollectAnalytics from '../../../ui/pages/Settings/SettingsAnalytics'
+
+function getCookie(name: string)
```
greg-nagy, line 4: We have a solution in the extension that solves the same problem but without the need for cookies. I will dig it up for reference in a bit.

greg-nagy (changes requested):
Wow, I am excited to see this land ✨🤩

Left some initial thoughts after a quick glance.

package.json
```diff
@@ -60,6 +60,7 @@
     "@tallyho/window-provider": "0.0.1",
     "buffer": "^6.0.3",
     "react": "^17.0.2",
+    "uuid": "^9.0.0",
```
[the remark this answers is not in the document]
zuuring: Thanks for your comments! Using this lib to easily be able to generate UUIDs on the website. However I think it can be totally removed from the extension as it's only needed on the website facing portion of the implementation :)

Shadowfiend:
Oho it is time! Got a few notes from looking through, but excited to see this take shape :)

Few big setup things:
- Please make sure you're using yarn for package management. This PR commits a `package-lock.json` file—if you can, please update that commit (https://github.com/tallycash/extension/blob/244c2a092f60e19ae6edf52979185dbff735c4fd/package-lock.json) via rebase to drop the file entirely from history.
- Please make sure you've set up pre-commit hooks (e.g. https://github.com/tallycash/extension/pull/2384/files#diff-bb2ba232569016c28a3e1afeb301ce20420881034e715b0c87f88c6fd5db410cR1293 was probably not hit by the linter).

Instructions for both of these are in the README, under [Quickstart](https://github.com/tallycash/extension#quickstart) and [installing pre-commit hooks](https://github.com/tallycash/extension#installing-pre-commit-hooks).

Some additional notes:
- I suggest keeping this PR as a draft until it's ready to merge (I'm assuming that isn't the case now since the uninstall URL is pointed to localhost:8000, but could be wrong!). Another thing that would be good to figure out before considering this ready to merge is how we're going to handle the API key on dev builds vs production builds, if we want to try to differentiate—as well as whether we should accept a blank key for local builds/the public default value, or if we should instead have a specific key that's flagged accordingly in Posthog.
- I would suggest putting the analytics behind a feature flag for now so it can be developed across multiple PRs and tested easily by the team, see below.
- When the demo is satisfactory and it's time to move to merging stuff, I suggest splitting this up into multiple PRs, in particular one introducing the feature flag + service stuff around analytics, and separate PR(s) to add actual tracking.
- `service/analytics` is probably not appropriate if we're not actually setting up an analytics service class. I think the correct architectural spot to put this is to funnel analytics via redux async thunk to the telemetry service, and have the telemetry service handle calls to Posthog—but that doesn't really need to be a blocker. We may also probably want to push stuff into Posthog from e.g. the internal Ethereum provider in certain cases.

background/lib/posthog.ts
```diff
@@ -0,0 +1,82 @@
...
+        Accept: "application/json",
+      },
+    })
+
+    if (!response.ok) {
+      throw new Error(`Error! status: ${response.status}`)
+    }
+    const result = (await response.json()) as HogResponse
+    // eslint-disable-next-line no-console
+    console.log("data: ", JSON.stringify(result, null, 4))
```
0xDaedalus, line 73: I assume this is here for debugging purposes - we'll probably want to remove this

greg-nagy:
> I think the correct architectural spot to put this is to funnel analytics via redux async thunk to the telemetry service, and have the telemetry service handle calls to Posthog—but that doesn't really need to be a blocker. We may also probably want to push stuff into Posthog from e.g. the internal Ethereum provider in certain cases.

The original idea as a first step was to put the trigger into the react components, because we wanted to measure weekly active users, so the analytics data can be accurate and not be distorted by failing redux migrations or anything that happens between the UI and the background. Otoh adding wallet and removing event could be put anywhere else.

manifest/manifest.json
```diff
@@ -32,7 +32,7 @@
     "default_title": "Tally Ho",
     "default_popup": "popup.html"
   },
-  "permissions": ["alarms", "storage", "unlimitedStorage", "activeTab"],
+  "permissions": ["alarms", "storage", "unlimitedStorage", "activeTab", "cookies", "storage", "<all_urls>"],
```
0xDaedalus, line 35: duplicate `storage` permission here

package.json
```diff
@@ -78,6 +78,7 @@
     "@types/react-router-dom": "^5.3.1",
     "@types/remote-redux-devtools": "^0.5.5",
     "@types/terser-webpack-plugin": "^5.0.3",
+    "@types/uuid": "^8.3.4",
```
0xDaedalus, line 81: I believe this is left-over from when the `uuid` library was in use.

background/lib/posthog.ts
greg-nagy: We should not have `console.log`s in the code at all, but I think this will be fixed with turning on the precommit hook.

zuuring: Removed all console logging for debugging!

background/lib/posthog.ts
```diff
@@ -0,0 +1,82 @@
...
+    const response = await fetch("https://app.posthog.com/capture/", {
+      method: "POST",
+      body: JSON.stringify({
+        // this is a safe public write only api key
+        // roll this key for demo
+        api_key: POSTHOG_API_KEY,
+        event: posthogEvent,
+        properties: {
+          distinct_id: userID,
+          data:  "This adds posthog events to Tally extension",
```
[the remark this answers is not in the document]
zuuring: For the `data:` entry? I was actually wondering the same! Would likely be useful for future posthog logging events. :)

background/lib/posthog.ts
```diff
@@ -0,0 +1,82 @@
...
+       console.log("UUID: ",cookie.value);
+       createEvent(eventName, cookie.value)
+     }
+     else {
+       console.log('No UUID found!');
+     }
+ });
+}
+
+export async function createEvent(eventName:string, userID:string): Promise<HogResponse> {
```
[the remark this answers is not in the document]
zuuring: Thanks for your comments on this! Going to relay to @puppycodes as we spoke about the possibility of setting up a posthog instance which will also remove our reliance on firing the event based on the analytics service.

zuuring:
> Oho it is time! Got a few notes from looking through, but excited to see this take shape :)
> 
> Few big setup things:
> 
> * Please make sure you're using yarn for package management. This PR commits a `package-lock.json` file—if you can, please update that commit (https://github.com/tallycash/extension/blob/244c2a092f60e19ae6edf52979185dbff735c4fd/package-lock.json) via rebase to drop the file entirely from history.
> * Please make sure you've set up pre-commit hooks (e.g. https://github.com/tallycash/extension/pull/2384/files#diff-bb2ba232569016c28a3e1afeb301ce20420881034e715b0c87f88c6fd5db410cR1293 was probably not hit by the linter).
> 
> Instructions for both of these are in the README, under [Quickstart](https://github.com/tallycash/extension#quickstart) and [installing pre-commit hooks](https://github.com/tallycash/extension#installing-pre-commit-hooks).
> 
> Some additional notes:
> 
> * I suggest keeping this PR as a draft until it's ready to merge (I'm assuming that isn't the case now since the uninstall URL is pointed to localhost:8000, but could be wrong!). Another thing that would be good to figure out before considering this ready to merge is how we're going to handle the API key on dev builds vs production builds, if we want to try to differentiate—as well as whether we should accept a blank key for local builds/the public default value, or if we should instead have a specific key that's flagged accordingly in Posthog.
> * I would suggest putting the analytics behind a feature flag for now so it can be developed across multiple PRs and tested easily by the team, see below.
> * When the demo is satisfactory and it's time to move to merging stuff, I suggest splitting this up into multiple PRs, in particular one introducing the feature flag + service stuff around analytics, and separate PR(s) to add actual tracking.
> * `service/analytics` is probably not appropriate if we're not actually setting up an analytics service class. I think the correct architectural spot to put this is to funnel analytics via redux async thunk to the telemetry service, and have the telemetry service handle calls to Posthog—but that doesn't really need to be a blocker. We may also probably want to push stuff into Posthog from e.g. the internal Ethereum provider in certain cases.
> 
> EDIT to add: the `yarn.lock` updates also look weird. Not sure what that's from, but probably worth checking out the lockfile from before adding the uuid types dependency, then `yarn install`ing at the project root again to make sure the changes are good.

Thank you so much for the extensive feedback on this! I have now pushed a few new commits which resolve all the errors and pass pre-commit checks. Which involved some refactoring. Now onto feature flags!

zuuring:
I have just pushed a new commit which adds Uninstall browser hook event, so there is no need to forward to a webpage at the moment.

Shadowfiend:
> The original idea as a first step was to put the trigger into the react components, because we wanted to measure weekly active users, so the analytics data can be accurate and not be distorted by failing redux migrations or anything that happens between the UI and the background.

I'm not sure what else could happen between the UI and background, but it seems like being given warning of a failing redux migration by our analytics would be a good thing, not a bad one. I don't think it's the end of the world to start with these direct calls, but I do think the correct *end state* is to move to funnel all of them through redux actions.

IMO we can find reasons to make exceptions for almost anything to work outside of the core data flow we have established through redux, but then we end up with 10 ways for data to actually get processed and explode complexity—that's what I'm pushing to avoid here. To do that we need to avoid skipping that data flow unless there are specific, compelling, and fairly critical reasons to do so (think user safety-level stuff). If this were some sort of security signal that we needed to make it out at all costs, I think it would make sense, but in this context I would argue the reasons don't rise to that level.

Another thing to note here is that we can attach events to clicks (e.g. the wallet added event) or to actual actions (e.g. when a new keyring is actually added to the background, when a new Ledger is added, etc). If we're handling those events in two different ways, it becomes more difficult to find what events are kicked off where, it reduces our ability to handle the events locally for additional purposes. One quick example of where we might want to process locally is to send information about whether a flow was finished without pushing identifying information to the server, vs letting the server correlate using a unique id that it's aware of.

background/lib/posthog.ts
```diff
@@ -0,0 +1,71 @@
+export const { POSTHOG_API_KEY } = process.env.POSTHOG_API_KEY
```
greg-nagy, line 1: You don't need to export this variable, and here you are trying to destructure a string that will fail.

manifest/manifest.json
```diff
@@ -32,7 +32,15 @@
     "default_title": "Tally Ho",
     "default_popup": "popup.html"
   },
-  "permissions": ["alarms", "storage", "unlimitedStorage", "activeTab"],
+  "permissions": [
+    "alarms",
+    "management",
+    "unlimitedStorage",
+    "activeTab",
+    "cookies",
```
greg-nagy, line 40: I couldn't make this work with cookies. I always got `Unchecked runtime.lastError: No host permissions for cookies at url: "https://tally.cash/".` 

We might need to test this a bit more.

greg-nagy (changes requested):
the PR does not work in it's current form 

let's move this back to draft and go through it as part of your onboarding. There is much here to unfold and it's not efficient to do this in PR reviews.

Shadowfiend:
We ended up pulling work from this into a few other places, and Posthog is now integrated into the wallet behind user opt-in, etc. Closing out!

PR #2390, opened 2022-10-05
Set networks to be `inactive` at extension initialization.

0xDaedalus:
In order to avoid querying for assetTransfers (by far our most expensive RPC query) and gas prices on all networks at application startup - this PR initializes all networks as inactive and only queries for transfers when a network becomes active.

The tradeoff for our gain in efficiency here is that a user will need to be active on a given network (use a dapp, open the extension) in order for those network's transfers to appear in the extension.

Since it is impossible to open the extension directly on the "Activity" tab - and since Activities are only something we care about in the wallet (dapps generally don't use this information) the tradeoff seems minimal.


greg-nagy (approved):
works well, and the thoughts on the description are just first class 🎓

let's 🚢

PR #2392, opened 2022-10-06
Prevent calls to CoinGecko from tests

hyphenized:
~Changed `fetchJson` to `fetch` so other tests don't accidentally send external requests.~
Added a jest mock for `fetchJson` so that it uses fetch if it's ever called within tests.
Also removes usage of `setTimeout` to check assertions in the indexing service tests due to it's inconsistent behaviour

## To test:
- [ ] Change `COINGECKO_API_ROOT` to a local server or use something like `webhook.site`
- [ ] Run tests, no requests should appear


background/lib/prices.ts
```diff
@@ -21,7 +20,7 @@ export async function getPrice(
 ): Promise<number | null> {
   const url = `${COINGECKO_API_ROOT}/simple/price?ids=${coingeckoCoinId}&vs_currencies=${currencySymbol}&include_last_updated_at=true`
 
-  const json = await fetchJson(url)
+  const json = await fetch(url).then((resp) => resp.json())
```
jagodarybacka, line 23: thank you for finding that. I'm not sure what was the reason behind using `fetchJson` in the first place 🤔

jagodarybacka (approved):
Please fix the conflict and I think we are good to go 🚀

background/lib/prices.ts
Shadowfiend: `fetchJson` does a lot more than fetch, including retries, timeouts, and handling rate limit responses (within reason). I believe the intent was to move *more* things to using it, not the other way around. 

`fetchJson` does use `fetch` under the covers in the browser, but not in node. What's the specific goal of switching away from it though? Is it eliminating retries so we don't end up with floating `setTimeout`s after tests complete?

greg-nagy (changes requested):
The realization and the learning that we have here is invaluable!

But this direction is not the one we should take. 

Blocking this pr to prevent confusion and will provide more guidance on the mocking that we should probably use. (Just need to collect resources first)

background/lib/prices.ts
mhluongo: Showing up late to this, but here's [a little more context](https://github.com/tallycash/extension/issues/2132) cc @jagodarybacka → we actually want to standardize on `fetchJson` and friends.

Your solve sounds right to me @hyphenized but I'll leave it to our existing reviewers to confirm.

mhluongo:
> I've also noticed that if a test involving the chain service runs for long enough, it also attempts to establish WebSocket connections (and fails the test).

Heh, adding tests after the fact is fun :grin: Glad you're leaning into this stuff

background/lib/prices.ts
hyphenized: @Shadowfiend Mocking http/https requires to explicitly call `jest.mock("https") in every test file that might use it because it is a build-in module, I've tried to reassign it but I couldn't get it to work. Maybe I'm not doing it the right way or in the right place 🤔

greg-nagy: > Mocking http/https requires to explicitly call `jest.mock("https") in every test file that might use it because it is a build-in module

To my best knowledge, this is the expected behavior

https://jestjs.io/docs/manual-mocks#mocking-node-modules

```
If we want to mock Node's build-in modules (e.g.: fs or path), 
then explicitly calling e.g. jest.mock('path') is required, 
because build-in modules are not mocked by default.
```

Mocking `fetchJson` seems to be the good balance here and doing so in the `__mocks__` so we don't have to worry about it in the tests. I think we can repeat this process for websockets as well and then we can be sure that we have covered all our outgoing requests.

greg-nagy (changes requested):
The way we should take this is to have a single mock in the `__mocks__` library which prevents any outgoing requests. 

In these mock setup sinon should be used for mocking. 

And as a final step we should clean up our test files from any unnecessary mocking. (`window.fetc` and the in place jest mock)

For the future: 

In a later step we can do the same dance to mock out the ethers providers. After a quick glance if we mock `JsonRpcProvider` and the `send` method we should be in a good spot. But we might need to mock in a different spot. This requires some more research.

background/services/indexing/tests/index.integration.test.ts
```diff
@@ -11,7 +11,19 @@ import ChainService from "../../chain"
 import PreferenceService from "../../preferences"
 import { getOrCreateDb as getIndexingDB } from "../db"
 
-const wait = (ms: number) => new Promise<void>((r) => setTimeout(r, ms))
+jest.mock("@ethersproject/web", () => {
```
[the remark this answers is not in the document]
hyphenized: I've made a few changes, let me know what you think 👀

I fear mixing sinon and jest in the manual mocks could cause some weird bugs but it seems to be working well so far

background/services/name/resolvers/rns.ts
```diff
@@ -8,29 +8,32 @@ import logger from "../../../lib/logger"
...
 import { normalizeEVMAddress } from "../../../lib/utils"
 
-const rskNetworkProvider = new JsonRpcProvider("https://public-node.rsk.co")
+export default function rnsResolver(): NameResolver<"RNS"> {
+  const rskNetworkProvider = new JsonRpcProvider("https://public-node.rsk.co")
 
-// REF: https://developers.rsk.co/rif/rns/architecture/registry/
-const RNS_REGISTRY_ADDRESS = "0xcb868aeabd31e2b66f74e9a55cf064abb31a4ad5"
+  // REF: https://developers.rsk.co/rif/rns/architecture/registry/
+  const RNS_REGISTRY_ADDRESS = "0xcb868aeabd31e2b66f74e9a55cf064abb31a4ad5"
```
hyphenized, line 15: Moved this inside the fn because it was causing a connection on loading `services/index.ts`

background/services/indexing/tests/index.integration.test.ts
```diff
@@ -11,7 +12,16 @@ import ChainService from "../../chain"
 import PreferenceService from "../../preferences"
 import { getOrCreateDb as getIndexingDB } from "../db"
 
-const wait = (ms: number) => new Promise<void>((r) => setTimeout(r, ms))
+const fetchJsonStub: SinonStub<
```
greg-nagy, line 15: I would like to have a solution to avoid this, but sadly jest mocking is not something that typescript is aware of :(

greg-nagy (approved):
thank you for this work, I feel a lot better with these approach laid out 💪

background/services/indexing/tests/index.integration.test.ts
hyphenized: A workaround was to import `fetchJson` directly from the mock, but required setting an esm export directive in the manual mock, I thought that seemed a bit confusing/hacky so I left this way

Been thinking maybe we could benefit from setting up some test specific overrides for typescript/etc if we moved tests files to a folder under the package root directory, what do you think? 👀

setupJest.ts
```diff
@@ -48,3 +48,12 @@ browser.tabs.getCurrent = jest.fn(() =>
 Object.defineProperty(Dexie.dependencies, "indexedDB", {
   get: () => indexedDB,
 })
+
+// Stub fetch calls
+Object.defineProperty(window, "fetch", {
+  writable: true,
+  value: (url: string) => {
+    // eslint-disable-next-line no-console
+    console.warn("Uncaught fetch call to: \n", url)
```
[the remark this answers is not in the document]
hyphenized: You're right, I tried to but I kept fighting with the linter's [import/no-extraneous-dependencies](https://github.com/import-js/eslint-plugin-import/blob/v2.25.4/docs/rules/no-extraneous-dependencies.md) rule. Moving the `sinon` dependency to the top level `package.json` caused the error to appear in tests files instead, and, although mocking the global fetch looked way cleaner when used within tests it's ultimately more prone to bugs. 

I noticed this was the case because some function calls (e.g. starting the service) could leave pending timers/async processing that ends up running after the test case ends, causing unexpected behaviour in other tests if these foreign calls interacted in some way with the global mock (I suppose this is also one of the reasons why globals are bad).

Using jest fake timers could prevent this by forcing all pending timers and async tasks to run but I've had issues with it in the past when using some libraries like: `swr` and `react-testing-library` (specially with the latter's helper function `waitFor`) 😕 

I think one way we could prevent this is to inject/introduce some kind of fetching service/helper that we can stub and be sure will be a unique instance between tests. If I had written the tests in a different way maybe I could've prevented these issues too, but, live and learn, I guess.

PR #2396, opened 2022-10-06
Fix non-deterministic indexing service test.

0xDaedalus:
This should fix a test that has been inconsistently passing and failing in CI.  See the comment for a detailed explanation..


background/services/indexing/tests/index.integration.test.ts
```diff
@@ -148,11 +148,12 @@ describe("IndexingService", () => {
...
-          indexingService
-            .getCachedAssets(ETHEREUM)
-            .map((assets) => assets.symbol)
-        ).toEqual(["ETH"])
+        // The order in which assets are emitted is non-deterministic
+        // since the `emit` function gets called as part of an unawaited
+        // series of promises (trackedNetworks.forEach in "internalStartService")
+        // Since we expect two asset emissions and we don't know which will
+        // be emitted first - we make our test assertions after the second
+        // emission in the event handler below this one.
```
Shadowfiend, line 156: Should we just have a simple length assertion here? Or is it also nondeterministic if the list will already have two?

Shadowfiend (commented):
Good sleuthing!!

background/services/indexing/tests/index.integration.test.ts
0xDaedalus: Hmmmnnn thats a good question - my guess is that its not deterministic simply because `emit` and `once` both return promises so it is technically possible that more than 1 asset gets persisted into indexedDB in between the time that the first asset persisted and the "assets" event is captured here.

jagodarybacka (approved):
This is a good solution IMO at least for now 👍

PR #2397, opened 2022-10-07
Implement global error screen

### LINKED ISSUE #1734: Implement global error screen (to prevent future green screen of death)
https://github.com/tallycash/extension/pull/1672#issuecomment-1169660695

> Suspecting this is prone to an infinite reload loop

I am using a `setTiimeout()` once, when the UI loads and don't restart it.

Otoh I was thinking about a more robust solution where this check is running continuously. 
With some considerations, eg the restart has been done 2 times then let's show an error screen on 3rd. 

This is something that's called watchdog in electric engineering and is used as a fail-safe in case 
anything goes off.

How would you feel about a solution like this? @henryboldi @Shadowfiend 

This could be the error screen, and the button would reload the whole extension like the refresh button.

CleanShot 2022-06-29 at 10 00 59

kkosiorowska:
Closes #1734 

This PR implements global error screen and fixes issue with an infinite reload loop. After running the function three times, the error component is rendered.

**Changes**
- remove an  infinite reload loop
- add a global error screen
- change reload logic 

**UI**
Screenshot 2022-10-07 at 12 14 52

**To test**
Apply the following patch:

```
diff --git a/ui/index.ts b/ui/index.ts
index a02a01aa..7907a351 100644
--- a/ui/index.ts
+++ b/ui/index.ts
@@ -32,7 +32,13 @@ export async function attachTabUIToRootElement(): Promise<void> {
 }
 
 export async function attachPopupUIToRootElement(): Promise<void> {
-  const store = await newProxyStore()
+  const storeType = await newProxyStore()
+
+  // Test 1
+  throw new Error("Forced error")
+
+  // Test 2
+  const store: typeof storeType = await new Promise(() => {})
 
   if (SUPPORT_TABBED_ONBOARDING) {
     const state = store.getState()

```

- [ ] check that the error component is visible after third reload


jagodarybacka:
Can you add testing steps - how to get this screen?

greg-nagy:
@VladUXUI this screen can be persistent for some. Should we add some instruction on how to get help? Eg join discord #bug-report channel etc?

VladUXUI:
The messaging and UX here need some love. 
Awesome suggestion Greg.
Will mark as blocked and get a quick design going tomorrow morning

VladUXUI:
I created a new ticket for this. Let's not block this with "new" things.

Good to merge from my point of view

kkosiorowska:
I have improved solutions and added patch file for reproduce issue. @greg-nagy, @jagodarybacka  Please make review.

src/popup.ts
```diff
@@ -1,11 +1,43 @@
...
+  "back_forward",
+  "prerender",
+}
 
-  if (!isAppRendered) {
-    window.location.reload()
+setTimeout(() => {
+  const state = window.history?.state || {}
+  let reloadCount = state?.reloadCount || 0
+  const navigationItem = performance.getEntriesByType(
```
jagodarybacka, line 17: This piece of code is not looking "clear" to me, can we add a comment on what is going on and why?

jagodarybacka: Oh I'm so sorry, I meant whole setTimeout here needs a comment, not this line 😂

src/popup.ts
```diff
@@ -1,11 +1,47 @@
...
+  "prerender",
+}
+
+// Prevents from the green screen. The solution checks if the first top-level element is rendered.
+// If this does not happen then reload the UI thread.  To prevent an infinity loop
+// the restart will be done max 3 times then a global error screen will be shown.
 
-  if (!isAppRendered) {
-    window.location.reload()
+setTimeout(() => {
```
greg-nagy, line 18: Why did you choose this solution instead of the original one? 

The only concern I had with that solution is the structure of the checks. I would prefer to check if the page has rendered before showing the error screen.

```
const timerId = setInterval(() => {
  counter += 1
  if (counter > 2) {
    clearInterval(timerId)
    attachUIToRootElement(GlobalError)
  } else {
    const isAppRendered = !!document.getElementsByClassName(
      "top_menu_wrap_decoy"
    ).length
    if (isAppRendered) {
      clearInterval(timerId)
    }
  }
}, 1000)

```

greg-nagy: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/type - this is a nice find though!

greg-nagy (approved):
This is indeed what I had in mind. I have another tweak in mind for this screen, which is using the `refreshBackgroundPage` as well, but that is certainly not in the scope of this PR. 

This is a really good improvement as is 🤩

Would like to understand the reason behind the code for the `setTimeout`, because I felt it hard to follow and understand. Otoh the code works, so the PR is good.

src/popup.ts
kkosiorowska: This solution isn't work well. I only noticed it after I have created the PR. `setInterval` will execute three times  but it doesn't refresh the page or call the method `attachPopupUIToRootElement`. This isn't what we had in mind.

greg-nagy: Just for the record — I do not expect the code to be reverted — it did work well for me. `setInterval` ran 3 times and attached the error screen on the 3rd run.

PR #2401, opened 2022-10-07
Fix bugged Ledger use on Optimism and Arbitrum

mhluongo:
... introduced via 7b7bbd3aaee8dfb5f4

# To Test

- [ ] Swap on Optimism with a Ledger.
- [ ] Swap on Arbitrum with a Ledger.

If both work, we're gucci!

### DIFF
--- a/background/services/ledger/index.ts
+++ b/background/services/ledger/index.ts
@@ -424,7 +424,7 @@ export default class LedgerService extends BaseService<Events> {
           throw new Error("Transaction doesn't appear to have been signed.")
         }
 
-        if (!isKnownTxType(tx.type)) {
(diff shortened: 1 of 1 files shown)


0xDaedalus (approved):

Shadowfiend:
https://github.com/tallycash/extension/commit/7b7bbd3aaee8dfb5f433998da7302cfed16cae24#diff-60319598208ed00f8064275b33b1795ee9a6d72502e628f583f992a81cbbd6faR190 is also incorrect in that commit. Legacy transactions are definitionally not EIP1559. Few other tidbits there that probably could use tightening up as well. Transaction types are gnarly!

PR #2412, opened 2022-10-12
Make feature flags switchable at runtime

### LINKED ISSUE #2300: Shadowy Super Doggos Panel: Make feature flags switchable at runtime
# Why?

So non dev folks (product, design, QA) can easily review any merged PR without the need for QA builds. 

# How?
- update feature flags to read the value from the localstorage in `features.ts`
- create a panel which is hidden by default but can be activated by clicking N times on something
  - HiddenDevPanel was used for this before (removed in https://github.com/tallycash/extension/pull/2247)
  - ideal would be a new panel that opens from the settings page and clicking 5-10 times the version string
    - make sure the cursor change doesn't reveal this feature
- add toggles for feature flags to the devpanel
- (optional) explore what else would make sense here 
  - see https://github.com/tallycash/extension/pull/2247#issuecomment-1251757837
  - triggering lock/unlock behavior

kkosiorowska:
Closes #2300 

This PR makes feature flags switchable at runtime.

**Changes**
- add a devpanel
- make the version label clickable
- read the feature flag values from the localstorage
- add toggles for feature flags to the devpanel

**UI**
Screenshot 2022-10-12 at 13 25 51

Screenshot 2022-10-18 at 13 09 53

https://user-images.githubusercontent.com/23117945/196415604-cedf811e-9f6a-4fa7-9aa3-2d5dcbe2ebc6.mov

https://user-images.githubusercontent.com/23117945/196418146-419a832c-9f67-49b4-adcf-f909734a3853.mov

**To Test**
Set up `SWITCH_RUNTIME_FLAGS` to `true`. Go to the settings view and click label with version 15 times.
- [ ] check that dev panel is visible
- [ ] hover over the name of the feature flag check that extra information is visible  
- [ ] check that only the runtime flags are visible in the dev panel
- [ ] check if BalanceReloader is working. Use for this `SUPPORT_ARBITRUM` flag (second video) 
- [ ] check if  feature flags are removed on extension startup. Switch `SWITCH_RUNTIME_FLAGS` to `false`, refresh background script and check Local Storage
- [ ] check that feature flags are switchable at runtime


mhluongo (changes requested):
There are a number of feature flags that this isn't appropriate for... I feel like that feedback keeps getting lost?

How are we planning to handle any feature flags that impact onboarding, for example? What about anything that changes the service startup lifecycle, or anything that has a strong performance impact?

I'd suggest we break feature flags into two kinds — those that are best flipped at build time (eg those I just listed) versus those that are fine to be flipped at runtime.

This approach also means we have no way to "pull back" a feature that has issues in a new release. For example, in our current feature flag setup, if the Arbitrum integration had a huge bug — we could just set it to false and ship a new release. In this setup, it'll require writing more code to pull anything back that we've initially flagged for release.

mhluongo:
I know this is a draft and you're just getting caught in product / engineering cross-fire here @kkosiorowska, but I want to make sure we have a discussion about this before the PR gets too much further. The idea and UX are solid, but there needs to be a release process plan and a little more architectural thought before we land something like this.

kkosiorowska:
> I know this is a draft and you're just getting caught in product / engineering cross-fire here @kkosiorowska, but I want to make sure we have a discussion about this before the PR gets too much further. The idea and UX are solid, but there needs to be a release process plan and a little more architectural thought before we land something like this.

You're absolutely right. Breaking feature flags into two kinds sound good.  I will think about release process plan for this.

background/features.ts
```diff
@@ -1,22 +1,21 @@
...
+  USE_BALANCE_RELOADER,
+  SUPPORT_MULTIPLE_LANGUAGES,
+  SUPPORT_ANALYTICS,
+  SUPPORT_TABBED_ONBOARDING,
+  SUPPORT_KEYRING_LOCKING,
+  SUPPORT_FORGOT_PASSWORD,
+  ENABLE_ACHIEVEMENTS_TAB,
+}
+
+export const isEnabled = (type: FeatureFlagTypes): boolean =>
```
hyphenized, line 19: I think maybe this could work if we add checks for `process.env.NODE_ENV === "production"` and return build time env vars and if that's not the case, local storage values. The unused code will get removed in production.

For QA to be able to use this with a production build we could also have another env var in the production check. Ideally this should happen in a `staging` environment but I don't know if that's easily doable with webpack.

Another thing that I think could improve this would be to add a way to reset background data / reload the background script and popup afterwards.

hyphenized (commented):
Left some (hopefully helpful) thoughts!

background/features.ts
kkosiorowska: > I think maybe this could work if we add checks for `process.env.NODE_ENV === "production"` and return build time env vars and if that's not the case, local storage values. The unused code will get removed in production.

I think that checking process.env.NODE_ENV === "production" is a very good idea. 

> For QA to be able to use this with a production build we could also have another env var in the production check. Ideally this should happen in a `staging` environment but I don't know if that's easily doable with webpack.
> 
> Another thing that I think could improve this would be to add a way to reset background data / reload the background script and popup afterwards.

I'll see what we can do about it.

jagodarybacka:
I think we should give them more "descriptive" labels for the UI because sometimes I forgot what they are about so will our users.

Also we should make sure to not put any feature flag here what will cause the migration - because there is no going back then

greg-nagy:
OOO ... so many awesome thoughts! Thank you all ❤️

Let me try to synthesize them and give a unified direction to this PR

A bit of background: The most important purpose for this PR is for us to get rid of the QA PRs. The vision for the next step here is to have this on in our builds from main/PR merges and turn it off in tagged releases. In the future, I would like this to be released to the public as well, but that's a non-concern for now.

- descriptive names from @jagodarybacka: this is an awesome suggestion, def wanted to do that with translations as well. Can be done in this PR but also ok to be in the following PR 
- feature flags with migrations from @jagodarybacka We can include them here for QA because we won't ship this for wide audience. We will need to think about this when we will ship this to the users. 
- separate the feature flags into runtime and build time by @mhluongo: this is definitely a good idea, and we should do it 
- please make sure that this feature can be turned off with a build time only feature flag
- please ensure that this is an extension of the existing build time feature flags and not a replacement. I did not communicate this part in the issue, sorry for that! (Implementation direction in the task list)
- this PR should have the above mentioned build time feature flag turned off. We will follow up with CI modification to turn off this flag for tagged builds. If that's in place, we can switch this to enabled, and our product, QA and design frenz will be able to use it easily 
- `process.env.NODE_ENV === "production"` idea by @hyphenized makes a lot of sense and would be a good direction, but I have an extended vision for this feature, and it would not work with that, so please skip that setup @kkosiorowska 
  - the reason for this is `yarn build` sets the `process.env.NODE_ENV` to `production`, and we would like to QA production releases, not only dev builds
- `staging` environment I think it would be best to skip this as long as we reasonably can. I find great value in trunk-based development and would like to keep the QA builds as close as possible to `main`
- background reload comment by @hyphenized we definitely should have the ability to reload the background. We have the `USE_BALANCE_RELOADER` feature flag and the `BalanceReloader` component with the `refreshBackgroundPage` async thunk. This should be in this panel, so we can restart the extension if needed for the feature flag to work. This should be a good next step, and we can iterate from there. 

I hope this clears up the doubts and concerns around this PR. If there are concerns with this approach please jump in with you next set of TLC :) 

So the task list:
- [ ] (nice to have) please add descriptive names to the feature flags with translations
- [ ] Please separate the feature flags into build and runtime categories 
- [ ] Please refactor the implementation so the runtime flags augment the build time feature flags and do not replace them. 
  - if the main build time only feature flag is on, THEN it reads the build time flags but also checks for the values in the localstorage. The value in localstorage overwrite the runtime flag.
  - the feature flag list should be cleaned up from the localstorage on extension start
    - it should be added only if the build time flag is on
    - if there was a previous value set and the build time flag is on, then it should keep the previous settings
- [ ] Please add a build time only feature flag for this feature and set it to be turned off
- [ ] Please add the `BalanceReloader` component to this screen so the background can be reloaded on feature flag change.

Shadowfiend:
> we would like to QA production releases, not only dev builds

Hmmm… When would a production release QA build require runtime feature flag toggling though? By definition a production release already has the correct flags enabled and disabled, right?

greg-nagy:
>> we would like to QA production releases, not only dev builds
> Hmmm… When would a production release QA build require runtime feature flag toggling though? By definition a production release already has the correct flags enabled and disabled, right?

Yes and no. :) 

Concrete situation: The definition of production release is ambiguous here. The context I wrote the quoted sentence was in terms of `process.env.NODE_ENV`. We would like that to be set to `production` on builds that we want to QA. Eg, When we ask others to QA, we don't want to worry about the performance hit of the extra redux middlewares. 

The long-term goal here is to create a new definition that will be the staging asset. We don't want dev/staging branches, but we want to give our non-dev frenz an easy way to QA. 

The process should go like this: 
- The staging asset should be built from main on PR merge 
- (ideally,  but not necessarily) using the production api keys 
- it would have the runtime feature flag switch enabled
- The staging asset would be uploaded to a static page and a `latest` would be created on every upload, which could be accessible through a well-known url. (The nondev team members could bookmark that url.) 

When the time comes for a QA the process would look like this: 
-  dev team says `Hey folks, please QA the keyring unlock`. 
- MH decides who should be involved in this round of QA 
- Everybody involves downloads the asset > installes the extension > switches the feature flag > reload > test the feature and ticket the findings

The problems this would solve (all of them happened more than once and caused us headaches)
- "Which asset should I test?" 
- "Ohh shoot, wrong asset?" 
- "Where can I find the asset?"
- "Hang on I need to create a QA PR"
- "This works in QA. Shoot this is broken with the prod api keys" 
- "How do I build the extension?" / "Which branch should I QA?"

kkosiorowska:
@greg-nagy  @jagodarybacka @mhluongo @Shadowfiend @0xDaedalus @hyphenized 

Thank you all for your wonderful comments. :heart:
I have updated the solution according to Greg's suggestions. Please look again. Important points:

- Note the break feature flags
- I added texts for feature flags in messages file. Please if you have any suggestions to make them more descriptive let me know. It was quite problematic for me.

src/background.ts
```diff
@@ -1,11 +1,30 @@
...
+    isEnabled(FeatureFlagTypes.SUPPORT_TABBED_ONBOARDING)
+  ) {
     const url = browser.runtime.getURL("tab.html#onboarding")
     browser.tabs.create({ url })
   }
+
+  if (
+    obj.reason === "update" &&
+    !isEnabled(FeatureFlagTypes.SWITCH_RUNTIME_FLAGS) &&
+    !!Object.keys(RuntimeFlag).find(
```
greg-nagy, line 20: We can remove this check from here. If the feature flag is not enabled it's ok to clean up the localstorage. Running this won't be expensive and the event already limits when this code will be run.

ui/components/HiddenDevPanel/FeatureFlagsPanel.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+function FeatureFlag(props: { name: string }): ReactElement {
+  const { name } = props
+  const { t } = useTranslation("translation", {
+    keyPrefix: "devPanel.featureFlags.flags",
+  })
+  const [value, setValue] = useState(isEnabled(FeatureFlagTypes[name]))
+
+  useEffect(() => {
+    setLocalStorageItem(name, String(value))
```
greg-nagy, line 21: Hm, so we initilize the runtime feature flags in localStorage when the user visits this page for the first time. 

image

greg-nagy (commented):
This is a wonderful PR and a lot of thought and care went into it. 🤩

There a few minor changes that would be good to be fixed.
None of them felt like a blocker but it worth another update before merging.

background/features.ts
```diff
@@ -1,22 +1,54 @@
...
+  ENABLE_ACHIEVEMENTS_TAB: process.env.ENABLE_ACHIEVEMENTS_TAB === "true",
+  HIDE_TOKEN_FEATURES: process.env.HIDE_TOKEN_FEATURES === "true",
+  SUPPORT_ARBITRUM: process.env.SUPPORT_ARBITRUM === "true",
+  SUPPORT_GOERLI: process.env.SUPPORT_GOERLI === "true",
+}
+
+/**
+ * Object with all feature flags. The key is the same as the value.
+ */
+export const FeatureFlagTypes: Record<string, string> = Object.keys({
```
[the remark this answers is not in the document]
kkosiorowska: Good point!

ui/pages/Settings.tsx
```diff
@@ -9,18 +9,61 @@ import {
...
+const NUMBER_OF_CLICKS_FOR_DEV_PANEL = 15
+
+function VersionLabel(): ReactElement {
+  const { t } = useTranslation()
+  const history = useHistory()
+  const [clickCounter, setClickCounter] = useState(0)
+  const [isHover, setIsHover] = useState(false)
+
+  useEffect(() => {
+    if (clickCounter === NUMBER_OF_CLICKS_FOR_DEV_PANEL && isHover) {
```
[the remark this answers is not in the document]
kkosiorowska: I will change it but in the feature, if we add more functions we will change it.

background/features.ts
```diff
@@ -1,22 +1,54 @@
...
+ * If value is not exist then is read from environment variables.
+ * The value for the build time flag is read from environment variables.
+ */
+export const isEnabled = (flagName: string): boolean => {
+  if (flagName in BuildTimeFlag) {
+    return BuildTimeFlag[flagName]
+  }
+  if (BuildTimeFlag.SWITCH_RUNTIME_FLAGS) {
+    const state = localStorage.getItem(flagName)
+    return state ? state === "true" : RuntimeFlag[flagName]
```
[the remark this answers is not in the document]
greg-nagy: eh, you are correct, which renders my comment mostly invalid.

greg-nagy (approved):
one step closer for testflight 🚀

mhluongo:
> A bit of background: The most important purpose for this PR is for us to get rid of the QA PRs.

@greg-nagy looks like I've missed my window, but to be clear: all that needed was some GitHub action magic to publish flagged builds per PR, plus an idea of a nightly or nightly-per-flag build. We've taken a dev process thing and fixed it by making the product less stable.

Next time, I recommend writing an RFC to focus on the underlying issue.

mhluongo:
What this does begin to enable is A/B testing feature flags and other OTA updates... but it also highlights the issues I continue to have with this approach. What about feature flag migrations? Etc.

PR #2416, opened 2022-10-12
RSK Integration

ahsan-javaiid:
## Description
- This pull request adds `RSK` integration into tally wallet. 
- User will be able to see balance, send and receive `RBTC` over rsk network
- ERC20 token (RIF) send / Receive also working

## Demo Video on Testnet

https://user-images.githubusercontent.com/104683677/194956528-f3edc460-1bba-46da-9097-d078137bbf8b.mp4

## Demo on Mainnet
Integration is tested on mainnet and below is the wallet view with actual funds.
Screenshot 2022-10-12 at 10 12 21 PM

## Type of Change
- [X] New feature

## Testing information
- Activate `SUPPORT_RSK` flag in .env file to enable `Rsk` integration
- Activate `HIDE_IMPORT_DERIVATION_PATH` UI to select the `Rsk` derivation path or use default derivation path like metamask
- Demo video attached 
- Tested on rsk mainnet 

## Checklist
- [X] Git hooks enabled
- [X] No lint errors
- [X] Git commit is signed


ahsan-javaiid:
cc: @alepc253

background/lib/nfts.ts
```diff
@@ -49,6 +49,7 @@ const SIMPLE_HASH_CHAIN_TO_ID = {
   ethereum: 1,
   optimism: 10,
   polygon: 137,
+  rsk: 30,
```
mhluongo, line 52: Does SimpleHash support NFTs on RSK?

background/redux-slices/0x-swap.ts
```diff
@@ -134,6 +135,7 @@ export const SWAP_FEE = 0.005
 
 const chainIdTo0xApiBase: { [chainID: string]: string | undefined } = {
   [ETHEREUM.chainID]: "api.0x.org",
+  [RSK.chainID]: "rootstock.api.0x.org",
```
mhluongo, line 138: The 0x API doesn't support RSK, we need to fall back here

background/services/chain/asset-data-helper.ts
```diff
@@ -48,6 +48,28 @@ export default class AssetDataHelper {
     }
 
     try {
+      // Using rpc provider for RSK instead of AlchemyProvider
+      if (
+        addressOnNetwork.network.chainID === RSK.chainID &&
+        smartContractAddresses
+      ) {
+        const balances = smartContractAddresses.map(async (token) => {
```
mhluongo, line 56: @0xDaedalus has an alternative here that should be quite a bit cleaner

background/services/chain/db.ts
```diff
@@ -106,6 +107,24 @@ export class ChainDatabase extends Dexie {
...
     })
 
+    if (SUPPORT_RSK) {
+      this.version(5).upgrade((tx) => {
+        tx.table("accountsToTrack")
+          .toArray()
+          .then((accounts) => {
+            const addresses = new Set<string>()
+
+            accounts.forEach(({ address }) => addresses.add(address))
```
mhluongo, line 117: I believe we've moved to a model where we don't add `accountsToTrack` until a user interacts with a network... @0xDaedalus to confirm

ui/_locales/en/messages.json
```diff
@@ -194,6 +194,18 @@
...
       "learnMore": "Learn More"
     },
+    "rsk": {
+      "title": "Network fees",
+      "header": "Gas fees on RSK work differently from Ethereum.",
+      "minGasPrice": "Minimum Gas Price",
+      "multiplier": "10%",
+      "percentage": "percentage",
+      "estimatedGas": "Estimated Gas",
+      "explainerOne": "The estimated gas cost for RSK transactions is calculated by adding 10% to Minimum Gas Price obtained from web3.eth.getBlock('latest').minimumGasPrice",
```
mhluongo, line 204: We should change up this copy... Tally Ho doesn't use `web3` for one, but mostly we need to make this user-facing.

Maybe something like

`"The estimated gas price for RSK transactions is calculated by adding 10% to the Minimum Gas Price obtained from the latest mined block."`

ui/components/TopMenu/TopMenuProtocolList.tsx
```diff
@@ -45,6 +47,19 @@ const productionNetworks = [
...
         },
       ]),
+  ...(SUPPORT_RSK
+    ? [
+        {
+          network: RSK,
+          info: i18n.t("protocol.rsk"),
+        },
+      ]
+    : [
```
mhluongo, line 57: This should be an empty array `[]`

mhluongo (changes requested):
Excellent start! I've left some initial feedback, and expect more from @0xDaedalus

ui/components/NetworkFees/FeeSettingsText.tsx
```diff
@@ -84,9 +85,10 @@ const estimateGweiAmount = (options: {
       transactionData.estimatedRollupGwei
   }
 
+  const decimalLength = RSK.chainID === network.chainID ? 2 : 0
```
Shadowfiend, line 88: Could you explain this particular one a little? Feels like a place we could generalize instead of special-case, but I want to understand where the special case is coming from.

ui/components/NetworkFees/FeeSettingsText.tsx
```diff
@@ -106,6 +108,11 @@ export default function FeeSettingsText({
   networkSettings = customNetworkSetting ?? networkSettings
   const baseFeePerGas =
     useBackgroundSelector((state) => {
+      if (transactionData && transactionData.chainID === RSK.chainID) {
```
Shadowfiend, line 111: This is another place we probably need to approach a little differently so we're not special-casing by chain id, and so we can avoid an `as`. Will dig in a little more hopefully tomorrow to see if there's a good suggestion to be had, unless someone else beats me to it 😁

Shadowfiend (commented):
Yay, delighted to see this!

background/services/chain/asset-data-helper.ts
mhluongo: Here's [a more efficient approach](https://github.com/tallycash/extension/pull/2423) that should be merged soon!

background/lib/nfts.ts
ahsan-javaiid: Looked into supported chains on https://simplehash.com/. Rsk is not there.
Reference: https://simplehash.readme.io/reference/supported-chains-testnets
Just removed the rsk from SIMPLE_HASH_CHAIN_TO_ID
Thanks for pointing it out.

background/redux-slices/0x-swap.ts
ahsan-javaiid: https://www.0x.org/ does not support Rsk. Updated Rsk swap endpoint to empty value. That means swap screen will be disabled for Rsk and the user cannot perform any action. 

I think it can be more generalized and we can show some warning or error message if a chain is not supported by 0x.org. 

Open to discuss what will be the good place and design for that or any other thoughts are most welcome.

ui/_locales/en/messages.json
ahsan-javaiid: Done!

ui/components/TopMenu/TopMenuProtocolList.tsx
ahsan-javaiid: Done!

ui/components/NetworkFees/FeeSettingsText.tsx
ahsan-javaiid: Mostly in case of L2 solutions gas prices are small so that's why need to consider decimal places. The `truncateDecimalAmount` function is truncating the decimal values. 

Here is Dry run of this scenario:
Consider estimatedSpendPerGas value = 0x387ee40 (59240000) and convert it into Gwei which is 0.05924

Now if we call truncateDecimalAmount with 0 decimal length it will discard all digits after decimal and the resulting value will be zero. 

I think we should consider a minimum of 2 or 3 decimal places or maybe more for smaller gas price values like 0.005924.  Any other thoughts are most welcome.

background/services/chain/db.ts
0xDaedalus: Confirmed - this migration should not be needed and a given account will be added to the db once the RSK network is selected via the in-wallet network switcher.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -117,7 +117,7 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
   // Functions that will create and initialize a new provider, in priority
   // order.
   private providerCreators: [
-    () => WebSocketProvider,
+    () => WebSocketProvider | JsonRpcProvider, // RSK does not support websockets yet so pass JsonRpcProvider as fallback
```
0xDaedalus, line 120: You can remove the comment here since this is functionality we want anyway - coincidentally I've made the same change in [this PR ](https://github.com/tallycash/extension/pull/2423/files#diff-92d7bac2ae34382b1d878f777db598f91ccbc12d4b6309739a8f5b956df7d80aR120)

background/services/chain/serial-fallback-provider.ts
```diff
@@ -158,7 +158,7 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
     // Internal network type useful for helper calls, but not exposed to avoid
     // clashing with Ethers's own `network` stuff.
     private evmNetwork: EVMNetwork,
-    firstProviderCreator: () => WebSocketProvider,
+    firstProviderCreator: () => WebSocketProvider | JsonRpcProvider, // RSK does not support websockets yet so pass JsonRpcProvider as fallback
```
0xDaedalus, line 161: As above - no need for comment.

ahsan-javaiid: Done!

background/services/chain/db.ts
ahsan-javaiid: Done!

background/services/chain/asset-data-helper.ts
```diff
@@ -48,6 +48,28 @@ export default class AssetDataHelper {
...
+          const balance = await getBalance(
+            provider,
+            token,
+            addressOnNetwork.address
+          )
+          return {
+            smartContract: {
+              contractAddress: token,
+              homeNetwork: FORK,
```
jagodarybacka, line 65: I don't think `homeNetwork: FORK` is correct here 🤔

jagodarybacka (commented):
Wow this PR is amazing, great job! 😍

background/services/chain/asset-data-helper.ts
ahsan-javaiid: Good catch. Thanks for pointing it out. 
Fixed! 👍

ui/components/NetworkFees/FeeSettingsText.tsx
Shadowfiend: Got it! I'm wondering if the right approach here is heuristic rather than hardcoded, in that case; I think we could reuse some of the logic in [`heuristicDesiredDecimalsForUnitPrice`](https://github.com/tallycash/extension/blob/c0bde96bf854eb3c64094911619eb89fba8365c6/background/redux-slices/utils/asset-utils.ts#L210-L250) for this, though it may be defining a similar function with completely independent logic.

We could also heuristically decide between gwei and wei instead—though if users expect to always see gwei I think we should stick to heuristic decimal decisions.

ui/components/NetworkFees/FeeSettingsText.tsx
Shadowfiend: Ok, so I think the ideal way forward here is actually going to do some internal specialization of the entire component, unfortunately 😅

In particular, there are three paths that are currently happening inside `FeeSettingsText` right now:
- EIP1559 transaction (Ethereum, Polygon).
- Legacy transaction with rollup fee (Optimism).
- Legacy transaction without rollup fee (Arbitrum).

These are all looked up and then some logic decides which one to show, and the flow of all that logic isn't really clear. A better approach would probably be to resolve up front which kind of transaction/network pair this is, and have one or two unexported components that handle the logic for displaying that type of transaction/network.

*Additionally*, we're doing a lot of actual math in this component, which should instead happen elsewhere (likely a redux selector or even directly in a redux slice/action) so the component can focus on looking up data.

-----

With that said, RSK seems like it should be using the same code path as Arbitrum, which doesn't reuse the `baseFeePerGas`. I think what we're seeing here is that maybe Arbitrum's summary fee settings text is incorrect in the first place, because the `estimateGweiAmount` function only reads legacy gas information for [rollups](https://github.com/tallycash/extension/blob/main/ui/components/NetworkFees/FeeSettingsText.tsx#L74-L85). I think for now the right thing to do may be to check the network id in *there*, and use `networkSettings.values.gasPrice` for RSK in that scenario until a proper refactoring of fee handling can be done that allows networks to either opt into a common fee handling scheme we've already built out, or define their own somehow.

ui/components/NetworkFees/FeeSettingsText.tsx
ahsan-javaiid: Done!!

ui/components/NetworkFees/FeeSettingsText.tsx
ahsan-javaiid: Done!!

background/services/chain/asset-data-helper.ts
```diff
@@ -48,6 +48,28 @@ export default class AssetDataHelper {
...
+          )
+          return {
+            smartContract: {
+              contractAddress: token,
+              homeNetwork: RSK,
+            },
+            amount: BigInt(balance.toString()),
+          }
+        })
+        const resolvedBalances = Promise.all(balances)
```
0xDaedalus, line 70: We'll go ahead and change this to use `multicall` once https://github.com/tallycash/extension/pull/2423 lands but I think in the meantime you may want to change this to a `Promise.allSettled` and filter to fulfilled promises.

Some spam and/or poisonous tokens do not implement `balanceOf` correctly and can throw an error - causing all balance lookups to fail.

background/redux-slices/0x-swap.ts
0xDaedalus: Unfortunately the user is still able to setup a swap and access the swap screen while they have RSK selected:

image

I think a good fix here would be to disable the swap page button if `rsk` is selected - and to redirect to a the `Wallet` page if the user happens to have the `Swap` page open when switching networks.

0xDaedalus (commented):
This looks like its just about mergeable - just the comment about the swap screen left to address.

background/services/chain/asset-data-helper.ts
ahsan-javaiid: Done!

background/redux-slices/0x-swap.ts
ahsan-javaiid: Done!

background/services/chain/asset-data-helper.ts
```diff
@@ -52,6 +52,36 @@ export default class AssetDataHelper {
...
+        const resolvedBalances = await Promise.allSettled(balances)
+        const fullfilledBalances = resolvedBalances.filter(
+          (value) => value.status === "fulfilled"
+        )
+
+        return fullfilledBalances.map(
+          (balance) =>
+            (balance as PromiseFulfilledResult<SmartContractAmount>).value
+        )
+      }
```
0xDaedalus, line 84: Looks like https://github.com/tallycash/extension/pull/2423 got merged over the weekend so we can remove all of this code since the newly imported `getTokenBalances` will fetch all of the token balances in a single RPC call.  I realize that I asked for this path to be updated in my last comments and thank you for doing that - didn't think that the multicall stuff would get merged as quickly as it was.

This should be the last thing blocking merge - oh so very excited to get this in!

0xDaedalus (approved):
Houston we have lift-off 🚀 🚀

PR #2420, opened 2022-10-13
Fix achievements list width

jagodarybacka:
Ref https://github.com/tallycash/extension/issues/2372

### What
Fixed the width of the list. 
I've reproduced this issue once, we have probably encountered a problem with `styled-jsx` not working as expected.


0xDaedalus (approved):
📈

PR #1576, opened 2022-06-06
Slip-a-ping: Add basic SLIP-0044 coin type info to base assets

Shadowfiend:
(Originally discussed as part of the still-feature-flagged [RNS work](https://github.com/tallycash/extension/pull/1366#discussion_r883694240).
Note that this PR does NOT update any name resolvers to use `coinType`.)

Network base assets now have a SLIP44 coinType[^1] field, and a very basic
mapping is added from symbol to coin type. Only coin types currently
referenced in network base assets are tracked.

As noted in the coin-types file, although the data could be imported
from an external package, it is deliberately inlined here as plain data
to avoid pulling in external executable code for something that
represents simple data.

[^1]: https://github.com/satoshilabs/slips/blob/master/slip-0044.md

PR #2423, opened 2022-10-13
Add path to query ERC-20 balances on any evm compatible chain.

0xDaedalus:
This PR adds the ability to query up to 500 erc-20 token balances via one `eth_call` rpc request.  This is much more resource and cost-effective than sending many individual calls to a given contract's `getBalance` function and should allow us to add additional providers that do not support [alchemy_getTokenBalances](https://docs.alchemy.com/reference/alchemy-gettokenbalances)

Highlights:
- SerialFallbackProvider now accepts a JsonRPCProvider as its primary provider.
- Added an additional fallback provider for polygon.
- Added a method to get all known (via tokenlist) tokens for a given network.
- The code paths added here do not exist any existing functionality since all of our currently supported networks are also supported by alchemy.

### To Test
- [x] replace the code at the top of `makeSerialFallbackProvider` with:
```
  if (network.chainID === POLYGON.chainID) {
    return new SerialFallbackProvider(
      network,
      () => new JsonRpcProvider("https://polygon-rpc.com")
    )
  }
  ```
- [ ] Reinstall wallet
- [ ] Import an account
- [ ] Select Polygon chain
- [ ] ERC-20 balances should load
- [ ] Inspect the network tab in devtools - you should see http requests going out to `https://polygon-rpc.com`

### Todo
Will follow up with testing in a separate PR.

### CODE REVIEW

**Comment by mhluongo** [comment]
:fire:

File: background/package.json [hunk 1]
```diff
@@ -51,17 +51,17 @@
     "dayjs": "^1.10.7",
     "dexie": "^3.0.3",
     "emittery": "^0.9.2",
+    "ethereum-multicall": "^2.15.0",
```
mhluongo, line 54: .... why? This has to be something we can do with `ethers`, right?

mhluongo (commented):
Love the approach. Will review more deeply soon, but wanted to kick off the discussion around whether we can avoid the `ethereum-multicall` lib

background/package.json
Shadowfiend: I think the main thing we get from the library is the set of deployed multicall addresses on various networks. Feels like something we can pull in directly, not unlike the decision around #1576's [SLIP-0044 data](https://github.com/tallycash/extension/pull/1576/files#diff-c39bd52887c7de83c4578b1121b894b6f51d906022e3ce80244cee62d02af941R1). Multicall is quite flexible and the package provides a lot of the infrastructural types to enable that flexibility, but because we're just calling a specific ERC20 function, I *think* we can do a lot of that work directly with Ethers with relatively little code.

0xDaedalus: Should be good for another look 👍

greg-nagy (changes requested):
I love this PR, love this solution and we definitely should use this! But I feel this is not the correct place for that and we also should have a convo about some essential things before we decide we go down this path 

I think we should explore the `getLogs` + event filter method first, then if needed tie into this solution after proper considerations.

- I would like to steer us toward a solution that relies on user's tokens first and tokenlists for decision making 
- with the `getLogs` method we would probably be able to calculate the balances without any external dependency — still need to test if this approach is feasible. If not, then we can use this method to get the actual balances for the tokens that we got from that

My questions for this solution
- Should we rely on the already deployed contracts? If so what's our threat model? 
- Or should we deploy the contracts on our own? 
- What are the limitations that this method has? 

I would like this PR moved back to `Draft` until we clear the above points. This is a fantastic POC to showcase how this works and we should definitely rely on this in the future.

0xDaedalus:
Happy to move back to draft - would love to see a poc of the `getLogs` solution sooner rather than later though so we can discuss the tradeoffs offered by each.

Responses to some of your thoughts here:
> I would like to steer us toward a solution that relies on user's tokens first and tokenlists for decision making

This solution does in fact rely on tokenlists (its what we feed into the multicall) - and I think that if we couple it with @mhluongo's open spam-token PR we'll have token discovery (on chains that support alchemy).

> Should we rely on the already deployed contracts? If so what's our threat model?

I think we can rely on the already deployed contracts for a few reasons:
1. The contract is not upgradeable.
2. It is created via `CREATE2` so we know that if the address is the same on a given EVM chain the contract's bytecode must also be the same.

> Or should we deploy the contracts on our own?

We certainly can if we want to be extra security-conscious (which is perhaps a good idea) but we would need to deploy it across dozens of chains and then deploy again for every new chain we want to support.

> What are the limitations that this method has?

One clear limitation is we need to have knowledge of a token either through token-lists or through user-addition in order to query its balance (this update does not deal with discoverability).

mhluongo:
> with the `getLogs` method we would probably be able to calculate the balances without any external dependency 

Log filtering can't get token balances, as per [ERC-20](https://eips.ethereum.org/EIPS/eip-20)... unless you replay the whole chain.

This is the correct approach to token balance checking without an indexing shortcut like Alchemy.

We shouldn't let experimental asset discovery get in the way of this functional, immediate improvement to the codebase and our goal of going cross-provider.

background/services/indexing/index.ts
```diff
@@ -749,7 +752,24 @@ export default class IndexingService extends BaseService<Events> {
...
       ).map(async (addressOnNetwork) => {
-        await this.retrieveTokenBalances(addressOnNetwork, activeAssetsToTrack)
+        const provider = await this.chainService.providerForNetworkOrThrow(
+          addressOnNetwork.network
+        )
+        const isAlchemyProvider =
+          provider instanceof AlchemyProvider ||
+          provider instanceof AlchemyWebSocketProvider
+
+        if (isAlchemyProvider) {
```
greg-nagy, line 762: The itch is so real to remove this :)

background/lib/erc20.ts
```diff
@@ -173,3 +180,52 @@ export function parseLogsForERC20Transfers(logs: EVMLog[]): ERC20TransferLog[] {
...
+  )
+
+  const balanceOfCallData = ERC20_INTERFACE.encodeFunctionData("balanceOf", [
+    address,
+  ])
+
+  // eslint-disable-next-line no-useless-catch
+  try {
+    const response = (await contract.callStatic.tryBlockAndAggregate(
+      false,
```
greg-nagy, line 202: Some hints as to what the parameters are would be helpful.

background/lib/erc20.ts
```diff
@@ -173,3 +180,52 @@ export function parseLogsForERC20Transfers(logs: EVMLog[]): ERC20TransferLog[] {
...
+
+      return {
+        amount: BigInt(BigNumber.from(data.returnData).toString()),
+        smartContract: {
+          contractAddress: tokenAddresses[i],
+          homeNetwork: network,
+        },
+      }
+    })
+  } catch (e) {
```
greg-nagy, line 226: I think this catch will be with us longer than we expect.

greg-nagy (approved):
I think this is a wonderful PR. Let's 🚢

background/lib/erc20.ts
```diff
@@ -173,3 +180,52 @@ export function parseLogsForERC20Transfers(logs: EVMLog[]): ERC20TransferLog[] {
...
+    MULTICALL_CONTRACT_ADDRESS,
+    MULTICALL_ABI,
+    provider
+  )
+
+  const balanceOfCallData = ERC20_INTERFACE.encodeFunctionData("balanceOf", [
+    address,
+  ])
+
+  // eslint-disable-next-line no-useless-catch
```
Shadowfiend, line 199: Not sure the value of having the try catch and a lint ignore just for the todo here, versus just having the todo without the extra line noise 🤔

background/services/chain/asset-data-helper.ts
```diff
@@ -59,10 +63,14 @@ export default class AssetDataHelper {
           smartContractAddresses
         )
       }
+      return await getTokenBalances(
```
Shadowfiend, line 66: The original thinking here was not to check the provider type, but rather to use the thrown exception to feature-detect the absence of the alchemy call and fall back, fwiw.

Shadowfiend (commented):
Couple of minor nits I wanted to mention.

PR #2446, opened 2022-10-17
Fix setting spend approval amount

### LINKED ISSUE #2408: In-wallet swaps - approval limit can't be changed
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/1029325326385422336

Chrome

Change the spend limit

### What did not work?

It always jumps back to infinite.

image

### Version

_No response_

### Relevant log output

_No response_

jagodarybacka:
Resolves #2408

### What
Allow to force reload transaction annotations.
Reload transaction annotation when user changes the spend limit. Add a loader to indicate that it is reloading.

### Testing

- [ ] allow asset spend on uniswap or in-wallet swap (example - USDT on OP)
- [ ] allow asset spend screen should show up
- [ ] click change limit and set a new limit
- [ ] limit should be reloaded and displayed on the screen

image


ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -104,14 +101,30 @@ export default function SignTransactionSpendAssetInfoProvider({
...
+  useEffect(() => {
+    setApprovalLimitString(
+      isMaxUint256(approvalLimit)
+        ? null
+        : fixedPointNumberToString({
+            amount: approvalLimit,
+            decimals: asset.decimals,
+          })
+    )
+    setIsLoading(false)
```
hyphenized, line 125: Does it not work if you set loading to false after awaiting the dispatch? 🤔

Shadowfiend:
Was this only an issue on in-wallet swaps? Wondering what they're doing differently if so, given that this has historically worked.

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
```diff
@@ -104,14 +101,30 @@ export default function SignTransactionSpendAssetInfoProvider({
...
       [spender.address, hexlify(bigintAmount)]
     )
+
     dispatch(
       updateTransactionData({
-        ...transactionDetails,
-        input: updatedInput,
+        transaction: {
+          ...transactionDetails,
+          input: updatedInput,
```
0xDaedalus, line 109: Not blocking and i know its a bit out of scope of the immediate bugfix but I wonder if we can clean up the `handleSaveClick` function a little bit.  As far as I can tell in ~20 lines we make the following hops: 
`approvalLimitInput` => `decimalAmount` => `bigintAmount` => `updatedInput`
which is just quite difficult to read inline and is further complicated by checking for falsy values after just about every conversion.

Shadowfiend (changes requested):
Going to block this.

The breakage here is from #2331, where we stopped passing transactions that were already annotated to enrichment. This was a temporary hack for prepopulated swap annotations, whose side effects we unfortunately missed when testing the PR. This is fixed in #2461. We should either wait to merge that work, or apply the relevant change now (https://github.com/tallycash/extension/commit/65422590fbd021f8773c46dc9cb5cc8f3a638abc#diff-bb2ba232569016c28a3e1afeb301ce20420881034e715b0c87f88c6fd5db410cL674-R678), break the asset swap annotations that we ignore in the current flow, and re-fix them once that PR lands.

ui/components/SignTransaction/SignTransactionSpendAssetInfoProvider.tsx
Shadowfiend: Going to request we hold on this so that we can fix it in the new signing flow, see #2331.

jagodarybacka:
> Was this only an issue on in-wallet swaps

@Shadowfiend no, every asset spend approval had this issue

>  either wait to merge that work, or apply the relevant change

Both options are fine with me. Would like to avoid duplicated work so idk which direction would you prefer @greg-nagy @0xDaedalus

greg-nagy:
> Both options are fine with me. Would like to avoid duplicated work so idk which direction would you prefer @greg-nagy @0xDaedalus

In this case please include the changes @jagodarybacka. The other PR is said to be ready for review but is still in Draft. Doesn't sound like a lot of effort and it could mean that this lands in the next release we cut on Friday.

greg-nagy (approved):
warms my heart 🔥

PR #2447, opened 2022-10-17
Don't inject provider on non html documents

hyphenized:
This should prevent weird script tags from appearing on xml docs

image

Ref: https://discord.com/channels/808358975287722045/888500174685614090/1031480622377615420

## To Test:
- [ ] [Open an xml document](https://www.w3schools.com/xml/note.xml), it should display correctly
- [ ] Check the script tag does not appear in other document formats

### DIFF
--- a/provider-bridge/index.ts
+++ b/provider-bridge/index.ts
@@ -67,6 +67,8 @@ export function connectProviderBridge(): void {
 }
 
 export function injectTallyWindowProvider(): void {
+  if (document.contentType !== "text/html") return
+
   try {
     const container = document.head || document.documentElement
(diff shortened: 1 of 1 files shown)


provider-bridge/index.ts
```diff
@@ -67,6 +67,8 @@ export function connectProviderBridge(): void {
 }
 
 export function injectTallyWindowProvider(): void {
+  if (document.contentType !== "text/html") return
```
greg-nagy, line 70:
```suggestion
  if (document.contentType !== "text/html") {
    return
  }
```

greg-nagy (approved):
\o/ wooo

PR #2457, opened 2022-10-19
Throttle CoinGecko API calls.

0xDaedalus:
This PR fixes an issue where if the product of (Addresses & Networks) was more than 60 - users would _always_ hit the coingecko ratelimit - and if it was less than that they were likely to when adding multiple addresses or networks at once (or while using coingecko.com).

This PR lays down some groundwork for https://github.com/tallycash/extension/issues/2134

There are two common scenarios happening that may rate-limit coingecko for users (and result in inconsistent prices).

1.  When the extension is first loaded - chain service emits every account on every network as part of the `newAccountToTrack` event - each of which triggers a call to `handlePricealarm`.

2.  When a user switches to a network for the first time - a `newAccountToTrack` event is emitted for every account in the wallet.

By throttling the `handlePriceAlarm` function - we ensure that these bursts of `newAccounttoTrack` events do not result in repeated duplicate calls to coinGecko - and mitigate the risk of the user exceeding the coingecko API limits.

### To Test
- [ ] Load up a new wallet, Open network tab in devtools.
- [ ] Prices should load for ethereum wallet.
- [ ] Switch to polygon - prices should load.
- [ ] Switch to optimism / arbitrum - prices should load.
- [ ] "Refresh" the wallet via - filter the devtools panel - you should see:
  - [ ] One request to `/simple/prices?ids=...`
  - [ ] N requests to `/simple/token_price/...` where N is the number of networks you have added.
  - [ ] Crucially you should not see multiple requests to the same URL within 5 seconds.

### DIFF
--- a/background/lib/prices.ts
+++ b/background/lib/prices.ts
@@ -17,27 +17,6 @@ import { USD } from "../constants"
 
 const COINGECKO_API_ROOT = "https://api.coingecko.com/api/v3"
 
-export async function getPrice(
-  coingeckoCoinId = "ethereum",
-  currencySymbol = "usd"
-): Promise<number | null> {
-  const url = `${COINGECKO_API_ROOT}/simple/price?ids=${coingeckoCoinId}&vs_currencies=${currencySymbol}&include_last_updated_at=true`
-
-  const json = await fetchJson(url)
-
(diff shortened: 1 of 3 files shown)


jagodarybacka (approved):
Works as expected

PR #2460, opened 2022-10-19
Display `Tally Ho` as a connection option on GMX and Uniswap

0xDaedalus:
https://user-images.githubusercontent.com/94649004/196811459-2e9c8b09-3a0f-43c2-a36f-22ce91534d21.mov

This PR is a first step towards increasing the number of dapps that display `Tally Ho` as a connection option for users of the wallet.  It introduces two bespoke handlers and introduces a pattern through which integrations for additional dapps can be made.

#### Highlights:
- Minimal overhead (handlers exit early and often - and only run on their given origin)
- An origin <> handler mapping through which future handlers can easily be added
- a new `tallySetAsDefault` property on `window.ethereum` which is a bit more idiomatic way of inferring that the user has set tally as default (as opposed to checking for `window.ethereum.isMetaMask`)

### To Test
- When Tally is the only wallet installed
  - [ ] Connecting to Uniswap should display `Tally Ho` as a connection option.
  - [ ] Connecting to gmx.io should display `Tally Ho` as a connection option.
- When both Metamask and Tally are installed
  - If the `Use Tally Ho as your default wallet` setting is not toggled
    - [ ] Connecting to Uniswap should display `Metamask` as a connection option.
    - [ ] Connecting to gmx.io should display `Metamask as a connection option.
  - If `Use Tally Ho as your default wallet` setting is toggled
    - [ ] Connecting to Uniswap should display `Tally Ho` as a connection option.
    - [ ] Connecting to gmx.io should display `Tally Ho` as a connection option.


window-provider/wallet-connection-handlers.ts
```diff
@@ -0,0 +1,111 @@
+const observeMutations = (handler: (node: Node) => void) => {
+  document.addEventListener("DOMContentLoaded", () => {
+    const observer = new MutationObserver(function monitorMutations(mutations) {
```
greg-nagy, line 3: I didn't see any perf hit, but running the observer on every DOM change feels a bit strong handed. 

Could we select the `Connect` element and listen to a click event? 
This would miss the eager loading scenario, but that could be handled with checking for the element on page load.

Not sure if this worth the effort though, if we don't generate any significant overhead.

window-provider/wallet-connection-handlers.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+const findAndReplaceUniswapInjectedOption = (addedNode: Node): void => {
+  if (
+    window.ethereum &&
+    Array.isArray(window.ethereum.providers) &&
+    window.ethereum.providers.length > 1
+  ) {
+    // If the user has more than 1 wallet installed
+    if (!window.ethereum.tallySetAsDefault) {
```
greg-nagy, line 22: We probably should expand this check filter the users who hasn't finished the onboarding. 
Or redirect to onboarding in the popup.

window-provider/wallet-connection-handlers.ts
```diff
@@ -0,0 +1,111 @@
...
+      subtree: true,
+      childList: true,
+    })
+  })
+}
+
+const findAndReplaceUniswapInjectedOption = (addedNode: Node): void => {
+  if (
+    window.ethereum &&
+    Array.isArray(window.ethereum.providers) &&
```
greg-nagy, line 18: This check should be refactored because [for some pages we remove the `providers` array alltogether](https://github.com/tallycash/extension/blob/784d3a55243ca5ee9c30567d3ad54fc2573f716c/src/window-provider.ts#L127).

We need to do that, because web3-react would search for MM in the providers array and use that as a connector, instead of `window.ethereum`. 
Which is 'meh', but it is what it is.

window-provider/wallet-connection-handlers.ts
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+    // If the user has more than 1 wallet installed
+    if (!window.ethereum.tallySetAsDefault) {
+      // And Tally is not set as the default - return
+      return
+    }
+  }
+
+  // Detect which version of web3-react
+  const maybeButton = addedNode?.childNodes?.[2]?.childNodes?.[0]
```
greg-nagy, line 29: maybeButton ❤️

jagodarybacka, line 29: is it possible to do it with querySelector?

jagodarybacka:
hmm, it is working the first time you open the connect modal on Uniswap but if you close it and open it again it shows injected for me :(

window-provider/wallet-connection-handlers.ts
greg-nagy:
```
const connectButton = document.querySelectorAll("button[data-testid='navbar-connect-wallet']")[0]
connectButton.addEventListener('click', (event) => {console.log("hey we are here")})
```

window-provider/wallet-connection-handlers.ts
```diff
@@ -0,0 +1,111 @@
...
+    ?.childNodes?.[0]?.childNodes?.[0] as Element
```
greg-nagy, line 30: We could use `document.getElementById("injected")`.
Not sure how these 2 approaches compare though.

window-provider/wallet-connection-handlers.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (
+    !!maybeButton &&
+    maybeButton.id === "injected" &&
+    maybeButton.parentElement?.getAttribute("data-testid") === "option-grid"
+  ) {
+    // Even though we can be reasonably sure that we've correctly identified
+    // a change related to web3-react.  It never hurts to double check.
```
greg-nagy, line 37: Is this a web3-react specific thing or uniswap?

greg-nagy (changes requested):
This is amazing 🤩

The checks for the `providers` array is a blocker, the rest is more of a discussion on the approach.

window-provider/wallet-connection-handlers.ts
0xDaedalus: Unfortunattely querySelector is O(N) where N is _all_ of the elements in a given element's tree.  This is uglier but much more efficient as it will stop traversing as soon as an expected leaf is not found.

window-provider/wallet-connection-handlers.ts
0xDaedalus: There are two issues with adding a listener that I encountered:

1) There are at least 3 "Connect" buttons on the main uniswap app screen - one of which has no id (of course we can grab a parent element and iterate through children - but still its a bit messy)
2) When using an `onClick` handler - the dom element we want to change may not be on the screen at the time the handler is called (especially if you emulate as a slower computer) - so we end up with a race condition we need to manage.

window-provider/wallet-connection-handlers.ts
0xDaedalus: This is a good point and should I'd like to address it before we merge

window-provider/wallet-connection-handlers.ts
0xDaedalus: Looks like this check was not even needed in uniswap's specific case (I think probably because of whats happening in the code you linked).  Will have to think about how to do this on other sites so that Tally does not override MetaMask when its not set as default - but we can cross that bridge when we get there.

window-provider/wallet-connection-handlers.ts
0xDaedalus: Alright - I've been won over.

Both approaches are brittle - but the additional check for truthyness `iconAndTextDiv` makes me feel better about not accidentally grabbing the wrong element (since it is not unreasonable for another element to accidentally be added with an `id` of `injected`).  Also we'll get O(1) lookup time with `getElementById` - so thats a nice plus.

window-provider/wallet-connection-handlers.ts
0xDaedalus: On second thought lets follow this up in a separate PR - proving to be a more involved change than I'd like it to be for a spike.

window-provider/wallet-connection-handlers.ts
0xDaedalus: uniswap - comment has been updated

0xDaedalus:
@greg-nagy This should be good for another look ✅

window-provider/wallet-connection-handlers.ts
greg-nagy: The issue for this: https://github.com/tallycash/extension/issues/2487

Added to next sprint for us to not forget about it, but this might be ok do move into this sprint.

PR #2462, opened 2022-10-20
Remove Balance Reloader feature flag and functionality.

0xDaedalus:
The button this feature flag enables has been off for months now - and was initially added as a temporary fix to address wallet balance issues which have since been resolved.

### To Test
- Load up the wallet
- Navigate to the `Wallet` page
- Page should render, buttons should be clickable, everything should work.


greg-nagy:
this should be updated after #2412 is landed 

we use the balance reloader component there, but the feature flag should be cleaned up

background/main.ts
```diff
@@ -172,6 +172,7 @@ const persistStoreState = debounce(persistStoreFn, 50, {
 const reduxCache: Middleware = (store) => (next) => (action) => {
   const result = next(action)
   const state = store.getState()
+  ;(window as any).store = store
```
jagodarybacka, line 175: please fix liner :)

0xDaedalus: whoopsie

0xDaedalus:
Hmmmn strong disagree here that we should continue having this feature flag and add this component back to the application.

Its unclear to me what value it brings to our users and the convenience it brings to a developer is quite easy to replicate in a few lines of code.  As noted in PR description - this was a component that was intended to be a temporary fix to a specific problem (that we don't have anymore).  In my mind this is just cruft we should get rid of.

greg-nagy:
> Hmmmn strong disagree here that we should continue having this feature flag 

We should not have this feature flag, and we should remove it. Just let's do it after the feature flag system has been refactored. 

> and add this component back to the application.

We have repurposed this component and used it on the developer panel to reload the background script after runtime feature flag change. 

CleanShot 2022-10-20 at 19 46 47


If, after this context, you are not convinced, we can remove it altogether and show everybody that they can reload the background through the dev toolbar.

mhluongo:
Calling out that I think the feature flag panel is a mistake that's going to bite us later, and adding this in the dev panel does next to nothing... a developer can already hit "refresh" in the extensions panel after all.

mhluongo:
Apologies, didn't mean to reopen. Standing by the comments though.

Shadowfiend:
Lord, fat fingers everywhere on this one. 

> going to bite us later

Can you say more about how it might bite us?

PR #2463, opened 2022-10-20
Oral Literature: Display asset transfers in activity item details

Shadowfiend:
Got tired of having the data and not seeing it lol. The implementation nominally reuses most of the existing design language for transaction details rather than trying to put these in their own place.

Screen Shot 2022-10-20 at 11 11 30


greg-nagy (approved):
this is 🔥

It would be awesome to see this on the list view. Wdy @VladUXUI?

PR #2464, opened 2022-10-20
Fix: weblate badge URL in readme

katspaugh:
The badge src was wrong:
Screenshot 2022-10-20 at 18 41 56

After the fix:

Screenshot 2022-10-20 at 18 42 43


0xDaedalus (approved):
Thank You ❤️

PR #2467, opened 2022-10-20
Extended Accounting: Handle longer account names gracefully

Shadowfiend:
Two big things here: account name styling fixes, and the addition of copy address to the account action menu.

First up, in the account dropdown and the selected account in the top bar both, longer account names have not been getting elided and rek the layout.

This PR fixes this in three places:
- The account selector.
- The top bar account name on the wallet.
- The top bar account name when signing.
- The account name on the dApp connection screen.

Next, the account action menu (where remove address and edit name exist) is now styled to match the design, with cleaner padding all around.

Lastly, there is a new Copy address button in the account action menu that… Copies the account address!

<details>
  <summary>Before screenshots</summary>
  Overflowing long ENS names in top bar and account selector
  Overflowing long ENS names in dApp connections
  Overflowing long ENS names in transaction signing
  Visually poor padding in account action menu
</details>

<details>
  <summary>After screenshots</summary>
  Properly elided ENS name in top bar and account selector
  Properly elided ENS name in dApp connections
  Properly elided ENS name in transaction signing
  Correct padding and new copy address account action
</details>

## Testing

- [ ] Add `theannuitypartnershipscrosschainunionreservefund.eth` as a read-only account.
  - [ ] Make sure that the selected account on the main wallet screens does not overflow.
  - [ ] Open the account selector and make sure the account in the list does not overflow.
  - [ ] Connect to a dApp and make sure the account in the dApp connection screen does not overflow.
  - [ ] Try to sign a dApp transaction and make sure the account on the signing screen does not overflow.
- [ ] Open the account selector and open the action menu for an account.
  - [ ] Make sure “Copy address” appears.
  - [ ] Try to copy the address and confirm that it copies the correct address for each account.
- [ ] Make sure the account selector is working as expected to switch accounts, add accounts, remove accounts, and edit names.


VladUXUI:
Couldn't we have the shared component for copying here as well? In the sense that clicking on address alone would copy the address?
Or did we go this extra button route because it's too many click areas around?

Shadowfiend:
That was my instinct, but we can def approach differently if you would prefer. Basically worried that the obvious visual area for “selecting” is the same as what we would be turning into an address-copy hit area instead.

Also totally forgot that adding that button makes this more than just a bug fix PR 🤦

greg-nagy (approved):
\( ﾟヮﾟ)/

PR #2468, opened 2022-10-20
May The Copy Be With You: Make overview addresses copyable

Shadowfiend:
Addresses on the overview page now look the same, but can be copied by clicking, as they now use the `SharedAddress` component to render.

## Testing

- [ ] Go to the overview/portfolio page.
  - [ ] Ensure both named and unnamed addresses can be copied by clicking.
  - [ ] Ensure there are no other styling changes.


greg-nagy (approved):
💅

PR #2469, opened 2022-10-20
Destination Resolution: Show resolved address for Send recipient

Shadowfiend:
In cases where the Send recipient is resolved from a name, the address is presented under the name as a copiable item. Very roughly aligns with existing designs around the contact list.

With resolved names, the full address is shown and can be clicked to copy:
Screen Shot 2022-10-20 at 18 22 56

Standard addresses are unchanged:
Unchanged address resolution

## Testing

- [ ] Go to the Send page and enter a resolvable name, e.g. `bnbspider2.crypto`.
  - [ ] Ensure the address appears below the name once it is resolved.
  - [ ] Ensure the address can be clicked to copy.
  - [ ] Ensure the address disappears if the text field changes.
  - [ ] Ensure the transaction is correctly addressed when you click Continue.
- [ ] Go to the Send page and enter a regular address.
  - [ ] Ensure there is no additional address below the one you have entered.
  - [ ] Ensure the transaction is correctly addressed when you click Continue.

### DIFF
--- a/ui/pages/Send.tsx
+++ b/ui/pages/Send.tsx
@@ -29,6 +29,8 @@ import { enrichAssetAmountWithMainCurrencyValues } from "@tallyho/tally-backgrou
 import { useHistory, useLocation } from "react-router-dom"
 import classNames from "classnames"
 import { ReadOnlyAccountSigner } from "@tallyho/tally-background/services/signing"
+import { setSnackbarMessage } from "@tallyho/tally-background/redux-slices/ui"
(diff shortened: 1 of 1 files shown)


VladUXUI:
Tested and looks/works awesomely!

greg-nagy (approved):
[Wubba Lubba dub-dub](https://rickandmorty.fandom.com/wiki/Wubba_Lubba_dub-dub)

PR #2474, opened 2022-10-21
Use a ternary to determine return value of `getAssetPricePoint`

0xDaedalus:
This is a bit more clear than the previous implementation.


background/redux-slices/utils/0x-swap-utils.ts
```diff
@@ -67,7 +67,9 @@ export async function getAssetPricePoint(
     await getTokenPrices(assetPricesNetworks, USD, network)
   )
 
-  return (unitPricePoint && getPricePoint(asset, unitPricePoint)) || undefined
+  return unitPricePoint === undefined
```
greg-nagy, line 70: I think it's time let go the `typeof unitPricePoint === "undefined"` pattern officially since we haven't been using that consistently throughout the codebase.

greg-nagy (approved):
let's 🚀

PR #2475, opened 2022-10-21
Transitional Fossil: Transition icon colors on hover where relevant

Shadowfiend:
In a few places, icons were being used alongside text whose color transitioned on hover. Because the icons are not colored the same way, the icon colors would change instantly instead of transitioning, which led to a jerkiness in the hover animation.

SharedIcon now has a transitionHoverTime property that sets the time it takes to transition the icon color on hover. This is applied appropriately for the account lock/unlock button in the accounts panel, as well as for the address copy button on the Send page.

## Testing

- [ ] Hover over the “Unlock signing” button in the accounts panel and note that the text and icon transition simultaneously instead of the icon switching colors immediately.
- [ ] Hover over the "Lock signing" button after unlocking, and see the same.
- [ ] Go to the Send page, enter an ENS or local account name. Hover over the resolved address below the name and note the text and icon transitioning simultaneously.


0xDaedalus (approved):
🚢

PR #2476, opened 2022-10-21
Mock Websocket Connections - Fix a number of issues relating to tests running inconsistently in CI

0xDaedalus:
This PR addresses a number of issues encountered while debugging tests failing in CI (but passing locally)

### Highlights
- Global mock (during tests) of the SerialFallbackProvider to avoid creating websocket connections during testing.
- [Upgraded from node12 to node 16 github actions.](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)
- Attempt local name resolution prior to using remote name resolvers.
- Explicitly stop chain after tests which start it are run.
- [Fix to an intermittently failing name service test.]

### To Test
- [ ] CHIEFLY - tests should pass locally without network connectivity (turn off wifi).
- [ ] `yarn test` should not fail the `NameService Looks up and returns names from built-in contracts` test
- [ ] Address name resolution should continue to work as expected in activities / send asset / read-only import.
- [ ] Our builds in github should no longer contain warnings about using an out-of-date version of Node.

### DIFF
--- a/setupJest.env.ts
+++ b/setupJest.env.ts
@@ -1,3 +1,19 @@
+import sinon from "sinon"
+import SerialFallbackProvider, * as serialFallbackProvider from "@tallyho/tally-background/services/chain/serial-fallback-provider"
+import { makeSerialFallbackProvider } from "@tallyho/tally-background/tests/factories"
+
+const sandbox = sinon.createSandbox()
+/* Reset IndexedDB between tests */
+beforeEach(() => {
+  sandbox.restore()
+  sandbox
+    .stub(serialFallbackProvider, "makeSerialFallbackProvider")
+    .callsFake(() => {
+      return makeSerialFallbackProvider() as SerialFallbackProvider
+    })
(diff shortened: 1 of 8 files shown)


setupJest.env.ts
```diff
@@ -1,3 +1,19 @@
+import sinon from "sinon"
+import SerialFallbackProvider, * as serialFallbackProvider from "@tallyho/tally-background/services/chain/serial-fallback-provider"
+import { makeSerialFallbackProvider } from "@tallyho/tally-background/tests/factories"
+
+const sandbox = sinon.createSandbox()
+/* Reset IndexedDB between tests */
+beforeEach(() => {
+  sandbox.restore()
+  sandbox
+    .stub(serialFallbackProvider, "makeSerialFallbackProvider")
```
greg-nagy, line 10: Would it make sense for this to be done in a jest `__mock__` auto mock?

greg-nagy (approved):
✨

PR #2477, opened 2022-10-21
Add build links to PR descriptions and read env from PR descriptions

Shadowfiend:
This PR implements the following 3 changes:
- Workflow build artifacts (but not release artifacts yet!) carry a suffix that is their PR number or, if unavailable, their commit SHA.
- The build workflow looks for a `## Testing Environment` (`Env` will work as well) marker in the PR description and pulls the code block that follows it for env variables, which are injected into the build's `.env` file. Env variables are only used if they are prefixed by `USE_`, `ENABLE_`, `SHOW_`, and `SUPPORT_` so as to minimize the possibility of environment injection by someone who may have PR description edit capabilities but not commit capabilities on a given branch. It's probably overly paranoid, but feels reasonable.
- After the build workflow completes, the `post-build-link` workflow triggers. This pulls the artifact from the build and adds a comment to the end of the PR with that information.

Notably, this PR also updates linting, prettier, and pre-commit to hit `.yml` files and all other files in the `.github` workflow directories.

## Goals

- The goal of the build artifact name is distinguishability of artifacts that may come from different source PRs or commits. This should reduce how error-prone testing from different artifacts is.
- The goal of the environment changes is to enable straightforward feature flag builds for flags that must be enabled at extension startup and cannot be toggled at runtime.
- The goal of the build link is to make it easy for someone trying to test to always be able to get the latest successful build from the PR description.

## Limitations

- Editing the PR description to change the testing environment will not trigger a new build. Rerunning the last build workflow after editing the description *should* pick up the new details.
- The `post-build-link` workflow trigger won't work until the `post-build-link` workflow is on `main` (for security reasons, these kinds of dependent workflows always run off the version on the default branch). In the meantime, I've hardcoded the workflow run id that is referenced by the build link, so it will always remain the same for now.

## Testing Environment

```
USE_UPDATED_SIGNING_UI=true
```

Latest build: [extension-builds-2477](https://github.com/tallycash/extension/suites/8912813689/artifacts/408242915) (as of Sun, 23 Oct 2022 01:36:41 GMT).


mhluongo:
🔥 This is more like it! Big fan of this approach, feels like what we should've reached for before a dev panel.

Are the builds of limited use with a bunch of the feature flags  moved to runtime? Or do the runtime flags also pick up build flags?

Shadowfiend:
My understanding of the feature flag panel is that it will show the current setting, so it should just reflect the updated env. I might be wrong though, need to look back at the code there.

greg-nagy:
Runtime feature flags are [initialized from the build time feature flags](https://github.com/tallycash/extension/blob/714ae25c3cd70eeac7c77a6a26c1d652fac1ebd9/background/features.ts#L12-L29) and they are designed to allow overwrites only if the build time feature flag is enabled for this feature. [`isEnabled` function](https://github.com/tallycash/extension/blob/714ae25c3cd70eeac7c77a6a26c1d652fac1ebd9/background/features.ts#L45-L54).

.github/workflows/build.yml
```diff
@@ -24,9 +24,19 @@ jobs:
...
         cache: "yarn"
     - run: yarn install --frozen-lockfile
+    - name: Detect env block
+      id: env-block
+      if: github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')
+      uses: actions/github-script@v6
+      with:
+        script: |
+          const detectEnvBlock = require("./.github/workflows/builds/detect-env-block.js")
+          return await detectEnvBlock({ github, context })
```
greg-nagy, line 34: 👀

Shadowfiend:
Got a +1 from @greg-nagy here and from @0xDaedalus elsewhere, so converting back to draft in order to clean up history + take care of any straggling tidbits.

Shadowfiend:
This PR is ready for review. Note that it is expected that the `post-link` workflow will not run in this branch at this point; it will not be able to run until this branch is merged to `main` due to the way that `workflow_run` and `workflow_dispatch` triggers operate (they must exist on `main` to trigger).

.tsconfig-eslint.json
```diff
@@ -1,5 +1,11 @@
 {
   "extends": "./tsconfig.json",
-  "include": [".eslintrc.js", "**/*.js", "**/*.ts", "**/*.tsx"],
+  "include": [
+    ".eslintrc.js",
+    "**/*.js",
+    "**/*.ts",
+    "**/*.tsx",
+    ".github/**/*.js"
```
0xDaedalus, line 8: Will this be already be caught by `**/*.js`?

.prettierignore
```diff
@@ -1,3 +1,4 @@
 dist
 # Ignore weblate json translation formatting
-ui/_locales/**/*.json
\ No newline at end of file
+ui/_locales/**/*.json
+!.github
```
0xDaedalus, line 4: Do we need to explicitly list this here and in the `.eslintrc` file because `.github` is a hidden folder?

0xDaedalus (approved):
Looks good to me - will leave open for a bit to let others comment

.tsconfig-eslint.json
Shadowfiend: Yeah, I expected it. It seems to need an explicit addition because of the hidden/dotfile directory. Same deal as with the `!.github` stuff.

.git-blame-ignore-revs
```diff
@@ -2,3 +2,5 @@
 23545c8ed87d8e9b1661ab02dfaefddaf3725f3e
 # Impose prettier formatting on JSON and Markdown.
 8a90e88658795d569ca525ecb7bced875922c59c
+# Impose prettier formatting on .github YAML files.
+4851e7989a3473c64b03adb4fbef16b3957ac2ab
```
greg-nagy, line 6: ❤️

greg-nagy (approved):
🤩

love the github yml fix :)

PR #2486, opened 2022-10-24
Add a Polygon to the supported network for Ledger signing

### LINKED ISSUE #2037: Connecting TH to Polygon POS not working with Ledger and Polygon Network
### **Build Info/PR/Commit:**
https://github.com/tallycash/extension/pull/2024

### **Reproduction Steps:**
(See Video)
1.  Have Ledger Connected, Have Polygon set as Network in TH
2. Attempt to Connect TH to Polygon POS

### **Expected Result:**

Connection Success

### **Actual Result:**
TH Closes, kinda thinks it's connected, but not really. Ledger Never gets the Sign.

https://www.awesomescreenshot.com/video/10734953?key=fb3551b6476beddf0159fd66c378bf76

kkosiorowska:
Closes #2037 

This PR fixes the issue with connecting TH to Polygon POS for Ledger. Polygon has been added to the supported network for signing Ledger.

**UI**

Before
https://user-images.githubusercontent.com/23117945/197497341-a86f8162-aa08-445f-9812-38265da85dd4.mov

After
https://user-images.githubusercontent.com/23117945/197497351-3e6f4177-6e09-4a1c-a863-eaae9fe7dc7d.mov

**To Test**
Set TH as the default wallet. Connect wallet to Ledger and use Polygon network. Go to [Polygon PoS](https://wallet.polygon.technology/). Attempt to Connect TH to Polygon POS
- [ ] check if the connection was successful


jagodarybacka:
hmm I can't connect with TH to the linked Polygon PoS website, can you confirm? we need to impersonate MM probably

kkosiorowska:
> hmm I can't connect with TH to the linked Polygon PoS website, can you confirm? we need to impersonate MM probably

@jagodarybacka 
I have no problem with that. Did you set TH as the default wallet ?

jagodarybacka (approved):
Successfully logged in with Ledger ✅ 

There is some problem with logging into the beta version of the polygon wallet website and it's redirecting users if they once selected the beta version. Will add a separate issue.

background/constants/networks.ts
```diff
@@ -112,6 +112,8 @@ export const TEST_NETWORK_BY_CHAIN_ID = new Set(
   [GOERLI].map((network) => network.chainID)
 )
 
+export const NETWORK_FOR_LEDGER_SIGNING = [ETHEREUM, POLYGON]
```
Shadowfiend, line 115: This shouldn't be a global type--it's really a ledger-specific setting, so it should live in the ledger service file. I also believe it's specifically the networks that the Ledger Ethereum app can sign for, which we should ideally reflect in the variable name, or at least in a comment. Lastly, it's possible the Ledger lib already exports some sort of information like this, in which case we may want to reuse--that might take some spelunking though.

PR #2488, opened 2022-10-24
🎳 Transmutation: Manifest v3

greg-nagy:
> Alchemy (錬金術 Renkinjutsu) is, as it is understood in the [Fullmetal Alchemist](https://fma.fandom.com/wiki/Fullmetal_Alchemist) series, the ancient metaphysical science/mystical art of manipulating and altering matter by using natural energy. This act is known as "Transmutation" (錬成 Rensei) and its sequence is usually described as:
> 1. Comprehension - Understanding the inherent structure and properties of the atomic or molecular makeup of a particular material to be transmuted, including the flow and balance of potential and kinetic energy within.
> 2. Deconstruction - Using energy to break down the physical structure of the identified material into a more malleable state so as to be easily reshaped into a new form.
> 3. Reconstruction - Continuing the flow of energy so as to reform the material into a new shape.

https://fma.fandom.com/wiki/Alchemy

# Testing
- load the extension and use normally
- [permission check](https://developer.chrome.com/docs/extensions/mv3/permission_warnings/#view_warnings)
  - clear dist folder
  - checkout main > build > package > load
  - you should see an extension can edit the pages permission modal
  - clear dist folder
  - checkout this branch > build > package (with the same private key) > load 
  - no new permission modal should pop up

## Todo
- [ ] dev time reloader script 
- [ ] usb handling and ledger

Ref #2694

Latest build: [extension-builds-2488](https://github.com/tallycash/extension/suites/9628539326/artifacts/459966499) (as of Fri, 02 Dec 2022 09:50:45 GMT).


background/lib/logger.ts
```diff
@@ -75,13 +76,8 @@ function purgeSensitiveFailSafe(log: string): string {
   )
 }
 
-const isBackgroundLogger = browser.extension.getBackgroundPage() === window
-// isPopupLogger will briefly be true even if we're in a tab, but should
-// quickly resolve to false.
-//
-// Note that as of this comment's writing the content scripts do not use the
-// logger, so defaulting to true should generally always produce the correct
-// result even if the following check never completes.
```
Shadowfiend, line 84: This still applies yes?

provider-bridge/index.ts
```diff
@@ -75,7 +73,7 @@ export function injectTallyWindowProvider(): void {
     // this makes the script loading blocking which is good for us
     // bc we want to load before anybody has a chance to temper w/ the window obj
     scriptTag.setAttribute("async", "false")
-    scriptTag.textContent = INJECTED_WINDOW_PROVIDER_SOURCE
+    scriptTag.src = browser.runtime.getURL("window-provider.js")
```
Shadowfiend, line 76: Is this for CSP reasons or we just don't have any direct script injection capabilities anymore?

ui/pages/Settings/SettingsExportLogs.tsx
```diff
@@ -7,11 +7,17 @@ import SharedPageHeader from "../../components/Shared/SharedPageHeader"
...
 export default function SettingsExportLogs(): ReactElement {
   const { t } = useTranslation()
+  const [base64LogData, setBase64LogData] = useState("")
 
-  const serializedLogs = serializeLogs()
-  const base64LogData = Buffer.from(
-    `${window.navigator.userAgent}\n\n\n${serializedLogs}` || ""
-  ).toString("base64")
+  useEffect(() => {
+    serializeLogs().then((serializedLogs) => {
```
Shadowfiend, line 13: We could use async/await here too for consistency fwiw.

background/lib/logger.ts
```diff
@@ -119,7 +115,7 @@ async function saveLog(
     .join("\n    ")
 
   const logKey = `logs-${level}`
-  const existingLogs = localStorage.getItem(logKey) ?? ""
+  const existingLogs = (await localStorageShim.getItem(logKey)) ?? ""
```
Shadowfiend, line 118: Very worried this approach will lose logs because it's async. I tried to move from local storage to this approach in the last logging update and got scared off by that 😕 Will ponder a bit.

Shadowfiend (commented):
Some quick thoughts.

provider-bridge/index.ts
greg-nagy: CSP reasons which are extra tough for manifest v3. If not for manifest v3, this wouldn't be an issue.

ui/pages/Settings/SettingsExportLogs.tsx
greg-nagy: I rewrote this 2 times back and forth ... :) 

non of them felt good. Will rewrite to async await

background/lib/logger.ts
greg-nagy: yeah, def scary, but it might worth it 

not sure yet tbh 

this is a dirty POC and my goal was to get something up and running to figure out the unknowns

greg-nagy: See #2823 for the production version fix for the problem.

PR #2490, opened 2022-10-24
Remove balance reloader feature flag

greg-nagy:
we have had that as a temporary fix and we are not using it for ~1 month now, so it's time

goodbye old friend

## Test
- everything should work normally


0xDaedalus (approved):
Goodnight, sweet prince.

PR #2494, opened 2022-10-24
Inject Tally Ho as a connection option on tofunft.com

0xDaedalus:
https://user-images.githubusercontent.com/94649004/197634480-2ca50e5c-39d3-4fda-83c6-9dc631d46cc0.mov

### To Test
- [ ] Navigate to [(https://tofunft.com/arbi](https://tofunft.com/arbi)
- [ ] Attempt to Connect Wallet
  - [ ] If Metamask is not installed - Tally Ho should be an option.
  - [ ] If Metamask is installed and Tally Ho is set as default - Tally Ho should be an option.
  - [ ] IF Metamask is installed and Tally Ho is not default - Tally Ho should not be an option.


jagodarybacka (approved):
Works fine ✅

PR #2505, opened 2022-10-25
Add support for React dev tools in development

hyphenized:
This should make debugging a bit easier
plugnplay

## To Test:
- [ ] Reload extension from dev panel
- [ ] Run `npx react-devtools`


webpack.config.ts
```diff
@@ -112,6 +112,30 @@ const modeConfigs: {
...
+            from: "node_modules/@tallyho/tally-ui/public/",
+            force: true,
+            priority: 1,
+            transform: (content, fileName) => {
+              if (fileName.endsWith("popup.html")) {
+                return content
+                  .toString("utf8")
+                  .replace(
+                    "<!-- INSERT_REACT_DEV_TOOLS_HERE -->",
+                    `<script src="http://localhost:8097"></script>`
```
0xDaedalus, line 127: Can we declare `http://localhost:8097` as a variable and use it here?   (maybe `REACT_DEVTOOLS_DEFAULT_PORT`) A bit of a nit but I think we should avoid magic strings whenever we can.

0xDaedalus (commented):
CC @greg-nagy I think I have a _very_ vague memory us not using react-devtools due to some weirdness with how we use redux - i might be totally wrong but thought I'd check with you.

greg-nagy:
> CC @greg-nagy I think I have a very vague memory us not using react-devtools due to some weirdness with how we use redux - i might be totally wrong but thought I'd check with you.

That's redux devtools and not react devtools. The issue there is that we have moved the redux to the background thread and for this we need to connect remotely to the redux instance. This makes the instance discovery quite brittle. Sometimes it works most of the time it doesn't.

webpack.config.ts
```diff
@@ -112,6 +112,30 @@ const modeConfigs: {
...
+                  )
+              }
+
+              return content
+            },
+          },
+        ],
+        // FIXME Forced cast below due to an incompatibility between the webpack
+        // FIXME version refed in @types/copy-webpack-plugin and our local
```
greg-nagy, line 136: Can we align the types with the actual package?

greg-nagy (dismissed):
I love this ❤️

The only thing I would like to ask is some documentation about the usage in the readme. 
But that can be a separate PR.

Tested and works well 🤩
Love your webpack magic skills. 

Oh, and please make sure you demo this on Friday!

Edit:
I decided not to merge because of the `REACT_DEVTOOLS_DEFAULT_PORT` variable  request, but left the approve.

webpack.config.ts
hyphenized: Good observation! I've added an env var and access to `dot-env` context from inside webpack

0xDaedalus (approved):

PR #2507, opened 2022-10-25
Fix tokenBalance filter to match alchemy API spec

0xDaedalus:
We were not filtering `0` balances correctly when fetching from alchemy - this PR correctly filters based on the shape of the response we get from alchemy.

### To Test
- [ ] Drop a console.log [here](https://github.com/tallycash/extension/blob/update-tokenbalance-filter/background/lib/alchemy.ts#L200:L200)
- [ ] Upon extension reload - ensure that no zero balances are logged.

Latest build: [extension-builds-2507](https://github.com/tallycash/extension/suites/9120639170/artifacts/423373959) (as of Thu, 03 Nov 2022 23:07:36 GMT).


background/lib/alchemy.ts
```diff
@@ -191,7 +191,11 @@ export async function getTokenBalances(
       // A hex value of 0x without any subsequent numbers generally means "no
       // value" (as opposed to 0) in Ethereum implementations, so filter it out
       // as effectively undefined.
-      .filter(({ tokenBalance }) => tokenBalance !== "0x")
+      .filter(
+        ({ tokenBalance }) =>
+          tokenBalance !==
+          "0x0000000000000000000000000000000000000000000000000000000000000000"
+      )
```
jagodarybacka, line 198: Maybe something more like:
```suggestion
      .filter(
        ({ tokenBalance }) =>
          Number(tokenBalance) !== 0
      )
```

jagodarybacka (dismissed):
Some tokens on the list have "bigger" zeros 😉 
image

background/lib/alchemy.ts
0xDaedalus: > Some tokens on the list have "bigger" zeros 😉

My eyes are bleeding

jagodarybacka:
I've discovered one more case :( 
image

background/lib/alchemy.ts
```diff
@@ -191,7 +191,7 @@ export async function getTokenBalances(
       // A hex value of 0x without any subsequent numbers generally means "no
       // value" (as opposed to 0) in Ethereum implementations, so filter it out
       // as effectively undefined.
-      .filter(({ tokenBalance }) => tokenBalance !== "0x")
+      .filter(({ tokenBalance }) => Number(tokenBalance) !== 0)
```
jagodarybacka, line 194:
```suggestion
      .filter(({ tokenBalance }) => tokenBalance !== "0x" && Number(tokenBalance) !== 0)
```

Shadowfiend:
I'm maybe 80% sure this is going back to a filter I already put in and that got undone a while ago... Any chance y'all have dug or can dig into commit history for this file and make sure we're not going in circles trying to fix something weird.

Shadowfiend:
Context: https://github.com/tallycash/extension/pull/500

Are we sure this is the place to filter 0 balances? Or should we be filtering elsewhere to explicitly decide we don't want 0 balances, vs undefined balances?

greg-nagy:
We are filtering these values anyway [here](https://github.com/tallycash/extension/blob/914596a398065a687b109671ba1df8b63b9141cd/background/services/indexing/index.ts#L490) so I think this is ok

We have talked about tracking assets that the user has ever had a balance with, but I think we left that idea behind.

greg-nagy (approved):
I think this is good to go

Shadowfiend:
> We have talked about tracking assets that the user has ever had a balance with, but I think we left that idea behind.

Fwiw my question was less about whether we want to *track* the 0 balances and more about where to draw the semantic distinction between 0 and "undefined" balances. If the Alchemy API distinguishes, it feels like our helper for it should also distinguish. If we want to then ignore those for the wallet, we can do it when we call the helper. This makes it so the logic that ignores zero balances can be in one place, perhaps annotated by a comment indicating that we can't reliably distinguish across lookup methods between a 0 balance and a token a user has never held. 

Also worth noting this is pretty nit picky, and I'm not trying to block any merges :)

greg-nagy (approved):
let's 🚢

greg-nagy:
> > We have talked about tracking assets that the user has ever had a balance with, but I think we left that idea behind.
> 
> Fwiw my question was less about whether we want to _track_ the 0 balances and more about where to draw the semantic distinction between 0 and "undefined" balances. If the Alchemy API distinguishes, it feels like our helper for it should also distinguish. If we want to then ignore those for the wallet, we can do it when we call the helper. This makes it so the logic that ignores zero balances can be in one place, perhaps annotated by a comment indicating that we can't reliably distinguish across lookup methods between a 0 balance and a token a user has never held.
> 
> Also worth noting this is pretty nit picky, and I'm not trying to block any merges :)

I am with @0xDaedalus here. I prefer to filter out the data we don't need as soon as possible. I would expect any and every data connector for us to work this way. Although I also understand your point. I would go with that approach if we would also want to develop a generic data connector that would be used in different projects. In that case, to be true to the underlying api is more important.

PR #2508, opened 2022-10-26
Fix insufficient balance error overlapping with currency amount

### LINKED ISSUE #2459: ,,Insufficient balance'' message is overlapping on the value in dollars
### Discord Discussion Link

_No response_

Chrome

Try to send more value than is in the account balance.

### What did not work?

The ,Insufficient balance'' message is overlapping on the value in dollars. 

Screenshot 2022-10-19 at 16 20 58

### Version

_No response_

### Relevant log output

_No response_

hyphenized:
Closes #2459

mypungamehasbeenweaklately

## To Test: 
- [ ] Enter an amount higher than your available balance, the currency amount should no longer display if there's an error


jagodarybacka (changes requested):
Can you fix the padding/margin so the "send to" field is not moving?

hyphenized:
> Can you fix the padding/margin so the "send to" field is not moving?

Good catch! Should be good now

0xDaedalus (approved):
looks 💯 to me

jagodarybacka (approved):
Looks good ✅

Shadowfiend:
Are we okay with the fact that this means if you have insufficient balance you can't see the balance you do have to adjust accordingly? Cc @VladUXUI

PR #2511, opened 2022-10-26
Cleaner handling of Generic / Alchemy RPC Providers

0xDaedalus:
This PR replaces the existing pattern of determining which url to use for a network:

```
network.chainID === RSK.chainID
        ? new JsonRpcProvider("https://public-node.rsk.co")
        : new AlchemyProvider(getNetwork(Number(network.chainID)), ALCHEMY_KEY)
```

With one that should - in the short term - scale to any networks that we choose to add - regardless of if they are supported by alchemy or not.  It also adds an extra fallback provider for each network we support (except arbitrum) so that we are somewhat protected in the case that alchemy experiences an outage.

## Highlights
- Wallet can now continue to function even if alchemy goes down
- Removed deprecated testnets
- Further centralized network-based constants (in preparation for future move to indexedDB)
- Readable multicall ABI

Things like per-request-routing-to-providers and per-provider-message-queueing are definitely something we will build towards - but are out of the scope of this PR.

### To Test
- [ ] Verify that the URLs in `CHAIN_ID_TO_RPC_URLS` are indeed proper RPC providers
- [ ] Switch the order of:
```
    ...alchemyProviderCreators,
    ...genericProviders,
```

to

```
    ...genericProviders,
    ...alchemyProviderCreators,
```
so that the extension prefers connecting to the public rpc endpoints.
- [ ] Load up the extension - click through to different networks - things like ENS name resolution, token balances, and transaction signing should work ✅


background/lib/erc20.ts
```diff
@@ -196,36 +196,27 @@ export const getTokenBalances = async (
...
-        },
-      }
-    })
-  } catch (e) {
-    // @TODO Handle failure case here for networks that don't have multicall deployed
-    // (e.g. local hardhat networks, brand new networks, etc..)
-    throw e
-  }
+  const response = (await contract.callStatic.tryBlockAndAggregate(
+    // false === don't require all calls to succeed
```
greg-nagy, line 200: ❤️

greg-nagy (approved):
love this warm and tidy feeling after this PR 

🛌

PR #2512, opened 2022-10-26
Cache provider networks.

0xDaedalus:
Since our architecture follows a pattern of using distinct provider instances per network - and we know that a given provider will never switch its network (rather - we will switch the provider the extension is using) - we can avoid eth_chainId RPC calls once the initial call is made and cached.

### To Test
- [ ] Open network tab
- [ ] Click the alchemy eth websocket connection => messages
- [ ] In the messages logs -  should observe one `eth_chainId` at provider startup and then none thereafter.


background/services/chain/serial-fallback-provider.ts
```diff
@@ -169,14 +170,29 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
+   * Since our architecture follows a pattern of using distinct provider instances
+   * per network - and we know that a given provider will never switch its network
+   * (rather - we will switch the provider the extension is using) - we can avoid
+   * eth_chainId RPC calls once the initial call is made and cached.
+   */
+  override async getNetwork(): Promise<Network> {
+    // eslint-disable-next-line no-underscore-dangle
+    if (this._network) {
+      // eslint-disable-next-line no-underscore-dangle
+      return Promise.resolve(this._network)
```
Shadowfiend, line 183: I'm a little nervous that this relies on a field that's not guaranteed to be stable across ethers versions. Might it be more future proof to implement our cache in `send` when handling `eth_chainId`?

0xDaedalus:
Should be good for another look ✅

tsconfig.json
```diff
@@ -8,7 +8,8 @@
     "jsx": "react-jsx",
     "strict": true,
     "resolveJsonModule": true,
-    "noEmit": true
+    "noEmit": true,
+    "noImplicitOverride": true
```
greg-nagy, line 12: ❤️

I am sure I will hate this at one point but this is also 🔥

pain that we love

greg-nagy (commented):
Love the override change.

We might be able to use ethers.js but if it doesn't feel correct than this is good 

Edit: lol, re this commit comment 

`Rather than rely on potentially shifting internals of `ethers` - this approach explicitly caches the chainId in a way that should be durable across future minor or major updates to ethers.`

I would have used ethers but I guess we are all paranoid in different places. My default is to always accept the more paranoid version.

greg-nagy (approved):
let's be paranoid together :)

PR #2514, opened 2022-10-27
rename L1 gas lookup naming > optimism specific

greg-nagy:
The L1 price lookup through price oracle implementation is Optimism specific, Arbitrum is using a different method. It makes more sense to keep it specific until we have another network with the same

## Test
- [x] gas lookup should work on all networks
- [x] send and swap should work on optimism


jagodarybacka (approved):
All good! ✅

PR #2516, opened 2022-10-27
❤️‍🩹 Nonces vs Mempool: Separate nonce behavior between chains with and without mempool

### LINKED ISSUE #2434: Nonce issue in case of approve + spend transaction
### Discord Discussion Link

na

Chrome

Swap on Velodrome, buy NFTs on looksrare

### What did not work?

**Looksrare:**
Try to buy two NFTs in a row on looksrare. 
- The first purchase succeeds (infinite approve spend + actual buying the nft)
- The second attempt (skips approve) will result in a 'the user rejected the request' error

**Velodrome**
Try to swap assets
- First attempt succeeded
- Second attempt, the 'user rejected the request' error

I believe it's a nonce handling issue, based on the logs.

[logs_v0_17_2__2022-10-16T09_23_00+02_00.txt](https://github.com/tallycash/extension/files/9793818/logs_v0_17_2__2022-10-16T09_23_00%2B02_00.txt)

### Version

v0.17.1

### Relevant log output

_No response_

greg-nagy:
## What 

If a chain doesn't have mempool then
the tx is either picked up or dropped immediately. Eg, on optimism
and arbitrum. This means that we should rely on the nonce that
we got from the chain, and no need for us to track lasSeenNonce.

The same distinction should also happen when releasing nonces.

Note: more robust nonce management that take into account
the nonces in the mempool and the fact that they might not be
linear will be a follow-up work.

## Future work 
- update eip-1559 nonce handling to consider pending txs 
- use `SerialFallbackProvider` when broadcasting tx

## Test
- [ ] send and swap on mainnet, polygon, optimism and arbitrum and they should work
- [ ] swap multiple times on Velodrome
- [ ] buy multiple NFTs on looksrare

Closes #2434


background/services/chain/index.ts
```diff
@@ -675,39 +675,54 @@ export default class ChainService extends BaseService<Events> {
...
+    // https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-getTransactionCount
+    let knownNextNonce = await provider.getTransactionCount(
+      transactionRequest.from,
+      "latest"
     )
 
+    // existingNonce handling only needed when there is a chance for it to
+    // be different from the onchain nonce. This can only happen if a chain has
+    // mempool or in other words implements EIP-1559.
+    if (EIP_1559_COMPLIANT_CHAIN_IDS.has(chainID)) {
```
Shadowfiend, line 687: This is a different thing, right? A chain can have a mempool and not be EIP-1559 compliant.

background/services/chain/index.ts
```diff
@@ -1034,7 +1052,6 @@ export default class ChainService extends BaseService<Events> {
               { ...transaction, status: 0, error: error.toString() },
               "alchemy"
             )
-            this.releaseEVMTransactionNonce(transaction)
```
Shadowfiend, line 1037: May want to leave a comment here that returning `Promise.reject` should release the associated nonce in the `catch` below.

background/services/chain/index.ts
```diff
@@ -675,39 +675,54 @@ export default class ChainService extends BaseService<Events> {
...
-
-    logger.debug(
-      "Got chain nonce",
-      chainNonce,
-      "existing nonce",
-      existingNonce,
-      "using",
-      knownNextNonce
+    // https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-getTransactionCount
+    let knownNextNonce = await provider.getTransactionCount(
```
Shadowfiend, line 679: This rename isn't really accurate. This isn't the known next nonce at this stage, it's speculative and depends on network.

I would suggest using `const chainTransactionCount` here, and keeping the naming below.

background/services/chain/index.ts
greg-nagy: https://github.com/tallycash/extension/pull/2516/commits/8e7ee962d8342cb7bd8204a5c819facc9cd7ead2

background/services/chain/index.ts
greg-nagy: https://github.com/tallycash/extension/pull/2516/commits/b533d1adbdd1f226143fa7c582cb0428ac50594f

background/services/chain/index.ts
greg-nagy: https://github.com/tallycash/extension/pull/2516/commits/d7afd22bd4c0c51f20a902e24fdae4cd4bf47b57

background/services/chain/index.ts
```diff
@@ -675,42 +676,58 @@ export default class ChainService extends BaseService<Events> {
...
+    // https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-getTransactionCount
+    const chainTransactionCount = await provider.getTransactionCount(
+      transactionRequest.from,
+      "latest"
     )
+    let knownNextNonce
+
+    // existingNonce handling only needed when there is a chance for it to
+    // be different from the onchain nonce. This can only happen if a chain has
+    // mempool or in other words implements EIP-1559.
```
jagodarybacka, line 688: `chain has mempool or in other words implements EIP-1559.` - is this correct? From @Shadowfiend comment - not every chain with mempool implements EIP1559. 
Does it work the other way around and do we have examples of such chains?

background/services/chain/index.ts
```diff
@@ -676,42 +677,59 @@ export default class ChainService extends BaseService<Events> {
...
-
-    logger.debug(
-      "Got chain nonce",
-      chainNonce,
-      "existing nonce",
-      existingNonce,
-      "using",
-      knownNextNonce
+    // https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-getTransactionCount
+    const chainTransactionCount = await provider.getTransactionCount(
```
0xDaedalus, line 681: Would it make sense to add the # of transactions returned from a call to [`alchemy_getPendingTransactions`](https://docs.alchemy.com/reference/alchemy-pendingtransactions) with a `fromAddress` of `transactionRequest.from` to get the `chainTransactionCount`?  (Assuming they are submitted with valid increasing nonces)

background/services/chain/tests/index.integration.test.ts
```diff
@@ -75,32 +77,40 @@ describe("ChainService", () => {
...
       validOptimismEVMTransaction
     )
+
+    // provider.once should be called inside of subscribeToTransactionConfirmation
+    // with the transaction hash
+    expect(onceSpy.called).toBe(true)
+
+    // on EIP-1559 chains we are not using evmChainLastSeenNoncesByNormalizedAddress
+    // but rely on nonce that we got from chain
+    if (!CHAINS_WITH_MEMPOOL.has(chainID)) {
```
0xDaedalus, line 97: I believe this will always return early and the `expect`'s following it will never be called - since the chainID will always be 10 - right?  I believe we'll still want to check that we persist the tx in the db - and we should definitely write a second test for nonce management on EIP-1559 chains (should mostly be the same as this one).

0xDaedalus (changes requested):
Looks fantastic - only blocker from me is the test comment - I think the way its currently written we're decreasing our test coverage slightly since the test is no longer testing tx persistance.

background/services/chain/index.ts
greg-nagy: The plan is similar but in a follow-up PR and without binding this to websockets. 

TL;DR: it's seems possible with `eth_getTransactionCount` / `provider.getTransactionCount()` but either ethers.js typing is lying to us or we need to update the version. 

https://github.com/ethers-io/ethers.js/discussions/3470 - the idea is similar to this, but this is not perfect either.
[`eth_getTransactionCount` on alchemy](https://docs.alchemy.com/reference/eth-gettransactioncount-astar) does support this. See `2. String - Integer block number, or the string 'latest', 'earliest' or 'pending'.` 

This has a chance to be the next heterodox idea of mine, so I definitely want to separate this from the current PR.

0xDaedalus: Seems like the `eth_getTransactionCount` with `pending` is just strictly a better solution here since it does not couple us to alchemy.

background/services/chain/tests/index.integration.test.ts
greg-nagy: [7cc1d5a](https://github.com/tallycash/extension/pull/2516/commits/7cc1d5afc2228a3ddb0f019bffc9b1c5a66df171)

greg-nagy: The original idea was to update tests after the pending changes are in place. But appreciate the pushback. Focusing on tests is always worth it.

PR #2517, opened 2022-10-27
This seems like an accident?

puppycodes:
remove some extra file named 1 in the root dir

### DIFF
--- a/1
+++ b/1
@@ -1,1 +1,0 @@
-/opt/homebrew/bin/yq


mhluongo:
Heh yep! I commented [in the original commit](https://github.com/tallycash/extension/commit/4ad80b2b5c1e3c11c30f5c556e600dcbc7383242#r86691369) but forgot to follow up

greg-nagy:
a missed shoutout :(

PR #2520, opened 2022-10-28
WIP: Requests over Port

hyphenized:
Description pending...


mhluongo:
If you're looking to replace Redux in the middle, this probably deserves an RFC?

Shadowfiend:
Going to close this for now as it's not really related to any of our core goals; we can always restore the branch + reopen the PR, or pull into a new one, if and when we want to pick it back up.

PR #2523, opened 2022-10-29
Allow renaming account sections

### LINKED ISSUE #2239: Ability to re-name account sections
### Discord Discussion Link

n/a

Be able to re-name the sections of accounts. 

Screen Shot 2022-09-19 at 09 49 12

hyphenized:
Closes #2239

## To Test:
- [ ] Create an account through the wallet, import an account, a ledger account and add a read-only account
- [ ] Go to the accounts panel, you should be able to change the name for every account group except read only accounts
- [ ] Add/delete derived addresses, name should not change
- [ ] Remove and Import the same account, the previous name should not have been persisted 
- [ ] Refactored dropdown menu should work 
- [ ] A warning should appear after attempting to remove the last address for a ledger account

Latest build: [extension-builds-2523](https://github.com/tallycash/extension/suites/9076343046/artifacts/420172390) (as of Tue, 01 Nov 2022 21:00:27 GMT).


background/services/name/resolvers/uns.ts
```diff
@@ -120,6 +120,7 @@ const isValidUNSDomainName = (s: string): boolean => {
       ".dao",
       ".zil",
       ".x",
+      ".blockchain",
```
0xDaedalus, line 123: 📈

ui/components/AccountItem/AccountItemOptionsMenu.tsx
```diff
@@ -1,11 +1,11 @@
 import { AccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
 import { setSnackbarMessage } from "@tallyho/tally-background/redux-slices/ui"
-import React, { ReactElement, useCallback, useRef, useState } from "react"
+import React, { ReactElement, useCallback, useState } from "react"
 import { useTranslation } from "react-i18next"
-import { useBackgroundDispatch, useOnClickOutside } from "../../hooks"
+import { useBackgroundDispatch } from "../../hooks"
+import { DropdownMenu } from "../Shared/SharedDropDown"
```
0xDaedalus, line 6: I think we have a more-or-less consistent pattern of calling using filename in component names of components from the `/Shared` directory.  So maybe `SharedDropDown`?

ui/components/AccountItem/AccountItemOptionsMenu.tsx
```diff
@@ -77,102 +72,47 @@ export default function AccountItemOptionsMenu({
...
+            className="icon_settings"
+            role="menu"
+            tabIndex={0}
+          />
+        }
+        options={[
+          {
+            key: "edit",
+            icon: "icons/s/edit.svg",
+            label: "Edit name",
```
0xDaedalus, line 88: `t('editName')`

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
```diff
@@ -33,6 +40,10 @@ import SharedAccountItemSummary from "../Shared/SharedAccountItemSummary"
 import AccountItemOptionsMenu from "../AccountItem/AccountItemOptionsMenu"
 import { i18n } from "../../_locales/i18n"
 import SharedIcon from "../Shared/SharedIcon"
+import { DropdownMenu } from "../Shared/SharedDropDown"
```
0xDaedalus, line 43: `SharedDropDown`

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
```diff
@@ -63,48 +74,271 @@ const walletTypeDetails: { [key in AccountType]: WalletTypeInfo } = {
   },
 }
 
+const EditSectionForm = ({
```
0xDaedalus, line 77: I think this can probably be in its own file.

0xDaedalus: We also prefer to enforce the `function EditSectionForm() {}` syntax over the `const EditSectionForm = () => {}` syntax.

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
```diff
@@ -63,48 +74,271 @@ const walletTypeDetails: { [key in AccountType]: WalletTypeInfo } = {
...
+      event:
+        | React.FormEvent<HTMLFormElement>
+        | React.MouseEvent<HTMLButtonElement, MouseEvent>
+    ) => {
+      event.preventDefault()
+      if (!newName) {
+        setTouched(true)
+        setError(t("noNameError"))
+      }
+      if (error) {
```
0xDaedalus, line 115: This won't catch an error set on line 113 since the `onSubmit` function will finish executing before the changes from `setError` are applied.

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
```diff
@@ -63,48 +74,271 @@ const walletTypeDetails: { [key in AccountType]: WalletTypeInfo } = {
...
+            }}
+          />
+        </form>
+      </div>
+      <div className="button_container">
+        <SharedButton
+          type="secondary"
+          size="medium"
+          onClick={() => {
+            onSubmitCallback(null)
```
0xDaedalus, line 165: It doesn't feel right to me that we're calling a function called `onSubmitCallback` on an action that is not a submission (clicking cancel).  Do you think it might make sense to just pass a separate `onCancelCallback` to this component instead?

ui/components/Shared/SharedDropDown.tsx
```diff
@@ -0,0 +1,255 @@
...
+              <button
+                type="button"
+                className="close_button"
+                aria-label="Close"
+                onClick={() => toggle()}
+              >
+                <div className="icon_close" />
+              </button>
+              <ul className="options">
+                {(options.filter(Boolean) as DropdownOption[]).map(
```
0xDaedalus, line 157: If we enforce correct options through types (`options: Array<DropdownOption>`) we don't have to filter here.  I'd like to avoid `x as y` whenever possible - its historically been a big source of bugs for us.

ui/components/Shared/SharedDropDown.tsx
```diff
@@ -0,0 +1,255 @@
...
+import AccountitemOptionLabel from "../AccountItem/AccountItemOptionLabel"
+
+type DropdownContextValue = {
+  isOpen: boolean
+  toggle: (value?: boolean) => void
+  styles: React.CSSProperties
+}
+
+// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+const DropdownContext = React.createContext<DropdownContextValue>(null!)
```
0xDaedalus, line 12: Is there any way around the non-null assertion here?  If not lets make sure to comment exactly why we're disabling the eslint rule

0xDaedalus (changes requested):
This looks awesome!  Took a quick look at the code but haven't had a chance to dive in and test yet.

One thing I did notice in the video is we probably want the cursor to change to a `pointer` when hovering over the settings cog.

Can-not-wait to get this in!

ui/components/Shared/SharedDropDown.tsx
hyphenized: Excellent catch! Replaced with a type guard

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
hyphenized: Thanks, good catch. By the way, is the error supposed to look like this?: cc @VladUXUI 


image

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
hyphenized: You're right, this might also end up confusing devs working on this

ui/components/Shared/SharedDropDown.tsx
hyphenized: This was an exception because the context is only accessed from within the Dropdown components and these only function when they have a non null context value. I've added a null safe check to the access hook so we don't have to disable a rule here.

0xDaedalus:
Spent a bit of time testing - some things I found:
- Importing a new keyring after naming a previously imported keyring resets all names
- Adding an address to a named keyring resets all keyring names.
- Should we enforce unique keyring names?  Currently its possible for me to name all of my accounts the same name e.g. `WALLET1` (cc @VladUXUI)

Also a general note about commit messages - in general we strive to write commit messages that are expressive as possible (after all we are an open source project) and tend to avoid labels like feat/chore/fix/ etc...  @Shadowfiend's PRs are by far the best example of this (I myself am still constantly working to improve the quality of my commit messages) and I encourage you to take a look at them ([for example this one](https://github.com/tallycash/extension/pull/2312/commits)).

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
VladUXUI: Hey hey, overall is looking awesome! 
But no, the error is under the input field

VladUXUI:
> * Should we enforce unique keyring names?  Currently its possible for me to name all of my accounts the same name e.g. `WALLET1` (cc @VladUXUI)

We aren't using this data to do anything really, so from a users perspective it isn't important to do so.
If it somehow helps code wise, that is up to you. but UX wise it's not a plus or a minus :)

ui/components/Shared/SharedDropDown.tsx
```diff
@@ -0,0 +1,270 @@
...
+    | React.ReactNode
+    | ((dropDownContext: {
+        toggle: DropdownContextValue["toggle"]
+      }) => React.ReactNode)
+}
+
+function Container({ children }: DropdownContainerProps): ReactElement {
+  const [isOpen, setIsOpen] = useState(false)
+
+  const styles: React.CSSProperties = {
```
jagodarybacka, line 40: I'm not sure about using another type of styles. I don't like `styled-jsx` package but I would rather use it consistently where it is possible.

ui/components/Shared/SharedDropDown.tsx
```diff
@@ -0,0 +1,270 @@
...
+    <Dropdown.Container>
+      <Dropdown.Toggler>{toggler}</Dropdown.Toggler>
+      <Dropdown.Content>
+        {({ toggle }) => (
+          <div style={{ position: "relative" }}>
+            <div className="options">
+              <button
+                type="button"
+                className="close_button"
+                aria-label="Close"
```
jagodarybacka, line 165: Do we want to add aria label to translations?

ui/components/Shared/SharedDropDown.tsx
```diff
@@ -0,0 +1,270 @@
...
+               line-height: 24px;
+               padding: 7px;
+               flex-direction: row;
+               width: 90%;
+               align-items: center;
+               height: 100%;
+               cursor: default;
+               justify-content: space-between;
+             }
+             .icon_close {
```
jagodarybacka, line 250: Can we use `SharedIcon` component here?

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
```diff
@@ -67,44 +77,97 @@ function WalletTypeHeader({
...
 }) {
   const { t } = useTranslation()
   const { title, icon } = walletTypeDetails[accountType]
+  const dispatch = useBackgroundDispatch()
+  const keyringMetadata = useBackgroundSelector(
+    (state) => state.keyrings.keyringMetadata
+  )
+
+  const sectionCustomName = keyringId
+    ? keyringMetadata[keyringId].customName
```
jagodarybacka, line 95: Let's make sure it won't throw when keyringMetadata will be undefined for some reason
```suggestion
    ? keyringMetadata[keyringId]?.customName
```

Shadowfiend (commented):
I'm extremely nervous that the preferences service isn't showing up here. That's where this stuff should get persisted, because redux isn't meant to be the core persistence mechanism for the extension, indexeddb is 😬 This is why account names are ultimately persisted there (and why default names also should be---something I skimped on for time reasons a while ago but which is already biting us in local name resolution).

hyphenized:
> I'm extremely nervous that the preferences service isn't showing up here. That's where this stuff should get persisted, because redux isn't meant to be the core persistence mechanism for the extension, indexeddb is 😬 This is why account names are ultimately persisted there (and why default names also should be---something I skimped on for time reasons a while ago but which is already biting us in local name resolution).

I see, I chose to store this on Redux because I didn't consider this data to be important enough to save it inside idb... 😕

Shadowfiend:
I think the current approach a reasonable starting point and probably worth merging, maybe behind a feature flag or maybe not. But ultimately all data in the system should either able to be looked up from a service or from IDB, by design. As an example of why we've approached things with this model in mind despite a lot of stuff living untouched in redux right now,  imagine a world where we only load account group metadata when the drop-down is displayed to minimize redux store size, or to accelerate time to first render.  Another version of this, of course, is the redux store only existing in the popup script (though there are a bunch of additional complications with that approach). The rule of thumb though is that anything we expect to persist between browser restarts ultimately should live in IDB. 

Like I mentioned, we've cut some corners here in the past---and if we're in a hurry to get this out it may be reasonable to cut this corner here too---but I wanted to make sure to articulate the underlying reasoning and what might be an ideal way to approach the solution here, so there's a sense that there is room for further improvement to the solution as it currently stands.

ui/components/AccountsNotificationPanel/AccountsNotificationPanelAccounts.tsx
hyphenized: I'll follow up with a fix for this in a different issue since I might have to check other places where this input is being used to make sure it's good

ui/components/Shared/SharedDropDown.tsx
hyphenized: > Spent a bit of time testing - some things I found:
> 
>     * Importing a new keyring after naming a previously imported keyring resets all names
> 
>     * Adding an address to a named keyring resets all keyring names.
> 
>     * Should we enforce unique keyring names?  Currently its possible for me to name all of my accounts the same name e.g. `WALLET1` (cc @VladUXUI)
> 
> 
> Also a general note about commit messages - in general we strive to write commit messages that are as expressive as possible (after all we are an open source project) and tend to avoid labels like feat/chore/fix/ etc... @Shadowfiend's PRs are by far the best example of this (I myself am still constantly working to improve the quality of my commit messages) and I encourage you to take a look at them ([for example this one](https://github.com/tallycash/extension/pull/2312/commits)).

Sorry for the late response, I agree, these commits are not really giving any context at all and are actually making it more difficult to review and understand the changes being made.  Looking back at [this comment](https://github.com/tallycash/extension/pull/2220#discussion_r972971856) I think I should have realized this sooner, thank you for the feedback!

hyphenized:
@0xDaedalus Sorry for the late response, I agree, these commits are not really giving any context at all and are actually making it more difficult to review and understand the changes being made. Looking back at this https://github.com/tallycash/extension/pull/2220#discussion_r972971856, I think I should have realized this sooner, thank you for the feedback!

@Shadowfiend Thanks for taking the time to explain, I think this was a bad assumption on my end because we are also trying to reduce the size of the Redux store. I've moved the metadata to the preferences db, but, there's still a piece of data in the Redux store that updates as a response to changes in the db. I feel like this is probable not the best approach to pass data down to the UI and have it react to changes, but, it's the most straightforward way I found to do it. 

> imagine a world where we only load account group metadata when the drop-down is displayed to minimize redux store size, or to accelerate time to first render

I do hope we can see something like this in the future

ui/components/Shared/SharedDropDown.tsx
hyphenized: Me neither, the idea here was to pass different styles based on some positioning prop but that ended up not being necessary

ui/components/Shared/SharedDropDown.tsx
hyphenized: Si

ui/components/Shared/SharedDropDown.tsx
hyphenized: We can!

background/redux-slices/ui.ts
```diff
@@ -179,6 +190,19 @@ export const setNewSelectedAccount = createBackgroundAsyncThunk(
   }
 )
 
+export const updateSignerSettings = createBackgroundAsyncThunk(
+  "ui/updateSignerSettings",
+  async (
+    [signer, title]: [
+      Exclude<AccountSigner, typeof ReadOnlyAccountSigner>,
```
0xDaedalus, line 197: Wonder if this should be extracted into its own type since its used in quite a few files

background/services/preferences/db.ts
```diff
@@ -233,6 +251,43 @@ export class PreferenceDatabase extends Dexie {
       .toCollection()
       .modify({ selectedAccount: addressNetwork })
   }
+
+  async updateSignerSettings(
```
0xDaedalus, line 255: The function signature here suggests that we might want to call this function `updateSignerTitle`

background/services/preferences/db.ts
```diff
@@ -233,6 +251,43 @@ export class PreferenceDatabase extends Dexie {
...
+    signer: Exclude<AccountSigner, { type: "read-only" }>,
+    title: string
+  ): Promise<AccountSignerSettings[]> {
+    const { accountSignersSettings } = await this.getPreferences()
+
+    const signerUISettings = accountSignersSettings.find(
+      ({ signer: storedSigner }) => isEqual(storedSigner, signer)
```
0xDaedalus, line 262: I'm admittedly not the biggest fan of lodash - but I do wonder if checking `keyringId` or `ledgerId` is enough of an equality comparison to where we can avoid using `isEqual` (assuming we keep `accountSignerSettings` as a property in the `preferences` table.

background/services/preferences/db.ts
```diff
@@ -208,6 +212,20 @@ export class PreferenceDatabase extends Dexie {
           })
       })
 
+    this.version(10)
+      .stores({
+        preferences: "++id",
```
0xDaedalus, line 217: I think we might be better served by defining a separate `accountSignerSettings` table here rather than appending another property into the `preferences` table.  Reasoning is:

- We can use the ledger/keyring id of a signer as a unique primary key to have a stronger guarantee we won't have duplicate accountSignerSettings.
- We can query / update / delete by that id - this should greatly simplify the `delete` logic.
- We won't need to stringify the entire preferences object and then query further in memory to access the account signer setting that we want.
- Much easier to observe / debug in devtools

0xDaedalus (commented):
Looks great and is certainly mergeable but curious to hear your thoughts on possibly creating a new table for the settings.

background/services/preferences/db.ts
hyphenized: Yes we can, I added a check without `isEqual` on the Header component, I'll refactor that out to `utils`

background/redux-slices/ui.ts
hyphenized: It is! Refactored into its own type!

background/services/preferences/db.ts
hyphenized: Future proofing too much 😅

background/services/preferences/db.ts
hyphenized: These are valid points, I don't believe we'll see a performance improvement here but it has definitely simplified the flow a lot. Great suggestion!

background/services/preferences/db.ts
```diff
@@ -6,7 +6,17 @@ import { AddressOnNetwork } from "../../accounts"
 import DEFAULT_PREFERENCES from "./defaults"
 import { AccountSignerSettings } from "../../ui"
 import { AccountSignerWithId } from "../../signing"
-import { isSameAccountSignerWithId } from "../../utils/signing"
+
+/**
+ * Returns a unique id for an account signer
+ * in the form of "signerType/someId" e.g. "ledger/deviceId"
+ */
+const getSignerRecordId = (
```
0xDaedalus, line 14: Not sure if we need this because i _believe_ we already prefix the ledger id with something like LEDGER_NANO_X_${id} but definitely better safe than sorry

0xDaedalus (approved):
👑 👑 👑 👑 👑 👑 👑 👑 

Amazing work!

Lets ship this bad boy.

PR #2526, opened 2022-10-31
Add `checksum` validation

ahsan-javaiid:
## Description
- This PR adds `checksum` validation when user enter address on `send` screen
- Purpose is to prevent users from sending funds to the wrong address
- Validation logic is provided by `@tallyho/hd-keyring` library which is already in use

## Demo 

Screenshot 2022-10-31 at 7 47 49 PM

## Type of Change
- [X] Address validation

## Testing information
- Activate `SUPPORT_RSK` env flag
- Go to send screen
- Enter the address according to network
- It should validate checksum according to network

## Checklist
- [X] Git hooks enabled
- [X] No lint errors
- [X] Git commit is signed

Latest build: [extension-builds-2526](https://github.com/tallycash/extension/suites/9127807837/artifacts/423937094) (as of Fri, 04 Nov 2022 10:34:49 GMT).


ahsan-javaiid:
cc: @0xDaedalus @alepc253

0xDaedalus:
I don't believe the functionality here is working as expected - I seem to be getting `bad checksum` errors on valid addresses.  For example if I try to send an asset to `0xd8da6bf26964af9d7eed9e03e53415d37aa96045` (vitalik.eth) on any network I get a `bad checksum` error message.

alepc253:
And, in fact, an address in lower case is valid for both ETH and RSK

ahsan-javaiid:
**Fixed**: 
Added a better UX to resolve this confusion. I think in the case of providing all lower-case or all upper-case, user should be warned, but user will still be able to hit continue button. 

**Demo**: 
Screenshot 2022-11-01 at 12 51 50 PM

**Note**: Now checksum validation will only trigger if address is mixed case.

Thanks for pointing this out: @0xDaedalus 👍

ui/hooks/validation-hooks.ts
```diff
@@ -157,10 +164,19 @@ export const useAddressOrNameValidation: AsyncValidationHook<
       const trimmed = newValue.trim()
 
       setErrorMessage(undefined)
+      setWarningMessage(undefined)
       if (trimmed === "") {
         onValidChange(undefined)
       } else if (isProbablyEVMAddress(trimmed)) {
-        onValidChange({ address: normalizeEVMAddress(trimmed) })
+        if (!isMixedCaseAddress(trimmed)) {
+          setWarningMessage("Address is not mixed-case")
```
Shadowfiend, line 172: I don't think this warning should land as-is without further discussion. 

Checking checksums on mixed case address seems more straightforward, though.

alepc253: I agree. I think lowercase addresses should be just valid without any message (I'm not sure about uppercase, we should double check the EIPs but I tend to say that we shouldn't show the message)

ahsan-javaiid: After checking couple of wallets and reading some links, seems like all lower case or all upper case addresses should be valid and may not need any warning message. 

So i just removed the warning message logic. Most welcome for any other thoughts.

0xDaedalus: Only remaining question here for me is do we want to add a small explainer about what `Bad checksum address` means?  Our demographic of users tends to trend towards more new-to-crypto users so I wonder if this will result "Why can't I send to this address" confusion

ahsan-javaiid: Yes, i think we can show more user friendly message. Couple of messages that comes to my mind. 

1) "be careful, the address doesn't seem to be of the network you want to send to"
2) "Address checksum is invalid"
3) "Not a valid address. Verify on explorer link"
4) "Not a valid RSK checksummed address. Please verify address"
5) "Address is not formatted properly. If you send to wrong address, funds cannot be recovered ". 

[More examples here
](https://www.google.com/search?q=wallet+bad+checksum+address&tbm=isch&ved=2ahUKEwjTlrzgpZL7AhXI_4UKHQXyCbEQ2-cCegQIABAA&oq=wallet+bad+checksum+address&gs_lcp=CgNpbWcQA1CeBFieBGDQB2gAcAB4AIABuQKIAdEEkgEFMi0xLjGYAQCgAQGqAQtnd3Mtd2l6LWltZ8ABAQ&sclient=img&ei=7dljY5PsGcj_lwSF5KeICw&bih=778&biw=1512&rlz=1C5CHFA_enPK1009PK1009)

Any other ideas, most welcome.

VladUXUI:
Awesome pr!
Quick note here from UX side.
The "Address is not mixed-case" is not that accurate if we are only warning for mixed-case.

Also if you are warning a user about something, it's good to also give a bit more context. I would advice we show a tooltip also (we have a component for this) with what the warning is about, with maybe something like:
"Address has a mix of lower and upper case, this is unusual so please make sure the address is correct"

VladUXUI:
Could we also maybe limit this to RSK chain for now? i would like to do a bit more research on it to better understand it before sending it off to all our users

alepc253:
@VladUXUI yes, we can add this only for RSK and see. I would go with something like "Be careful! This address may not be an RSK one"

ahsan-javaiid:
Done!

0xDaedalus (approved):
LGTM 👍

PR #2532, opened 2022-11-01
Gas price bids on Arbitrum

jagodarybacka:
Resolves https://github.com/tallycash/extension/issues/2367
Resolves https://github.com/tallycash/extension/issues/2373
Resolves https://github.com/tallycash/extension/issues/2428

### Background
Arbitrum has its own unique gas rules so we can't reuse EIP-1559 compatible gas settings implementation. This PR is a refactor of our old code, so it is aligned with what Arbitrum transactions expect.

One problem that was not solved here - I could not find a good price oracle for medium and high gas price bids. The one in the first commit has too low rate limit for us. I would like to avoid looking for another paid provider. The manual calculation follows the rules I see in Debank's price oracle.

### What 

This PR fixes problems with the Arbitrum gas selector.
- Allow changing the gas price bid for Arbitrum transactions
- Gas price is calculated manually as we don't have a good price oracle for now. 
- Reuse legacy network settings component

image

### Testing
- setup transaction on Arbitrum (send ETH), check the gas estimate, it should be close to the estimate on [L2Fees](https://l2fees.info/)
- change network settings and select a higher gas option
- send the transaction and check if the gas bid was the one you selected and if it was higher than the price you paid for gas
image

 

Latest build: [extension-builds-2532](https://github.com/tallycash/extension/suites/9073161831/artifacts/419958519) (as of Tue, 01 Nov 2022 17:58:31 GMT).


background/lib/gas.ts
```diff
@@ -80,6 +91,73 @@ const getPolygonGasPrices = async (price: bigint): Promise<BlockPrices> => {
...
 
+const getArbitrumPrices = async (
+  baseFeePerGas: bigint,
+  blockNumber: number
+): Promise<BlockPrices> => {
+  // let estimatedPrices: BlockEstimate[]
+  // try {
+  //   // TODO: this is not good enough - low rate limit
+  //   const { data: gasEstimates } = (await fetchJson(
+  //     "https://api.debank.com/chain/gas_price_dict_v2?chain=arb"
```
jagodarybacka, line 102: I wanted to find a price oracle like we have for Polygon and Eth but I can't find any free one that will be good enough. Not sure how much we care about it.

jagodarybacka:
Commented code should be removed before merging, left it for now as maybe we can came up with some solution during the code review

background/lib/gas.ts
```diff
@@ -80,6 +91,73 @@ const getPolygonGasPrices = async (price: bigint): Promise<BlockPrices> => {
...
+  //   ]
+  //   throw Error("No Arbitrum oracle")
+  // } catch (e) {
+  //   logger.warn(
+  //     "Error getting block prices from Arbitrum, fallback to manual calculation",
+  //     e
+  //   )
+  // }
+
+  return {
```
greg-nagy, line 132: I think this is good enough. 

Most sources give the same gas fee for all 3, and dework seems to have stolen your algorithm for their price estimations :)

background/lib/gas.ts
```diff
@@ -80,6 +91,73 @@ const getPolygonGasPrices = async (price: bigint): Promise<BlockPrices> => {
...
+    network: ARBITRUM_ONE,
+    blockNumber,
+    baseFeePerGas,
+    estimatedTransactionCount: null,
```
greg-nagy, line 136: We can probably drop this property. This is something that blockchain has and we don't really use it anywhere.

background/lib/gas.ts
0xDaedalus: I think this is perfectly OK tbh - oracle shouldn't be required its just potentially a nice-to-have

background/lib/gas.ts
```diff
@@ -80,6 +91,73 @@ const getPolygonGasPrices = async (price: bigint): Promise<BlockPrices> => {
...
+        confidence: 99,
+        maxPriorityFeePerGas: 0n, // priority fee doesn't make sense for Arbitrum
+        maxFeePerGas: 0n, // max fee doesn't make sense for Arbitrum
+        price: baseFeePerGas * 3n,
+      },
+      {
+        confidence: 95,
+        maxPriorityFeePerGas: 0n,
+        maxFeePerGas: 0n,
+        price: baseFeePerGas * 2n,
```
0xDaedalus, line 148: Wonder if we should be less aggressive with our increments here estimates look reasonable when we're at 0.1 gwei but if we go to even single digit gwei's an x2 multiplication might be quite the USD difference.

Perhaps 10% traunches?  Something like `(basefeePerGas * 11n) / 10n`?

background/lib/gas.ts
```diff
@@ -120,6 +198,10 @@ export default async function getBlockPrices(
     }
   }
 
+  if (network.chainID === ARBITRUM_ONE.chainID) {
+    return getArbitrumPrices(baseFeePerGas ?? 0n, currentBlock.number)
```
0xDaedalus, line 202: should the default here be `0.1 gwei` rather than `0n`?

greg-nagy (commented):
I think this is good after the removal of the commented out code blocks 

It seems that it doesn't really matter what we set, the chain will use gas according to what the tx needs.

See sample tx

background/redux-slices/transaction-construction.ts
```diff
@@ -60,7 +60,6 @@ export type TransactionConstruction = {
   transactionLikelyFails: boolean
   estimatedFeesPerGas: { [chainID: string]: EstimatedFeesPerGas | undefined }
   customFeesPerGas?: EstimatedFeesPerGas["custom"]
-  lastGasEstimatesRefreshed: number
```
0xDaedalus, line 63: Why are we getting rid of this?

background/lib/gas.ts
0xDaedalus: Sorry - what I mean is that its perfectly ok to exclude an oracle and this code is OK to remove.

0xDaedalus (changes requested):
Looks like the gas cost prediction is still too high [(I paid $0.27 for a tx that was estimated to be $0.95)](https://arbiscan.io/tx/0x8dd8573f54099abba395b359dbe44913d8a14b0adf20d6cf5f91d1bc3ef901fc) - I imagine its because we're not separating the L1 fee and the L2 fee in our calculations.  Increasing the gwei of the L2 portion of the tx (the only thing we are able to affect on arbitrum) only increases a portion of the gas cost on Arbitrum.  The cost formula is as follows:

[`L2 gas used + ( L1 calldata price * L1 calldata size) / (L2 gas price)`](https://medium.com/offchainlabs/understanding-arbitrum-2-dimensional-fees-fd1d582596c9)

I think this is likely the root cause of us overestimating and I would imagine the way forward here is adjust how we do the dollar cost calculations a bit specifically for Arbitrum (and future Rollups as well).

background/lib/gas.ts
jagodarybacka: I only did that because that's what the debank's API endpoint was doing in the past couple days :P 
10% is I think too low (0.1 gwei, 0.11 gwei...)

jagodarybacka:
> [L2 gas used + ( L1 calldata price * L1 calldata size) / (L2 gas price)](https://medium.com/offchainlabs/understanding-arbitrum-2-dimensional-fees-fd1d582596c9)

I know, I've read all these articles but I don't see where we can get this data from. Looking at the ETH send - our cost calculation is the same as on other websites.

> [(I paid $0.27 for a tx that was estimated to be $0.95)](https://arbiscan.io/tx/0x8dd8573f54099abba395b359dbe44913d8a14b0adf20d6cf5f91d1bc3ef901fc)

I believe that's because you had the "fastest" option selected so 0.3gwei/gas - this was a bid but actually you paid only 0.1gwei/gas as more was not needed and it was refunded.

greg-nagy (approved):
let's 🚢

Resisted the urge to include this in a quick follow up patch release ... but wanted to soooo bad ... :) 
Anyhow @0xDaedalus  it's all yours

0xDaedalus:
Thanks for your careful reviews and responses guys - looks like my confusion did indeed come from the fact that arbitrum will allow you to pay a tx cost lower than your bid (which is interesting...). 

LGTM - lets 🚢

PR #2539, opened 2022-11-01
Smoother Links: Fix a few small issues with the post-build-link workflow

Shadowfiend:
The fixes are:
- Using commit ids for extension artifacts that don't come from PRs. They were getting a suffix of `-Object` due to a misunderstanding of the GitHub ref data structure.
- Skipping the build link stuff if the Build workflow ran for something that wasn't a PR (as determined by the artifact suffix looking like a commit, roughly). If there are issues looking up the build link information for a PR build, we now throw errors instead of logging them and completing successfully.
- Handling null PR descriptions. When PR descriptions are left empty, the value returned by the GitHub API is a `null`. The script now handles that as if it were a blank string, and inserts the build link normally.

Latest build: [extension-builds-2539](https://github.com/tallycash/extension/suites/9079460433/artifacts/420393927) (as of Wed, 02 Nov 2022 01:15:16 GMT).


.github/workflows/builds/post-build-link.js
```diff
@@ -71,8 +76,7 @@ module.exports = async function postBuildLink({ github, context }) {
   })
 
   if (pullLookupStatus !== 200) {
-    console.error("Failed to fetch PR body :( Status", pullLookupStatus, ".")
-    return
+    throw new Error("Failed to fetch PR body :( Status", pullLookupStatus, ".")
```
hyphenized, line 79: This error will show `Failed to fetch PR body :( Status` because the Error constructor doesn't concatenate parameters like `console.error` right?

Shadowfiend: Whoops, well spotted 🤦 Will fix this evening. Surprised TypeScript didn't slap me with a trout on this one, so will check that piece out as well.

Shadowfiend:
Fix't.

.github/workflows/builds/post-build-link.js
Shadowfiend: Fix't @ https://github.com/tallycash/extension/pull/2539/commits/a89b5da7b65ec62e3b580a4120a7f854d3e97f0e

0xDaedalus (approved):
📈

PR #2545, opened 2022-11-02
Redirect to onboarding if user has no accounts

### LINKED ISSUE #2432: Removing last read only address is not redirecting to onboarding
### What
If you remove the last address app should redirect to onboarding but it stays on the wallet page
image

### Steps to reproduce
- install new wallet
- add read only account
- go to swaps page
- remove address

Expected: app should redirect to the onboarding page
Current: app stays in invalid state with removed account

https://user-images.githubusercontent.com/20949277/195885989-22d50357-8dc7-4c28-b47c-fa0380853cf4.mov

hyphenized:
Previously this check was only performed in the main wallet page, leaving users who removed their last account stranded on the other pages until they went back to the wallet page.

Removed Location internal type in favour of the one provided by react router as well

Closes #2432 

## Testing Env

```
SUPPORT_TABBED_ONBOARDING=true
```

To Test: 
----
- [ ] Remove any type of account on the swap page, you should be redirected to onboarding
- [ ] With tabbed onboarding enabled, the popup should close.

Latest build: [extension-builds-2545](https://github.com/tallycash/extension/suites/9195167673/artifacts/428851896) (as of Tue, 08 Nov 2022 17:41:37 GMT).


ui/pages/Popup.tsx
```diff
@@ -200,8 +210,16 @@ export function Main(): ReactElement {
                     >
                       <TopMenu />
                     </div>
-                    {/* @ts-expect-error TODO: fix the typing when the feature works */}
                     <Switch location={transformedLocation}>
+                      {
+                        // If an account doesn't exist, display onboarding
```
jagodarybacka, line 215:
```suggestion
                        // If there are no existing accounts, display onboarding
```

Because if there are accounts loaded then it should switch to another address, not display onboarding

background/redux-slices/selectors/signingSelectors.ts
```diff
@@ -25,25 +32,39 @@ export const selectAccountSignersByAddress = createSelector(
...
+              type: "keyring",
+              keyringID: keyring.id,
+            },
+          ]
+        }
       )
       .filter(isDefined)
 
-    return Object.fromEntries([
+    const readOnlyEntries: [string, typeof ReadOnlyAccountSigner][] =
```
jagodarybacka, line 54: I think now you can refactor this selector too https://github.com/tallycash/extension/blob/main/background/redux-slices/selectors/signingSelectors.ts#L54

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -45,6 +45,11 @@ import {
   ReadOnlyAccountSigner,
   SignerType,
 } from "../../services/signing"
+import getAccountState from "./accounts/getAccountState"
+import getAllAddresses from "./accounts/getAllAddresses"
+
+export { default as getAccountState } from "./accounts/getAccountState"
+export { default as getAllAddresses } from "./accounts/getAllAddresses"
```
jagodarybacka, line 52: I think this is a good direction, I hate these huge files 😂  but can we either refactor whole `accountsSelector` file at once or make a task to not forget about it?

mhluongo:
Not sure this is a good solution considering we're moving to tabbed onboarding.

ui/pages/Popup.tsx
hyphenized: Good catch, I'll update the comment

background/redux-slices/selectors/signingSelectors.ts
hyphenized: Nice call, found a few other places to refactor this as well 🚀

background/redux-slices/selectors/accountsSelectors.ts
hyphenized: We can include that in this PR, though, I wanted to check if the team had anything to say about this before making the change. cc: @tallycash/extension-devs

jagodarybacka:
@mhluongo @VladUXUI so we plan to not include any popup onboarding at all? We have to handle situations like this anyway - the user can either remove all addresses or skip onboarding and open the wallet. I think full onboarding with introduction slides is not necessary, only the "select type of account to add" screen is needed here.

mhluongo:
>  so we plan to not include any popup onboarding at all?

From Element, that's the plan... we can open the tabbed onboarding whenever a wallet is in a non-onboarded state.

That's actually what my feature-flagged implementation does already, though it requires the extension button to be hit. Here, I think we should do the same thing, but it will need an additional trigger when the last account is removed.

mhluongo:
Also generally.... the existing in-extension onboarding code is pretty rough. I very much vote for removing the tech debt and snipping the code once tabbed onboarding is flipped on.

ui/pages/Popup.tsx
```diff
@@ -200,8 +210,22 @@ export function Main(): ReactElement {
                     >
                       <TopMenu />
                     </div>
-                    {/* @ts-expect-error TODO: fix the typing when the feature works */}
                     <Switch location={transformedLocation}>
+                      {
+                        // If there are no existing accounts, display onboarding
+                        // (if we're not there already)
+                        //
+                        // TODO: FeatureFlags.SUPPORT_TABBED_ONBOARDING
```
mhluongo, line 218: @hyphenized maybe you can support `SUPPORT_TABBED_ONBOARDING` now with a `window.open()` to get this to merge-able?

hyphenized: Sure, onboarding tab now opens after the last account is removed

background/redux-slices/selectors/accountsSelectors.ts
greg-nagy: ## Thought for this change to be in this PR:
This change and this discussion should happen outside of this PR. I think the tension is valid and we should discuss this but in a separate conversation. I recommend opening a PR/issue dedicated for the introduction of this pattern and using the [light RFC/propsal style](https://coda.io/d/Engineering-Thesis_duruoW1C74B/RFC-Process_su2dT#_luU0q). For an example [generic RPC](https://github.com/tallycash/extension/pull/2547) or the [CI refactor](https://github.com/tallycash/extension/pull/2477)

## Thoughts for the discussion/idea:
I have felt the pain of putting everything into a separate file in angular land. That can become unwidely very quickly. Breaking up big files can be a good idea if we find the right separations based on a well-defined concern/domain. 

If you/team feels strongly about putting everything into its file, I think we can give it a try. But I feel strongly about not using the default exports for this. I prefer named exports because otherwise, manual renaming will be necessary, and that's error-prone and makes the code harder to follow.  

Another thought: I have seen the domain-based separation working well, which groups code together based on what concern it belongs to and not what type it is. Eg `redux-slices/accounts/` and inside that folder would be the `account.slice.ts`, `account.selectors.ts`, `account.utils.ts`. Note: The duplication in the dir name and file name feels unnecessary at first, but I have found that useful.

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -398,32 +403,9 @@ function getNetworkAccountTotalsByCategory(
...
-        accounts,
-        assets,
-        accountSignersByAddress,
-        keyringsByAddresses,
-        sourcesByAddress,
-        mainCurrencySymbol,
-        network
-      )
-    }
+  (state: RootState) => state,
```
greg-nagy, line 406: Does this change not break the selector memoization/caching? 
https://redux.js.org/usage/deriving-data-selectors#passing-input-parameters

greg-nagy (commented):
This PR is lookin' good and fancy :) 

I think the selector restructure should be in a separate PR/proposal/discussion. 

Not sure about the selector input refactor either, but I might not have the full picture.

greg-nagy:
This breaks the dApp connection window — which was broken already :) See #2545 

The change this PR introduces is that we see a green screen without any error now when the user hasn't finished the onboarding OR removed his/her last account. 

CleanShot 2022-11-07 at 18 50 20

Not sure if this is a blocker for this PR, because that screen was already really bad.

background/redux-slices/selectors/accountsSelectors.ts
hyphenized: Unfortunately both of these are bad because some of the input selectors return non primitives and the default equality check for selectors memoization is `===` over the output selector params. The previous selector is a bit better because it's dependencies are passed directly to `createSelector`, but it does not make a significant difference in performance 😕

hyphenized:
Nice catch @greg-nagy! I've added that route to the redirect exception list

jagodarybacka (changes requested):
We are not switching tabbed onboarding flag yet right? Because it is not working correctly with `SUPPORT_TABBED_ONBOARDING=false`

https://user-images.githubusercontent.com/20949277/200557646-2eb0c4c0-6d37-46f3-99f2-4479fdcca4d6.mov

hyphenized:
@jagodarybacka Ahhh, looks like I missed the new password route. Should be good now, thanks!

jagodarybacka:
@hyphenized found another thing:
setup:
image
remove account: 
image

jagodarybacka (approved):
Looks good, would be nice to get it merged as I'm building https://github.com/tallycash/extension/pull/2578 on top of this. 
I'll let @greg-nagy take another look before merging.

background/redux-slices/selectors/accountsSelectors.ts
greg-nagy: https://github.com/reduxjs/reselect#createselectorcreatormemoize-memoizeoptions that behavior can be changed 

I don't think that this is a blocker now, but [we have random reloads without store changes](https://github.com/tallycash/extension/issues/2550) probably partially because of selector memoization is not working properly.

hyphenized: Hmmm yeah it does sound like that could be the cause, thanks for sharing that issue

PR #2546, opened 2022-11-02
🤖Show Me The Data: Introducing the analytics service

greg-nagy:
# Show Me The Data!

CleanShot 2022-11-04 at 17 03 42
CleanShot 2022-11-04 at 17 02 19

## What 

This PR introduces the analytics service. Implemented events
- background start 
- extension install
- address on network added for tracking 
- UI opened

Not implemented functionality 
- website-related functionality
  - uninstall tracking: It is blocked on a goodbye website page 
  - UTM code tracking with the website: blocked on privacy policy and cookie usage discussion
- data deletion: blocked on tech discussion 
- swaps tracking 
- dApp connections tracking
- signing tracking
- analytics user notification
- wallet usage time > length of time the wallet was open

Note: Tests will be in a follow-up PR.

## How 

Implemented an analytics service, which will be the background connection point. It's responsible for stateful information handling — like checking if the service is turned on or storing/generating the user's uuid — and listening for events. 

The preferences for the analytics service are stored in the preferences service. The data flow is a bit complex because we need to sync the redux and the background state. The settings in the preferences service are handled as the source of truth. This is propagated to the UI on startup and when change happens. Redux state is only updated when the background state is known to be good. 

The posthog-specific concerns are put in a lib, collecting every service-specific implementation detail.

The `USE_ANALYTICS_SOURCE` is meant to signal the context the analytics event was sent from. We can use this field to filter events only interesting to us. You can overwrite it with whatever you want in your local environment. This will help find only events that originated from your setup. 

## Test

Note: ⚠️ The analytics is turned off by default. Testing is more straightforward if you set [`defaultPreferences.analytics.isEnabled`](https://github.com/tallycash/extension/blob/7936d5fce3b2722bf5a0306c92397286dd884a91/background/services/preferences/defaults.ts#L33) to `true`.

- download the asset or change the environment variables and build locally 
- import a seed phrase 
- go to settings and turn on the feature
- reload the extension
- wait 1+ min 
- check the uuid in devtools > Application > indexedDB > `tally/analytics`
- go to https://app.posthog.com/events > Live events
  - if you need an account reach out to me or @puppycodes 
- you should see`New Install` and a `Bakckground start` event 
- open the extension: `UI open` should arrive

## Testing Env

```
SUPPORT_ANALYTICS=true
SHOW_ANALYTIC_MENU=true
SHOW_DELETE_DATA_BUTTON=false
```

Latest build: [extension-builds-2546](https://github.com/tallycash/extension/suites/9338393780/artifacts/439117915) (as of Wed, 16 Nov 2022 13:37:33 GMT).


background/services/chain/index.ts
```diff
@@ -829,8 +829,14 @@ export default class ChainService extends BaseService<Events> {
   }
 
   async addAccountToTrack(addressNetwork: AddressOnNetwork): Promise<void> {
-    await this.db.addAccountToTrack(addressNetwork)
-    this.emitter.emit("newAccountToTrack", addressNetwork)
+    const isNewAccountOnNetwork = await this.db.getTrackedAccountOnNetwork(
```
hyphenized, line 832: Naming is a bit confusing here

hyphenized (commented):
Nice work, code looks good. Verified that analytic events are being sent 🚀

background/services/chain/index.ts
greg-nagy: https://github.com/tallycash/extension/pull/2546/commits/66224e62c4cdcee49dfad4f5750689492eb8173d

background/lib/posthog.ts
```diff
@@ -0,0 +1,77 @@
+import { v4 as uuidv4 } from "uuid"
```
0xDaedalus, line 1: I'd love to find a way to have the functions in this file depend on the analytics service (without the analytics service depending on them).  With this current implementation I think its quite easy to accidentally import `sendPosthogevent` directly and bypass the analytics service completely.

background/services/preferences/index.ts
```diff
@@ -213,6 +222,21 @@ export default class PreferenceService extends BaseService<Events> {
     this.emitter.emit("updatedSignerSettings", await updatedSignerSettings)
   }
 
+  async getAnalyticsSettings(): Promise<Preferences["analytics"]> {
```
0xDaedalus, line 225: I think we should probably cache this in memory even though that means we'll be maintaining the same state in two places.  Firing off analytics should be as lightweight as possible so lets try to avoid i/o on the hot path when possible.

background/services/analytics/db.ts
```diff
@@ -0,0 +1,33 @@
...
+      // Let's use an incremental id. If we need to change the UUID
+      // for any reason the migration will be trivial.
+      // Note: we are using outbound primary index here, which means
+      // that the id property won't be included in the stored object.
+      // https://dexie.org/docs/inbound#example-of-outbound-primary-key
+      analyticsUUID: "++,uuid",
+    })
+  }
+
+  async getAnalyticsUUID(): Promise<string | undefined> {
```
0xDaedalus, line 23: We should cache this as well - especially considering it should never change once set.

background/services/analytics/db.ts
```diff
@@ -0,0 +1,33 @@
...
+    return (await this.analyticsUUID.reverse().first())?.uuid
+  }
+
+  async setAnalyticsUUID(uuid: string): Promise<void> {
```
0xDaedalus, line 27: I might even consider getting rid of this method and implicitly generating and saving a uuid inside of `getAnalyticsUUID` (if one does not already exist).  Not very CRUD and we _are_ talking about db methods here so perhaps not an ideal path but since we never want to change it once its set I'm not sure we need to expose adding records to other parts of the application.

background/services/analytics/db.ts
```diff
@@ -0,0 +1,33 @@
+import Dexie from "dexie"
+
+export interface AnalyticsUUID {
+  uuid: string
+}
+
+export class AnalyticsDatabase extends Dexie {
```
0xDaedalus, line 7: This should be pretty straightforward to test - could be nice to get some test coverage around analytics especially since its an area of the codebase that I'm sure will grow in the coming months.

background/services/analytics/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.db.close()
+
+    await super.internalStopService()
+  }
+
+  async sendAnalyticsEvent(
+    eventName: string,
+    payload?: Record<string, unknown>
+  ): Promise<void> {
+    // @TODO: implement event batching
```
0xDaedalus, line 60: Great TODO 👍 - def something that we will want in the future

background/redux-slices/ui.ts
```diff
@@ -47,6 +48,7 @@ export type Events = {
   newSelectedAccountSwitched: AddressOnNetwork
   userActivityEncountered: AddressOnNetwork
   newSelectedNetwork: EVMNetwork
+  updateAnalyticsPreferences: AnalyticsPreferences
```
0xDaedalus, line 51: I believe we need a redux migration for this

background/services/analytics/db.ts
```diff
@@ -0,0 +1,33 @@
...
+      // for any reason the migration will be trivial.
+      // Note: we are using outbound primary index here, which means
+      // that the id property won't be included in the stored object.
+      // https://dexie.org/docs/inbound#example-of-outbound-primary-key
+      analyticsUUID: "++,uuid",
+    })
+  }
+
+  async getAnalyticsUUID(): Promise<string | undefined> {
+    return (await this.analyticsUUID.reverse().first())?.uuid
```
0xDaedalus, line 24: Rather than reversing and limiting to the first uuid - I wonder if we should just ensure that there is only ever 1 uuid in the table.

background/services/analytics/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    private chainService: ChainService,
+    private preferenceService: PreferenceService
+  ) {
+    super()
+  }
+
+  protected override async internalStartService(): Promise<void> {
+    await super.internalStartService()
+
+    this.sendAnalyticsEvent("Background start")
```
0xDaedalus, line 45: Could this be something a bit more expressive like `Extension Turned On`?

background/services/preferences/index.ts
```diff
@@ -213,6 +222,21 @@ export default class PreferenceService extends BaseService<Events> {
...
+
+  async updateAnalyticsPreferences(
+    analyticsPreferences: AnalyticsPreferences
+  ): Promise<void> {
+    await this.db.updateAnalyticsPreferences(analyticsPreferences)
+    const { analytics } = await this.db.getPreferences()
+
+    // This step is not strictly needed, because the settings can only
+    // be changed from the UI
+    this.emitter.emit("updateAnalyticsPreferences", analytics)
```
0xDaedalus, line 237: Maybe rename this to `analyticsPreferencesUpdated` so it does not share the same name as the originating event?

background/lib/posthog.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export function createPosthogPayload(
+  personUUID: string,
+  eventName: string,
+  payload?: Record<string, unknown>
+): string {
+  return JSON.stringify({
+    // See posthog Data model: https://posthog.com/docs/how-posthog-works/data-model
+    // ID of the event
+    uuid: uuidv4(),
```
0xDaedalus, line 33: Struggling to understand why posthog don't assign uuid's to events themselves - lol

background/lib/posthog.ts
0xDaedalus: Not that this is not a blocking comment - just musing on my part

.env.defaults
```diff
@@ -9,7 +9,7 @@ HIDE_IMPORT_DERIVATION_PATH=true
 HIDE_SWAP_REWARDS=true
 GAS_PRICE_POOLING_FREQUENCY=120
 ETHEREUM_NETWORK=mainnet
-WEBSITE_ORIGIN=https://tally.cash
+WEBSITE_ORIGIN=https://tallyho.org
```
greg-nagy, line 12: zxdgf

greg-nagy, line 12: asdf

greg-nagy, line 12: asdf

greg-nagy, line 12: sadfasdfsad

.env.defaults
```diff
@@ -28,3 +27,9 @@ SUPPORT_ACHIEVEMENTS_BANNER=false
 SWITCH_RUNTIME_FLAGS=false
 UNS_API_KEY="4a77c949-511b-4c16-9862-6edfb6ae6012"
 REACT_DEVTOOLS_DEFAULT_PORT=8097
+SUPPORT_ANALYTICS=false
```
greg-nagy, line 30: asdfsdaf

background/lib/posthog.ts
mhluongo: If we don't set the UUID, we have to wait for PostHog to ACK before we can know the ID of the event... rather than fire and forget.

I doubt it matters much for our use case, but it matters a lot for others... and is common in [event sourcing](https://martinfowler.com/eaaDev/EventSourcing.html) to have clients / event creators set the event ID.

background/services/preferences/index.ts
greg-nagy: Will do in the next PR

.github/workflows/build.yml
```diff
@@ -36,14 +36,17 @@ jobs:
         if: github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')
         run: |
           echo ${{ steps.env-block.outputs.result }} > .env
+          echo 'USE_ANALYTICS_SOURCE="BETA"' >> .env
```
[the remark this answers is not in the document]
greg-nagy: this is a confusing state, because I didn't want to introduce the BETA build in this PR 

in a follow up PR the `DEV` will be used in dev mode and in PRs,  `BETA` will be the builds from main and `PROD` will be built from a tag

https://github.com/tallycash/extension/issues/2612

background/lib/posthog.ts
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    isEnabled(FeatureFlags.SUPPORT_ANALYTICS) &&
+    !!POSTHOG_URL &&
+    !!POSTHOG_API_KEY
+  )
+}
+
+export function createPosthogPayload(
+  personUUID: string,
+  eventName: string,
```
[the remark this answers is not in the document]
greg-nagy: not in this PR 

in this PR I wanted to let the event names roam freely

background/services/analytics/db.ts
greg-nagy: almost went this way tbh :) 

but then realized that for UTM code tracking, we will need to check the website for existing UUID and use that if there is. Handling that concern in the db layer felt really off.

background/services/analytics/db.ts
greg-nagy: I want to do that in a separate PR — see the description and https://github.com/tallycash/extension/issues/2613

(I expect things to change rapidly in the next couple of PRs)

background/services/analytics/index.ts
greg-nagy: probably should be handled in the lib tough. Still debating

background/redux-slices/ui.ts
greg-nagy: This is just an event that we fire off. The underlying `collectAnalytics` flag has been added by @kkosiorowska when she implemented the UI. 

The other thing here is that we initialize this flag from the background and in those cases I believe we don't need a migration. I will double-check this, just to be safe.

background/services/analytics/db.ts
greg-nagy: I was pondering over this a bit. This `reverse().first()` method makes indexedDB usage similar to `localStorage` without losing information at any time. Don't think the overhead is huge so it felt like a comfy tradeoff. 

Your idea aka `let's use indexedDB the same way as localStorage` occurred to me, but felt off to do that. 0 real reasons though.

background/lib/posthog.ts
greg-nagy: 👆

Posthog does generate a UUID, it just felt safer and cleaner not to rely on this to achieve fire and forget

background/services/preferences/index.ts
greg-nagy: I felt that using the same name is a pro, and not a con, because in every place it's obvious what's the source (`uiSliceEmitter` or `ProviderService`)

I tried to use different event names to closely related events in the past for dapp connections and I regret doing so to this day, because it's confusing and resulted in bad naming. 

Open to be convinced, if you feel strongly about this.

background/services/analytics/index.ts
greg-nagy: for sure - will tackle this in follow up pr

background/lib/posthog.ts
greg-nagy: CleanShot 2022-11-15 at 19 57 52

the face that makes you realize how little you know at least once during your career :D

puppycodes: > Struggling to understand why posthog don't assign uuid's to events themselves - lol

They do it on the client side with their JS lib, which we aren't using on purpose.

background/lib/posthog.ts
puppycodes: > I'd love to find a way to have the functions in this file depend on the analytics service (without the analytics service depending on them). With this current implementation I think its quite easy to accidentally import `sendPosthogevent` directly and bypass the analytics service completely.

Yeah def what we are doing really isn't specific to PostHog but could very much generalized to emitting a KV somewhere

background/services/analytics/db.ts
puppycodes: not to beat a dead horse but remind me why do we need to import an entire lib just to store a KV accessible via chrome.storage.local.set()?

puppycodes: Another good reason i'm thinking of using the chrome storage api instead is chrome.storage.sync() api which means we can implement the same UUID across devices logged in with the same google account

background/services/analytics/index.ts
0xDaedalus: The only thing I can add is that `Background start` is probably meaningless to non engineers - and we will likely have plenty of non engineers looking at our analytics data - so we might want to tune our nomenclature to be a bit more general

background/services/analytics/db.ts
greg-nagy: > FWIW I am more concerned with enforcing a pattern that guarantees there is only 1 uuid per extension install rather than overhead anything to do with overhead.

I am not sure this will be the case in the long run but for now, it's enforced in `getOrCreateAnalyticsUUID` flow. 
Even if we want to enforce this, it probably shouldn't be done in the db layer or by creating a solution where we can lose the previous uuid (if we need to have more than 1 for any reason)

background/services/analytics/index.ts
greg-nagy: Open to ideas, for sure. 
I don't have anything better without distorting what's happening :(

background/services/analytics/db.ts
greg-nagy: > why do we need to import an entire lib just to store a KV accessible via chrome.storage.local.set()? isn't that a lot simpler than using indexdb? 

Yes and no :) 

We could use `chrome.storage.local` if we want to. The reasons I chose not to
- our services rely on indexedDB and we have a well-established pattern for using them. It's ok to diverge from this if there is a value in it. In this case, the difference doesn't feel significant so I erred on the side of keeping our patterns
- I expect to have some future requests — analytics support while the user is offline, analytics bulk send — and all of these will require a database
- `chrome.storage.local` is not accessible through the devtoolbar 

> Another good reason i'm thinking of using the chrome storage api instead is chrome.storage.sync() 

This idea scares me on many levels. I feel strongly about not sharing our uuid with any 3rd party especially google and we should do whatever we can to avoid this.

.env.defaults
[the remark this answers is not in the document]
greg-nagy: > how do yall feel about the practice of separating out commits in a PR that don't have to do with the PR subject itself?

We def feel good about this, and this is what we practice — at least, we do our best. 

This change is related because, at one point, the analytics will need to get the UUID from the website for UTM tracking and to be able to do that, the extension needs to know the proper webpage url. As such, it felt appropriate to be included in the initialization PR.

But I also get your point. This could have also been only in the PR that handles that specific concern.

background/redux-slices/ui.ts
0xDaedalus: Whoops - right you are.

0xDaedalus (approved):
Approving this so we can get it tested and into Fridays release but just want to re-emphasize the need for a follow-on PR here and that I'm looking forward to this part of our codebase growing :)

greg-nagy:
> Approving this so we can get it tested and into Fridays release but just want to re-emphasize the need for a follow-on PR here and that I'm looking forward to this part of our codebase growing :)

https://github.com/tallycash/extension/issues/2637

background/services/analytics/db.ts
puppycodes: @greg-nagy Gotcha! Thanks so much! 
Appreciate you! Makes sense 🔥

PR #2547, opened 2022-11-03
[Proposal] - Prefer generic RPC providers when possible.

0xDaedalus:
In order to cut down on costs we need to have an adjustable way of routing rpc calls between free and premium rpc providers.

This PR is intended as an illustration of one possible way for us to start moving towards segmenting our RPC call usage.  It is meant to spark discussion about the overall approach and is not yet in a mergeable state.

An ideal solution is one that preserves our provider fallback logic and reconnect logic while still allowing us to send alchemy-specific rpc calls regardless of what type of provider is set as "currentProvider".

Highlights
- [ ] Generic RPC providers are now the "default" providers we connect to
- [ ] alchemy specific calls still always get routed to alchemy
- [ ] A rudimentary method to tune what percentage of our traffic is hitting alchemy
- [ ] Fixed some bugs where we were not properly using the `SerialFallbackProvider`'s `send` method where we should have been.

Detailed comments are in the commit messages - looking forward to feedback!

Latest build: [extension-builds-2547](https://github.com/tallycash/extension/suites/9212911094/artifacts/430107606) (as of Wed, 09 Nov 2022 13:39:48 GMT).


Shadowfiend:
Love to see it :) It could be useful to be explicit about the problem being solved ("why") and the goal of this solution (half of "how") in the description!

background/services/chain/asset-data-helper.ts
```diff
@@ -52,13 +48,9 @@ export default class AssetDataHelper {
     }
 
     try {
-      // FIXME Allow arbitrary providers?
-      if (
-        provider.currentProvider instanceof AlchemyWebSocketProvider ||
-        provider.currentProvider instanceof AlchemyProvider
-      ) {
+      if (provider.supportsAlchemy) {
```
greg-nagy, line 51: This is the problem that I have just a vague idea of how to handle. — This is a discussion not a PR review

Namely that we need to do feature detection in 2 places. Once a layer above the transport layer — `SerialFallbackProvider` — and once in the Transport layer. My problem with that is that this will limit what we can and can not do, and puts the same concern in 2 places.

How would you feel about handling this concern only on `SerialFallbackProvider` layer and below? 

We would define a structure that makes sense for us — probably based alchemy api — and use a `tally_getTokenBalances` + define a transformer method. 

Then in the [`SerialFallbackProvider` where you check for `alchemy_` in the method](https://github.com/tallycash/extension/blob/ce9e54ce052cecfaa576b1018ec00b7cd038218c/background/services/chain/serial-fallback-provider.ts#L236) we could 
1) make the decision which provider we want to use that supports this type of action — eg loadbalance with quicknode/others or own implementations 
2) lookup the method for the provider and transformation logic from our lookup list 
3) setup the transformation logic to be applied to the response before letting it propagate to higher layers. 

This would make the whole load balancing invisible to the rest of the wallet. 

Your current implementation is more gentle with our codebase and fits better, but in this case, a more heavy-handed approach might be acceptable.

greg-nagy (commented):
I love this PR and that we have started to go in this direction!

The other thought here on which I assume we are on the same page but would like to verify. 

Do you intend to introduce a queue into the solution? We could use that for call batching and combine that with the multicall solution. Or that's not part of this work?

background/services/chain/asset-data-helper.ts
0xDaedalus: This is a very interesting idea but I have a few reservations:
- What if we need different params for an alchemy call and a generic rpc call that are not hardcodeable?  We could of course send all params that _might_ be needed in a call to `provider.send('tally_doTheThing')` but this seems like it might cause us more headaches than it alleviates.
- I agree that its not ideal to have the same check in 2 places - but I actually think the separation of concerns is quite nice.  I am a big fan of having transformation logic live separately from message passing logic - and the pattern of using `.supportsAlchemy` lets us easily accomplish that.  IMO the closer `SerialFallbackProvider` can be to a standard `JsonRPCProvider` the easier our code will be to reason about (both for us and for outside contributors).

I also want to point out that in this proposal the alchemy methods are specifically _not_ load balanced because I think its quite reasonable to assume that alchemy api calls will be more responsive and have better uptime than publicly available rpc servers - so we always use them if they are available.

I do agree though that making all of this invisible to consumers of `SerialFallbackProvider` would be quite nice - I just don't know if thats something we need to do at this moment to cut down on our infra costs.

0xDaedalus:
> Do you intend to introduce a queue into the solution? We could use that for call batching and combine that with the multicall solution. Or that's not part of this work?

I think this is out of scope for this particular proposal - and almost certainly deserves its own separate proposal.

background/services/chain/asset-data-helper.ts
greg-nagy: ## Important notes
First: I am OK with this PR as is if you want to merge
Second: The final decisions and how you want to progress are in your court! I am sharing my thoughts because of the proposal and you have asked me to. 
Third: I do not consider my approach better or this one bad! Different tradeoffs, different visions. 

## Engineer convo

Your reservations are good, insightful, and appreciated. Thank you

> What if we need different params for an alchemy call and a generic rpc call that are not hardcodeable? We could of course send all params that might be needed in a call to provider.send('tally_doTheThing') but this seems like it might cause us more headaches than it alleviates.

If we discuss the concrete methods, the tradeoff can be better weighted. I expect that the arguments will be very similar. I might be off here, so let's get back to this only when we consider this approach. 

> I agree that its not ideal to have the same check in 2 places - but I actually think the separation of concerns is quite nice. I am a big fan of having transformation logic live separately from message passing logic - and the pattern of using .supportsAlchemy lets us easily accomplish that.

My understanding is that both approaches fulfill the separation of message passing and transformation logic. It's more like `f(g(x))` vs `g(f(x))` what we are talking about here. We need to send the message and we need to transform the response. 

> IMO the closer SerialFallbackProvider can be to a standard JsonRPCProvider the easier our code will be to reason about (both for us and for outside contributors).

Are we talking about interface or implementation? This approach breaks the `JsonRPCProvider` interface — which is fine — but with the approach, we could implement the interface, and only the implementation would differ. I think it's ok not to differ from `JsonRPCProvider` in our implementation, but the point confused me. 

> I also want to point out that in this proposal the alchemy methods are specifically not load balanced because I think its quite reasonable to assume that alchemy api calls will be more responsive and have better uptime than publicly available rpc servers - so we always use them if they are available.

The idea wasn't to load balance between paid and public service but the ability to load balance between alchemy, quicknode, and whatever provider we can strike a good deal with in the future. And to be able to do so without this change propagating to higher layers. 

> I just don't know if thats something we need to do at this moment to cut down on our infra costs.

My thoughts were formed with the `Proposal` nature in mind — eg this is a lightweight RFC to a big change in our architecture, and we are discussing what should be the end state. There are multiple ways for cutting down infra costs. Your approach is definitely the quickest way to do that. But I thought we are talking about the desired architecture here. If that's not the case, please consider my thoughts moot, and let's treat this as a normal PR. 

I think what we are talking about here is parallel to what we have gone through with the signing ui. We started with a solution that resonates with your approach here and then we migrated to a similar solution to what I suggested. 

The other parallel I feel is good here: If we would have an ideal infrastructure provider, how would we use it? 

With all this said: These are my thoughts on why I would prefer another approach to the current one. My preference is not better or worse than this. Different tradeoffs, different considerations. These thoughts were meant to spark more thoughts. I am fine with the current approach, the final decision is not in my court as I am not owning this work item. If you decide that you want to treat this as a normal PR please consider all my thoughts above irrelevant. This is the quickest way to check off our KR.

greg-nagy:
> > Do you intend to introduce a queue into the solution? We could use that for call batching and combine that with the multicall solution. Or that's not part of this work?
> 
> I think this is out of scope for this particular proposal - and almost certainly deserves its own separate proposal.

This was meant as a "what's your vision" question and definitely not in the scope of this proposal.

background/services/chain/asset-data-helper.ts
0xDaedalus: Lets think of this as a proposal about how we can lower infra costs while also accomplishing the goal of moving closer to splitting up calls per rpc provider.

Lets go ahead and move forward wit this (there's still some work that needs to be done on this PR) and keep iterating through future PRs / proposals.

0xDaedalus:
Closing this PR so we can maintain a tidy place for the Proposal Discussion.  We can continue to discuss the high level stuff in here and discuss code-level implementation details in the upcoming implementation PR.

PR #2558, opened 2022-11-04
Fix errors caught by TypeScript checking in the post-build-link workflow 

hyphenized:
`// @ts-check` 🚀

Latest build: [extension-builds-2558](https://github.com/tallycash/extension/suites/9155621081/artifacts/426076485) (as of Sun, 06 Nov 2022 22:54:15 GMT).


.github/workflows/builds/post-build-link.js
```diff
@@ -44,7 +44,7 @@ module.exports = async function postBuildLink({ github, context }) {
 
   if (matchArtifact === undefined || matchArtifact === null) {
     throw new Error(
-      `Failed to find extension artifact :( Artifacts were ${JSON.strignify(
```
mhluongo, line 47: Herp-a-derp! I wonder if we can us `ts-node` for these?

hyphenized: That should be possible, might be worth a try!

package.json
```diff
@@ -73,6 +73,7 @@
     "webextension-polyfill": "^0.7.0"
   },
   "devDependencies": {
+    "@actions/github": "^5.1.1",
```
Shadowfiend, line 76: Not a huge fan of pulling this in at the top level for something only the workflows need. Not sure if we can drop a package.json in .github or something...

Shadowfiend (commented):
This is awesome. Left a quick note about dependencies though....

package.json
Shadowfiend: It may be that to get these in a separate package json we'll have to split them to a separate action unfortunately.

hyphenized: Good call, I've moved this dependency away to a separate `.github/package.json`

.github/workflows/builds/post-build-link.js
```diff
@@ -73,15 +83,15 @@ module.exports = async function postBuildLink({ github, context }) {
...
 
   if (pullLookupStatus !== 200) {
-    throw new Error("Failed to fetch PR body :( Status", pullLookupStatus, ".")
+    throw new Error(`Failed to fetch PR body :( Status ${pullLookupStatus}.`)
   }
 
-  const baseUrl = context.payload.repository.html_url
-  const artifactUrl = `${baseUrl}/suites/${checkSuiteId}/artifacts/${matchArtifact.id}`
+  const baseUrl = context.payload?.repository?.html_url
+  const artifactUrl = `${baseUrl}/suites/${checkSuiteId}/artifacts/${matchArtifact?.name}`
```
Shadowfiend, line 94: Pretty sure this is incorrect. The artifact URL uses the id, not the name.

.github/workflows/builds/post-build-link.js
Shadowfiend: Don't know that it's worth the extra effort now that we've added TypeScript type checks anyway.

.github/workflows/builds/post-build-link.js
```diff
@@ -1,12 +1,22 @@
 // @ts-check
 /* Allow console output for debug information in Actions output. */
 /* eslint-disable no-console */
+
+/**
+ * @param {object} ctx Context
+ * @param {ReturnType<import("@actions/github")["getOctokit"]>} ctx.github
+ * @param {import("@actions/github")["context"] & {inputs?: Record<string,unknown>}} ctx.context
```
Shadowfiend, line 8: Going to flag that it bothers me that we're having to `&` something in here, and it bothers me that we're having to pull the `getOctokit` type.

That said, I think this'll give us more accurate types than any other obvious approach atm, so it's Good Enough™.

.github/workflows/builds/detect-env-block.js
```diff
@@ -1,5 +1,13 @@
+// @ts-check
 /* Allow console output for debug information in Actions output. */
 /* eslint-disable no-console */
+
+/**
+ * @param {object} ctx Context
+ * @param {ReturnType<import("@actions/github")["getOctokit"]>} ctx.github
+ * @param {import("@actions/github")["context"] & {inputs?: Record<string,unknown>}} ctx.context
```
Shadowfiend, line 8: We don't really need the `inputs` here, so I would suggest leaving it off altogether.

Shadowfiend (dismissed):
Blocking this on the note regarding the artifact URL. Otherwise looks good, left a few nits but I don't think they're super actionable or urgent and they are definitely not blocking.

.github/workflows/builds/post-build-link.js
hyphenized: Yeah I don't think it's the best solution, but if someone needs to make changes to the workflows these types should help

.github/workflows/builds/post-build-link.js
hyphenized: Thanks, didn't notice the jsdoc comment, thought this was a legit fix 😅

mhluongo (approved):
:rocket:

Issue #2482, opened 2022-10-21
Unlock signing screen visual issue


0xDaedalus:
<img width="384" alt="image" src="https://user-images.githubusercontent.com/94649004/197281816-9956ded2-c075-4aa5-bdfc-782308c0340e.png">

To recreate
- Ensure wallet is locked
- Set up an in-wallet swap on Ethereum
- Attempt to execute swap
- See the above screen

Shadowfiend:
Is this with or without the new signing flow? Smells like the new flow, which implements this part differently--but if it isn't the new flow might fix for the same reason lol.

PR #2560, opened 2022-11-05
Signature Reversal: Flip the USE_UPDATED_SIGNING_UI flag to true

Shadowfiend:
This will switch away from the old signing UI code path and use the updated signing UI for all signing by default.

## Testing

- [x] Ensure that each type of transaction signing has its summary data displayed correctly.
  - [x] ETH transfer (can be from the extension).
  - [x] ERC20 transfer (can be from the extension).
  - [x] Approval transaction (the Tally test wallet does not have SUSHI approved, so an easy way is to try to swap SUSHI).
  - [x] Normal contract interaction (a swap with an approved asset will work here).
- [x] For all of the above, make sure fee changes are applied correctly and, if there are other internal form elements (e.g., the approval transaction amount), that these are also applied correctly.
- [ ] Ensure that each type of data signing has its summary data displayed correctly.
  - [x] Sign In With Ethereum. Can be checked at https://login.xyz/ under Get Started.
  - [x] Check regular message signing still looks visually consistent. Can be checked at https://mintkudos.xyz/ on Polygon by hitting Connect Wallet.
- [x] Ensure that Ledger accounts can sign.
  - [x] Ensure that, after clicking `Sign` or `Sign Data` for each of the above scenarios, Tally displays the exact same fields and values as the Ledger.
  - [x] Ensure the Ledger displays that it is connected on the signing screen for transactions, data, and typed data when the Ledger is connected, unlocked, and has blind data signing enabled. If any of those is not true, ensure that the Ledger displays as disconnected (and the Check Ledger button replaces the Sign button).
- [ ] Ensure that keyring accounts can sign and broadcast normally.
  - [x] If the keyring is unlocked, ensure signatures work correctly for each of the scenarios above without additional steps after clicking `Sign` or `Sign Data`.
  - [x] If the keyring is locked, ensure that the unlock step comes after clicking `Sign` or `Sign Data`, not before, for each of the scenarios above.

Latest build: [extension-builds-2560](https://github.com/tallycash/extension/suites/9175383348/artifacts/427480699) (as of Mon, 07 Nov 2022 21:33:32 GMT).


mhluongo:
I've found a few small issues around scroll behavior, but the outcome is still a real improvement over the current UI. Going to open a couple follow-up issues

mhluongo (approved):
I've been able to test Ledger extensively. I could use backup testing regular accounts and typed data signing

Shadowfiend:
Pushed an updated version of the fix for #2482, which makes scrolling space that doesn't cut off messages and transaction data when signing.

mhluongo:
Hell yeah, looks great!

0xDaedalus:
>  Check typed data signing still looks visually consistent. Can be checked at https://app.uniswap.org/#/swap trying to swap from USDC using the Tally test wallet by clicking “Allow the Uniswap Protocol to use your USDC”. Do not confirm this signature so that it remains a viable test path for later. Ensure addresses can be clicked to be copied, and ensure the Etherscan icon is not orange unless hovered.

Typed data visualization looks good but it does not look like addresses are clickable and the etherscan icon is orange regardless of hover status.

image

Shadowfiend:
If that matches the status quo would suggest not blocking. If it doesn't, might be worth blocking.

mhluongo:
I'd vote to not block. Opened a [follow-up issue](https://github.com/tallycash/extension/issues/2579)

0xDaedalus (approved):
Man oh man does it feel good to get this one in.  As excited I am about merging this and seeing the fruits of the work that has been put into updating the signing screen - I think I'm even more excited about the benefits we'll get from this pattern in the coming months / years 😄 

📈 📈 📈

ui/components/Signing/index.tsx
```diff
@@ -66,13 +66,9 @@ function SigningLoaded<T extends SignOperationType>({
...
             z-index: 5;
           }
-          section:has(footer) {
-            /* Make room for the footer. */
-            height: calc(100% - 80px);
-          }
           section :global(h1.title) {
             color: var(--trophy-gold);
-            font-size: 36px;
+            font-size: ;
```
jagodarybacka, line 71: why is this font-size empty?

Shadowfiend: Absolutely a bug from https://github.com/tallycash/extension/commit/6dacf2ef4fb91213b4c8ad4cb855ad2d299d5f00, would fix if I weren't away but probably a good idea for a quick fix to land. Good catch and thanks for the thorough look!

Also obv this is already merged but please make sure @VladUXUI has had a look and everything looks clear on the design end 🙏 Some changes to better align with designs did land here and I want to make sure nothing funky snuck in.

VladUXUI:
Loving this!!! 

This data is looking a bit wonky tho'
image

PR #2561, opened 2022-11-07
Fix UI inconsistencies in Swap Page 

### LINKED ISSUE #2522: Inconsistent Swap price preview
https://user-images.githubusercontent.com/94649004/198730023-1213cd48-53f1-4015-851c-da061829facc.mov

hyphenized:
This PR adds a few React hooks and refactors most of the code that handles swap quotes in the Swap page. Quote requests are now debounced.

[ImprovedBartering.webm](https://user-images.githubusercontent.com/28708889/200248932-482759d3-3b7f-4206-ae2a-a0089f817eb9.webm)

Closes #2522

## To Test:
Head to the Swap Page, 
- [ ] Select two assets, input an amount in either field, once you stop typing, the quote should start loading. If you make a correction during load, the last quote data should be reflected on the UI.
- [ ] After flipping the assets, the quote should refresh
- [ ] Changing the target asset should quote the source asset against the target asset amount and vice versa
- [ ] Changing the network should reset both assets
- [ ] Coming from a different page should correctly set initial assets

Latest build: [extension-builds-2561](https://github.com/tallyhowallet/extension/suites/9990416967/artifacts/484825760) (as of Wed, 21 Dec 2022 12:41:33 GMT).


mhluongo:
cc @Shadowfiend :smile:

ui/pages/Swap.tsx
```diff
@@ -41,50 +38,154 @@ import { selectSlippageTolerance } from "@tallyho/tally-background/redux-slices/
...
+        // Finish loading once the last quote is fulfilled
+        if (requestId.current === id) {
+          setLoading(false)
+        }
+      }
+    },
+    [dispatch, requestContextRef, mountedRef]
+  )
+
+  const [debouncedRequest] = useState(() => {
```
0xDaedalus, line 166: Have you considered using [`useDebounce`](https://github.com/tallycash/extension/blob/main/ui/hooks/dom-hooks.ts#L95:L95) here?

ui/pages/ApproveQuoteBtn.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react"
```
0xDaedalus, line 1: This component should probably not be in the `/pages` directory.

ui/pages/Swap.tsx
```diff
@@ -41,50 +38,154 @@ import { selectSlippageTolerance } from "@tallyho/tally-background/redux-slices/
...
+  isSameAsset,
+} from "../utils/swap"
+
+const UPDATE_SWAP_QUOTE_DEBOUNCE_TIME = 500
+
+const REFRESH_QUOTE_INTERVAL = 5 * SECOND
+
+const useSwapQuote = (useSwapConfig: {
+  savedQuoteRequest?: SwapQuoteRequest
+  initialSourceAsset?: FungibleAsset | SmartContractFungibleAsset
```
0xDaedalus, line 80: Should this be `NetworkBaseAsset | SmartContractFungibleAsset`?

0xDaedalus (commented):
Quick glance with a few small comments here to get some discussion on this PR going.

I'd also like to flag [that one of the commits](https://github.com/tallycash/extension/pull/2561/commits/1af2fbef619eb126f40af3e78b45e532116ed9b7) has over 600 lines of code changed which is somewhat larger than the typical diff size we would expect for a single commit (although I realize that there is moving of code going on).

I know it is not always convenient to keep commit sizes small (especially when working in a part of the codebase that may be unfamiliar) but I would encourage you to strive for smaller commits as they makes both reviewing and reasoning about the code quite a bit easier.  Looking at the commit message itself it seems like there are certainly some bullet points in there that could have been their own commits.

ui/pages/Swap.tsx
```diff
@@ -41,50 +38,154 @@ import { selectSlippageTolerance } from "@tallyho/tally-background/redux-slices/
...
+              AsyncThunkFulfillmentType<typeof fetchSwapPrice>
+            >,
+          network: requestContext.network,
+          settings: transactionSettings,
+        })
+
+        if (!mountedRef.current) return
+
+        setQuoteRequestState(result)
+      } finally {
```
0xDaedalus, line 156: I think we might want a `catch` with some logging here that does not throw instead of a `finally` - I believe this current pattern will silently swallow errors

ui/pages/Swap.tsx
```diff
@@ -41,50 +38,154 @@ import { selectSlippageTolerance } from "@tallyho/tally-background/redux-slices/
...
+
+  const requestId = useRef(0)
+  const [loading, setLoading] = useState<QuoteType | boolean>()
 
-// FIXME Unify once asset similarity code is unified.
-function isSameAsset(asset1: AnyAsset, asset2: AnyAsset) {
-  if (typeof asset1 === "undefined" || typeof asset2 === "undefined") {
-    return false
+  type RequestQuoteUpdateConfig = {
+    type: "getSourceAmount" | "getTargetAmount"
```
0xDaedalus, line 106: Do you think we should stick with our existing nomenclature of `buyAmount` and `sellAmount`?

ui/utils/swap.ts
```diff
@@ -0,0 +1,143 @@
...
+    slippageTolerance: number
+    networkSettings: NetworkFeeSettings
+  }
+  sourceAsset: FungibleAsset | SmartContractFungibleAsset
+  targetAsset: FungibleAsset | SmartContractFungibleAsset
+  quoteRequest: SwapQuoteRequest
+  timestamp: number
+}
+
+export const fetchQuoteUpdate = async ({
```
0xDaedalus, line 31: What is the difference between a quote and a quoteUpdate?   Is the initial fetch different from subsequent ones?

ui/utils/swap.ts
```diff
@@ -0,0 +1,143 @@
...
+    sourceAsset,
+    targetAsset,
+    timestamp: Date.now(),
+  }
+
+  if (updatedQuoteData) {
+    if (
+      settings.networkSettings.gasLimit !== BigInt(updatedQuoteData.quote.gas)
+    ) {
+      requestResult.swapTransactionSettings = {
```
0xDaedalus, line 87: Any particular reason we're favoring immutably updating properties here rather than assigning them directly?

0xDaedalus (changes requested):
Lots of good stuff here - with some small changes to address.

I'd also like someone else (cc @jagodarybacka @kkosiorowska @greg-nagy) to take a thorough look due to the size of this PR and the fact that it touches the most important part of our application.

I also think that the `SUPPORT_SWAP_QUOTE_REFRESH` flag is awesome and certainly a step in the right direction - but needs a product discussion before we flip it on.

ui/pages/Swap.tsx
hyphenized: Yes, but the aim of `useDebounce` seemed to be around debouncing state updates.

ui/pages/Swap.tsx
hyphenized: Yeah... I changed it because I found it confusing with the order of the inputs, but I'm willing to go with what the team feels makes more sense here.

ui/utils/swap.ts
hyphenized: Good catch! There's no need, was just being extra careful with the original implementation here.

ui/utils/swap.ts
0xDaedalus: maybe we just call this `fetchQuote` then?

ui/pages/Swap.tsx
hyphenized: Good catch, I reviewed the code and refactored this type accordingly

hyphenized:
@0xDaedalus Thank you very much for the feedback! I've rebased this PR to make it easier to review (and understand for future contributors!) and, moved the auto update functionality to a different PR 😄

0xDaedalus (commented):
So sorry for the late review here - I had left comments but forgot to click `Submit Review` - completely my fault 🙇

ui/pages/Swap.tsx
```diff
@@ -587,50 +425,76 @@ export default function Swap(): ReactElement {
...
+                amountMainCurrency={
+                  targetAmount
+                    ? quote?.priceDetails?.buyCurrencyAmount
+                    : undefined
+                }
+                priceImpact={quote?.priceDetails?.priceImpact}
+                isPriceDetailsLoading={loadingQuote}
                 showPriceDetails
-                // FIXME Merge master asset list with account balances.
+                // FIXME: Merge master asset list with account balances.
```
jagodarybacka, line 478: Are we going to ever do this? I'm not even sure if I understand what this comment is about

jagodarybacka (changes requested):
Lots of good code here 😍 

Some behaviours I don't like:
- if I enter "0" amount manually the old $ quote is still visible
image
- if there are no assets selected I think we can hide "no price info" text
image
- please update this branch with main so we can see how loading doggo behaves here 🐶

ui/hooks/react-hooks.ts
```diff
@@ -0,0 +1,85 @@
...
+  const valueRef = useRef<T>(value)
+  valueRef.current = value
+  return valueRef
+}
+
+/**
+ * Runs a callback on mount, if the callback returns a function,
+ * it will be called on unmount
+ */
+export function useOnMount<Fn extends (...args: unknown[]) => unknown>(
```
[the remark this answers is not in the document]
hyphenized: It's very similar but that hook runs on first render, before the component has mounted

ui/pages/Swap.tsx
hyphenized: Yes, this has already been done by @dills122 in #2628 🚀

jagodarybacka:
There is a possibility to be stuck in a "Approve asset" state - find some token that will cause "Approve" button to be displayed and change a network

https://user-images.githubusercontent.com/20949277/207590062-cb622810-5839-4782-a80d-83d0f04eca9d.mov

jagodarybacka (approved):
Works nice! ✅ 

ah no I've found one more thing: I think we should block the button if one of the assets is `0` - but I think there are even more small improvements that could be made here so up to you if you want to do it in this PR or add a task 
Screenshot 2022-12-16 at 12 50 25

jagodarybacka (approved):
Now it works great 🎉

PR #2573, opened 2022-11-07
Simulation Stimulation: Prototype a transaction simulation tab

mhluongo:
The prototype is frontend-only, text-only, and makes a new API call every time the tab is clicked. Nevertheless, it's pretty cool to see every transaction get a sim.

It's built off @Shadowfiend 's new [signature page work](https://github.com/tallycash/extension/pull/2560) to keep the code clean.

Screenshot from 2022-11-07 17-26-30

## Testing

- [ ] Try different transactions on mainnet, Optimism, Arbitrum, Polygon, etc
- [ ] Before signing a transaction, click the "Simulation" tab
- [ ] Review the text simulation and give me feedback!

## Testing Environment

```
SUPPORT_SIMULATION_TAB=true
```

Latest build: [extension-builds-2573](https://github.com/tallycash/extension/suites/9176600449/artifacts/427563467) (as of Mon, 07 Nov 2022 22:46:41 GMT).


mhluongo:
@VladUXUI @mr-michael sharing this for discussion, and to get the juices flowing. Note that...

* simulations are more useful for complex interactions
* we should come up with a stronger visual representation
* we can also show this level of detail on mined transaction detail pages, ala Etherscan. It'll take elbow grease, but we can use the design work in both places

mhluongo:
Also note this doesn't support RSK, BSC, Gnosis, Avax, or other upcoming chains (except Astar). If we like this approach we can make those happen pretty quickly though.

greg-nagy:
let's not do this anymore gh ... please ....

Shadowfiend:
Closing this for now; we've got a rework coming up that will pull in transaction simulation from a design perspective, and may or may not integrate in the same way. We will, of course, be able to reference this code when the time comes.

PR #2580, opened 2022-11-09
Normalize token lists contract addresses

### LINKED ISSUE #1898: Don't lose image metadata or token lists when merging assets
Sometimes assets will "flicker" between having a logo and not, or having a token list citation or not. In particular, WETH, NU, and KEEP all have this issue.

hyphenized:
This PR normalizes tokens addresses for `tokenLists` and assets stored in the indexing service db. This fixes a few issues with tokenlists handling that slipped by the token lists JSON schema. While the schema for token lists specifies that tokens should define a checksummed address, the validation uses a regex pattern that is unable to distinguish whether an address is indeed checksummed or not. 

It also fixes an issue with the logic to retrieve balances that caused tokens without metadata to be stored in the db and, prevents the loss of image urls when merging similar assets data in cases where the token retrieved from the tokenlist does not have a `logoURI`.

Closes #1898

Latest build: [extension-builds-2580](https://github.com/tallycash/extension/suites/9384944073/artifacts/442428834) (as of Fri, 18 Nov 2022 14:46:58 GMT).


mhluongo:
So excited for this one! Likely also an opportunity to use a nominal typing technique @Shadowfiend brought up recently... reviewing now

background/services/indexing/db.ts
```diff
@@ -143,6 +144,58 @@ export class IndexingDatabase extends Dexie {
...
+
+      const selectInvalidOrDuplicateRecords = (
+        record: SmartContractFungibleAsset & {
+          chainId?: unknown
+          address?: string
+        }
+      ) => {
+        const normalizedAddress = normalizeEVMAddress(record.contractAddress)
+        const isInvalidRecord =
+          typeof record.chainId !== "undefined" &&
```
0xDaedalus, line 159: Should this be `typeof record.chainId === "undefined" || typeof record.address === "undefined"`?

I would expect that either one of those missing would mean the given record is invalid - no?

background/services/indexing/index.ts
```diff
@@ -251,20 +251,21 @@ export default class IndexingService extends BaseService<Events> {
...
+
+    const searchResult = knownAssets.find(
+      (asset): asset is SmartContractFungibleAsset =>
+        isSmartContractFungibleAsset(asset) &&
         asset.homeNetwork.name === network.name &&
-        asset.contractAddress === contractAddress
+        normalizeEVMAddress(asset.contractAddress) ===
+          normalizeEVMAddress(contractAddress)
     )
-    return found as SmartContractFungibleAsset
```
0xDaedalus, line 267: `as` strikes again.

background/services/indexing/index.ts
```diff
@@ -551,28 +551,25 @@ export default class IndexingService extends BaseService<Events> {
     addressOnNetwork: AddressOnNetwork,
     contractAddress: string
   ): Promise<void> {
+    const normalizedAddress = normalizeEVMAddress(contractAddress)
```
0xDaedalus, line 549: Love the explicit variable naming

background/services/indexing/index.ts
```diff
@@ -128,13 +127,6 @@ export default class IndexingService extends BaseService<Events> {
         },
         handler: () => this.handleBalanceAlarm(true),
       },
-      forceBalance: {
```
0xDaedalus, line 131: rather than removing this - I think it makes more sense to change `runAtStart` to false.

The code written as is erroneously checks for balance updates at start - but also forces balance updates every 12 hours after start.  Wether we need the latter functionality or not is definitely debatable - but lets wait until we have some analytics data before getting rid of it entirely.

background/services/indexing/index.ts
```diff
@@ -760,7 +760,9 @@ export default class IndexingService extends BaseService<Events> {
         } else {
           await this.retrieveTokenBalances(
             addressOnNetwork,
-            await this.db.getAllKnownTokensForNetwork(addressOnNetwork.network)
+            this.cachedAssets[addressOnNetwork.network.chainID].filter(
```
0xDaedalus, line 763: Lets use the `getCachedAssets` method here to stay consistent with the rest of the codebase

background/services/indexing/db.ts
```diff
@@ -143,6 +144,58 @@ export class IndexingDatabase extends Dexie {
...
+        .filter(selectInvalidOrDuplicateRecords)
+        .delete()
+
+      await tx
+        .table("customAssets")
+        .filter(selectInvalidOrDuplicateRecords)
+        .delete()
+
+      const normalizeAssetAddress = (record: SmartContractFungibleAsset) => {
+        Object.assign(record, {
```
0xDaedalus, line 183: Spicy 🌶️

background/services/indexing/db.ts
mhluongo: I can confirm that `chainId` is often `undefined` / missing in the wild, and in that case should be defaulted to mainnet

hyphenized: > I would expect that either one of those missing would mean the given record is invalid - no?
Yeah I was just being safe here. I was not aware of `chainId` possibly being `undefined` in some cases 👀

mhluongo: It could be in an older version of the schema, perhaps? Just confident I've seen it in real-world token lists :upside_down_face:

greg-nagy: paranoid is good

background/services/indexing/index.ts
```diff
@@ -166,10 +169,10 @@ export default class IndexingService extends BaseService<Events> {
         await this.cacheAssetsForNetwork(network)
         this.emitter.emit("assets", this.cachedAssets[network.chainID])
       })
-    })
 
-    // Kick off token list fetching in the background
-    this.fetchAndCacheTokenLists()
+      // Force a balance refresh on service start
+      tokenListLoad.then(() => this.handleBalanceAlarm())
```
greg-nagy, line 174: Why mixing await and then? `await tokenListLoad; this.handleBalanceAlarm()` should also work here, right?

background/services/indexing/index.ts
```diff
@@ -252,20 +255,21 @@ export default class IndexingService extends BaseService<Events> {
    * @param network - the home network of the asset
    * @param contractAddress - the address of the asset on its home network
    */
-  async getKnownSmartContractAsset(
+  getKnownSmartContractAsset(
```
greg-nagy, line 258: I am always torn with this, but in the end I think this the right path.

background/services/indexing/db.ts
greg-nagy: and no need for eslint disable ... :)

greg-nagy (commented):
Love this PR and the fix!

I was testing the migrations:
- checkout main 
- yarn start
- import seed
- click all networks to activiate them 
- checkout branch
- restart build 
- reload extension 

I would like to get another set of eyes on this before merge. 

The code looks good. 

The PR description would be good to have the `what, how, why, test` structure.

Love the clean commit history 🙇‍♂️

background/services/indexing/index.ts
0xDaedalus: I think this implementation is correct - we don't want loading the token list to block initializing the indexing service

background/services/indexing/db.ts
0xDaedalus: My point here is that a record with a `chainId` of `1` and an address of `0x2d...` would be marked as invalid with the current boolean logic right?

```
typeof 1 !== "undefined" // type is number, true
&&
typeof "0x2d..." !== "undefined" // type is string, true
```

0xDaedalus (changes requested):
Tested - functionality looks good!  I think we need to change the boolean logic (or variable naming) mentioned above though in order to avoid confusion down the line.  Lets make sure to re-test after we do.

background/services/indexing/db.ts
hyphenized: Ahh I think I get where you're going, it would be marked as invalid but that should be the case for every token list asset that hasn't been transformed through [`tokenListToFungibleAssetsForNetwork`](https://github.com/tallycash/extension/blob/8ffb7c969efd8cedb826d803b28b4f7df46fa2ef/background/lib/token-lists.ts#L61) right?

background/constants/currencies.ts
```diff
@@ -37,33 +37,33 @@ export const RBTC: FungibleAsset & CoinGeckoAsset & NetworkBaseAsset = {
...
-      tokenLists: [],
-      websiteURL: "https://ethereum.org",
-    },
-  }
+export const OPTIMISTIC_ETH: NetworkBaseAsset = {
+  name: "Ether",
+  symbol: "ETH",
+  decimals: 18,
+  coinType: coinTypesByAssetSymbol.ETH,
+  contractAddress: "0xdeaddeaddeaddeaddeaddeaddeaddeaddead0000",
```
jagodarybacka, line 45: shouldn't the contract address be normalized?

hyphenized: Yes, I normalized it to be consistent with the new token list changes

0xDaedalus (approved):
📈

PR #2582, opened 2022-11-09
Support routing provider messages between alchemy and generic RPCs

0xDaedalus:
This PR is another step closer to our vision of a truly multi chain and multi RPC wallet.  While the primary impetus of this PR is to cutting down on infra spend - it also lays some groundwork for how we might handle chains added via `wallet_addEthereumChain`.

Relevant discussion in https://github.com/tallycash/extension/pull/2547

### What this PR does:
- Lays groundwork for networks that are not supported by Alchemy to be added to the wallet.
- Moves [pending block subscriptions](https://github.com/tallycash/extension/blob/support-generic-rpcs/background/services/chain/serial-fallback-provider.ts#L388:L388) off of alchemy providers by default
- [Prioritizes generic RPC providers](https://github.com/tallycash/extension/blob/support-generic-rpcs/background/services/chain/serial-fallback-provider.ts#L779:L779) as the "primary" providers to use in `SerialFallbackProvider`
- [Adds a mechanism](https://github.com/tallycash/extension/blob/support-generic-rpcs/background/services/chain/serial-fallback-provider.ts#L42:L42) that lets us route a certain percentage of calls to alchemy providers.
- Reduces our alchemy load by at least 50% - although the actual reduction will likely be more due to the aforementioned pending block subscription changes.

### What this PR does not do:
- Decouple us from alchemy (if alchemy endpoints go down for Ethereum / Optimism / Arbitrum / Polygon - we do not yet have a way to failover our `alchemy` calls to their generic rpc equivalents).
- Fully load balance between alchemy and generic rpc's (`alchemy_` rpc calls will still always go to alchemy providers).

### To Test
Load up a fresh wallet - navigate to different chains and ensure that on each chain:
  - [x] Asset balances load
  - [x] Activities load
  - [x] Transaction submission works
  - [x] Historical activities load 
  - [x] Once submitted - transactions properly show as `pending` in the activities list and then resolve once they are accepted.

Latest build: [extension-builds-2582](https://github.com/tallycash/extension/suites/9387846100/artifacts/442637134) (as of Fri, 18 Nov 2022 17:19:24 GMT).


greg-nagy (commented):
The solution looks good on the high leve, but need to get more into testing it. 

I am concerned about the all or nothing rollout of this change.
I think it's possible to set a % rollout in chrome store, which could probably mitigate this problem. 

Related concern: Do we know that their infrastructure won't break when we switch this on?

greg-nagy (approved):
looks good, works well, let's merge and include in the next release!

👑

Issue #2583, opened 2022-11-10
Polygon in-wallet swaps to WETH


Naxsun:
### Discord Discussion Link

https://discord.com/channels/808358975287722045/1037670242421321769/1040170198088556575

### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

In-wallet swap to WETH

### What did not work?

**When swapping to WETH on Polygon, an in-wallet swap swaps to**
- Wrapped Ethereum (PoS) https://polygonscan.com/token/0xae740d42e4ff0c5086b2b5b5d149eb2f9e1a754f
- Example in-wallet swap tx: https://polygonscan.com/tx/0xd99f973810754dbfb125016ebcd1ea0134acbf4f4fbbc864740fcefcf719fb06

**A swap on Matcha to WETH, swaps to**
- Wrapped Ether (Wormhole) :https://polygonscan.com/token/0x11cd37bb86f65419713f30673a480ea33c826872
- Example tx: https://polygonscan.com/tx/0xd54f2e851cee8c8533a5f0c4ab8fd7cb37aee6d76a6a0800ce76ba8dd716f8f1

**A swap on Uniswap to WETH** swaps to
- Wrapped Ether: https://polygonscan.com/token/0x7ceb23fd6bc0add59e62ac25578270cff1b9f619
- Example tx: https://polygonscan.com/tx/0x1105337e2a4394a0771289ab96011c870f472c3fd59307524884bb030898f984
### Version

v0.17.7

PR #2584, opened 2022-11-10
Add token contract links to the asset selector component

jagodarybacka:
Ref #2583
### What
Let's add a link to the scan website to make it easier for users to know what token it is - to avoid confusion where there are multiple tokens with the same or very similar names on the list.

https://user-images.githubusercontent.com/20949277/201067230-deca9eea-542f-4b2f-a915-00d83f73f088.mov

### Testing

- [ ] open asset selector (send or swap page) and check if links are working correctly
- [ ] base asset won't have link to the contract as it is not a token

Latest build: [extension-builds-2584](https://github.com/tallycash/extension/suites/9232166466/artifacts/431486585) (as of Thu, 10 Nov 2022 10:33:34 GMT).

### DIFF
--- a/ui/components/Shared/SharedAssetInput.tsx
+++ b/ui/components/Shared/SharedAssetInput.tsx
@@ -50,6 +50,7 @@ const symbolPriority = Object.fromEntries(
   ])
 )
 interface SelectAssetMenuContentProps<AssetType extends AnyAsset> {
+  currentNetwork: EVMNetwork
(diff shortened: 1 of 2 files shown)


VladUXUI:
Small update, but maybe for later would be to also alight the base asset value to the other values. even if they don't have icon

VladUXUI (approved):
Looking good! 🚀

greg-nagy (approved):
love this ❤️

PR #2589, opened 2022-11-10
Remove duplicated WETH from assets on Polygon

### LINKED ISSUE #2588: Duplicated WETH on Polygon
After migration there is a problem with duplicated WETH balance

image

jagodarybacka:
Resolves #2588
### What

Migration from 0.17.7 to 0.17.9 had side effect that the WETH was duplicated on the assets list on Polygon.
This PR removed duplicated asset.

### Testing

- [x] reinstall extension from 0.17.7 tag
- [x] add some accounts, check WETH and ETH balances on Polygon
- [x] change to this branch and reload extension
- [x] on Polygon - what was WETH should be POS - WRAPPED ETH, what was ETH is now WETH
- [x] check Wallet balance, combined balance, and Portfolio balances
- [x] test the same scenario with no accounts and with Polygon not activated

Latest build: [extension-builds-2589](https://github.com/tallycash/extension/suites/9241324928/artifacts/432110067) (as of Thu, 10 Nov 2022 18:10:58 GMT).


0xDaedalus (approved):
Tested and retested - lgtm 👍

PR #2590, opened 2022-11-10
Stop marking account activity on dapp connection.

0xDaedalus:
The "lastUserActivityOnNetwork" and "lastUserActivityOnAddress" properties are exclusively used to gate pull-based lookups for user assets / activites on networks.  markAccountActivity is responsible for updating these properties - but there is no utility in them being updated if the wallet is not open.  Since we already mark account activity via the "userActivityEncountered" and "newSelectedAccount" events when a user opens the wallet and selects a new account (respectively) there is no need for us to use a background action like connecting to a dapp to mark account activity.

### To Test
Use the application, navigate between dapps, between networks, add accounts, everything should work as before.

Latest build: [extension-builds-2590](https://github.com/tallycash/extension/suites/9320436077/artifacts/437857387) (as of Tue, 15 Nov 2022 17:45:15 GMT).


jagodarybacka (approved):
Code is fine, but I'm wondering about the case where the account was inactive and then the user connects it to dapp - one thing `markAccountActivity()` is doing is updating account balance - so may it be the case that user sees balance that is not up to date? 
Doesn't seem like a big problem to me tbh.

mhluongo:
Seems like we'd prefer the old behavior for a good UX... maybe we hold off on this until we have more data? Connecting to dApps is likely an indicator that the wallet will open soon.

0xDaedalus:
> so may it be the case that user sees balance that is not up to date?

One nice thing about how dapps generally interact with wallets is they never request information directly "from" the wallet.  For the most part they just use the wallets `provider` as an api for interacting directly with a given chain and use data queried from the chain directly.  There are of course exceptions to this ([we have one](https://github.com/tallycash/extension/blob/main/background/services/internal-ethereum-provider/index.ts#L202:L202)) but the strong suit of this pattern as a general rule is that the source of truth that dapps use is on-chain-data rather than cached data inside of the wallet.

> Connecting to dApps is likely an indicator that the wallet will open soon.

This was the original thinking when implementing this and I agree that it may be an indicator the wallet is gonna be open soon but I don't know if that necessarily translates to UX gains.  My _hypothesis_ is that for the majority of our relatively-new-to-crypto users, they themselves are initiating chain state changes (via submitting tx's) that are relevant to them.  Extrapolated this means is that if their wallet is not open (or hasn't been open for N minutes) state changes relevant to it are probably not happening.

I do agree that holding off on this change until we have more data would for sure let us make the best decision but I'm not sure if / when this data will be available.

mhluongo:
If we don't have a pressing product reason, and we don't have data... feels like a premature optimization.

> I do agree that holding off on this change until we have more data would for sure let us make the best decision but I'm not sure if / when this data will be available.

Would it help to have a place to write stuff like this down? "Questions we want answered with data" or similar, then we can figure out how to get it with PostHog. Might help us move toward a hypothesis / experiment flow

hyphenized:
I think we might want to keep this, `eth_requestAccounts` is also emitted when a user reloads the dApp page or performs certain actions within it, so this is correctly marking activity on the address/network pair. 

If we want users to see updated balances/assets transfers more often, we could also make `markAccountActivity` sync assets/balances on activity if a certain amount of time has passed since the last sync, as the current logic will only force a sync if 5 minutes+ have passed since the last “account activity”.

0xDaedalus:
> If we don't have a pressing product reason

I think the most salient product reason here is to reduce the amount of resources the wallet is using when it is in the background and irrelevant to the user

>  so this is correctly marking activity on the address/network pair.

I would argue that activity in a dapp is orthogonal to activity within the wallet.  E.g. I check aave and euler about 5x a day to check on rates of current assets I have in there without opening my wallet.  Another example would be browsing opensea for 20 minutes looking for an nft to buy before signing any kind of transaction.

> If we want users to see updated balances/assets transfers more often,

I think this concern is also not related here.  My goal with this change is to have users see updated balances/ asset transfers when it is relevant to them - and those things are _only_ relevant when the wallet is open since dapps requests that data on-demand using the wallet as a proxy

greg-nagy:
I am with @0xDaedalus on this one. His [points here](https://github.com/tallycash/extension/pull/2590#issuecomment-1314149137) are spot on.

An important thing here is that the dApps are NOT using our internal balances state. To the dApps, we are simply a communication channel responsible for signing transactions. Sometimes not even that. (But this is mentioned above.)

Also, another angle: Since we have introduced the inactivity concept, the extension is working in a mode where the extension state is updated on user action. The forced refetch happens every 12 hour > the user is likely to have stale data on action, then a quick refetch and use the wallet. I am not aware that it has caused any issue to our users so far. 

I think the bottom line is that we are not introducing something new, or breaking something that has worked well before, but increasing the usage of a pattern that has proven to be efficient in reducing our provider usage without causing user issues.

The code looks good, but not merging this so we can have an open conversation.

hyphenized:
> I would argue that activity in a dapp is orthogonal to activity within the wallet. E.g. I check aave and euler about 5x a day to check on rates of current assets I have in there without opening my wallet. Another example would be browsing opensea for 20 minutes looking for an nft to buy before signing any kind of transaction.

That's true but these events tell us the user is interacting with a dapp and could open the wallet anytime.

So we could update balances once the wallet is opened or, use these events to force syncs periodically to avoid showing outdated balances altogether.

I do agree that this will help reduce the amount of resources the wallet uses while in the background and I think a small delay loading balances is worth it so, it depends whether we want to prevent showing the user outdated balances straight away or not.

PR #2603, opened 2022-11-11
Update method for getting pending transactions via alchemy

0xDaedalus:
`alchemy_filteredNewFullPendingTransactions` is [deprecated](https://docs.alchemy.com/changelog/deprecated-support-for-alchemy_newfullpendingtransactions-and-alchemy_filterednewfullpendingtransactions) - this PR updates our extension to use the new recommended method.

### To Test
- [ ] Submit a transaction on Ethereum & Polygon - the activity should be picked up and resolved when transaction is mined.
- [ ] Send ETH or MATIC via metamask to your tally wallet address - balance should be updated as soon as tx is mined.

Latest build: [extension-builds-2603](https://github.com/tallycash/extension/suites/9284250991/artifacts/435247150) (as of Mon, 14 Nov 2022 05:10:44 GMT).

### DIFF
--- a/background/services/chain/serial-fallback-provider.ts
+++ b/background/services/chain/serial-fallback-provider.ts
@@ -644,7 +644,7 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
   }
 
   /**
-   * Attempts to subscribe to full pending transactions in an Alchemy-specific
+   * Attempts to subscribe to pending transactions in an Alchemy-specific
    * way. Returns `subscribed` if the subscription succeeded, or `unsupported`
    * if the underlying provider did not support Alchemy-specific subscriptions.
    */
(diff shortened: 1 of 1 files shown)


background/services/chain/serial-fallback-provider.ts
```diff
@@ -655,7 +655,10 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
     try {
       await this.subscribe(
         "filteredNewFullPendingTransactionsSubscriptionID",
-        ["alchemy_filteredNewFullPendingTransactions", { address }],
+        [
+          "alchemy_pendingTransactions",
+          { fromAddress: address, toAddress: address },
```
greg-nagy, line 660: wow, happy that this works well

I thought we will need to have 2 subscriptions based on how the filters work

Issue #2606, opened 2022-11-12
User balances slow to update


mr-michael:
User balances appear slow to update after a variety of actions are taken (bridge, swap, send, receive, etc.) that impact balance

Examples:

https://discord.com/channels/808358975287722045/888500174685614090/1040934220710826074
https://discord.com/channels/808358975287722045/888500174685614090/1040894186553102346
https://discord.com/channels/808358975287722045/888500174685614090/1040881421557964912
https://discord.com/channels/808358975287722045/888500174685614090/1040849194585444473
https://discord.com/channels/808358975287722045/888500174685614090/1040813236733349948
https://discord.com/channels/808358975287722045/888500174685614090/1040840667452145714
https://discord.com/channels/808358975287722045/888500174685614090/1040857200840691753

greg-nagy:
Merging #2485 

### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/1032600715610308638

### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

View balance

### What did not work?

User:
"Please add a refresh option to the wallet or set the UI to update the portfolio. All assets are old until we connect to a dapp. This image shows that I have weth in tally but the dapp says no amount of it. You don't have it. The problem is when I want to do an internal swap in the Tally wallet."

![image](https://user-images.githubusercontent.com/71384922/197483535-b1c3910e-9d17-497c-bc07-d3ad5b76b023.png)

![image](https://user-images.githubusercontent.com/71384922/197483550-1a5a948a-d6a1-4bb1-bef4-de55c46ee139.png)

Logs:
[logs_v0_17_3__2022-10-20T14_08_2803_00 (1).txt](https://github.com/tallycash/extension/files/9850109/logs_v0_17_3__2022-10-20T14_08_2803_00.1.txt)



### Version

v0.17.3

### Relevant log output

_No response_

Naxsun:
I don't know if this is new info or old info. 

But a significant portion of the cases involves bridging, where the bridged asset shows up on the new network, but also remains as balance on the previous network. 

E.g.
https://discord.com/channels/808358975287722045/888500174685614090/1041787272414117898

chlolands:
[logs_v0_17_9__2022-11-14T16_55_07-05_00.txt](https://github.com/tallycash/extension/files/10007408/logs_v0_17_9__2022-11-14T16_55_07-05_00.txt)
[...]

PR #2607, opened 2022-11-14
Fix ghost balances

hyphenized:
Account balance updates weren't clearing up tokens no longer being held because the updated balances payload did not include zero amount balances

Refs: #2606

## To Test:
- [ ] Get X amount of tokens 
- [ ] Bridge these X tokens
- [ ] The balance should now disappear from the network you bridged the assets from

Latest build: [extension-builds-2607](https://github.com/tallycash/extension/suites/9317863059/artifacts/437670843) (as of Tue, 15 Nov 2022 15:41:46 GMT).


background/main.ts
```diff
@@ -845,27 +846,48 @@ export default class Main extends BaseService<never> {
   async connectIndexingService(): Promise<void> {
     this.indexingService.emitter.on(
       "accountsWithBalances",
-      async (accountsWithBalances) => {
+      async ({ balances, addressOnNetwork }) => {
         const assetsToTrack = await this.indexingService.getAssetsToTrack()
+        const trackedAccounts = await this.chainService.getAccountsToTrack()
+        const allTrackedAddresses = new Set(
+          trackedAccounts.map((account) => account.address)
```
jagodarybacka, line 853: Are these addresses normalized in chainService? 
```suggestion
          trackedAccounts.map((account) => normalizeEVMAddress(account.address))
```

background/services/indexing/index.ts
```diff
@@ -528,7 +537,10 @@ export default class IndexingService extends BaseService<Events> {
     )
 
     await this.db.addBalances(accountBalances)
-    this.emitter.emit("accountsWithBalances", accountBalances)
+    this.emitter.emit("accountsWithBalances", {
+      balances: accountBalances,
+      addressOnNetwork: addressNetwork,
```
0xDaedalus, line 542: I'm not sure that we need to emit this since we already have the address and network available [in the AccountBalance](https://github.com/tallycash/extension/blob/main/background/accounts.ts#L14:L14).

I see where you're coming from with this change though - the account balances are only implicitly coupled with address networks and this changes gives us explicit coupling through the api to emit them. hhhhmmmmnnn

background/main.ts
hyphenized: Good question, all the wallet account addresses are normalized before reaching the chain db [here](https://github.com/tallycash/extension/blob/272e38f90c0e3dff95ad78f4de7c42ec5f89ada5/background/redux-slices/accounts.ts#L418) with the exception of ledger addresses which are normalized after derivation.

background/services/indexing/index.ts
hyphenized: It's because there is an edge case where if a user has one token left and then swaps it, the redux action will receive an empty array leaving us unable to tell which address/network has no token balances left

background/main.ts
```diff
@@ -845,27 +846,48 @@ export default class Main extends BaseService<never> {
...
+        )
+
+        if (
+          !allTrackedAddresses.has(
+            // REVIEW: Address emitted from indexing service might not be normalized
```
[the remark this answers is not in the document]
hyphenized: This comment was actually for the reviewers hahah. I've checked the indexing service and removed the comment

background/services/indexing/index.ts
0xDaedalus: Awesome - can we add a note about this to the comment on line 58?

background/services/chain/index.ts
```diff
@@ -823,8 +840,20 @@ export default class ChainService extends BaseService<Events> {
       dataSource: "alchemy", // TODO do this properly (eg provider isn't Alchemy)
       retrievedAt: Date.now(),
     }
-    this.emitter.emit("accountsWithBalances", [accountBalance])
-    await this.db.addBalance(accountBalance)
+
+    // Don't emit or save if the account isn't tracked
+    if (allTrackedAddresses.has(normalizedAddress)) {
+      this.emitter.emit("accountsWithBalances", {
```
greg-nagy, line 846: This is likely to be a huge win for perf

greg-nagy:
A note, that I believe this behavior is coming from https://github.com/tallycash/extension/pull/2507, and we should find a better place for that filter than revisit this PR 

I think @0xDaedalus is already on this

0xDaedalus (approved):
Echoing Greg that there might be a way to accomplish this same functionality with less code but lets get this bugfix in before the release and iterate in a follow-on PR.

That being said this PR is fantastic - and has some sweet performance improvements and edge-case catches outside the scope of the original bug ticket - bravo!.

🚢

PR #2628, opened 2022-11-16
added balance data to buy assets in swap view

dills122:
## Description:

Swap view only showed the balance of the token you were swapping from, not for the one you were swapping to.

## Fix 

Added balance for token being swapped to, if the user has any of that token currently.

## Demo

https://user-images.githubusercontent.com/15662762/202069032-dee37c6e-75bb-4bf8-82b9-0c60229abadc.mov

Related to: https://github.com/tallycash/extension/issues/974

Latest build: [extension-builds-2628](https://github.com/tallycash/extension/suites/9570521344/artifacts/455699473) (as of Tue, 29 Nov 2022 19:19:01 GMT).


mhluongo:
Ooh, I like this — @VladUXUI what do you think, is there risk of confusion here?

VladUXUI:
Thanks @dills122! This looks good and I don't think there will be any confusion here.
The pattern for the swap "from" is kept, so users will know already that the small amount is current balance

VladUXUI:
@dills122 i see you've been picking up a few UI things(super excited about that!)
Some of these tickets are quite old and some might even be outdated (we need to groom our backlog)

But for the time being, if you first want to ask if "this ticket is still something we want to pick up?" i'll gladly answer as soon as possible.
This is to avoid you doing anything that we might have reconsidered or had another fix for.

Totally up to you

VladUXUI:
Also i'm assuming the exchange rate in the video is because of local build right?
$6 to $101 is not bad 😅

VladUXUI:
Or... is swap amount = balance after swap? (so current balance + this balance?)
If that is the case, that would indeed be confusing

dills122:
> Also i'm assuming the exchange rate in the video is because of local build right? $6 to $101 is not bad 😅

@VladUXUI Yeah I think the crazy numbers are just from some of the testnet coins. I did a quick check using `LINK` on test net and it looks like the values are the same with & without this fix.


Without these changes:

tally-ui-974-before

With these changes:

tally-ui-974-after

0xDaedalus:
Hey @dills122 - wondering if this PR is good to be taken out of Draft and ready for code review?

dills122:
> Hey @dills122 - wondering if this PR is good to be taken out of Draft and ready for code review?

Hey @0xDaedalus, yep this should be ready for a look. Slipped my mind with the holiday last week.

ui/components/Shared/SharedAssetItem.tsx
```diff
@@ -6,24 +6,19 @@ import SharedAssetIcon from "./SharedAssetIcon"
...
-  | {
-      asset: T
-      amount: bigint
-      localizedDecimalAmount: string
-    }
+export type AnyAssetWithOptionalAmount<T extends AnyAsset> = {
+  asset: T
+  amount?: bigint
+  localizedDecimalAmount?: string
+}
```
hyphenized, line 13: I don't believe this is the same, marking these properties as additional prevents us from narrowing the correct type by checking for the existence of either `amount` or `localizedDecimalAmount` because these are now marked as being possibly `undefined`.

hyphenized:
@dills122 We've made some changes to the Swap Page, could you please update this branch? Thank you 🙏

jagodarybacka:
Closing due to inactivity

PR #2635, opened 2022-11-16
Turn on RNS name resolution

0xDaedalus:
Turns on RNS name resolution.

### To Test
- [ ] try to import the the address `riverplate.rsk` via read-only import.

Latest build: [extension-builds-2635](https://github.com/tallycash/extension/suites/9344983858/artifacts/439558195) (as of Wed, 16 Nov 2022 18:40:58 GMT).


hyphenized (approved):
I'm unable to import the address through onboarding



But it works fine after switching to RSK



We can fix it here or create a follow-up issue

0xDaedalus:
Hmmmn this is a good point - I'll go ahead and make a follow-on ticket.

I do see though that sending to RNS names works - I think thats good enough to flip this on since we have the `(beta)` flag on RSK anyhow.

image

Would you mind confirming you can recreate the above screenshot?

hyphenized:
> Would you mind confirming you can recreate the above screenshot?

Yup that's working
image

PR #2636, opened 2022-11-16
🤖💡Tell Me You Get Me: Implement default on for analytics and notification

### LINKED ISSUE #2609: Default on + notification + opt-out

greg-nagy:
## Mandatory bad joke
<details>
  <summary>I get you ... the short story of Twilight</summary>
  
  CleanShot 2022-11-16 at 16 34 54
</details>

## What

Let's introduce the `default on` feature flag and behavior with notification for analytics. When this feature flag is flipped 
1) the analytics collection should start 
2) the user should see a notification 
3) the analytics menu item should become visible in the settings

The `SUPPORT_ANALYTICS` is a global feature flag that affects the sending of the analytics events on the lowest level. 
This is a user context flag that enables the feature. 

## How 

We have a settings flag in the preference service to detect when the `ENABLE_ANALYTICS_DEFAULT_ON` is turned on. This way, we can ship analytics but not collect any data before the users are notified.

The check happens [in main](https://github.com/tallycash/extension/blob/9b3e2b7ff895795639bbd05482b2b4710c7c0fa6/background/main.ts#L1422-L1433)

## Test

- start and load extension as is 
- no event should come through to `https://app.posthog.com/events`
- flip `SUPPORT_ANALYTICS=true` and restart the build
- no event should come through to `https://app.posthog.com/events`
- no notification should be visible
- flip `ENABLE_ANALYTICS_DEFAULT_ON=true` and set `USE_ANALYTICS_SOURCE=<myname>` and restart build 
- notification should be displayed
- Settings menu should appear
- events should start coming with a new uuid (there is a 60s delay)
- use the `X` on the notification > notification should disappear and stay that way
- in the settings turn off analytics
- events should stop flowing

## Testing Env

```
SUPPORT_ANALYTICS=true
ENABLE_ANALYTICS_DEFAULT_ON=true
```

## UI

CleanShot 2022-11-16 at 16 17 50

Closes #2609

Latest build: [extension-builds-2636](https://github.com/tallycash/extension/suites/9358511602/artifacts/440549854) (as of Thu, 17 Nov 2022 10:54:08 GMT).


ui/components/Wallet/WalletAnalyticsNotificationBanner.tsx
```diff
@@ -0,0 +1,101 @@
...
+        .container {
+          margin: 0 16px;
+          max-height: 200px;
+          width: 352px;
+        }
+        .container.hide {
+          max-height: 0;
+          pointer-events: none;
+          opacity: 0;
+          transition: all 500ms ease;
```
hyphenized, line 73: hmmm I think this transition should be in the parent `.container`

ui/components/Wallet/WalletAnalyticsNotificationBanner.tsx
```diff
@@ -0,0 +1,101 @@
...
+              iconSmall="settings"
+              iconPosition="left"
+            >
+              {t("settingsLink")}
+            </SharedButton>
+          </div>
+          <SharedIcon
+            onClick={() => dispatch(setShowAnalyticsNotification(false))}
+            icon="icons/s/close.svg"
+            ariaLabel="close"
```
hyphenized, line 56: `t("shared.close")` 🌐

ui/components/Wallet/WalletAnalyticsNotificationBanner.tsx
```diff
@@ -0,0 +1,101 @@
...
+    <div
+      className={classNames("container", {
+        hide: !showNotification,
+      })}
+    >
+      <SharedBanner>
+        <div className="content_container">
+          <SharedIcon
+            icon="icons/m/notif-correct.svg"
+            ariaLabel="correct-icon"
```
hyphenized, line 35: I think we can remove this since it's not clickable or change `SharedIcon` to set a label on the `<i>`

hyphenized (dismissed):
Super! Left some comments

background/services/preferences/defaults.ts
```diff
@@ -30,7 +31,10 @@ const defaultPreferences: Preferences = {
   },
   accountSignersSettings: [],
   analytics: {
-    isEnabled: false,
+    isEnabled: isEnabled(FeatureFlags.ENABLE_ANALYTICS_DEFAULT_ON),
+    hasDefaultOnBeenTurnedOn: isEnabled(
```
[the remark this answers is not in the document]
greg-nagy: Good catch 

Not in a dexie migration, but in the check from main.

These sholuld be false by default and should be turned on when the check fires in main. 
Wouldn't have caused issues, because we were planning to release with feature flags false first so the behavior would have been the same.
But that was a very implicit behavior and could have also been a source of errors if the release order is not waht we expect. 

🙏

ui/components/Wallet/WalletAnalyticsNotificationBanner.tsx
greg-nagy: [c12ba84](https://github.com/tallycash/extension/pull/2636/commits/c12ba846f08f7e767e57e58f3c2bb501f6f8bc96)

ui/components/Wallet/WalletAnalyticsNotificationBanner.tsx
greg-nagy: [4ab40ab](https://github.com/tallycash/extension/pull/2636/commits/4ab40ab0033b965718574f7ae31a447f46a86672)

ui/components/Wallet/WalletAnalyticsNotificationBanner.tsx
greg-nagy: [ba893a8](https://github.com/tallycash/extension/pull/2636/commits/ba893a80c02a11dc907994d6a4afe306bf5b6ed9)

0xDaedalus (approved):
📈

PR #2639, opened 2022-11-16
Add across.to to MetaMask impersonation list

pxrl:
Many users have reported issues connecting to Across with Tally installed. A number of these errors are due to conflicting extensions, but there are still cases where some users are unable to connect despite having disabled all other extensions and restarted their browser.

This commit is therefore a last-ditch effort to flush out some other underlying issue, and to provide users with a way to reliably use Tally on across until the underlying efforts are identified and corrected.


pxrl:
Closing after discussion with @Naxsun.

PR #2640, opened 2022-11-17
Add NFTs page to the wallet 

### LINKED ISSUE #2595: Add NFTs page to the wallet
- [ ] create new feature flag and add new changes based on the flag
- [ ] add a new tab to the bottom ribbon
image

- [ ] remove swaps page from bottom ribbon, move it to the Wallet page (switch between new and old versions based on the flag)
image

- [ ] fix the order of pages in the bottom ribbon

kkosiorowska:
Closes #2595 

This PR adds an NFTs page to the wallet.

Changes:
- add a new NFTs tab to the bottom ribbon
- move the swap page to the Wallet page

**UI**

Before
Screenshot 2022-11-18 at 19 05 19

After
Screenshot 2022-11-18 at 19 04 23

Screenshot 2022-11-18 at 19 30 33

**To test**
Set up `SUPPORT_NFT_TAB` to `true`.
- [ ] check that the NFTs tab is added to the bottom ribbon
- [ ]  check if the swap tab is removed from the bottom ribbon
- [ ] check that the swap page is moved to the Wallet page

Latest build: [extension-builds-2640](https://github.com/tallycash/extension/suites/9462540285/artifacts/448001089) (as of Wed, 23 Nov 2022 09:03:25 GMT).


.env.defaults
```diff
@@ -25,6 +25,7 @@ SUPPORT_FORGOT_PASSWORD=false
 ENABLE_ACHIEVEMENTS_TAB=true
 SUPPORT_ACHIEVEMENTS_BANNER=false
 SWITCH_RUNTIME_FLAGS=false
+HIDE_NFTS=true
```
mhluongo, line 28: We should prefer "SHOW_NFTS=false" as the default, or alternatively something like "SUPPORT_NFT_TAB".

ui/components/Wallet/WalletAccountBalanceControl.tsx
```diff
@@ -65,25 +68,38 @@ export default function WalletAccountBalanceControl(
...
-                  <SharedButton
-                    onClick={handleClick}
-                    iconSmall="receive"
-                    size="medium"
-                    type="tertiary"
-                    iconPosition="left"
-                  >
-                    {t("wallet.receive")}
-                  </SharedButton>
+                <div className="actions_button_wrap">
```
greg-nagy, line 71: This should also depend on the feature flag.

jagodarybacka (approved):
Non-blocking comment, I think it is fine :)

Will wait for @VladUXUI to take a look too

jagodarybacka (approved):
✅

greg-nagy (changes requested):
This looks good, and works well! 

Nice job here 🚀

After blockers are fixed this is good to merge.

ui/components/Shared/SharedSquareButton.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    onClick?.(event)
+    if (linkTo) {
+      setNavigateTo(linkTo)
+    }
+  }
+
+  if (navigateTo && navigateTo === linkTo) {
+    return <Redirect push to={linkTo} />
+  }
```
hyphenized, line 51: We can let the parent component handle navigation through the `onClick` prop

e.g.
```tsx
function BiggerComponent(){
  ...
  const history = useHistory();
  ...
  return <...>
    <SharedSquareButton onClick={()=> history.push("/somewhere") }>somewhere</SharedSquareButton>
  </...>
}

File: ui/components/Wallet/WalletAccountBalanceControl.tsx [hunk 4]
```diff
@@ -65,25 +68,38 @@ export default function WalletAccountBalanceControl(
...
+                  <div className="button_wrap">
```
hyphenized, line 72: I think these buttons should only change if the feature flag is enabled

hyphenized (changes requested):
Great work! Left some comments.

mhluongo:
@VladUXUI @mr-michael this is the sort of change that could grow or shrink swap volume significantly. Wanted to make sure you both were sure that's where we want the swap page, and that we're keeping track of rough per-install swap volume so we can revert if it's negative.

VladUXUI:
> Non-blocking comment, I think it is fine :)
> 
> Will wait for @VladUXUI to take a look too

Looking good! yes the button on the asset cards can remain

VladUXUI:
> @VladUXUI @mr-michael this is the sort of change that could grow or shrink swap volume significantly. Wanted to make sure you both were sure that's where we want the swap page, and that we're keeping track of rough per-install swap volume so we can revert if it's negative.

Swaps had a very different account/network context than the rest of the pages, so imo had to be moved. The placement is one that users are very familiar with as this is a pattern employed by most extensions or application wallets.
I don't think it's more difficult to find there then it would be in our main menu. We also have swaps button in more places than your usual extension/app.
Fairly confident it won't negatively affect our swaps volume, but good point about tracking! Will do that and see if we need to revert

VladUXUI (approved):

ui/components/Shared/SharedSquareButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  return (
+    <button type="button" aria-label={ariaLabel} onClick={handleClick}>
+      <div className="content_wrap">
```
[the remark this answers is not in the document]
kkosiorowska: Let's make it simple.

greg-nagy (approved):
looks good, let's 🚢

PR #2641, opened 2022-11-17
[FIX] Importing addresses via RNS

ahsan-javaiid:
## Description
RNS names not resolving when user import via "Read-Only address" import because by default `ethereum` network is selected.

1) Refer to the issue https://github.com/tallycash/extension/issues/2638
2) Refer to the pull request: https://github.com/tallycash/extension/pull/2635 

## Root Cause
**Case 1**:
If `rootstock` network is selected then "Read-only address" import works fine.  This is the scenario when user already created or imported wallet account and connected to rootstock network.
**Case 2**:
When user is coming first time (No wallet) then default selected network will be `ethereum`. In this case "Read-only address" import field uses `ethereum` network due to which `rns` name is getting rejected by wrong network (ethereum). 

## Possible Solutions
- Provide UI to select the network (I think it's best suited ) on address import screen
- Auto detect the network (May work out but user may get confused -> From which network it get imported)

## Demo 
Screenshot 2022-11-17 at 4 31 05 PM

## Type of Change
- [X] Fix

## Testing information
- Try importing RNS name in by click of "Read-Only address"
- Follow cases given in root cause section above

## Checklist
- [X] Git hooks enabled
- [X] No lint errors
- [X] Git commit is signed

Latest build: [extension-builds-2641](https://github.com/tallycash/extension/suites/9363298910/artifacts/440869723) (as of Thu, 17 Nov 2022 14:58:01 GMT).


ahsan-javaiid:
cc: @0xDaedalus @alepc253

hyphenized:
Looks good, but I think having the user manually select the network is going hurt the UX. We could also remove the network check from the RNS resolver for now since no other networks have `.rsk` domains

alepc253:
@hyphenized this approach (removing the network selector) will make much more sense if we finally apply the right derivation path to every network. Regarding the network check removal, yes, `.rsk` might be enough to assume that the network is Rootstock.

ahsan-javaiid:
Done 👍

hyphenized (approved):
Works well!

VladUXUI:
Hey all, this needs a better solve, as content and CTA are not aligned here.
Also the spacing is a bit wrong here.
Let's please revert and do this right

hyphenized:
@VladUXUI the UI changes were reverted in favor of changing the logic for name resolution

VladUXUI:
image

Here is a quick version that doesn't change our current flow but still enables users to use RSK. This also keeps the content more accurate to the chain.

We do need a better fix long term, but let's please change to this for now

Issue #2619, opened 2022-11-14
Create NFTs service and database


jagodarybacka:
## Redux state

Redux should be treated as a cache for collections and NFTs. We aim to keep it as small as possible and not let it grow without control.

- [x] redux should be initialized from the database - to avoid state migrations
- [x] update collections data
- [x] select collections

## Service

- [x] initialize redux on start
- [x] handle fetching data from providers
- [x] providers should be easily configurable
- [x] handle refetching and updating data

## Database

- [x] save NFTs data
- [x] select NFTs from given collection
- [x] update NFTs data

jagodarybacka:
Closing as most of this is already implemented, missing tasks will be added

PR #2642, opened 2022-11-17
NFTs background

jagodarybacka:
Ref #2619
### What

- This PR introduces NFTs service, database and redux slice with basic functionality:
  - fetching collections by account and tracked networks, saving them to database
  - fetching and saving nfts from given collections
  - caching collections in redux store
- Adds basic types: NFT, NFTCollection
- Adds new nfts providers map
- Handles fetching collections and NFTs from SimpleHash and POAP APIs

Based on the `SUPPORT_NFT_TAB`:

- if flag is `true`: create service and database, create `nfts` and `collections` tables, initialize redux slice based on the database contents
- if flag is `false`: create service but don't connect it to other services, create database but don't create any tables, don't initialize redux slice at all

### Testing

- with `SUPPORT_NFT_TAB = false` 
  - reload extension, check if there is no database `tally/nfts` visible in DevTools Application >  IndexedDB
  - check if no `nftsUpdate` redux slide was created using script:  
```javascript
    const state = JSON.parse(stateString, null, 2)
    console.log(state)
})
```
- with `SUPPORT_NFT_TAB = true` 
  - reload extension, redux slice should be created and collections should be fetched for accounts that are added (check db and redux slice)

Latest build: [extension-builds-2642](https://github.com/tallycash/extension/suites/9448736594/artifacts/446992982) (as of Tue, 22 Nov 2022 16:13:38 GMT).


0xDaedalus:
draft comment for a draft PR - I we'll definitely want to break this up into multiple PRs - and probably breakup https://github.com/tallycash/extension/pull/2642/commits/9e957a8f04d95ae746bc3743a7034ff77cbb045e into multiple commits as well.  Happy to commit (!) to prioritizing PR reviews so that we don't get stuck with a waterfall of PRs all depending on each other.

jagodarybacka:
Updated commits.

Are we ok with merging state like this - with ugly, duplicated files (two nfts redux slices, duplicated types etc)? We can't remove it before we are not in the same place with new redux slice as with the previous implementation.

greg-nagy (commented):
nice work so far! 

Agree with @0xDaedalus here > let's split this up to multiple PRs and we are going to review in a timely manner so you don't get stuck 

A suggestion on how to break this up:
- introduce a feature flag
  > this can be a single PR in itself, or you can add the skeleton work for redux slice + service. 
- create the service and make it's initialisation dependent from the feature flag 
    - > this will be a bit annoying, but worth it, because then we don't need to worry about db migrations etc during development 
- create the new redux slice and make it dependent on the feature flag  + wire the parts together 
    -  > it's easy to do in the `redux-slices/index.ts`

greg-nagy:
> Are we ok with merging state like this - with ugly, duplicated files (two nfts redux slices, duplicated types etc)? We can't remove it before we are not in the same place with new redux slice as with the previous implementation.

We should have the duplications and the "ugliness" behind the feature flag.

jagodarybacka:
@greg-nagy I think the current solution will do the trick - service is created but it is not doing anything, database is created but no tables are added, redux slice is not created.  

> duplications 

I want to keep the `_update` files separate as it is already quite a lot of code and it would be hard to maintain within the same file with the old code

greg-nagy:
@jagodarybacka are you sure that you want to start maintaining migrations for redux and db from the point in time that this PR is shipped in a release?

jagodarybacka:
@greg-nagy with the flag off there should be no redux slice and no database tables so there should be nothing that would need a migration

greg-nagy:
> @greg-nagy with the flag off there should be no redux slice and no database tables so there should be nothing that would need a migration

sorry, I missed the changes and replied quickly at the end of the day

background/lib/nfts.ts
```diff
@@ -116,6 +130,7 @@ function poapNFTModelToNFT(original: PoapNFTModel): NFT {
   }
 }
 
+// eslint-disable-next-line import/prefer-default-export
```
greg-nagy, line 133: ❤️

background/lib/nfts_update.ts
```diff
@@ -0,0 +1,103 @@
...
+} from "../nfts"
+
+function groupChainsByAddress(accounts: AddressOnNetwork[]) {
+  return accounts.reduce((result, account) => {
+    const {
+      address,
+      network: { chainID },
+    } = account
+    if (CHAIN_ID_TO_NFT_METADATA_PROVIDER[chainID]) {
+      result[address] ??= [] // eslint-disable-line no-param-reassign
```
greg-nagy, line 24: If you use a special name for the variable like `acc`, `res` you don't need the eslint disable.

I cand't find the exact eslint rule, but maybe @Shadowfiend can.

background/lib/nfts_update.ts
```diff
@@ -0,0 +1,103 @@
...
+export function getNFTCollections(
+  accounts: AddressOnNetwork[]
+): Promise<NFTCollection[]>[] {
+  const chainIDsByAddress = groupChainsByAddress(accounts)
+
+  return Object.entries(chainIDsByAddress).flatMap(
+    async ([address, chainIDs]) => {
+      const collections: NFTCollection[] = []
+      if (
+        chainIDs.filter((chainID) =>
```
greg-nagy, line 85: same as above

background/lib/nfts_update.ts
```diff
@@ -0,0 +1,103 @@
...
+        const { nfts: poapNFTs } = await poapGetNFTs(address)
+        nfts.push(...poapNFTs)
+      }
+
+      const simpleHashChains = chainIDs.filter((chainID) =>
+        NFT_PROVIDER_TO_CHAIN.simplehash.includes(chainID)
+      )
+
+      if (simpleHashChains.length) {
+        await Promise.all(
```
greg-nagy, line 56: Here `Promise.allSettled()` is probably a better fit.

background/services/nfts/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    private chainService: ChainService
+  ) {
+    super()
+  }
+
+  protected override async internalStartService(): Promise<void> {
+    await super.internalStartService()
+
+    // Nothing else beside creating a service should happen when feature flag is off
+    if (isEnabled(FeatureFlags.SUPPORT_NFT_TAB)) {
```
greg-nagy, line 38: For services when you put them behind a feature flag wrapping the `const nftsService = NFTsService.create(chainService)` with it is usually the easiest thing to do. 

But this is also perfectly fine.

background/services/nfts/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    )
+  }
+
+  async fetchCollections(accounts?: AddressOnNetwork[]): Promise<void> {
+    const accountsToFetch =
+      accounts ?? (await this.chainService.getAccountsToTrack())
+
+    await Promise.all(
+      getNFTCollections(accountsToFetch).map(async (request) =>
+        request.then(async (collections) => {
```
greg-nagy, line 69: It might be better to break these off into multiple steps instead of promise chaining.

greg-nagy (commented):
lot of changes, lot of neat things here ✨

I have signaled which issues are blockers, the rest are nice to haves.

I expect that multiple rounds of review will be here, because of the size.

Anyhow this is looking really amazing

background/lib/simple-hash_update.ts
```diff
@@ -0,0 +1,262 @@
...
+    "https://api.simplehash.com/api/v0/nfts/collections_by_wallets"
+  )
+  requestURL.searchParams.set("chains", getChainIDsNames(chainIDs))
+  requestURL.searchParams.set("wallet_addresses", address)
+
+  const headers = new Headers()
+  headers.set("X-API-KEY", process.env.SIMPLE_HASH_API_KEY ?? "")
+
+  try {
+    const result = (await (
```
hyphenized, line 246: We can also avoid type assertions here
```suggestion
    const result: SimpleHashCollectionsByWalletAPIResponse = (await (
```

hyphenized (commented):
Awesome work 💯! Gave it a quick look and left a couple of comments

background/services/nfts/index.ts
jagodarybacka: Yeah it's kinda rough now, will think of better way later during development

background/services/nfts/db.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  private collections!: Dexie.Table<NFTCollection, number>
+
+  constructor() {
+    super("tally/nfts")
+
+    // No tables are created when feature flag is off
+    if (isEnabled(FeatureFlags.SUPPORT_NFT_TAB)) {
+      this.version(1).stores({
+        nfts: "&[id+collectionID+owner+network.chainID],id,collectionID,owner,network.chainID",
```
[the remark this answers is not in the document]
jagodarybacka: hmmm, I will think about it later!

hyphenized (approved):
This is awesome! Let's get this merged! 🚀

background/services/nfts/db.ts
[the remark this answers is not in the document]
greg-nagy: [This is how it looks like when it needs to happen](https://github.com/tallycash/extension/blob/8baa9770a5db40ec54da889fb4415d4e66771c4c/background/services/provider-bridge/db.ts#L12-L56)

PR #2649, opened 2022-11-18
Speed up transaction enrichment.

0xDaedalus:
Closes https://github.com/orgs/tallycash/projects/5

### The Problem
Users with slow internet speeds or high connection latency have a very degraded experience using Tally Ho as compared to their experience using MetaMask.  This is primarily manifested through long (sometimes exceeding multiple minutes) loading times when loading the Sign Transaction screen.

The root cause of this is the sheer number of network requests we make when enriching a transaction combined with the fact that we do not display a transaction for the user to sign until it has fully completed enrichment.  On a connection with e.g. 25 kbps download speed and 200ms latency this enrichment can sometimes take up to 20s.

The ideal solution here is probably to decouple the ability to sign transactions from their enrichment status.  This is a change that requires a somewhat large re-architecture so is a nice-to-have down the line but not something we can immediately implement.

### The Solution
This PR is somewhat of a band-aid fix and the underlying issues will be addressed as part of the [Enrichment Refactor Milestone](https://github.com/tallycash/extension/milestone/18) - but at a high level the fixes here:
- Eliminate calls to the RNS name resolver when enriching transactions not on the Rootstock network
- Sets a 10 second timeout for transaction signature enrichment
- Adds caches to two calls we tend to make successively.

### To Test
- Load extension
- Set network throttling to `3G Slow`
- Open up multiple networks - browse some dapps
- Attempt to sign different kinds of transactions - enrichment should never take more than 10s
- Open up the network tab and compare network traffic on `main` and this branch - there should be a notable decrease.

Latest build: [extension-builds-2649](https://github.com/tallycash/extension/suites/9442814909/artifacts/446589718) (as of Tue, 22 Nov 2022 11:31:25 GMT).


greg-nagy (approved):
I love this PR :) 

Haven't tested, I would like to ask for help with that.
The code changes are looking good!

Will check back on this tomorrow morning and merge if nobody beats me to it.

background/lib/utils/index.ts
```diff
@@ -168,3 +168,5 @@ export const hexToAscii = (hex_: string): string => {
     str += String.fromCharCode(parseInt(hex.substr(i, 2), 16))
   return str.replace("\x00", "")
 }
+
+export const wait = (ms: number) => new Promise<void>((r) => setTimeout(r, ms))
```
jagodarybacka, line 172: Can we get rid of this lint warning?

jagodarybacka (approved):
I can't get Uniswap to work with that slow internet but in-wallet swaps are working ok I think. Not the best experience but it's better than nothing 😉

greg-nagy:
> I can't get Uniswap to work with that slow internet but in-wallet swaps are working ok I think. Not the best experience but it's better than nothing 😉

uniswap is blocked by an unrelated issue with ankr rpcs — we are working on that but until it's fixed the testing of this PR is not really feasible — or you need to comment out all the ankr rpcs in `CHAIN_ID_TO_RPC_URLS`

PR #2653, opened 2022-11-20
Skip filtering/deleting zero balances

### LINKED ISSUE #2652: Swapping temporarily clears out erc-20 token balances
### Notes
- Token balances eventually return to normal (probably once the `getTokenBalances` alarm runs
- Probably related to https://github.com/tallycash/extension/pull/2607 - [this block](https://github.com/tallycash/extension/blob/main/background/redux-slices/accounts.ts#L278:L284) is probably a good place to start looking
- An idea here might be to revert some of the changes introduced in 2607 and instead adopt a solution similar to [this untested diff](https://github.com/tallycash/extension/compare/main...foo-11/15).

https://user-images.githubusercontent.com/94649004/202914556-d95a2fc0-ecda-4e77-a8e9-3fa9c64c8fe7.mov

hyphenized:
After transaction enrichment, a token refresh would cause a partial balance update to be sent to Redux, removing token balances not included in the payload.

Closes #2652 

## To test
- [ ] Bridge/Send assets
- [ ] Import an account, balances should not disappear while activities load

Latest build: [extension-builds-2653](https://github.com/tallycash/extension/suites/9409737625/artifacts/444261580) (as of Sun, 20 Nov 2022 19:52:25 GMT).


Shadowfiend:
This also introduced an explicit behavior change that a token that has a balance and no longer does will never disappear, ie if you've ever held a token it will always be listed in your balances. Want to make sure that's cleared with @VladUXUI since the PR description doesn't make that clear--I believe the current behavior of being willing to drop a token once it's balance goes to 0 was in fact intentional (as evidenced by the special casing of base assets).

greg-nagy:
> This also introduces an explicit behavior change that a token that has a balance and no longer does will never disappear, ie if you've ever held a token it will always be listed in your balances. Want to make sure that's cleared with @VladUXUI since the PR description doesn't make that clear--I believe the current behavior of being willing to drop a token once it's balance goes to 0 was in fact intentional (as evidenced by the special casing of base assets).

I think we have introduced this behavior in #2507 because we didn't understand what you meant. 

This PR actually fixes this, because it undoes the filtering from #2507 and moves the behavior to a reducer. The if check `updatedAccountBalance.assetAmount.amount === 0n && existingAccountData.balances[updatedAssetSymbol] === undefined` filters balance updates only when we didn't have a balance for the asset previously. This feels correct and this was the original intention behind #2507 if memory serves.

Tested the change, and if a balance goes to 0 then it disappears.

greg-nagy (approved):
the reaction time is 🤯 appreciate this a lot 🙏

looks good, works well! Let's 🚀

Shadowfiend:
> I think we have introduced this behavior in #2507 

Yes but it's not clear if we checked with Vlad that this was the right target behavior either 😁

VladUXUI:
Thanks for asking!
Asset disappearing if balance if bellow 0 is what we want indeed.
This should follow the "show balance bellow $2 setting/rule"

Shadowfiend:
@VladUXUI  Just to clarify, if the "show balance bellow $2" setting is on, an asset that we have previously held that is now at 0 should or shouldn't be displayed? I believe the change here makes it so that it isn't displayed in that scenario.

PR #2654, opened 2022-11-21
Add auto-update to Swap quotes

hyphenized:
This should improve UX and prevent stale quotes, the quote **should not** refresh automatically if it's too recent or if an input has focus.

## Testing Env

```
SUPPORT_SWAP_QUOTE_REFRESH=true
```

To Test: 
----
- [ ] Go to Swaps page, pick two assets
- [ ] Quote should update every 5~ seconds unless there has been a recent update within the last < 3 seconds

Latest build: [extension-builds-2654](https://github.com/tallyhowallet/extension/suites/10201629952/artifacts/499499316) (as of Thu, 05 Jan 2023 14:35:35 GMT).


jagodarybacka (approved):
The code looks good and it works as designed but I feel like we need to confirm the UX is fine (cc @VladUXUI). 
Also, I'm afraid of how it will work with the doggo loader - probably we would exclude that loader from this update to avoid moving the interface too much. Or maybe we can have a "sitting" version of the dog to leave him in place all the time 😄

hyphenized:
> Or maybe we can have a "sitting" version of the dog to leave him in place all the time 😄

💯 

 That would be awesome! If the quote loads quickly the doggo would switch between states too fast though 🤔

jagodarybacka (changes requested):
Hmm something broke here, if you'll remove "sell" value then "buy" value is not updating correctly

VladUXUI:
Heyyyy, sorry totally missed these pings.
I need to think about this a little bit, not the biggest fan of the button moving around, or the gif always being there.

A quick fix here would be to move the button down, so it doesn't move when gif appears. In the same time i think that the gif appearing every 5 seconds is a bit distracting

VladUXUI:
Another interesting thing i saw in that preview is that button says "approve assets" gif appears and the button changes to "get final quote"

VladUXUI:
how often do we get a new quote from 0x?

hyphenized:
> Hmm something broke here, if you'll remove "sell" value then "buy" value is not updating correctly

Hmmm I changed it so that the auto refresh works if either amount is available but I wonder if we should clear both inputs if one is empty (or zero) on blur 🤔 

> A quick fix here would be to move the button down, so it doesn't move when gif appears. In the same time i think that the gif appearing every 5 seconds is a bit distracting

I agree, there's also the possibility of the quote loading too fast for the animation to start. I've set a fixed height on the loading container so the footer button doesn't move around before/after loading. Looks like this now: 

image


> Another interesting thing i saw in that preview is that button says "approve assets" gif appears and the button changes to "get final quote"

This is a bug, fixed in the other PR

>i also noticed that the quote doesn't actually change, of if it does it's too little to see on screen

It does change but I guess there's not a lot of volatility for most assets, if you try to swap 3 ETH for MATIC then you should see the quote changing more often.

> how often do we get a new quote from 0x?

Quote should update every 5 seconds, I've provisionally changed it to 10 secs

jagodarybacka:
Screenshot 2022-12-16 at 12 44 52Screenshot 2022-12-16 at 12 48 58

IMO it would look nicer if loader was below the button - so both can stay in place regardless if it's loading or not

ui/hooks/react-hooks.ts
```diff
@@ -0,0 +1,127 @@
...
+ */
+export function useValueRef<T>(value: T): React.MutableRefObject<T> {
+  const valueRef = useRef<T>(value)
+  valueRef.current = value
+  return valueRef
+}
+
+/**
+ * Runs a callback on mount, if the callback returns a function,
+ * it will be called on unmount
```
Shadowfiend, line 49: What is the difference between this and `useEffect`? Explaining it in the docs is a good idea so people will know when/why they might want to use it--or why it's being used in a certain place.

ui/hooks/react-hooks.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  useEffect(() => {
+    valueRef.current = value
+  }, [value])
+
+  return valueRef.current
+}
+
+/**
+ * Useful for "batching" state changes, Similar API to the old Component.setState
```
Shadowfiend, line 82: It feels like the key distinguishing factor here is that the setter can be called with a function that will return the desired stare? Again, understanding why you would use this over other options (like a simple useState with an object and maybe some destructing) feels like it might be useful.

Shadowfiend (commented):
Had a minute to look so left a few small questions and comments, completely non blocking.

ui/hooks/react-hooks.ts
hyphenized: Functionality wise it's the same, otherwise it's a bit more declarative and removes the need to `//eslint-disable`

ui/hooks/react-hooks.ts
hyphenized: This is useful for managing "related" state after/inside an async callback. Since the React version we're using does not batch state updates inside timeouts/microtasks, if we use separate `useState` hooks for e.g. `loading` and `result`, each update will trigger a new render and commit it's results, causing strange behaviour with logic or effects that have both `loading` and `result` as dependencies. 

Using a simple `useState` with an object would work too, but we'd have to be more explicit with each state update (that might be better as it would make state changes easier to track within the code 👀 )

jagodarybacka (approved):
Confirmed this is ok design-wise, lets ship 🚢

PR #2656, opened 2022-11-21
Prefer alchemy provider for signing

hyphenized:
- Fallback provider `getSigner` was returning a public RPC signer, causing some transactions to fail.
- Added another Arbitrum public RPC, since the previous one had somewhat strict rate limits.

## To Test: 
- [ ] Switch to arbitrum, try to swap, you should not see "Transaction failed to broadcast"

Latest build: [extension-builds-2656](https://github.com/tallycash/extension/suites/9425282592/artifacts/445345547) (as of Mon, 21 Nov 2022 16:03:04 GMT).

### DIFF
--- a/background/constants/networks.ts
+++ b/background/constants/networks.ts
@@ -145,7 +145,10 @@ export const CHAIN_ID_TO_RPC_URLS: {
     "https://optimism-mainnet.public.blastapi.io",
   ],
   [ETHEREUM.chainID]: ["https://rpc.ankr.com/eth"],
-  [ARBITRUM_ONE.chainID]: ["https://rpc.ankr.com/arbitrum"],
+  [ARBITRUM_ONE.chainID]: [
+    "https://arb1.arbitrum.io/rpc",
+    "https://rpc.ankr.com/arbitrum",
+  ],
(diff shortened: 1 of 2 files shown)


0xDaedalus:
I think maybe a slightly safer way forward here is to comment out the ankr provider rather than adding a new one and overriding `getSigner`.  This will basically revert arbitrum to the functionality it had prior to this change and give us a bit more time to test the new provider we are adding as well as the `override` change.

hyphenized:
@0xDaedalus Good call! Disabled it here: #2657

greg-nagy:
@0xDaedalus Should this be closed, or do you want to iterate on it?

0xDaedalus:
Lets not close it because in the current state of `main` we do not have a generic rpc provider for arbitrum - which is undesirable.

greg-nagy:
Instead of this direction should we extend `SerialFallbackProvider` [similar to `alchemy_` check](https://github.com/tallycash/extension/blob/059c23187b66e7e0ca179a6b62c6572d351bdb5a/background/services/chain/serial-fallback-provider.ts#L247)

I mean we know precisely the method calls and for the modifying calls preferring alchemy seems reasonable.

mhluongo:
The approach in #2662 is a clean alternative :ok_hand:

PR #2657, opened 2022-11-21
Disable Arbitrum generic RPC

hyphenized:
This RPC is causing swaps to fail, disabling it as a quick fix to use alchemy while we find and test a different provider

Refs: #2656

## To Test: 
- [ ] Switch to Arbitrum, try to swap, you should not see "Transaction failed to broadcast"

Latest build: [extension-builds-2657](https://github.com/tallycash/extension/suites/9426857886/artifacts/445460044) (as of Mon, 21 Nov 2022 17:18:19 GMT).


0xDaedalus (approved):
📈

PR #2662, opened 2022-11-22
🗺 Show Me the Way: Prefer alchemy provider for sensitive/important rpc method calls

greg-nagy:
## Why 

For important rpc calls we should prefer the reliable provider we have. 

## Test 
- import seed
- check balances, activities
- change wallet to optimism 
- open uniswap + connect
- swap on optimism - should not see `Inssufficient liquidity` filckering
- swap on every network on uniswap
- go to 1inch
- swap on every network 
- test internal swaps on every network 
- test send on every network

Latest build: [extension-builds-2662](https://github.com/tallycash/extension/suites/9545265656/artifacts/453904954) (as of Mon, 28 Nov 2022 17:41:31 GMT).


background/services/chain/serial-fallback-provider.ts
```diff
@@ -178,6 +178,14 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
     once: boolean
   }[] = []
 
+  private rpcMethodsForAlchemy = [
```
mhluongo, line 181: It would be nice if these were regex, or if it was clear that they are RPC method name prefixes (eg "alchemy_")

background/services/chain/serial-fallback-provider.ts
```diff
@@ -178,6 +178,14 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
+    "alchemy_", // alchemy specific api calls start with this
+    "eth_sendRawTransaction", // broadcast should always go to alchemy
+    "eth_call", // this is causing issues on optimism with ankr and is used heavily by uniswap
```
mhluongo, line 184: Could we only make the switch here per-network? `eth_call` is really common everywhere

mhluongo: That would require a restructure from this array into a small `alchemyOrDefaultProvider(chainID, method)` function, but I suspect that will pay dividends over the next few releases

mhluongo (commented):
Quick thoughts! Love that we're getting more sophisticated here.

background/constants/networks.ts
```diff
@@ -129,10 +129,32 @@ export const CHAIN_ID_TO_RPC_URLS: {
...
+    "alchemy_", // alchemy specific api calls start with this
+    "eth_sendRawTransaction", // broadcast should always go to alchemy
+    "eth_subscribe", // generic http providers do not support this, but dapps need this
+    "eth_estimateGas", // just want to be safe, when setting up a transaction
+  ],
+  [OPTIMISM.chainID]: [
+    "eth_call", // this is causing issues on optimism with ankr and is used heavily by uniswap
+  ],
+  [ARBITRUM_ONE.chainID]: [
+    "eth_call", // this is causing issues on optimism with ankr and is used heavily by uniswap
```
jagodarybacka, line 158:
```suggestion
    "eth_call", // this is causing issues on arbitrum with ankr and is used heavily by uniswap
```

background/constants/networks.ts
```diff
@@ -129,10 +129,32 @@ export const CHAIN_ID_TO_RPC_URLS: {
...
+ * This allows us to use "namespaces" for providers eg `alchemy_...` or `qn_...`
+ *
+ * The structure is network specific with an extra `everyChain` option.
+ * The methods in this array will be directed towards alchemy on every network.
+ */
+export const RPC_METHOD_PROVIDER_ROUTING = {
+  everyChain: [
+    "alchemy_", // alchemy specific api calls start with this
+    "eth_sendRawTransaction", // broadcast should always go to alchemy
+    "eth_subscribe", // generic http providers do not support this, but dapps need this
```
0xDaedalus, line 151: I believe this will default us to subscribing to pending transactions via alchemy which will revert a significant amount of the cost savings we got from switching to generic providers

background/constants/networks.ts
```diff
@@ -129,10 +129,32 @@ export const CHAIN_ID_TO_RPC_URLS: {
...
+ * prefer alchemy provider over the generic ones.
+ *
+ * The method names can be full or the starting parts of the method name.
+ * This allows us to use "namespaces" for providers eg `alchemy_...` or `qn_...`
+ *
+ * The structure is network specific with an extra `everyChain` option.
+ * The methods in this array will be directed towards alchemy on every network.
+ */
+export const RPC_METHOD_PROVIDER_ROUTING = {
+  everyChain: [
```
0xDaedalus, line 148: I'd suggest using `Set`s here so we avoid iterating through an array each time we send a message

jagodarybacka (dismissed):
Did some swaps and all is fine, idk if we want to basically do the whole release checklist on this PR alone (swapping on each network, in wallet and on dapps)

background/constants/networks.ts
greg-nagy: Yep. But how would we have subscriptions on http providers? 

My understanding is that subscriptions work only on websocket providers and our public endpoints are http only. 
What am I not missing here?

background/constants/networks.ts
hyphenized: Agree but I think we still have to iterate if we want to do partial checks like `startsWith`

hyphenized (commented):
Works well, judging "how expensive" a method is by the compute units used and the amount of calls made, we shouldn't have issues with other public RPCs methods.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -109,6 +110,27 @@ function isConnectingWebSocketProvider(provider: JsonRpcProvider): boolean {
...
 
+/**
+ * Return the decision whether a given RPC call should be routed to the alchemy provider
+ * or the generic provider.
+ *
+ * Checking whether is alchemy supported is a non concern for this function!
+ *
+ * @param chainID string chainID to handle chain specific routings
+ * @param method the current RPC method
+ * @returns true | false whether the method on a given network should routed to alchemy or can be sent over the generic provider
```
mhluongo, line 121: For booleans I try to word `@returns` as...

"true if the method should be routed to an Alchemy provider, false otherwise"

to make it really clear which state means which.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -109,6 +110,27 @@ function isConnectingWebSocketProvider(provider: JsonRpcProvider): boolean {
...
+ */
+function alchemyOrDefaultProvider(chainID: string, method: string): boolean {
+  return (
+    RPC_METHOD_PROVIDER_ROUTING.everyChain.some((m: string) =>
```
mhluongo, line 125: Very nice

mhluongo (commented):
Couple quick thoughts, love the changes. It'd be good to see you and @0xDaedalus aligned on the performance & product impact of subscribes before we merge though

background/constants/networks.ts
0xDaedalus: We have 3 alchemy-specific methods that we use and I (hope) that the list won't grow much beyond that so I think its prefectly reasonable to not use `startsWith` and just enumerate the methods we want to allow.  Should be straightforward to make it typesafe as well to prevent future contributors from adding an alchemy call somewhere without adding it to this list.  That being said I don't want to block on this nor pre-optimize - onwards!

background/constants/networks.ts
```diff
@@ -129,10 +129,32 @@ export const CHAIN_ID_TO_RPC_URLS: {
...
+    "alchemy_", // alchemy specific api calls start with this
+    "eth_sendRawTransaction", // broadcast should always go to alchemy
```
0xDaedalus, line 150: Privacy & Censorship implications here but we can cross that bridge when we get there

0xDaedalus (approved):
📈 🚀 🚢

PR #2663, opened 2022-11-22
Fidget Spinners: Simplify DOM and eliminate strange render loops in Popup

Shadowfiend:
The goal here is to simplify scroll handling, clarify ownership of styles,
remove render warnings and glitches, and fix types in `Popup`.

- `CorePage` is now used by page components, not forcibly wrapped in the
  router.
- `CorePage` owns, through props, rendering or not rendering the top and
  tab bars.
- Transition handling is drastically simplified, involving no `useState`
  and simply relying on CSS classes to determine transition direction.
- Proper typing is used in `Popup.tsx`; `ts-expect-error` usage is gone.
- Fixed positioning is largely removed for anything other than popovers
  and slideovers.
- Less z-indexing.
- Wallet asset list element structure more clearly reflects the quick
  links to swap and send as distinct, non-hierarchically-descendant of
  the asset item link.

Still need to check the onboarding flow; also the functionality that resets the tab selection in wallet activity
when the NFT availability changes is currently not working, but no such
networks exist by default. Some more work to be done there.

Latest build: [extension-builds-2663](https://github.com/tallycash/extension/suites/9455038633/artifacts/447435047) (as of Tue, 22 Nov 2022 22:06:13 GMT).

### DIFF
--- a/background/redux-slices/ui.ts
+++ b/background/redux-slices/ui.ts
@@ -18,6 +18,12 @@ const defaultSettings = {
   hideBanners: false,
 }
 
+export interface Location {
+  pathname: string
+  key?: string
+  hash: string
+}
+
 export type UIState = {
   selectedAccount: AddressOnNetwork
   showingActivityDetailID: string | null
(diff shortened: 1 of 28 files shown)


Shadowfiend:
The drift here has been very serious, and current cycles probably don't suffice to bring this up to par. We'll keep the idea alive, but for now going to close the PR out.

PR #2664, opened 2022-11-23
Refactor tabs data

hyphenized:
- Internationalized tab button labels
- Refactored tab data from simple strings to objects, since a string won't always match the route name, the title and the icon for the tab

## To Test
- [ ] Navigation between tabs should work exactly as before

Latest build: [extension-builds-2664](https://github.com/tallycash/extension/suites/9472810629/artifacts/448699199) (as of Wed, 23 Nov 2022 17:44:12 GMT).


ui/utils/tabs.ts
```diff
@@ -1,16 +1,43 @@
...
+export type TabInfo = {
+  title: string
+  path: string
+  icon: string
+}
+
+const tabs: TabInfo[] = [
+  {
+    path: "/wallet",
+    title: t("tabs.wallet"),
```
Shadowfiend, line 13: Just a gut thing, but it feels like this may be best kept as the key with the actual value resolved by calling `t` in the component. Dropping a potentially dynamic internationalized value into a global constant feels like it could have unexpected consequences.

ui/components/TabBar/TabBarIcon.tsx
```diff
@@ -1,30 +1,22 @@
...
+  icon: string
   isActive: boolean
+  onClick?: React.MouseEventHandler<HTMLButtonElement>
 }
 
-export default function TabBarIcon(props: Props): ReactElement {
-  const history = useHistory()
-
-  const { name, isActive } = props
+export default function TabBarIcon(props: TabBarIconProps): ReactElement {
```
Shadowfiend, line 11: Feels like this isn't really an icon, right? The component is maybe best named TabBarButton or so? That would make it less weird that the tab bar icon is being passed an icon prop 😁

Shadowfiend (commented):
Couple of small notes since I've been working adjacent to this recently.

ui/utils/tabs.ts
hyphenized: Solid observation! There are quite a few other places where we applied this pattern, I'll change this in this PR and create an issue for the other cases

ui/components/TabBar/TabBarIcon.tsx
hyphenized: Good catch, the naming does feel off. I'll change it to `TabBarIconButton` for the time being until there's a better suggestion.

ui/_locales/en/messages.json
```diff
@@ -533,7 +533,7 @@
       "bannerTitle": "New Odyssey week!",
       "emptyBannerContent": "Check out Arbitrum Odyssey campaign",
       "notificationTitle": "Odyssey notifications",
-      "notificationDismissInfo":"Want to receive future Odyssey notifications? We will show you one every week that you can dismiss.",
+      "notificationDismissInfo": "Want to receive future Odyssey notifications? We will show you one every week that you can dismiss.",
```
greg-nagy, line 536: Should we remove these files from `.prettierignore` ? 🤔

ui/_locales/i18n.ts
```diff
@@ -34,6 +34,20 @@ declare module "react-i18next" {
   }
 }
 
+/**
+ * Returns all keys that hold string values
+ */
+type StringValueKeys<Dict> = {
```
greg-nagy, line 40: 🤩

ui/components/TabBar/TabBar.tsx
```diff
@@ -1,37 +1,49 @@
...
 import { NETWORKS_SUPPORTING_SWAPS } from "@tallyho/tally-background/constants/networks"
-import TabBarIcon from "./TabBarIcon"
-import tabs from "../../utils/tabs"
+import { EVMNetwork } from "@tallyho/tally-background/networks"
+import { useTranslation } from "react-i18next"
+import TabBarIconButton from "./TabBarIconButton"
+import tabs, { defaultTab, TabInfo } from "../../utils/tabs"
 import { useBackgroundSelector } from "../../hooks"
 
+const isTabSupportedByNetwork = (tab: TabInfo, network: EVMNetwork) => {
```
greg-nagy, line 12: Love this direction and refactor ❤️

greg-nagy (approved):
looks good, works well 

I like a lot how you think about encapsulating the concerns


Let's 🚢

PR #2668, opened 2022-11-23
🤖⏱ Quality Time: update UI open event to track open length not date

### LINKED ISSUE #2610: Extension open time interval tracking

greg-nagy:
> Quality time is an expression referring to how an individual [proactively](https://en.wikipedia.org/wiki/Proactively) interacts with another while they are together, regardless of the duration.
https://en.wikipedia.org/wiki/Quality_time

## Why 

Because quality time is an important love language for any relationship. 
This is true for our users with our extension. 

It's more important to know how long a wallet modal was open than when it was opened.

## Test 

- switch the env variables 
- start, install, import
- open UI > close UI
- check posthog that the `UI shown` event appears, and the payload feels correct

## Testing Env

```
SUPPORT_ANALYTICS=true
ENABLE_ANALYTICS_DEFAULT_ON=true
USE_ANALYTICS_SOURCE="PR"
```

Closes #2610 

Latest build: [extension-builds-2668](https://github.com/tallycash/extension/suites/9473827809/artifacts/448772946) (as of Wed, 23 Nov 2022 18:44:45 GMT).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1502,9 +1502,16 @@ export default class Main extends BaseService<never> {
   private connectPopupMonitor() {
     runtime.onConnect.addListener((port) => {
       if (port.name !== popupMonitorPortName) return
-      this.analyticsService.sendAnalyticsEvent("UI open")
+
+      const openTime = Date.now()
 
       port.onDisconnect.addListener(() => {
+        this.analyticsService.sendAnalyticsEvent("UI shown", {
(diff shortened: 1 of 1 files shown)


hyphenized (approved):
Looks great! These events will also be sent when the popup opens for signing transactions/connects to dapps/etc right? Wonder if it'd be worth to filter those :shipit:

PR #2671, opened 2022-11-24
Tabs for NFTs page

### LINKED ISSUE #2597: Add NFTs and Badges tabs
image

- [ ] add NFTs and Badges tabs and let them display different components

kkosiorowska:
Closes #2597 

This PR adds NFTs and Badges tabs.

## UI
Screenshot 2022-11-25 at 16 08 03

Screenshot 2022-11-25 at 16 07 47

Screenshot 2022-11-25 at 15 35 12

## Testing Env

```
SUPPORT_NFT_TAB=true
```

## To Test
- [ ] go to the NFTs page and check the tabs
- [ ] set up `hasNFTs` to `false` and check tabs again

Latest build: [extension-builds-2671](https://github.com/tallycash/extension/suites/9507475768/artifacts/451163203) (as of Fri, 25 Nov 2022 15:15:30 GMT).


ui/pages/NFTs.tsx
```diff
@@ -1,24 +1,62 @@
...
+
+const PANEL_NAMES = ["NFTs", "Badges"]
 
 export default function NFTs(): ReactElement {
   const { t } = useTranslation("translation", {
     keyPrefix: "nfts",
   })
+  const [panelNumber, setPanelNumber] = useState(0)
+
+  const renderTabContent = useCallback(
```
jagodarybacka, line 16: Do we get any advantage from using callback and switch here? It's not the way we are using panels anywhere else in the codebase.

Also if we will want to pass props to the components in the tabs it will be more complicated to use.

kkosiorowska: I don't like the components becoming large, because it can be more difficult to understand the logic. I was thinking that wrapping render tab logic in useCallback can take more benefits when the component will grow up. Tabs management will be in one place which will make scrolling through the component easier. But of course, I understand your point and I can change it.

jagodarybacka: We can leave it like this and try it out, easy to change later :)

greg-nagy (dismissed):
🚀

The changes are looking good. Please resolve the conflict and this will be good to merge

jagodarybacka (approved):
✅

Issue #2661, opened 2022-11-22
Update Portfolio page with NFTs stats


jagodarybacka:
- [ ] remove tabs (Assets, NFTs, Achievements)
- [ ] add NFT stats section 
<img width="498" alt="image" src="https://user-images.githubusercontent.com/20949277/203311213-9f20bfe8-1792-416a-a40f-c509b305f606.png">

- [ ] update headers of Accounts, Networks and Assets to match the design
- [ ] add Assets count to the header
<img width="477" alt="image" src="https://user-images.githubusercontent.com/20949277/203311357-1f0a0a49-8d3e-41e9-b633-cadd0367e4c8.png">

jagodarybacka:
Resolved by https://github.com/tallycash/extension/pull/2703

PR #2674, opened 2022-11-25
Overview NFT updates

hyphenized:
Refs: #2661

## Testing Env

```
SUPPORT_NFT_TAB=true
```

## To Test
- [ ] Import an account
- [ ] Check portfolio overview

Latest build: [extension-builds-2674](https://github.com/tallycash/extension/suites/9513598769/artifacts/451592314) (as of Sat, 26 Nov 2022 02:32:26 GMT).


ui/_locales/en/messages.json
```diff
@@ -280,6 +280,15 @@
     "accounts": "Accounts",
     "assets": "Assets",
     "networks": "Networks",
+    "nfts": "<0>{{nfts}}</0> $t(overview.units.nft, {'count': {{nfts}} }) in <2>{{collections}}</2> $t(overview.units.collection, {'count': {{collections}} }) & <4>{{badges}}</4> $t(overview.units.badge, {'count': {{badges}} })",
```
jagodarybacka, line 283: Hmm, I'm not sure about it. Looks overcomplicated compared to what we want to get here as we just want to translate `NFTs in`, `Collections` and `Badges`. I'm afraid how it will look like in weblate UI for translators.

hyphenized: Should be more readable now 🙂

kkosiorowska (approved):
It's looking good! Let's wait for @jagodarybacka  feedback about translations.

jagodarybacka (approved):
I would prefer even simpler version but let's leave it like that 😛 Looks good!

PR #2680, opened 2022-11-28
Lower Network Polling Timeout to 2 Minutes

0xDaedalus:
This PR Lowers the duration for which we poll RPC providers after a user has conducted some activity on a network.  Concretely what this means is if a user has the wallet popup open for more than 2 minutes without interacting with it the information presented to them may be stale.

The gain we get from this change is a lower probability of useless RPC calls (with the previous timeout of 5 minutes if a user opened the wallet for 1 minute and then closed it - we would still unnecessarily poll for data for 4 minutes).

This is admittedly a band-aid fix (the right path here might be to track when the extension closes and stop polling for data when it does) but is a good-enough intermediate step until we have bandwidth for a more robust solution.

### Acceptance Criteria
- [ ] Normal wallet functionality & dapp interactions should still work as they did before.  
- [ ] Wallet balances & asset transfers for all active networks should correctly load on application startup.

Latest build: [extension-builds-2680](https://github.com/tallycash/extension/suites/9592630938/artifacts/457293963) (as of Wed, 30 Nov 2022 18:22:39 GMT).


background/services/chain/index.ts
```diff
@@ -84,7 +84,8 @@ const BLOCKS_FOR_TRANSACTION_HISTORY = 128000
 // OpenEthereum with tracing to catch up to where we are.
 const BLOCKS_TO_SKIP_FOR_TRANSACTION_HISTORY = 20
 
-const NETWORK_POLLING_TIMEOUT = MINUTE * 5
+// Add a little bit of wiggle room
+const NETWORK_POLLING_TIMEOUT = MINUTE * 2.05
```
greg-nagy, line 88: Why `.05` ? :)

ui/pages/Popup.tsx
```diff
@@ -102,10 +102,15 @@ export function Main(): ReactElement {
   const currentAccount = useBackgroundSelector(selectCurrentAddressNetwork)
   // Emit an event when the popup page is first loaded.
   useEffect(() => {
+    /**
+     * Marking user activity every time this component is rerendered
```
greg-nagy, line 106: Just a note, that we have a lot of rerenders coming from empty updates. 

Note affecting the functionality here.

greg-nagy (approved):
🚀

background/services/chain/index.ts
0xDaedalus: To account for possible lag in retrieving the message.

PR #2682, opened 2022-11-29
Add Initial Arbitrum Nova Support

0xDaedalus:
Full support is blocked by https://github.com/tallycash/token-list/pull/72 but we can merge this and add the tokenlist in a later PR.

### Acceptance Criteria
- [x] Sending ETH to wallets should work
- [x] In-wallet swaps should not be possible
- [x] The NFTs tab should not be in the `Wallet` page.
- [x] Transactions should work (you can use www.arbswap.io - although tokens will not show up in the wallet.  Activities should properly go from pending to resolved once a tx is accepted.

There is also a [follow-up issue](https://github.com/tallycash/extension/issues/2681) of using the correct icon for the network.

Latest build: [extension-builds-2682](https://github.com/tallycash/extension/suites/9554515838/artifacts/454551076) (as of Tue, 29 Nov 2022 04:44:07 GMT).


greg-nagy (approved):
this is looking good

Some things I have found 
- CleanShot 2022-11-30 at 17 26 43 - the network selector looks a bit wonky 
- multicall is not deployed on arbi nova
- for gas change I think we should use the optimism modal combined with the normal gas estimation method

PR #2699, opened 2022-11-30
Add NFTs list to the assets selector

### LINKED ISSUE #2644: Add NFTs list to the assets selector
- [ ] Add NFTs tab to the Assets selector
- [ ] Show NFTs tab in the Assets selector while selecting assets on Send page
- [ ] Create NFTs collections list
- [ ] Allow expanding collections
- [ ] Display NFTs owned by current account in a given collection
- [ ] Add links to collections contracts
- [ ] Allow selecting NFTs

image

hyphenized:
https://user-images.githubusercontent.com/28708889/204843967-eee6c8fc-2a05-4d61-b884-063810307b24.mov

Closes #2644

Latest build: [extension-builds-2699](https://github.com/tallycash/extension/suites/9692258525/artifacts/464329544) (as of Tue, 06 Dec 2022 09:32:58 GMT).


background/nfts.ts
```diff
@@ -44,6 +44,7 @@ export type NFT = {
   name: string
   description: string
   thumbnail: string
+  previewURL?: string
```
jagodarybacka, line 47: I think we can make `thumbnail` and `previewURL` follow the same rules so default to either `undefined` or empty string.

And make `thumbnail` -> `thumbnailURL`

ui/components/Send/NFTCollectionAccordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+              const url = `${
+                scanWebsite[collection.network.chainID].url
+              }/token/${contract}`
+
+              window.open(url, "_blank")?.focus()
+            }
+          }}
+          customStyles={`
+                        margin: 12px 0 auto;
+                      `}
```
jagodarybacka, line 113:
```suggestion
          customStyles="margin: 12px 0 auto;"
```

ui/components/Send/NFTCollectionAccordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+          }
+
+          .nft_collection {
+            display: flex;
+            align-items: center;
+            gap: 8px;
+            flex-grow: 1;
+          }
+
+          .nft_collection :global(img) {
```
jagodarybacka, line 136: Can we add `styles`/`customStyles` prop to the component instead of these global styles?

ui/components/Send/NFTCollectionListItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+        fallback={noPreviewImg}
+        loading="lazy"
+        alt={name}
+      />
+      <button type="button" className="label" onClick={onClick}>
+        <p className="ellipsis">{name}</p>
+        <SharedIcon height={16} width={16} icon="icons/s/continue.svg" />
+      </button>
+      <style jsx>{`
+        div :global(img) {
```
jagodarybacka, line 31: Same here - I feel safer with passing styles than using styled-jsx globals as we had some weird buggy styles conflicts in the past

ui/pages/Send.tsx
```diff
@@ -196,10 +209,20 @@ export default function Send(): ReactElement {
...
-              selectedAsset={selectedAsset}
+              onSelectNFT={(nft) => {
+                setSelectedNFT(nft)
+                setAssetType("nft")
+              }}
+              selectedAsset={selectedAsset ?? undefined}
+              selectedNFT={(assetType === "nft" && selectedNFT) || undefined}
               amount={amount}
+              NFTCollections={
+                isEnabled(FeatureFlags.SUPPORT_NFT_TAB)
```
jagodarybacka, line 220: It should be `SUPPORT_NFT_SEND` but I will change it when my PR with this flag will be merged :)

background/lib/simple-hash_update.ts
```diff
@@ -140,32 +140,39 @@ function simpleHashNFTModelToNFT(
...
-    previews?.image_large_url ||
-    previews?.image_medium_url ||
     previews?.image_small_url ||
+    previews?.image_medium_url ||
+    previews?.image_large_url ||
+    fullsizeURL ||
     ""
+
+  const previewURL =
+    (previews?.image_medium_url || previews?.image_large_url || fullsizeURL) ??
```
jagodarybacka, line 150: I think in the worst case we can fallback to small instead of undefined

ui/components/Send/NFTCollectionAccordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+      />
+      {collection.owner && (
+        <SharedIcon
+          icon="icons/s/new-tab.svg"
+          width={16}
+          color="var(--green-40)"
+          hoverColor="var(--trophy-gold)"
+          onClick={() => {
+            if (!isLoading) {
+              const { contract } = collection.nfts[0] ?? {}
```
jagodarybacka, line 102: If SH doesn't include contract on collection level then it maybe means we shouldn't show it here. Is contract always the same for every nft in collection? 

For now can we visually disable this icon if it is not clickable?

jagodarybacka (changes requested):
This PR is so awesome 💖 A lot of improvements that I will reuse 🙇 

I don't want you to spend a lot more time on this right now as Send is not the first prio so let's make some quick improvements

- [x] please fix padding here 
image

- [x] max button doesn't make sense for nfts 
image

ui/components/Send/NFTCollectionAccordion.tsx
hyphenized: > Is contract always the same for every nft in collection
Yes, a contract may have multiple collections too. It might be different in other networks 🤔

background/lib/simple-hash_update.ts
hyphenized: I agree but if we don't receive the `fullsizeURL` then I think we won't have a small size preview either

ui/components/Send/NFTCollectionListItem.tsx
hyphenized: Yeah that sounds like a much better idea, these global styles could also be easily missed

background/nfts.ts
hyphenized: Good call, I've made both `undefined` and refactored `thumbnail` in `NFTCollection` to be consistent with this.

jagodarybacka (changes requested):
Please test it with `SUPPORT_NFT_TAB=false` as this is the default for now - if you will reinstall the extension with this flag off then the `Send` page is throwing an error because it tries reading from the new NFTs Redux slice which is not yet created

ui/components/Send/NFTCollectionAccordion.tsx
jagodarybacka: Will investigate it later, let's leave it like that for now

hyphenized:
> Please test it with `SUPPORT_NFT_TAB=false` as this is the default for now - if you will reinstall the extension with this flag off then the `Send` page is throwing an error because it tries reading from the new NFTs Redux slice which is not yet created

Thanks for catching this! Feature flagged the selector code to retrieve NFTs within the Send page

jagodarybacka (approved):
All good now, let's 🚢 
(let's not bother @VladUXUI for now as this is not finished, we can do one big feature's design review later)

PR #2700, opened 2022-11-30
Handle special routing for alchemy only if there's an alchemy provider

hyphenized:
We previously applied this to every chain even if it wasn't supported by alchemy

## To Test:
- [ ] Switch to Avalanche
- [ ] Try sending assets

Latest build: [extension-builds-2700](https://github.com/tallycash/extension/suites/9592667866/artifacts/457296589) (as of Wed, 30 Nov 2022 18:24:29 GMT).


greg-nagy (approved):
❤️

PR #2711, opened 2022-12-02
Fix price point decimals accuracy

hyphenized:
Previously if we had two assets with the same symbol and different decimals count on different networks, we would calculate balances incorrectly

## To Test
- [ ] Go to etherscan and bsc, balances should match estimates shown in the wallet main and overview pages

Latest build: [extension-builds-2711](https://github.com/tallycash/extension/suites/9735237297/artifacts/467119752) (as of Wed, 07 Dec 2022 23:51:53 GMT).


jagodarybacka (commented):
It seems to be working correctly but it would be nice to check it on BNB - could you fix the conflicts so we will have BNB available on this branch 😅

background/redux-slices/assets.ts
```diff
@@ -198,44 +193,54 @@ export const transferAsset = createBackgroundAsyncThunk(
...
 
     if (pricedAsset) {
-      const pricePoint = pricedAsset.recentPrices[pairedAssetSymbol]
-      const { pair, amounts, time } = pricePoint
-
-      if (pair[0].symbol === assetSymbol) {
-        return pricePoint
+      let pricePoint = pricedAsset.recentPrices[pairedAssetSymbol]
+
+      // Flip it if the price point looks like USD-ETH
```
0xDaedalus, line 211: Can we add some context here as to why we flip?

background/redux-slices/assets.ts
```diff
@@ -198,44 +193,54 @@ export const transferAsset = createBackgroundAsyncThunk(
...
-    const pricedAsset = (
-      chainID
-        ? assets.filter(
-            (asset) =>
-              "homeNetwork" in asset && asset.homeNetwork.chainID === chainID
-          )
-        : assets
-    ).find(
+  [selectAssetsState, selectAsset, selectPairedAssetSymbol],
+  (assets, assetToFind, pairedAssetSymbol) => {
```
0xDaedalus, line 197: Non-blocking but can we follow up this PR with some tests please?

0xDaedalus (commented):
Code looks good - will test manually to confirm

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -82,8 +82,7 @@ const computeCombinedAssetAmountsData = (
   assets: AssetsState,
   mainCurrencySymbol: string,
   currentNetwork: EVMNetwork,
-  hideDust: boolean,
-  useCurrentNetwork = false
```
greg-nagy, line 86: 🎉

0xDaedalus (approved):
🚀

greg-nagy (approved):
let's not block this one now, but please have tests next time :) 

We should get into testing every PR

background/redux-slices/assets.ts
0xDaedalus: @hyphenized would you mind creating a follow-up issue here and assigning it to yourself?

Issue #2569, opened 2022-11-07
Add Native Binance Smart Chain Support to Tally Ho


0xDaedalus:
### Requirements
- [x] Send / Receive Assets
- [x] Swaps
- [ ] Gas Price Preview
- [x] Dapp connection / tx signing
- [x] Ledger Support
- [x] Coming Soon page for Gas Price Adjust
- [x] If #2547 or similar has been merged - preference for a generic rpc provider.
- [x] Assets showing up in Overview
- [x] ENS Name Resolution
- [x] Ensure support for popular dapps  (cc @mr-michael @alsoknownaslj @Naxsun  to provide list in a separate issue).

### Stretch Goals
- [ ] [Gas Price Adjust](https://github.com/tallycash/extension/issues/2571) (cc @VladUXUI do we have an updated design for a simple gas price + gas limit gas adjustment screen?  This is basically pre-EIP 1559 adjustment)
- [x] NFT Support (If we can find a free api)

### Useful References
- https://github.com/tallycash/extension/pull/2287/files
- https://github.com/tallycash/extension/pull/1347/files

### Notes
- Support for NFTs is not a hard requirement.
- I expect we'll want to open multiple PRs to ultimately close out this ticket - trying to tackle this in one PR might result in a large amount of code to review.
- PRs and issues open for avalanche-specific bugs encountered while implementing avalanche support should be added to the [`Binance Smart Chain Support`](https://github.com/tallycash/extension/milestone/15) milestone.

kkosiorowska:
Ledger is working only for the ETH app for now. BSC support for Ledger should be added when this [PR](https://github.com/tallycash/extension/pull/2577) will be merged.

PR #2715, opened 2022-12-05
Add support for popular dapps for BSC

kkosiorowska:
Refs #2569 

This PR adds support for popular dapps for the Binance Smart Chain. 

## UI
Screenshot 2022-12-05 at 13 16 38

## To Test
Check support for the most important dapps:

- [ ] Celer 
- [ ] Stargate 
- [ ] Multichain
- [ ] GALXE
- [ ] PankcakeSwap
- [ ] Venus
- [ ] Alpaca Finance

### DIFF
--- a/window-provider/wallet-connection-handlers.ts
+++ b/window-provider/wallet-connection-handlers.ts
@@ -1,6 +1,9 @@
 const TALLY_ICON_URL =
   "https://tally.cash/icons/icon-144x144.png?v=41306c4d4e6795cdeaecc31bd794f68e"
 
+const TALLY_NAME = "Tally Ho"
+const META_MASK = "MetaMask"
+
 const observeMutations = (handler: (node: Node) => void) => {
   document.addEventListener("DOMContentLoaded", () => {
     const observer = new MutationObserver(function monitorMutations(mutations) {
@@ -54,7 +57,7 @@ const findAndReplaceUniswapInjectedOption = (): void => {
       // Replace the `Injected` text with `Tally Ho`
       iconAndTextDiv.innerHTML = iconAndTextDiv.innerHTML.replace(
         "Injected",
-        "Tally Ho"
+        TALLY_NAME
       )
     }
   }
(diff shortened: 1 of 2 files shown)


window-provider/wallet-connection-handlers.ts
```diff
@@ -232,13 +235,207 @@ function findAndReplaceAboardMetamaskOption(addedNode: Node): void {
...
 }
 
+function findAndReplacePancakeSwapInjectedOption(addedNode: Node): void {
+  if (moreThanOneWalletInstalledAndTallyIsNotDefault()) {
+    return
+  }
+
+  const text = "Injected"
+  if (
+    addedNode.textContent?.includes(text) &&
```
hyphenized, line 245: Unless it's giving you issues, I think we want to use `innerText` since it ignores text from hidden and non user readable elements.

hyphenized (approved):
Works well, couldn't test `app.multchain.cn` 👀 

Non blocking, Venus shows this after clicking on your address
Screen Shot

jagodarybacka (commented):
Are we ok with this state when TH is not default? (and there are no other wallets - so TH is still able to connect correctly) 
image

kkosiorowska:
@hyphenized 
After changes:

Screenshot 2022-12-06 at 15 17 33

kkosiorowska:
@jagodarybacka 
> Are we ok with this state when TH is not default? (and there are no other wallets - so TH is still able to connect correctly) 

After changes:

Screenshot 2022-12-06 at 15 23 59

window-provider/wallet-connection-handlers.ts
```diff
@@ -1,6 +1,10 @@
 const TALLY_ICON_URL =
   "https://tally.cash/icons/icon-144x144.png?v=41306c4d4e6795cdeaecc31bd794f68e"
 
+const TALLY_NAME = "Tally Ho"
+const META_MASK = "MetaMask"
```
jagodarybacka, line 5: I would expect it to be named `METAMASK` as I think they are always using it as a single word :P

jagodarybacka (approved):
https://app.multchain.cn.com/ was fixed 👌 I'll let @hyphenized take a look before merging

window-provider/wallet-connection-handlers.ts
kkosiorowska: Good Point! I will fix it in the next PR.

PR #2724, opened 2022-12-06
Periodically cleanup the balance and hasCode caches on SerialFallbackProvider

0xDaedalus:
While the cacheing of eth_getBalance and eth_hasCode significantly improves performance and reduces rpc usage on initial wallet load - its current implementation leaves room for the possibility for the cache to grow large enough as to cause the user to run out of memory.  A periodic check that both reports the size of the cache (for the purposes of debugging user issues) and reduces it should help mitigate this issue until we get to a point where we no longer need the cache.

### To Test
- [ ] Set CACHE_CLEANUP_INTERVAL to a low number (10 seconds should do it)
- [ ] Use the wallet, use a few dapps like uniswap - observe logs about cache getting cleaned

Latest build: [extension-builds-2724](https://github.com/tallycash/extension/suites/9694561653/artifacts/464484591) (as of Tue, 06 Dec 2022 11:34:34 GMT).


jagodarybacka (approved):
Cache cleanup seems to be working fine but I'm wondering what are these addresses.
image
My console.log was:
image
And the only account I had imported was `0x5ca...`

greg-nagy:
no we don't

greg-nagy (approved):
let's 🎸

greg-nagy:
> Cache cleanup seems to be working fine but I'm wondering what are these addresses....

@jagodarybacka The contract [addresses that are cached](https://github.com/tallycash/extension/blob/6930887503f04947fb5ca67ee5ae33d1fb2857d9/background/services/chain/serial-fallback-provider.ts#L426)

smhmehdi:
Best

PR #2728, opened 2022-12-07
Remove provider-wide pendingTransaction subscriptions.

0xDaedalus:
Our current method of pending transaction subscription does not seem to be providing much benefit while resulting in either:

- a) Large infrastructure costs when subscribing via alchemy (their pricing model seems to charge us for every pending transaction on a given chain - not just the transactions that match the filter we specify) 
- b) Large CPU load when subscribing via a generic provider (we need to poll for every transaction in the mempool using its transaction hash and then serialize it).

The main benefit we get out of subscribing to pending transactions in the mempool is we are able to detect and display transactions that relevant to a given account even if those transactions were not initiated by the Tally Ho wallet (e.g. a user has their account imported to both tally ho and rainbow and make a transaction using rainbow).

It seems reasonable to assume that this scenario is both somewhat rare and already taken care of by our periodic polling with alchemy_getAssetTransfers (although in that case the transactions can take up to 2 minutes to show up).

With the above points in mind and keeping in mind that we subscribe to transactions sent from the wallet via provider.once - we've chosen to remove our subscription to pending transactions until we find a use case that justifies the infra cost and/or cpu overhead generated by the subscription.

An open question here is what - if anything - do we need to extract from the [`handlePendingTransaction`](https://github.com/tallycash/extension/blob/main/background/services/chain/index.ts#L1662:L1662) function to make sure our nonce is fresh when submitting a transaction.  Specifically I'm wondering if recent nonce improvements by @greg-nagy make our nonce management inside of `handlePendingTransaction` no longer relevant.

### To Test
- [ ] Use the wallet - transactions initiated by you (send / swap) should immediately show up in wallet activity after being sent
- [ ] Send / swap from metamask while having tally side-loaded - the transaction should eventually show up in activities (from testing seems like it takes less than 2 minutes) 

Latest build: [extension-builds-2728](https://github.com/tallycash/extension/suites/9712065533/artifacts/465601352) (as of Wed, 07 Dec 2022 02:50:28 GMT).

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -154,11 +154,6 @@ export default class ChainService extends BaseService<Events> {
     evm: {},
   }
 
-  subscribedAccounts: {
-    account: string
-    provider: SerialFallbackProvider
-  }[]
-
   subscribedNetworks: {
     network: EVMNetwork
     provider: SerialFallbackProvider
@@ -296,7 +291,6 @@ export default class ChainService extends BaseService<Events> {
       ),
     }
 
-    this.subscribedAccounts = []
     this.subscribedNetworks = []
     this.transactionsToRetrieve = []
 
@@ -317,10 +311,6 @@ export default class ChainService extends BaseService<Events> {
     Promise.allSettled(
       accounts
         .flatMap((an) => [
-          // subscribe to all account transactions
-          this.subscribeToAccountTransactions(an).catch((e) => {
-            logger.error(e)
-          }),
           // do a base-asset balance check for every account
           this.getLatestBaseAccountBalance(an).catch((e) => {
             logger.error(e)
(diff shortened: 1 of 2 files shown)


jagodarybacka (commented):
I had a fresh install of TH and waited ~5min so some activities are loaded.
- Made a swap in-wallet - showed up instantly in activities. 
- I made a swap on Uniswap with MM on Polygon and it wasn't visible in Tally after 8min. 
- I reloaded the extension and that transaction was loaded after ~1-2min. 

Can someone else test the same thing? Maybe I should have waited full 10min for this tx to be fetched...

jagodarybacka (changes requested):
I've checked it today again and it hasn't added swaps from Uniswap&MetaMask after 20min, on Polygon and Optimism. What's interesting - extension reload did nothing this time and I still don't see these transactions in activities 😞

0xDaedalus:
Moving to draft while I investigate 😭

Shadowfiend:
Going to go ahead and close this out for now—we've made improvements in a few other places in our resource usage, and the underlying functionality is useful. We can always come back to it if it becomes relevant again.

PR #2733, opened 2022-12-08
[Proposal] Ravioli services: Breakout functional blocks into smaller services

greg-nagy:
Check the [Light RFC process first](https://coda.io/d/Engineering-Thesis_duruoW1C74B/RFC-Process_su2dT#_luU0q)

## Background

Our services are getting big. It's getting harder and harder to reason about them and work with them, and it also results in coupling between parts that could/should be encapsulated on their own. One example is the tx queue.

But cutting out parts now is hard. We need a broad understanding in the team about how to approach this problem. 
This proposal implements one approach to the problem to start the discussion.

### Goal

Let's start separating chunks of the services that are meaningful on their own. 
This should make it easier to reason about them, test them and generally work with them.

### Implementation

The current implementation instantiates the tx queue service in the chain service and uses 
the emitter in the queue service to communicate back to the chain service. 

#### Limitations 

Using the emitter for communication can result in complexity, harder to follow code, and performance issues. The best would be to use method calls, but that would result in a dependency cycle without a bigger refactor. 

## Future work

After we agree, I will put a doc in the repo and start systematically capturing proposals and decisions with the outcome.

## Discussion points
Should we do this at all? If yes, is this a good approach? If not, what is? 

⚠️ The current separation is clean, but it's a happy coincidence. The `retrieveTransaction` is only used in this code path. Should that be part of the `TransactionRetrieveQueueService`?  If so, should the queue service use the chain db? Does it worth it to build out multiple events to glue them together?

- Should we share dbs between services or not?
- Where should services be created? In the service, it belongs to or in the main? Where should its code be? In the main service's folder?
- Should we create new services or more functions/classes? How do we decide when to use which? What is the definition of a service? How do we define its scope of it?
- Should we go with services in these cases or use factory functions?
- What is the best way to break the dependency cycle? Service1 <> Service2 method calls. Should we use 
  - the emitter to communicate backward to the original service?
  - should we pass in callbacks? 
  - should we refactor services even more and handle this with a logical grouping of the code? [See this article](https://medium.com/angular-in-depth/how-to-break-a-cyclic-dependency-between-es6-modules-fd8ede198596)

Latest build: [extension-builds-2733](https://github.com/tallycash/extension/suites/9745042596/artifacts/467762836) (as of Thu, 08 Dec 2022 11:04:10 GMT).


background/services/types.ts
```diff
@@ -110,3 +110,20 @@ export type ServiceCreatorFunction<
...
   Deps extends Promise<Service<ServiceLifecycleEvents>>[]
 > = (...serviceDependencies: [...Deps, DexieOptions?]) => Promise<ServiceType>
+
+/**
+ * This type is the same as the ServiceCreatorFunction except it also expects
+ * a database to be passed into the service creator function.
+ *
+ * The goal here is to unlock partitioning certain services into smaller
+ * services meanwhile they can share a common database. The concept is similar
+ * to pods in kubernetes.
```
mhluongo, line 120: Ehh, the concept is similar to all sorts of other things. As a k8s person this analogy is confusing

mhluongo (commented):
Quick thoughts:

I agree that services are getting too big, but it's not clear to me from this example why we should prefer children versus sibling services, or even just splitting services into more modules.

Services are currently made up a in-memory data, a database, timers, and an emitter. All code in a service shares a lifecycle, and can block other code in the same service based on that lifecycle.

Do we want the tx retrieval queue's startup to potentially block the rest of the chain service? And if not, wouldn't it make more sense as a sibling?

mhluongo:
For background: The original services idea was partially inspired by [`component`](https://github.com/stuartsierra/component), a declarative Clojure DI library.

It's all about smaller components to isolate side-effects, eg using a "service" to wrap all access to a particular API, and another for a database, etc. This push for smaller and better-scoped services is worthwhile IMO, just unsure whether this is the way to cut it.

mhluongo:
> What is the best way to break the dependency cycle? Service1 <> Service2 method calls.
> Should we use
the emitter to communicate backward to the original service?

Yes. The idea here was that services can start having one-to-many relationships with data easily with events. If we use method calls, then the calling service has to update multiple other services with the results of a call... rather than those services describing what they want to pay attention to via event matching.

> should we pass in callbacks?

Please, no, anything but that 😅

greg-nagy:
Thanks for the thoughts, I hope for more to come 🤞

> but it's not clear to me from this example why we should prefer children versus sibling services, or even just splitting services into more modules.

I don't think we should. We should have a team-wide understanding of how to make these decisions. I am hoping to have a discussion that will lead to this. 

My 2c: 
- no alarm, no db dependency -> use a module 
- need to share an existing db -> use a child service
- no need to share db, but needs alarm AND/OR database -> use a sibling service

> And if not, wouldn't it make more sense as a sibling?

No we don't, and In its current form, this should be a sibling service 💯. If we change the scope to also include `retrieveTransaction` this can change. How should we make this decision in the future? (I chose the simple cut path here and backed off from what felt a better separation of concerns)

> re:component / Cloujure DI 

I love this whole concept a lot! This is literally where I want to get to. All side effects in a service, services are small testable and composable.

background/services/types.ts
greg-nagy: [af35d2e](https://github.com/tallyhowallet/extension/pull/2733/commits/af35d2e061d29b32fae3919d6841bf481f30d3f3)

kkosiorowska:
I think breaking up the blocks into smaller services is a great idea. ✨ However, in this case, we should specify exactly how should be divided and what services should contain. I would like to avoid a situation where it would come to the division of blocks into too many services, which may create a mess.

In the process of breaking down services, I think it is good to create a new service diagram. A diagram describing in more detail what the service contains, what do and what connection has. I mean, maybe this way the division will turn out to be easier and cleaner. 🤔 The answer to the questions may be simpler. This can later be used for documentation

> * Should we share dbs between services or not?

I think not ? It is then difficult to determine what should be where. In addition, in such a case there should be rules on what can be shared and what should not. 

> * Should we create new services or more functions/classes? How do we decide when to use which? What is the definition of a service? How do we define its scope of it?

I think to answer this question a diagram may help. Seeing services from the side.

> * What is the best way to break the dependency cycle? Service1 <> Service2 method calls. Should we use

I like this article and the idea of refactoring. But I think it may be difficult to obtain. The services are already extensive and it may take a while.

Shadowfiend:
Going to close this out for the time being. It's not top of mind, we didn't reach clear actionable guidance here, and the bigger pain points after discussing with the teams are perhaps more focused around `Main` than around all services across the board.

PR #2750, opened 2022-12-13
Reload NFTs without a loader

### LINKED ISSUE #2737: Mark NFTs data as stale after the delay
- Let's not reload NFTs every time user enter the page, let's mark them as stale after 5min.
- Let's add a button to reload NFTs manually

jagodarybacka:
Resolves #2737

### What
On NFTs reload let's not show the page loader.
NFTs should be updated without huge layout shifts. Let's show the loader only when NFTs are fetched for the first time for some account - as we don't know what collections should we expect there.

### Testing

- [ ] add a new account and go to the NFTs page quickly - you may see the "doggo loader"
- [ ] go to Wallet pge and go back to the NFTs page - there should be no loader but NFTs are updated in the background. If the account receives a new NFT then it should be visible on the next NFTs page visit.

### DIFF
--- a/background/redux-slices/nfts_update.ts
+++ b/background/redux-slices/nfts_update.ts
@@ -68,7 +68,7 @@ function updateCollection(
   acc.nfts[chainID][ownerAddress][collection.id] = {
     id,
     name,
-    nftCount,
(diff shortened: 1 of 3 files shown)


greg-nagy (approved):
looks good, works well

PR #2754, opened 2022-12-13
Remove the assign issues to project flow from github actions

greg-nagy:
We don't use this — otherwise we would
have fixed the problem with it.

It's probably not a good idea to assign
issues automatically to a project because
it can result in a messy back-log.

This should be revisited when we will have
the new owner for project management.

Latest build: [extension-builds-2754](https://github.com/tallycash/extension/suites/9840916912/artifacts/474207013) (as of Tue, 13 Dec 2022 14:52:52 GMT).


jagodarybacka (approved):
Ty, it was so annoying 😂 Will wait for one more person to agree with removing it

hyphenized (approved):
> It's probably not a good idea to assign
issues automatically to a project because
it can result in a messy back-log.

Agree, let's remove this

PR #2788, opened 2022-12-15
[Proposal]: Refactor CI build flows into smaller chunks by function

greg-nagy:
⚠️ This is not mergeable! This code wasn't tested at all and can break many things. ⚠️
At this point, this is an illustration of the concept.

If this idea interferes with the new CI design we can close this PR without any further ado.

## Why

It has started to become increasingly difficult to reason about what is happening exactly in which workflow. 

## What 

PRs, main and release workflows have different triggers and different steps. Instead of unifying all the triggers and then separating steps conditionally, let's create a separate workflow for every context.

PRs should build, upload asset, detect env blocks, test and lint Main should build, upload asset, test, lint, run e2e tests Release should build, and create release

test and build is redundant between PR and main. This is intentional. In theory we never push directly to main but if we do — which we shouldn't — let's make sure lint and tests are passing.

Release is not running tests and lint. This might not be the best idea but seems reasonable enough for now.

## Future work

lint and test steps should be refactored and reused throughout the workflows

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,12 +1,9 @@
-name: Build
+name: Main flow
 
 on:
-  pull_request: {}
   push:
     branches:
       - main
-    tags:
-      - v*
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


hyphenized (commented):
> PRs, main and release workflows have different triggers and different steps. Instead of unifying all the triggers and then separating steps conditionally, let's create a separate workflow for every context. 

Great proposal! I agree and I think this is a step in the right direction. Not only are the workflows simpler and easier to modify, but, this also gives us a safety net by eliminating the possibility of accidentally making undesired changes to e.g. the release workflow.

We can move lint and test steps to a reusable workflow too

Shadowfiend:
Haven't had a chance to chime in here, but: concept is sound, current implementation duplicates a bunch of stuff. I would focus on reusable workflows connected to each other rather than repeating steps along the way whenever possible.

Shadowfiend:
This isn't top of mind right now, so closing the PR for the time being while noting it for reference if and when we want to circle back to this kind of refactor.

PR #2792, opened 2022-12-15
Updated CODEOWNERS for new org name

Shadowfiend:
tallycash has become tallyhowallet, so the old team name did not resolve.

Latest build: [extension-builds-2792](https://github.com/tallyhowallet/extension/suites/9896250759/artifacts/478063092) (as of Thu, 15 Dec 2022 21:09:12 GMT).


mhluongo (approved):
👌

PR #2797, opened 2022-12-16
Merge intersection observer hooks into one

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2766

### What 
As we have multiple places using the intersection observer hook let's merge them into one hook.

- I had to make some compromises as the [NFTs Collection component](https://github.com/tallyhowallet/extension/blob/unite-intersection-observers/ui/components/NFTS_update/NFTCollection.tsx#L49-L94) is passing to the intersection observer callbacks with some dependencies - so we have to allow callback to change over time. 
- I think creating element ref in the hook and returning it is a nice thing to have so we don't have to create element ref outside of the hook

Latest build: [extension-builds-2797](https://github.com/tallyhowallet/extension/suites/9992690730/artifacts/484975261) (as of Wed, 21 Dec 2022 14:39:59 GMT).

### DIFF
--- a/ui/hooks/dom-hooks.ts
+++ b/ui/hooks/dom-hooks.ts
@@ -5,6 +5,7 @@ import {
   useState,
   useCallback,
   useMemo,
+  useLayoutEffect,
 } from "react"
 
 export const useOnClickOutside = <T extends HTMLElement = HTMLElement>(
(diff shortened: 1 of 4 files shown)


ui/hooks/dom-hooks.ts
```diff
@@ -139,24 +140,27 @@ export function useLocalStorage(
...
-  const observer = useMemo(
-    () => new IntersectionObserver(callback, options),
-    [callback, options]
-  )
+  const optionsRef = useRef(initialOptions)
+  const elementRef = useRef<T>(null)
+
+  const observer = useMemo(() => {
+    return new IntersectionObserver(callback, optionsRef.current)
+  }, [callback, optionsRef])
```
hyphenized, line 151: We might miss updates if we drop and recreate the observer if a memoized callback changes. If we want to ensure that the callback is called with updated state references on observer events, we should change the hook implementation to hold the callback in a ref, update this ref's value on each render with the latest callback, and, call the ref's current value within the observer callback.

This would also relieve us from having to `useCallback` the handler that's passed as an argument.

hyphenized (changes requested):
This looks great! I also think it's a nice thing to not have to create the element ref outside of the hook, but, I think passing the ref as a parameter would offer us more flexibility in cases where we have multiple hooks iterating on an element's ref.

ui/hooks/dom-hooks.ts
jagodarybacka: You are right that we possibly can lose some updates that way but we also want to avoid unnecessary or doubled calls to the callback function.
I haven't noticed any problems with the current version - I'm mostly looking at the NFTs collections list as this is the most complicated use case we have right now. I feel like the solution you describe would break this place - but maybe I don't understand the idea fully. We can pair on this later 🤓

hyphenized: I feel like changing these implementations could cause the `backdrop-filter` bug to resurface (The intersection observer that applied the "fix" for that bug intentionally used `useEffect` instead of `useLayoutEffect`). I've created a [sandbox](https://codesandbox.io/s/black-worker-e954dl) that displays the issue I was referring to in my previous comment.


Related: https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md

jagodarybacka:
Stale, let's leave it as it is

PR #2799, opened 2022-12-16
Load NFT images with higher resolution if available

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2790

### What

On the NFTs Preview slide-up we would like to show images with higher resolution than on the list because they are bigger and poor quality starts to become visible.
As we already have loaded the smaller image (for the list items) it can be displayed instantly while the higher resolution image has some load time.
Let's progressively display images with better resolution on the preview while showing smaller images first for seamless UX.

### Testing

- load some accounts with NFTs, scroll the list, check out the preview slide-ups - images should be updated to higher resolution (barely visible most of the times I've checked)

Latest build: [extension-builds-2799](https://github.com/tallyhowallet/extension/suites/9986920660/artifacts/484584991) (as of Wed, 21 Dec 2022 09:15:58 GMT).

### DIFF
--- a/ui/components/NFTS_update/NFTImage.tsx
+++ b/ui/components/NFTS_update/NFTImage.tsx
@@ -1,5 +1,6 @@
 import classNames from "classnames"
-import React, { ReactElement, useState } from "react"
+import React, { ReactElement, useEffect, useState } from "react"
+import noop from "../../utils/noop"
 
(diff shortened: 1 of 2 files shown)


kkosiorowska (approved):
LGTM ✅

PR #2800, opened 2022-12-16
☀️ Daylight Abilities Spike 

0xDaedalus:
This PR is the result of a 2 day spike centered around bringing information from daylight.xyz into the wallet.  All functionality 

image

image

image

Everything should be properly feature flagged (UI elements, redux store creation, indexedDB creation) and should not interfere with existing functionality.  This is meant to be a jumping-off point for the abilities feature rather than a complete implementation.  

Some functionality is missing (properly reporting spam, filter menu) but the meat of the feature is there.  Completing the functionality and honing down to the design in the figma specification can be handled in subsequent PRs.

Latest build: [extension-builds-2800](https://github.com/tallyhowallet/extension/suites/9993095157/artifacts/485003341) (as of Wed, 21 Dec 2022 15:00:34 GMT).


background/main.ts
```diff
@@ -1572,6 +1590,20 @@ export default class Main extends BaseService<never> {
...
+    return this.abilitiesService.markAbilityAsCompleted(address, abilityId)
+  }
+
+  async markAbilityAsRemoved(
+    address: HexString,
+    abilityId: string
+  ): Promise<void> {
+    return this.abilitiesService.markAbilityAsRemoved(address, abilityId)
+  }
+
```
jagodarybacka, line 1606: I think it would be nice to decide what should be the approach to that kind of methods (that are just connectors between redux and services) - should we add methods to the `main` class (like you did here), or should we call services directly in redux slice (not sure if we have this approach anywhere) or should we make redux slice's event emitters (like I did for NFTs)

background/services/abilities/daylight.ts
```diff
@@ -0,0 +1,100 @@
...
+
+// eslint-disable-next-line import/prefer-default-export
+export const getDaylightAbilities = async (
+  address: string
+): Promise<DaylightAbility[]> => {
+  const response: AbilitiesResponse = await fetchJson(
+    // Abilities whose deadline has not yet passed - we will probably
+    // want to turn this on once the feature is ready to go live
+    // `${DAYLIGHT_BASE_URL}/${address}/abilities?deadline=set&type=mint&type=airdrop&type=access`
+    `${DAYLIGHT_BASE_URL}/${address}/abilities?type=mint&type=airdrop&type=access`
```
jagodarybacka, line 96: Is API key not needed?

background/services/abilities/db.ts
```diff
@@ -0,0 +1,72 @@
...
+      return true
+    }
+    return false
+  }
+
+  async getAbility(
+    address: string,
+    abilityId: string
+  ): Promise<Ability | undefined> {
+    return this.abilities.get({ address, abilityId })
```
jagodarybacka, line 36: Are you making sure somewhere that the address will have the same format as the one that was added to the DB?

background/services/abilities/index.ts
```diff
@@ -0,0 +1,200 @@
...
+import {
+  DaylightAbility,
+  DaylightAbilityRequirement,
+  getDaylightAbilities,
+} from "./daylight"
+import { AbilitiesDatabase, getOrCreateDB } from "./db"
+import ChainService from "../chain"
+import { FeatureFlags } from "../../features"
+
+export type AbilityType = "mint" | "airdrop" | "access"
```
jagodarybacka, line 13: You have that type already in the [background/services/abilities/daylight.ts](https://github.com/tallyhowallet/extension/pull/2800/files#diff-75c3a55a7960518dffb7cb5b4667f019b326e98842c3adc0fb78232eb1797d30) right?

background/services/abilities/index.ts
```diff
@@ -0,0 +1,200 @@
...
+    abilityId: string
+  ): Promise<void> {
+    return this.db.markAsRemoved(address, abilityId)
+  }
+
+  async abilitiesAlarm(): Promise<void> {
+    const accountsToTrack = await this.chainService.getAccountsToTrack()
+    const addresses = accountsToTrack
+      .map((account) => account.address)
+      .filter((address, index, self) => self.indexOf(address) === index)
```
jagodarybacka, line 191: Isn't it just `Set`?

background/services/abilities/index.ts
```diff
@@ -0,0 +1,200 @@
...
+    private chainService: ChainService
+  ) {
+    super({
+      abilitiesAlarm: {
+        schedule: {
+          periodInMinutes: 60,
+        },
+        runAtStart: true,
+        handler: () => {
+          this.abilitiesAlarm()
```
jagodarybacka, line 125: If I remember correctly user has an option to opt-out of the abilities right? Do we plan to stop that alarm then?
Do we even need an alarm for this? Can we fetch abilities based on user's actions in the wallet?

ui/components/Shared/SharedSlideUpMenu.tsx
```diff
@@ -92,7 +92,7 @@ export default function SharedSlideUpMenu(props: Props): ReactElement {
             height: var(--menu-height);
             overflow: hidden;
             border-radius: ${isFullScreen ? "0" : "16px 16px 0 0"};
-            background-color: var(--green-95);
+            background-color: #13302e;
```
jagodarybacka, line 95: We probably shouldn't do it 😄  Can we just add a new color variable?

ui/pages/Abilities/AbilityRemovalConfirm.tsx
```diff
@@ -0,0 +1,141 @@
...
+        .header {
+            height: 24px;
+            font-family: 'Segment';
+            font-style: normal;
+            font-weight: 600;
+            font-size: 18px;
+            line-height: 24px;
+        
+            
+            color: #D6EAE9;
```
jagodarybacka, line 95:
```suggestion
            line-height: 24px;     
            color: #D6EAE9;
```

jagodarybacka (commented):
Spike looks great ☀️ I've made some comments going through the code:

background/services/abilities/daylight.ts
0xDaedalus: there is a free tier that seems to be more than enough for our needs 😄

background/services/abilities/index.ts
0xDaedalus: yup but i made this type separate for two reasons:

1. We don't want to support every daylight type in the app (there is a lot of spam)
2. To decouple our internal data model from daylight's (what if we want to add a 'tally' ability in the future)

background/services/abilities/index.ts
0xDaedalus: Indeed - brain fart 😅

background/services/abilities/index.ts
0xDaedalus: > Do we plan to stop that alarm then?

I think we can add a check in the alarm that prevents it from doing anything

> Do we even need an alarm for this?

The current working plan is to eventually have some type of notification pop up when you are eligible for a new ability.  Decided to init the alarm here with a 60 minute timer with that eventual state in mind.

ui/components/Shared/SharedSlideUpMenu.tsx
0xDaedalus: Hmmmn looking at the figma file again it looks like `var(--green-95)` is indeed the correct color here - although I SWEAR it was not when I was working on this last week.  Probably a gap in my knowledge of how to use figma - will revert.

background/main.ts
0xDaedalus: Agreed - I have heard complaints that the event emitter pattern is a bit confusing (although I don't personally feel that way) which is why I went with this pattern.  I would _love_ to standardize how we do this - shall we start a thread in #Tech?

background/services/abilities/db.ts
0xDaedalus: Good catch - this was happening implicitly (and we were not normalizing) - so added normalization as well as a way for us to check for these kinds of inconsistencies at compile time going forward.

ui/pages/Abilities/AbilityCard.tsx
```diff
@@ -0,0 +1,173 @@
...
+import React, { ReactElement, useState } from "react"
+import SharedButton from "../../components/Shared/SharedButton"
+import SharedIcon from "../../components/Shared/SharedIcon"
+import SharedSlideUpMenu from "../../components/Shared/SharedSlideUpMenu"
+import SharedTooltip from "../../components/Shared/SharedTooltip"
+import { useBackgroundDispatch } from "../../hooks"
+import AbilityCardHeader from "./AbilityCardHeader"
+import AbilityRemovalConfirm from "./AbilityRemovalConfirm"
+
+const AbilityCard = ({ ability }: { ability: Ability }): ReactElement => {
```
hyphenized, line 12:
```suggestion
function AbilityCard({ ability }: { ability: Ability }): ReactElement {
```

hyphenized (changes requested):
image 

Daylight is here! It's looking good, left some comments!

background/main.ts
jagodarybacka: We can, I'll set a reminder to go back to this topic next year

ui/components/Overview/AbilitiesHeader.tsx
```diff
@@ -0,0 +1,95 @@
...
+          align-items: flex-end;
+          background: #0d2321;
+          border-radius: 17px;
+          padding: 4px 8px 4px 8px;
+          color: ${newAbilities > 0 ? "var(--success)" : "var(--green-40)"};
+          font-weight: 500;
+          cursor: pointer;
+        }
+        .icon_daylight {
+          background: url("./images/assets/daylight.png");
```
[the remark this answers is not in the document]
0xDaedalus: Happy to do in a followup

hyphenized (approved):
Great work on the spike, looks like this is good to go!

PR #2802, opened 2022-12-19
Display an error message from an RPC request instead of 'the user rejected the request'

### LINKED ISSUE #2672: tx-error: 'the user rejected the request'
### Discord Discussion Link

na

Chrome

Set-up a transaction

### What did not work?

Some users have reported receiving an error message when setting up a transaction. "The user rejected the request". 

It seems to not be user specific, but apply to all users for affect dApps, or transaction types on a specific dApp. With some dapps it consistently fails the 1st time, but when a user repeats tx set-up it does succeed.

There could be a relation with transactions that require an approval first, or where an approval tx is expected but approval was already given earlier on.

Reports / to reproduce:

**User: Nax**
**Swap on Velodrome (Optimism)**
Logs:
[logs_v0_18_2__2022-11-24T11_11_41+01_00.txt](https://github.com/tallycash/extension/files/10083365/logs_v0_18_2__2022-11-24T11_11_41%2B01_00.txt)

Screen Shot 2022-11-24 at 11 11 18

After receiving this error message, hit swap again, and it'll go through
Screen Shot 2022-11-24 at 11 12 28

**User: KrasniyVladi | Sui Global#3627**
**Bridge on https://app.chainhop.exchange/**

"
sup guys, cant make bridge with chainhop from optimism to arbitrum. I click to sign the transaction and nothing happens. I updated the site, reconnected - nothing works.
[2022-11-19 18:34:33.360][log][main.6855edfc.chunk.js:1 ()]  err: {code: 4001, message: 'The user rejected the request.'}
"
Logs:
[logs_v0_18_0__2022-11-19T18_39_3803_00 (2).txt](https://github.com/tallycash/extension/files/10083429/logs_v0_18_0__2022-11-19T18_39_3803_00.2.txt)

Note Nax: User used 0.18.0, I've failed to reproduce it with 0.18.2

### Version

v0.18.2

### Relevant log output

_No response_

kkosiorowska:
Closes #2672 

This PR propagates a real error message to the user. The issue described in the task no longer occurs or happens sporadically and is difficult to catch. PR is to help define the issues with transaction failure. During testing the solution, please try to execute several transactions to define possible issues, which will be fixed in subsequent tasks. 

## UI
208640296-75a6fb80-958c-45de-9739-47bab8dec15d

Screenshot 2022-12-19 at 10 51 22

## Steps to Recreate
Use [revoke.cash](https://revoke.cash/) to revoke the spend approval for the velodrome contract if needed.
- Swap on [Velodrome](https://app.velodrome.finance/swap) (Optimism)
- Make sure that transaction need an approval first 
- Make sure that the approval process passes. (Also, in this case, an error may occur.)
- Next, try to sign the transaction.
- Check if a propagated error is displayed.

Latest build: [extension-builds-2802](https://github.com/tallyhowallet/extension/suites/10662750685/artifacts/533034497) (as of Mon, 30 Jan 2023 14:15:13 GMT).


0xDaedalus:
This is technically not in line with EIP-1193 but i definitely see the value in it.  Curious to hear others thoughts on this.

jagodarybacka (commented):
So do I understand it correctly that because of this change we will be responding to dapps with a message with a possibly modified error code? Isn't it going to break something or cause wrong error handling in the dapp itself?

0xDaedalus:
> So do I understand it correctly that because of this change we will be responding to dapps with a message with a possibly modified error code?

The error code _must_ stay the same - but [the spec](https://eips.ethereum.org/EIPS/eip-1193#rpc-errors) dictates that:
```
message SHOULD adhere to the specifications in the Error Standards section below
````
which does specify the description sent along with code 4001.

Especially with the wording of "should" I'm thinking its ok for us to bend the rules here a little bit.

**Comment by kkosiorowska** [comment]
> So do I understand it correctly that because of this change we will be responding to dapps with a message with a possibly modified error code? Isn't it going to break something or cause wrong error handling in the dapp itself?

Only the message is modified. The change is intended to describe the issues more precisely and it shouldn't break anything. The message is extracted from the error and then displayed in the dapp.

File: background/services/provider-bridge/index.ts [hunk 1]
```diff
@@ -452,7 +452,17 @@ export default class ProviderBridgeService extends BaseService<Events> {
         }
       }
     } catch (error) {
-      logger.log("error processing request", error)
+      if (typeof error === "object" && error !== null) {
+        logger.log("error processing request", error)
+        // eslint-disable-next-line default-case
```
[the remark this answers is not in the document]
kkosiorowska: > It seems to me like it is not a "switch" if there is no default case + we have to use `true`

It seemed to me that this notation would be clearer and easier to understand. But of course, I can change this into a simple` if` structure.

jagodarybacka (dismissed):
I'm ok with bending the rules a little bit if it is not breaking other stuff

0xDaedalus (dismissed):
Lets use this PR as an opportunity to write our first test for the provider-bridge.

kkosiorowska:
> Lets use this PR as an opportunity to write our first test for the provider-bridge.

I created some first simple tests for provider-bridge.

background/services/provider-bridge/index.ts
```diff
@@ -452,7 +452,15 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
-      logger.log("error processing request", error)
+      if (typeof error === "object" && error !== null) {
+        logger.log("error processing request", error)
+        if ("body" in error) {
+          return getRPCErrorResponser(error)
+        }
+        if ("error" in error) {
+          return getRPCErrorResponser((error as { error: string }).error)
+        }
+      }
```
hyphenized, line 463: Some of the switch cases can throw errors that are already instances of `EIP1193Error` e.g.: `EIP1193Error(EIP1193_ERROR_CODES.unauthorized)`. If we're not intentionally shadowing those with `EIP1193_ERROR_CODES.userRejectedRequest` we should also handle them correctly in this catch statement.

kkosiorowska: Let's also handle this correctly.

background/services/provider-bridge/tests/index.unit.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  eth_sendTransaction: [
+    {
+      from: accountAddress,
+      data: Date.now().toString(),
+      gasPrice: "0xf4240",
+      to: "0x1111111111111111111111111111111111111111",
+    },
+  ],
+}
+describe("Provider bridge", () => {
```
hyphenized, line 43:
```suggestion
describe("ProviderBridgeService", () => {
```

hyphenized (changes requested):
Nice work, I think this is almost ready to be merged. Left a couple of additional comments.

background/services/provider-bridge/utils.ts
```diff
@@ -23,3 +26,20 @@ export const keyPermissionsByChainIdAddressOrigin = (
   })
   return map
 }
+
+export function getRPCErrorResponser(error: unknown): {
```
Shadowfiend, line 30: Because we're only calling this from our own code, what do you think about making `error` already be a `{ body: string }`, and letting the calling code handle both *checking* that this is the case, and casting to `{ body: string }`? My thinking here is that right now the caller does both the property checks and some of the parsing, so we should bring the casting close to those checks.

background/services/provider-bridge/utils.ts
```diff
@@ -23,3 +26,20 @@ export const keyPermissionsByChainIdAddressOrigin = (
...
+  code: number
+  message: string
+} {
+  const parsedError = JSON.parse((error as { body: string }).body)?.error
```
Shadowfiend, line 34: Similarly to above, just because we've determined this is a `{ body: string }` at this point, doesn't mean we know that it's parsable as JSON. We think it is, but the compiler won't help us avoid breaking this invariant in the future.

What do you think about wrapping all of this in a `try`/`catch`, and if the JSON parsing fails (or if `parsedError` is `undefined` or `null`!), returning `undefined`. Then the caller can surface a generic error if it can't properly extract the error.

Shadowfiend (commented):
Left a couple of notes on the casts and type safety here, but it's also worth noting that historically I don't think we've correctly surfaced underlying RPC errors (i.e., if we send an `eth_sendTransaction` and it returns with an error). It feels like that is part of what this is meant to address—but these should be returned *with the underlying RPC error code*! EIP-1193 states:

```
ProviderRpcError codes and messages SHOULD follow these conventions, in order of priority:

1. The errors in the [Provider Errors](https://eips.ethereum.org/EIPS/eip-1193#provider-errors) section below
2. Any errors mandated by the erroring RPC method’s specification
3. The [CloseEvent status codes](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent#Status_codes)
```

(2) is what these errors fall under, and they shouldn't produce a “The user rejected the request” error code. This is a longstanding bug in our implementation of the bridge.

From a “get things done” perspective, I would suggest finishing this off as-is, then seeing if there are error codes we should be passing through instead of replacing with the 4001 code. Super super excited to see this being tackled!

jagodarybacka (commented):
No manual tests yet but couple comments on the code, looking good btw 😊

PR #2805, opened 2022-12-20
Update data after transferring last NFT from collection

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2789

### What
When the last NFT from a given collection is transferred we have to remove the collection from the database and redux slice.

### Testing
- transfer NFTs (you can use OpenSea) between accounts, pick collection that has only one NFT
- check if everything is updated correctly - you can't see the collection or NFT that you transferred

Latest build: [extension-builds-2805](https://github.com/tallyhowallet/extension/suites/10158162594/artifacts/496426456) (as of Tue, 03 Jan 2023 14:00:30 GMT).


background/services/nfts/db.ts
```diff
@@ -79,12 +79,24 @@ export class NFTsDatabase extends Dexie {
       .delete()
   }
 
-  async removeNFTsByID(removedNFTs: TransferredNFT[]): Promise<void> {
-    await this.nfts
-      .filter((nft) =>
-        removedNFTs.some((transferred) => transferred.id === nft.id)
-      )
+  async removeByTransferredNFTs(removedNFTs: TransferredNFT[]): Promise<void> {
```
0xDaedalus, line 82: This method name is a bit confusing - unclear that someone would be able to tell whats going on in this method by reading the method name

jagodarybacka: Do you have a better idea? Maybe `cleanupAfterTransfers`?

background/redux-slices/nfts_update.ts
```diff
@@ -237,16 +237,32 @@ const NFTsSlice = createSlice({
...
-          if (collection) {
-            collection.nfts = collection.nfts.filter((nft) => nft.id !== id)
+            if (collection) {
+              const hasTransferredNFT = collection.nfts.some(
+                (nft) => nft.id === nftID
+              )
+
+              if (hasTransferredNFT) {
+                if (collection.nfts.length === 1) {
+                  delete immerState.nfts[chainID][normalizedAddress][
```
kkosiorowska, line 254: We should probably also remove the collection from the filter list

background/services/nfts/db.ts
```diff
@@ -79,12 +79,24 @@ export class NFTsDatabase extends Dexie {
...
+    const nftsToRemove = this.nfts.filter((nft) =>
+      removedNFTs.some((transferred) => transferred.id === nft.id)
+    )
+
+    const collectionIdsToRemove = (await nftsToRemove.toArray()).reduce(
```
greg-nagy, line 87: collection ids should be available in the [api's `collection` property.](https://docs.simplehash.com/reference/nfts-by-owners)

Using that value could simplify this part

background/services/nfts/db.ts
```diff
@@ -79,12 +79,24 @@ export class NFTsDatabase extends Dexie {
...
+      (acc, nft) => {
+        acc.add(nft.collectionID)
+        return acc
+      },
+      new Set<string>()
+    )
+
+    await this.collections
+      .filter((collection) => collectionIdsToRemove.has(collection.id))
```
greg-nagy, line 96: Blocker

How can we be sure that all the NFTs have been removed from the collection? 

Some tests would be useful.

background/services/nfts/db.ts
greg-nagy: (Soft blocker)

The db layer shouldn't care/know about the fact that these nfts were transfered. 
The db layer should handle only the concern of removing NFTs. 

I would use `removeNFTsByIDArray` and using only the ids array as params.

greg-nagy (changes requested):
The remove from db function might be buggy and it should have some tests.

The separation of concerns between the db and the service layer is a very nice to have. 
Seems nounced but these small tidbits can add up over time.

background/services/nfts/db.ts
jagodarybacka: I think [SH takes care](https://docs.simplehash.com/reference/nft-model) about making NFT ids unique:
> Unique identifier for an NFT (chain.contract_address.token_id for EVM chains & Flow, chain.mint_address for Solana)

Even if we remove an NFT that is actually still owned by a given account the next step in the whole "NFTs data update" process is to refetch collections - so accidentally removed NFT/collection should be added back to our local data.

background/services/nfts/db.ts
jagodarybacka: The problem is that we don't really know if all NFTs were removed as there is no info about the collection state at this point (requesting transfers from SH). In the database, we may not have all NFTs from the given collection (they are not fetched yet).

The idea behind this code was that we will remove the collection if any NFT from this collection was removed - because the next step will be to [refetch all collections](https://github.com/tallyhowallet/extension/blob/nft-detect-transfers-fixes/background/services/nfts/index.ts#L97) so they will be back in the database. 

I see now that this is not clear looking at this method alone so I will think about improving the structure, feel free to give some suggestions 😃

background/redux-slices/nfts_update.ts
jagodarybacka: You found a 🐛 !

greg-nagy (approved):
looks, good, let's 🚚

PR #2807, opened 2022-12-21
[FIX] Rootstock address validation on transaction passed from external dapp

ahsan-javaiid:
## Description

Tally is rejecting Rootstock transactions passed from external dapps. For example, if an external dapp
passes checksummed `from` and `to` fields in transaction payload then Tally process dies with below error log.

```
logger.ts:211 Error signing transaction Error: bad address checksum (argument="address", value="0xE6671D3846C57de1c559Ff2Da7bc0f4B4fEa9521", code=INVALID_ARGUMENT, version=address/5.5.0)
    at Logger.makeError (index.js:185:1)
    at Logger.throwError (index.js:194:1)
    at Logger.throwArgumentError (index.js:197:1)
    at getAddress (index.js:78:1)
    at Formatter.address (formatter.js:188:26)
    at SerialFallbackProvider.<anonymous> (base-provider.js:1559:1)
    at Generator.next (<anonymous>)
    at fulfilled (base-provider.js:5:43)
```

## Reason

Because ether.js library rejects Rootstock addresses. Below is script to justify this:

```
import { ethers } from "ethers"
import { toChecksumAddress } from "@tallyho/hd-keyring"

const validRSKAddress = "0xE6671D3846C57de1c559Ff2Da7bc0f4B4fEa9521"

void (() => {
  try {
    const addr = ethers.utils.getAddress(validRSKAddress)
    console.log("Address validation passed",  addr)
  } catch (e) {
    console.log("ether rejected the address", e)
  }

// Solution: Validate using chainId is a correct way to validate Rootstock addresses
  console.log(
    "Is really valid rootstock address: ",
    isValidChecksumAddress(validRSKAddress, 30)
  ) // True
})()

function isValidChecksumAddress(address: string, chainId?: number): boolean {
  return toChecksumAddress(address, chainId) === address
}
```

## Suggested solution

- Validate Rootstock transaction address fields using `isValidChecksumAddress` function at the point when
  dapp passes transaction to tally for signature
- After that make `to` and `from` fields to lowercase to skip the ether.js address validation
  Note: Above two points will only apply if network is `Rootstock`

## Type of Change

- Fix

## Checklist

- [x] Git hooks enabled
- [x] No lint errors
- [x] Git commit is signed

Latest build: [extension-builds-2807](https://github.com/tahowallet/extension/suites/12850672175/artifacts/692679448) (as of Fri, 12 May 2023 08:38:33 GMT).


ahsan-javaiid:
cc: @0xDaedalus @alepc253

ahsan-javaiid:
Hi Tally team 👋
Just mentioning one other approach for this pull request. 

**Alternative approach**:
1) Validate address fields using chainId approach for all the chains like in this PR but remove `Roostock` check
2) Lowercase the address fields in `tx` payload regardless of network (Remove Rootstock check in this PR)
Eventually ehter.js address validation will be bypassed for all the chains and addresses will be validated as per chainId approach

Let me know whichever approach makes more sense to the team and feel free to share further thoughts.

background/services/internal-ethereum-provider/index.ts
```diff
@@ -328,16 +351,19 @@ export default class InternalEthereumProviderService extends BaseService<Events>
         data: transactionRequest.input,
         ...transactionRequest,
         gasLimit: transactionRequest.gas, // convert gas -> gasLimit
+        // Etherjs rejects Rootstock checksummed addresses so convert to lowercase
+        from: isRootstock
+          ? transactionRequest.from?.toLowerCase()
```
jagodarybacka, line 356: Would the` normalizeEVMAddress()` function be helpful here?

jagodarybacka (commented):
What is the best dapp to test this PR with?

background/services/internal-ethereum-provider/index.ts
ahsan-javaiid: Yes it can be used in place of `toLowerCase ` as the output is same. 👍

alepc253:
> What is the best dapp to test this PR with?

Hi @jagodarybacka, we found this while porting an old dApp to make it work with Tally and Ledger but it was in dev/local environment (anyway, I'll try to find a productive example to share). In general, we take the approach of turning to lowercase in the dApp before sending the transaction but it would be great if this is also considered in Tally (because we can prevent errors and, in addition to this, remove a possible friction making Tally the easiest wallet to integrate your dApp on Rootstock).

ahsan-javaiid:
cc: @0xDaedalus for review.

ahsan-javaiid:
Hi Tally team:

Any thought's on considering this pull request. 🙏🏻

cc: @0xDaedalus

background/services/internal-ethereum-provider/index.ts
Shadowfiend: > See the difference by passing all uppercase address to normalizeEVMAddress function. // output 0x.

This isn't true as far as I can tell. This will only happen if the passed address isn't valid hex, which I think makes `normalizeEVMAddress` the safer option. `normalizeEVMAddress` (via the hd-keyring `normalizeHexAddress` call) parses the hex to bytes and pads to an even length before converting back to hex, rather than simply lowercasing the string that's passed (see https://github.com/tahowallet/hd-keyring/blob/main/src/utils.ts#L20-L28).

```
> const keyring = require("@tallyho/hd-keyring")
undefined
> keyring.normalizeHexAddress
[Function: normalizeHexAddress]
> keyring.normalizeHexAddress("0x52908400098527886E0F7030069857D2E4169EE7")
'0x52908400098527886e0f7030069857d2e4169ee7'
> keyring.normalizeHexAddress("0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA")
'0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
> keyring.normalizeHexAddress("0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAE")
'0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaae'
> keyring.normalizeHexAddress("0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAG")
'0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' // last byte is chomped as invalid
> keyring.normalizeHexAddress("0xGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGG")
'0x' // all invalid bytes are chomped
```

ahsan-javaiid: - Perfect and thanks for pointing it out 👍
- Updated the implementation to use `normalizeHexAddress` ✅
- Feel free to reconsider the pull request 🙏

Shadowfiend:
Will be working to get this merged early this week, thanks for the quick turnaround @ahsan-javaiid !

background/services/internal-ethereum-provider/index.ts
```diff
@@ -419,16 +442,20 @@ export default class InternalEthereumProviderService extends BaseService<Events>
         data: transactionRequest.input,
         ...transactionRequest,
         gasLimit: transactionRequest.gas, // convert gas -> gasLimit
+        // Etherjs rejects Rootstock checksummed addresses so convert to lowercase
+        from: isRootstock
+          ? transactionRequest.from &&
+            normalizeHexAddress(transactionRequest.from)
+          : transactionRequest.from,
```
Shadowfiend, line 449:
```suggestion
        from: isRootstock && transactionRequest.from !== undefined
          ? normalizeHexAddress(transactionRequest.from)
          : transactionRequest.from,
```

Let's move the existence check for `from` to the top here.

Shadowfiend: Really hate that we have to handle `from` here when we validate its existence further down… To be tackled another day.

ahsan-javaiid: ✅

Shadowfiend (approved):
This looks good in principle. I haven't had a chance to test drive it (I would probably send a raw RPC request to the provider from the debugger in a dApp tbh). 

Could you give us a test dApp that *doesn'* do this where we might intercept a signing request and change the casing?

alepc253:
@Shadowfiend this is the dapp where we found the issue: https://gnosis.rsk.co/
Video example attached:
https://user-images.githubusercontent.com/3451282/234598148-6d28f8b6-b69d-464a-91d5-6952c801e1fb.mov

jagodarybacka (approved):
Tested with that gnosis rsk dapp, works fine, no error anymore ✅ 
Regression tests for signing and sending txs passed ✅

kkosiorowska (approved):
Let's merge it :rocket:

PR #2813, opened 2022-12-23
Don't refresh NFTs collections if there are no accounts

jagodarybacka:
### What

Omit first NFTs refresh after extension installation

### Testing

Reinstall extension, open background dev tools - there should be no error in the console from the NFTs service

Latest build: [extension-builds-2813](https://github.com/tallyhowallet/extension/suites/10095829481/artifacts/492153849) (as of Thu, 29 Dec 2022 07:59:23 GMT).


kkosiorowska (approved):
LGTM ✨

PR #2814, opened 2022-12-23
NFT collection expanding

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2720

### What
Expanding NFT collection as an overlay above other items on the list.
The direction of expansion depends on the side of the list the given element occupies.

https://user-images.githubusercontent.com/20949277/212049141-dbad73a1-2fd1-4b21-bc73-0bfcbf90e415.mov

## Testing Env

```
SUPPORT_NFT_TAB=true
```

Latest build: [extension-builds-2814](https://github.com/tallyhowallet/extension/suites/10329676442/artifacts/508473384) (as of Thu, 12 Jan 2023 10:58:16 GMT).


jagodarybacka:
@VladUXUI please take a look - it needs some improvements - for example bottom line of the expanded collection is on top of other items which looks weird - but overall it feels pretty nice to me.
Not overlaying items below wasn't working as nicely when I tried it but if you won't like this version we can try again.
We have to stick to the CSS-only solutions as this is already heavy for the browser and any JS-initialised animations were awful.

VladUXUI:
Looking and working nice! still a bit iffy on how they move down, but i couldn't find a better design solution yet.
Let's add a box shadow to the expanded view, so it's a bit more clear it's an overlay
box-shadow: 0 3px 7px rgb(0 20 19 / 54%), 0 14px 16px rgb(0 20 19 / 54%), 0 32px 32px rgb(0 20 19 / 20%);

VladUXUI:
One "bug" although it's probably something you know, is that if i open multiple, they overlap.
Would be good to be able to have only one open.

Another extra good thing would be if we could automatically close it as soon as user scrolls out of view

greg-nagy (approved):
ok, this is so beautiful that everything else looks like pixel art from the 90's ... 😍

code looks good 

🚀

PR #2822, opened 2022-12-30
Create a priority queue of transaction hashes

### LINKED ISSUE #2046: New account wallet activity "starvation"
To reproduce...

- [ ] Load up a read-only account with a lot of activity
- [ ] Load up a read-only account without much activity

The second account won't have any wallet activity show up until *every single transaction* in the first account has been pulled, sometimes taking hours — because we're relying on a FIFO queue of transaction hashes per network rather than some sort of priority queue.

Thankfully, assets appear to still be discovered quickly, as they're inferred before transaction receipts are retrieved.

kkosiorowska:
Closes #2046 

When we add a second account, the activity won't appear on it until every transaction on the first account has been retrieved, which takes a very long time. For this reason, add priorities to the queue. 

This PR creates a priority queue of transaction hashes. The first 25 transactions for an account have a higher priority than the rest. Transactions with priority for individual accounts will keep the order of loading from adding accounts.

## UI

Before

https://user-images.githubusercontent.com/23117945/210052072-f9583258-2fe3-4fc4-97aa-b5c120b50502.mov

After

https://user-images.githubusercontent.com/23117945/210052148-cb81f681-1943-4816-aef9-510ea062d9d5.mov

## To Test
- Load up a read-only account with a lot of activity
- Load up a read-only account without much activity

Latest build: [extension-builds-2822](https://github.com/tallyhowallet/extension/suites/10409723765/artifacts/514296480) (as of Tue, 17 Jan 2023 07:58:15 GMT).


background/services/chain/index.ts
```diff
@@ -1024,11 +1059,16 @@ export default class ChainService extends BaseService<Events> {
...
-    return this.transactionsToRetrieve.some(
-      ({ hash, network }) =>
-        hash === txHash && txNetwork.chainID === network.chainID
-    )
+  isTransactionHashPriorityQueued(
+    txNetwork: EVMNetwork,
+    txHash: HexString
+  ): boolean {
+    return this.transactionsToRetrieve
+      .map((transaction) => transaction[0])
```
jagodarybacka, line 1067: I think mapping the whole thing is not necessary here

background/services/chain/index.ts
```diff
@@ -133,6 +139,11 @@ export type QueuedTxToRetrieve = {
   hash: HexString
   firstSeen: UNIXTime
 }
+/**
+ * The first element in the array is the transaction info the second is the priority.
+ * The priority value is a number in the range 0-1. Where 0 value is the highest priority.
```
jagodarybacka, line 144: Not blocking but I feel like higher number - higher priority would be more intuitive approach

background/services/chain/index.ts
```diff
@@ -133,6 +139,11 @@ export type QueuedTxToRetrieve = {
...
+ */
+export type PriorityQueuedTxToRetrieve = [QueuedTxToRetrieve, number]
```
jagodarybacka, line 146: Idea: maybe instead of an array, it would be easier to use an object to eliminate index-based references.
```suggestion
export type PriorityQueuedTxToRetrieve = { transaction: QueuedTxToRetrieve, priority: number }
```

background/services/chain/index.ts
```diff
@@ -1004,17 +1019,37 @@ export default class ChainService extends BaseService<Events> {
    * @param firstSeen The timestamp at which the queued transaction was first
    *        seen; used to treat transactions as dropped after a certain amount
    *        of time.
+   * @param priority The priority of the transaction in the queue to be retrieved
    */
-  queueTransactionHashToRetrieve(
+  priorityQueueTransactionHashToRetrieve(
```
0xDaedalus, line 1024: Not sure if there is value in describing the implementation method of the queue in the method name.  Might be more idiomatic to just leave it as `queueTransactionHashToRetrieve`

background/services/chain/index.ts
```diff
@@ -1004,17 +1019,37 @@ export default class ChainService extends BaseService<Events> {
...
+    ]
+    const seen = this.isTransactionHashPriorityQueued(network, txHash)
     if (!seen) {
       // @TODO Interleave initial transaction retrieval by network
-      this.transactionsToRetrieve.push({ hash: txHash, network, firstSeen })
+      if (this.transactionsToRetrieve.length === 0) {
+        this.transactionsToRetrieve.push(newElement)
+      } else {
+        let isAdded = false
+        for (let i = 0; i < this.transactionsToRetrieve.length; i += 1) {
```
0xDaedalus, line 1041: I think we can use a `find` here rather than this for loop - right?

greg-nagy (commented):
Wow! This is the very same problem that I got scarred off and instead started the #2733 and wanted to finish after the queue logic is encapsulated. 

I think this is a good PR in the sense that is does fix the problem outlined in the issue. Because of this I am not strictly against merging it. The follow up work can be handled in a follow up PR/issue. 

Some background: 
- At the moment this starvation exists for both networks and wallets. 
- txs are constantly removed and added to the queue > this can change the priority of a tx
  - > but this change is probably not that significant
- loading txs from the indexed db can also be a heavy path depending on how many tx we have queried already when the extension was shut down
  - > altough this is not that likely

The good part in this PR is that it handles the most problematic path. 
But it also couples the priority definition to the `loadAssetTransfers` and the alchemy api response. Which feels off given the content of the queue can change and there are multiple factors that should be taken into account. 

The generic solution I had in mind is a retrive function that encapsulates the logic of balancing the requests between networks and wallets. For this to be also performant the queue structure also needs to change to something like `{[address+network]: QueuedTxToRetrieve}` and the txs should be put in the correct "bucket" when added to the queue. With this structure it's easy to always grab the next element without constantly iterating through the queu. Otoh this requires some administration logic.

I will leave with you wheter you want to merge this PR with this approach — after going through the review OR want to change to the solution outlined above OR you want to wait for #2733 and implement it on top of it. (Note: #2733 is likely to change and probably will take some time to be merged. Please take this into account)

kkosiorowska:
> Sorry for the confusion. In the previous review I wanted to outline the overall picture but mixed in some comments to the current implementation that are blockers. I leave this request change to specifically note the issues with the current implementation.

@greg-nagy 
I have looked into this problem more. I think this PR should be the beginning of changes. I will implement in it the improvements from the review. Then I will try to implement your suggestions from the first comment in the next PR. We could agree together on what we want to change/add. 
And also I think we shouldn't wait for #2733 to do it because as you said it may take some time.

background/services/chain/index.ts
kkosiorowska: Good point!

background/services/chain/tests/index.unit.test.ts
```diff
@@ -320,5 +325,60 @@ describe("Chain Service", () => {
...
+            // First PRIORITY_MAX_COUNT items from the hashesForSecondAccount
+            else if (idx - PRIORITY_MAX_COUNT < PRIORITY_MAX_COUNT)
+              hash = hashesForSecondAccount[idx - PRIORITY_MAX_COUNT]
+            // The rest of the items for hashesForFirstAccount
+            else if (idx < NUMBER_OF_TX + PRIORITY_MAX_COUNT)
+              hash = hashesForFirstAccount[idx - PRIORITY_MAX_COUNT]
+            // The rest of the items for hashesForSecondAccount
+            else {
+              hash = hashesForSecondAccount[idx - NUMBER_OF_TX]
+            }
```
0xDaedalus, line 377: There is a bit too much going on in this test - I don't see a way for us to easily debug this should the test fail (the assertion will just tell us that the hashes are not equal and then we'll have to go around digging in the code and/or test itself to figure out what is wrong).

Since these are unit tests - generally this means that we're trying to assert too much behavior in one test.  Can we please break this test out into multiple ones - ideally each one testing _only one_ aspect of the function we're testing.

0xDaedalus (changes requested):
Functionality is there and seems to work 🎉  - I think we need to improve the test coverage though as this code is definitely in the hot-path and is complicated enough to warrant robust tests.

0xDaedalus (approved):
🚀🚀🚀

PR #2834, opened 2023-01-03
Persist EVM Networks in Chain Service DB

0xDaedalus:
Closes https://github.com/tallyhowallet/extension/issues/2759

A first step in allowing dapps to add networks via [EIP-3085](https://eips.ethereum.org/EIPS/eip-3085).  This PR sets up the foundation for chain service to communicate over networks sent in through `wallet_addEthereumChain`.

### To Test
- [ ] Load up an account
- [ ] Balances should load on every network, sending transactions should work as well.

Latest build: [extension-builds-2834](https://github.com/tallyhowallet/extension/suites/10163103326/artifacts/496790283) (as of Tue, 03 Jan 2023 18:22:56 GMT).


background/services/chain/index.ts
```diff
@@ -170,7 +163,7 @@ export default class ChainService extends BaseService<Events> {
 
   private lastUserActivityOnNetwork: {
     [chainID: string]: UNIXTime
-  }
+  } = {}
```
greg-nagy, line 166: ❤️

background/services/chain/index.ts
```diff
@@ -363,6 +331,24 @@ export default class ChainService extends BaseService<Events> {
     )
   }
 
+  async initializeNetworks(): Promise<void> {
```
greg-nagy, line 334: So this is the reason you didn't populate the db on migration 💡

background/services/chain/db.ts
```diff
@@ -165,6 +176,25 @@ export class ChainDatabase extends Dexie {
...
   }
 
+  async getAllEVMNetworks(): Promise<EVMNetwork[]> {
+    return this.networks.where("family").equals("EVM").toArray()
+  }
+
+  async initializeEVMNetworks(): Promise<void> {
+    const existingNetworks = await this.getAllEVMNetworks()
+    await Promise.all(
+      DEFAULT_NETWORKS.map(async (defaultNetwork) => {
```
greg-nagy, line 186: nit: `DEFAULT_NETWORKS` feels strange here. This should probably come from the service as a param.

greg-nagy (changes requested):
a very welcome step indeed 

let's fix the tests and remove the console.log then, it's good to go

the code is good and works well

greg-nagy (approved):
🤩

PR #2837, opened 2023-01-04
Add manifest v3 feature

### LINKED ISSUE #2827: Add feature flag for manifest v3 and switch between `manifest.json` versions

aurelsziebert:
Closes #2827

Makes possible to use V3 version of the manifest file.

Note that the project is broken with this feature on, because there are a few necessary changes in order to fulfil the migration from V2 to V3, what is out of the scope of this PR.

These changes are listed in issue #2694, and will be shipped in further PRs soon.

## To test

- [ ] Put `SUPPORT_MANIFEST_V3=true` into your .env file.

Latest build: [extension-builds-2837](https://github.com/tallyhowallet/extension/suites/10193524764/artifacts/502599142) (as of Sun, 08 Jan 2023 15:08:29 GMT).


webpack.config.ts
```diff
@@ -252,9 +280,32 @@ export default (
...
+                    assets.findIndex(
+                      (a) =>
+                        a.sourceFilename ===
+                        asset.sourceFilename.replace(
+                          ".json",
+                          `${overrideFlag}.json`
+                        )
+                    ) > -1
+                  return overrideFound ? acc : [...acc, asset]
+                }
```
hyphenized, line 303: I think that rather than dropping the base file if an override is present, we could get rid of these extra checks if we had two base files (e.g. `manifest.v2.json` & `manifest.v3.json`) and changed the base prefix (currently `manifest`) depending whether `SUPPORT_MANIFEST_V3` is enabled or not.

We'd then have something like the following:

- `manifest.v3.json`
- `manifest.v3.chrome.json`
- `manifest.v3.development.json`
- `manifest.v2.json`
- `manifest.v2.chrome.json`
- `manifest.v2.development.json`

This would allow us to keep the explicit file names and have a cleaner file structure to represent version and mode specific overrides.

hyphenized (changes requested):
Hi @aurelsziebert, thank you so much for jumping into this! Left you a couple of comments

webpack.config.ts
mhluongo: Seconded.

aurelsziebert: That would be simpler, I agree. Sometimes I tend to overcomplicate things.
Although the goal is reached on both way, and this is already a working solution, I'll take your advice, and make the change, because it makes the code cleaner, and safer. And it doesn't take much time anyway.
Thanks for the comment!

manifest/manifest.chrome.json
```diff
@@ -1,5 +1,5 @@
 {
-  "minimum_chrome_version": "74",
+  "minimum_chrome_version": "88",
```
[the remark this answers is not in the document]
aurelsziebert: Pushed, thanks for the comment.

Shadowfiend:

PR #2840, opened 2023-01-04
Persist BaseAssets in IndexedDB 

### LINKED ISSUE #2760: Persist BaseAssets in IndexedDB
In order to allow users to add / remove evm chains as they please - we need a way to persist and retrieve base assets of the chains they have added.  

Per [EIP-3085](https://eips.ethereum.org/EIPS/eip-3085) it looks like the combination of an asset's name/symbol/decimals is all we can use to determine the uniqueness of a given asset.

Futhermore - we will probably need to change up our [NetworkBaseAsset](https://github.com/tallycash/extension/blob/main/background/networks.ts#L21:L21) type and dependent code since we will certainly not get `coinType` or `metadata` when importing a wallet via `wallet_addEthereumChain`.

In addition to base asset persistence / retrieval - a strong solution here would:

- Make use of cacheing to minimize reads & write to indexedDB
- Initialize in a way that would not require us to write migrations when adding base assets of new "supported-by-default" networks in the future.

kkosiorowska:
Closes #2760 

 This PR sets up base assets in IndexedDB. The goal in the future is to allow users to add/remove chains.

Screenshot 2023-01-04 at 10 35 52

### To Test
- [ ] check that the base assets have been added to IndexedDB

### DIFF
--- a/background/constants/currencies.ts
+++ b/background/constants/currencies.ts
@@ -1,5 +1,6 @@
 import { FiatCurrency } from "../assets"
 import { NetworkBaseAsset } from "../networks"
+import { DEFAULT_BASE_ASSETS_BY_SYMBOL } from "./default-currencies"
 import { coinTypesByAssetSymbol } from "./coin-types"
 
 export const USD: FiatCurrency = {
@@ -14,9 +15,7 @@ export const FIAT_CURRENCIES_SYMBOL = FIAT_CURRENCIES.map(
 )
 
 export const ETH: NetworkBaseAsset = {
-  name: "Ether",
-  symbol: "ETH",
-  decimals: 18,
+  ...DEFAULT_BASE_ASSETS_BY_SYMBOL.ETH,
   coinType: coinTypesByAssetSymbol.ETH,
   metadata: {
     coinGeckoID: "ethereum",
@@ -26,9 +25,7 @@ export const ETH: NetworkBaseAsset = {
 }
 
 export const RBTC: NetworkBaseAsset = {
-  name: "RSK Token",
-  symbol: "RBTC",
-  decimals: 18,
+  ...DEFAULT_BASE_ASSETS_BY_SYMBOL.RBTC,
   coinType: coinTypesByAssetSymbol.RBTC,
   metadata: {
     coinGeckoID: "rootstock",
@@ -38,9 +35,7 @@ export const RBTC: NetworkBaseAsset = {
 }
 
 export const OPTIMISTIC_ETH: NetworkBaseAsset = {
-  name: "Ether",
-  symbol: "ETH",
-  decimals: 18,
+  ...DEFAULT_BASE_ASSETS_BY_SYMBOL.ETH,
   coinType: coinTypesByAssetSymbol.ETH,
   contractAddress: "0xdeaddeaddeaddeaddeaddeaddeaddeaddead0000",
   metadata: {
@@ -51,9 +46,7 @@ export const OPTIMISTIC_ETH: NetworkBaseAsset = {
 }
 
 export const MATIC: NetworkBaseAsset = {
-  name: "Matic Token",
-  symbol: "MATIC",
-  decimals: 18,
+  ...DEFAULT_BASE_ASSETS_BY_SYMBOL.MATIC,
   coinType: coinTypesByAssetSymbol.MATIC,
   contractAddress: "0x0000000000000000000000000000000000001010",
   metadata: {
@@ -64,9 +57,7 @@ export const MATIC: NetworkBaseAsset = {
 }
 
 export const AVAX: NetworkBaseAsset = {
-  name: "Avalanche",
-  symbol: "AVAX",
-  decimals: 18,
+  ...DEFAULT_BASE_ASSETS_BY_SYMBOL.AVAX,
   coinType: coinTypesByAssetSymbol.AVAX,
   metadata: {
     coinGeckoID: "avalanche-2",
@@ -76,9 +67,7 @@ export const AVAX: NetworkBaseAsset = {
 }
 
 export const BNB: NetworkBaseAsset = {
-  name: "Binance Coin",
-  symbol: "BNB",
-  decimals: 18,
+  ...DEFAULT_BASE_ASSETS_BY_SYMBOL.BNB,
   coinType: coinTypesByAssetSymbol.BNB,
   metadata: {
     coinGeckoID: "binancecoin",
@@ -88,9 +77,7 @@ export const BNB: NetworkBaseAsset = {
 }
 
 export const BTC: NetworkBaseAsset = {
-  name: "Bitcoin",
-  symbol: "BTC",
-  decimals: 8,
+  ...DEFAULT_BASE_ASSETS_BY_SYMBOL.BTC,
   coinType: coinTypesByAssetSymbol.BTC,
   metadata: {
     coinGeckoID: "bitcoin",
--- a/background/networks.ts
+++ b/background/networks.ts
@@ -18,7 +18,9 @@ export type NetworkFamily = "EVM" | "BTC"
 
 // Should be structurally compatible with FungibleAsset or much code will
 // likely explode.
-export type NetworkBaseAsset = FungibleAsset &
+export type NetworkDefaultBaseAsset = FungibleAsset
+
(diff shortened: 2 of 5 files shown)


background/constants/currencies-with-coin-types.ts
```diff
@@ -0,0 +1,94 @@
...
+  coinType: coinTypesByAssetSymbol.RBTC,
+  metadata: {
+    coinGeckoID: "rootstock",
+    tokenLists: [],
+    websiteURL: "https://www.rsk.co/",
+  },
+}
+
+const OPTIMISTIC_ETH: BaseAssetWithCoinType = {
+  ...DEFAULT_BASE_ASSETS_BY_SYMBOL.ETH,
```
jagodarybacka, line 34: Shouldn't it be OPTIMISTIC_ETH?

background/constants/currencies-with-coin-types.ts
```diff
@@ -0,0 +1,94 @@
...
+import { coinTypesByAssetSymbol } from "./coin-types"
+import { BASE_ASSETS } from "./currencies"
+
+const SYMBOLS = BASE_ASSETS.map(({ symbol }) => symbol)
+type BaseAssetSymbol = typeof SYMBOLS[number]
+
+const DEFAULT_BASE_ASSETS_BY_SYMBOL = BASE_ASSETS.reduce(
+  (sum, asset) => ({ ...sum, [asset.symbol]: asset }),
+  {}
+) as { [key in BaseAssetSymbol]: NetworkBaseAsset }
```
jagodarybacka, line 11: I feel like it's too complicated compared to what we want to achieve as these assets are exported individually too, can't we just import these assets to one variable and get them by name? Or rename them during import?

background/redux-slices/0x-swap.ts
```diff
@@ -159,7 +160,10 @@ const get0xAssetName = (asset: SwappableAsset, network: EVMNetwork) => {
   if (
     network.name === OPTIMISM.name &&
     "contractAddress" in asset &&
-    asset.contractAddress === OPTIMISTIC_ETH.contractAddress
+    asset.contractAddress ===
+      BASE_ASSETS_WITH_COIN_TYPE.find(
+        ({ symbol }) => symbol === OPTIMISTIC_ETH.symbol
+      )?.contractAddress
```
jagodarybacka, line 166: Can we just export these assets so we can avoid `find`? Or if this is not something we want then maybe add new util?

background/services/chain/db.ts
```diff
@@ -180,6 +187,25 @@ export class ChainDatabase extends Dexie {
...
+  async initializeBaseAssets(): Promise<void> {
+    const existingBaseAssets = await this.getAllBaseAssets()
+    await Promise.all(
+      BASE_ASSETS.map(async (defaultAsset) => {
+        if (
+          !existingBaseAssets.some(
+            (asset) => asset.symbol === defaultAsset.symbol
+          )
+        ) {
+          await this.baseAssets.put(defaultAsset)
```
jagodarybacka, line 203: Mapping assets and then doing `bulkPut` would be probably more efficient

background/networks.ts
```diff
@@ -18,7 +18,9 @@ export type NetworkFamily = "EVM" | "BTC"
 
 // Should be structurally compatible with FungibleAsset or much code will
 // likely explode.
-export type NetworkBaseAsset = FungibleAsset &
+export type NetworkBaseAsset = FungibleAsset
```
0xDaedalus, line 21: I think this will cause us quite a lot of headaches down the line.  We need a reliable way to programmatically distinguish base assets from fungible assets - especially in the context of the limited information we get about the asset we get from `wallet_addEthereumChain`

background/networks.ts
```diff
@@ -18,7 +18,9 @@ export type NetworkFamily = "EVM" | "BTC"
...
+
+export type BaseAssetWithCoinType = NetworkBaseAsset &
```
0xDaedalus, line 23: Rather than defining this as a new type - would it make sense to change the `coinType` and `metaData` properties in the existing type to be optional?

kkosiorowska: Let's make these properties optional.

background/constants/currencies-with-coin-types.ts
kkosiorowska: Thanks for catching this!
Optimism uses `ETH` as a base asset. So I remove the unnecessary `OPTIMISTIC_ETH` from the list of base assets.

background/constants/currencies-with-coin-types.ts
kkosiorowska: I reverted to our earlier logic. I don't want the base assets mixed with their extended versions therefore I don't want to export them. Importing these assets to one variable and getting them by name is a good plan. The above variable does just that. However, I didn't notice that we already use something like this. I will take an example from this and create a similar function in a separate file for the base assets.

Let me know what you think about it.

background/redux-slices/0x-swap.ts
kkosiorowska: After making changes, we use our assets again.  They were expanded by the base character. This change is outdated.

background/services/chain/db.ts
kkosiorowska: I changed the logic so that we only execute the `bulkPut` method once. What do you think about this?

hyphenized (changes requested):
Looks great! Left a couple of comments.

I think we can add `chainId` to network base assets so we can identify them without `coinType` and additionally know which network they belong to. There's also the case where we have two base assets with the same symbol but different decimals, though, that might be too much of an edge case 😅

background/constants/currencies.ts
```diff
@@ -99,14 +86,23 @@ export const BTC: NetworkBaseAsset = {
   },
 }
 
-export const BASE_ASSETS = [ETH, BTC, MATIC, RBTC, OPTIMISTIC_ETH, AVAX, BNB]
+export const BASE_ASSETS_WITH_COIN_TYPE = [
```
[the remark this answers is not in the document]
kkosiorowska: I think it's a good idea 😀

kkosiorowska:
> I think we can add `chainId` to network base assets so we can identify them without `coinType` and additionally know which network they belong to. There's also the case where we have two base assets with the same symbol but different decimals, though, that might be too much of an edge case 😅

hmm I think that's a good point.  We are able to get `chainId` from receiving data with `wallet_addEthereumChain` so lets' do it. This may make our work easier in the future.

kkosiorowska:
Ready for another look ✅

background/redux-slices/utils/asset-utils.ts
```diff
@@ -249,3 +249,16 @@ export function heuristicDesiredDecimalsForUnitPrice(
...
+}
+
+export function getBuiltInNetworkBaseAsset(
+  symbol: string,
+  chainID: string
+): (NetworkBaseAsset & Required<CoinGeckoAsset>) | undefined {
+  return BUILT_IN_NETWORK_BASE_ASSETS.find(
+    (asset) => asset.symbol === symbol && asset.chainID === chainID
+  )
+}
```
jagodarybacka, line 264: I feel like `existsInNetworkBaseAssets` `getBuiltInNetworkBaseAsset` and `isBuiltInNetworkBaseAsset`  will be easily confused but idk how to do it better, maybe we don't need to do `some` if we are doing `find` on the same array? 🤔 also, we are sometimes checking just symbol and sometimes chainID too (in `getBuiltInNetworkBaseAsset`)

background/constants/base-assets.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+const BNB: NetworkBaseAsset = {
+  chainID: "56",
+  name: "Binance Coin",
+  symbol: "BNB",
+  decimals: 18,
+}
+
+const BTC: NetworkBaseAsset = {
```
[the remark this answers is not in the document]
kkosiorowska: hmm 🤔  I saved all base assets to indexedDB because didn't want to change or add other logic for them. At this point, this PR transfers base assets to db.

@hyphenized  @0xDaedalus What do you think about whether we need BTC? I'm not sure if in that case, we should remove this asset

0xDaedalus (approved):
Looking great - BTC removal can be addressed in a followup - lets 🚢 !

PR #2841, opened 2023-01-04
Prepare logger to work with Manifest v3

### LINKED ISSUE #2828: Update `logger.ts` to use the async `browser.storage.local` api

aurelsziebert:
Closes #2828 

It now uses the `browser.storage.local` api, while the api of the logger kept sync.

It handles multiple logger instances, so no any log can overwrite an another one.

A suggestion: the entries of the export result are grouped by kind (debug, warn, etc.), and are ordered by time only inside these groups. It can't be read all the logs from the beginning to the end linearly by time. 
It would be nice to have all the log entries ordered regardless its kind, or source.
To achieve this, we should store every single log entries in serialized arrays, instead of storing all of it in finalized, formatted, joined strings. So ordering, and joining could happen at export time. And also regex splitting at `logger.ts@366` could be omitted, so the code would be simpler, and more elegant.

Latest build: [extension-builds-2841](https://github.com/tallyhowallet/extension/suites/10174954713/artifacts/500956359) (as of Fri, 06 Jan 2023 15:03:24 GMT).

### DIFF
--- a/background/lib/logger.ts
+++ b/background/lib/logger.ts
@@ -225,11 +225,13 @@ const logger = () => {
   }
 
   async function clearStorage() {
-    localStorage.removeItem(groupKey(LogLevel.debug))
-    localStorage.removeItem(groupKey(LogLevel.log))
-    localStorage.removeItem(groupKey(LogLevel.info))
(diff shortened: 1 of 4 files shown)


Shadowfiend:

PR #2842, opened 2023-01-04
Handle removing NFTs filters

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2835
Resolves https://github.com/tallyhowallet/extension/issues/2819

### What

- fix wrapping description text
- remove accounts and collections from filters when the account is removed or when the last NFT from the collection was transferred

### Testing
1.
- add two accounts, load their NFTs, check redux state
- remove one of the accounts, check redux state - filters related to this account should be removed
2.
- transfer last NFT from collection to the address that is not imported in the wallet
- check if that collection is no longer visible on the filters list


kkosiorowska (approved):
LGTM ✅

PR #2845, opened 2023-01-05
Replace window global to self in service worker environment

### LINKED ISSUE #2832: Handle window not existing in the service worker context

aurelsziebert:
Closes #2832 

Service workers don't have access to DOM elements like window, so need to use self or globalThis.

I tried to define custom rule in eslintrc, but I didn't find a nice way to exclude one item from the default blacklist. Plus we should allow self only in the SW environment, not in the whole project. This why I decided to turn warnings off locally for the affected lines.

Latest build: [extension-builds-2845](https://github.com/tallyhowallet/extension/suites/10193417355/artifacts/506854480) (as of Wed, 11 Jan 2023 12:26:03 GMT).


Shadowfiend:

PR #2847, opened 2023-01-05
Refactor all local storage usage to async

### LINKED ISSUE #2830: Refactor all `localStorage` usage into `browser.storage.local`

aurelsziebert:
Closes #2830 

Several problems emerged. This PR is not ready yet, I put the issues into comments above on the Changes section.

Latest build: [extension-builds-2847](https://github.com/tallyhowallet/extension/suites/10196672021/artifacts/506855827) (as of Wed, 11 Jan 2023 12:27:06 GMT).

### DIFF
--- a/background/features.ts
+++ b/background/features.ts
@@ -74,6 +74,7 @@ export const isEnabled = (flagName: FeatureFlagType): boolean => {
   }
 
   if (BuildTimeFlag.SWITCH_RUNTIME_FLAGS) {
+    // FIXME use browser.storage.local instead of localStorage
     const state = localStorage.getItem(flagName)
     return state !== null ? state === "true" : RuntimeFlag[flagName]
   }
--- a/ui/_locales/i18n.ts
+++ b/ui/_locales/i18n.ts
@@ -3,7 +3,6 @@ import { initReactI18next, TFuncKey, TFuncReturn } from "react-i18next"
 import { browser } from "@tallyho/tally-background"
 import _ from "lodash"
 import { resources } from "./index"
-import { setLocalStorageItem } from "../hooks"
 
 const LANGUAGE_KEY = "lang"
 const DEFAULT_LANGUAGE = "en"
@@ -34,9 +33,8 @@ i18n
 
 const getLanguage = (): string => i18n.language
 const setLanguage = (lang: string): void => {
-  i18n
-    .changeLanguage(lang)
-    .then(() => lang && setLocalStorageItem(LANGUAGE_KEY, lang))
+  i18n.changeLanguage(lang)
+  browser.storage.local.set({ [LANGUAGE_KEY]: lang })
 }
 
 declare module "react-i18next" {
--- a/ui/_locales/i18n.ts
+++ b/ui/_locales/i18n.ts
@@ -1,7 +1,9 @@
 import i18n from "i18next"
 import { initReactI18next, TFuncKey, TFuncReturn } from "react-i18next"
+import { browser } from "@tallyho/tally-background"
+import _ from "lodash"
 import { resources } from "./index"
-import { getLocalStorageItem, setLocalStorageItem } from "../hooks"
+import { setLocalStorageItem } from "../hooks"
 
 const LANGUAGE_KEY = "lang"
 const DEFAULT_LANGUAGE = "en"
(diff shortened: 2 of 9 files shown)


background/features.ts
```diff
@@ -74,6 +74,7 @@ export const isEnabled = (flagName: FeatureFlagType): boolean => {
   }
 
   if (BuildTimeFlag.SWITCH_RUNTIME_FLAGS) {
+    // FIXME use browser.storage.local instead of localStorage
```
aurelsziebert, line 77: The whole isEnabled function should be turned async, if we want to keep supporting this feature. 
This will cause lots of headache at the usages.

ui/_locales/i18n.ts
```diff
@@ -20,12 +21,20 @@ i18n
...
   })
+;(async () => {
+  const storedLang = _.get(
+    await browser.storage.local.get(LANGUAGE_KEY),
+    LANGUAGE_KEY
+  )
+  if (storedLang && storedLang !== DEFAULT_LANGUAGE) {
+    i18n.changeLanguage(storedLang)
+  }
+})()
```
aurelsziebert, line 32: I am not sure if the ui updates automatically if the language changes here. 
If so, then maybe it is an acceptable solution, because this code will probably run very fast.

Anyway this is not an urgent problem since we have only one language.

ui/components/HiddenDevPanel/FeatureFlagsPanel.tsx
```diff
@@ -16,10 +15,15 @@ function FeatureFlag(props: { name: RuntimeFlagType }): ReactElement {
...
+  const [isSaving, setIsSaving] = useState(false)
   const [value, setValue] = useState(isEnabled(FeatureFlags[name]))
 
   useEffect(() => {
-    setLocalStorageItem(name, String(value))
+    ;(async () => {
+      setIsSaving(true)
+      await browser.storage.local.set({ [name]: String(value) })
+      setIsSaving(false)
+    })()
```
aurelsziebert, line 26: This is a nice example how can we handle the async local storage.

It would be great to use the useLocalStorage hook here, but I am afraid that it is impossible to adapt it to use the asyc api.

ui/hooks/dom-hooks.ts
```diff
@@ -115,24 +116,14 @@ export const useDebounce = <T>(initial: T, wait = 300): [T, (v: T) => void] => {
...
-  defaultValue: string
-): string => localStorage.getItem(key) || defaultValue
-
 export function useLocalStorage(
   key: string,
   initialValue: string
 ): [string, React.Dispatch<React.SetStateAction<string>>] {
-  const [value, setValue] = useState(() => {
-    return getLocalStorageItem(key, initialValue)
-  })
```
aurelsziebert, line 132: It is not possible to give the stored value as default.

ui/hooks/dom-hooks.ts
```diff
@@ -115,24 +116,14 @@ export const useDebounce = <T>(initial: T, wait = 300): [T, (v: T) => void] => {
...
+  const [value, setValue] = useState(initialValue)
 
   useEffect(() => {
-    setLocalStorageItem(key, value)
+    browser.storage.local.set({ [key]: value })
```
aurelsziebert, line 126: Saving should be async but it can't be, because we have to keep the returned setValue function sync.

ui/hooks/dom-hooks.ts
```diff
@@ -115,24 +116,14 @@ export const useDebounce = <T>(initial: T, wait = 300): [T, (v: T) => void] => {
...
 
-export const setLocalStorageItem = (key: string, value: string): void =>
-  localStorage.setItem(key, value)
-
-export const getLocalStorageItem = (
-  key: string,
-  defaultValue: string
-): string => localStorage.getItem(key) || defaultValue
-
 export function useLocalStorage(
```
aurelsziebert, line 119: I am afraid, that useLocalStorage hook can't work with the async local storage api.

There is a workaround here with good ideas: https://stackoverflow.com/questions/63432520/useasyncstorage-custom-hook-with-usestate-hook-in-react-native but still no solution for the problems wrote below.

Shadowfiend:

PR #2856, opened 2023-01-10
Sort NFTs within one collection according to selected order

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2849

### What

Selecting order in the NFTs setting should be sorting collections and NFTs within the collection. This should apply to the collections list and badges list as well.

For NFTs owned in a given collection only "Newest added" and "Oldest added" order make sense, as the rest is applied on the collection level. 

### Testing

- test "Newest added" and "Oldest added" sort options on the NFTs page - they should be sorting collections and NFTs inside collections as well. Check if badges are also sorted (within one collection - badges are grouped by collections but they are not collapsed, this can be improved later).

Latest build: [extension-builds-2856](https://github.com/tallyhowallet/extension/suites/10306008307/artifacts/506767568) (as of Wed, 11 Jan 2023 11:17:33 GMT).


background/redux-slices/utils/nfts_update.ts
```diff
@@ -127,5 +158,6 @@ export const getFilteredCollections = (
         isEnabledFilter(collection.owner, filters.accounts)
     )
     .sort((collection1, collection2) =>
-      sortNFTS(collection1, collection2, filters.type)
+      sortCollections(collection1, collection2, filters.type)
     )
+    .map((collection) => sortNFTs(collection, filters.type))
```
kkosiorowska, line 163: Sorting collections works by sorting nfts. I have a feeling that we are doing it twice. One time for collections the other time for nfts. 

I think the `map` function should happen before `sort`. Then we get ordered nfts. We will be able to take the first element from the array to sort the collection. Correct me if I'm wrong.

kkosiorowska (commented):
Looks great ✨! One small comment.

kkosiorowska (approved):
✅

PR #2859, opened 2023-01-11
Remove unused `node-fetch`

greg-nagy:
We don't use node-fetch anywhere I know of. So intead of upgrading in #2851 let's remove.

Latest build: [extension-builds-2859](https://github.com/tallyhowallet/extension/suites/10308037276/artifacts/506905783) (as of Wed, 11 Jan 2023 13:05:24 GMT).


Shadowfiend (approved):
Looks like this was originally introduced as part of very early setup work and just didn't get cleaned up. Let's do it.

PR #2873, opened 2023-01-16
Align Portfolio page NFTs stats with NFTs page stats

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2866

### What
Apply filters from NFTs page to Portfolio stats about NFTs, add warning and change copy
image

### DIFF
--- a/ui/components/NFTS_update/Filters/NFTsFilters.tsx
+++ b/ui/components/NFTS_update/Filters/NFTsFilters.tsx
@@ -109,6 +109,7 @@ export default function NFTsFilters(): ReactElement {
   return (
     <SharedSlideUpMenuPanel header={t("title")}>
       <div className="filters">
+        <span className="simple_text filter_warning">{t("warning")}</span>
         <div>
           <span className="simple_text filter_title">{t("sortTypeTitle")}</span>
           {RADIO_BTNS.map(({ value, label }) => (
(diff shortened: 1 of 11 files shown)


ui/components/NFTS_update/Filters/NFTsFilters.tsx
```diff
@@ -153,13 +154,18 @@ export default function NFTsFilters(): ReactElement {
...
         }
         .filter_title {
           display: inline-block;
           margin-bottom: 4px;
           width: 100%;
         }
+        .filter_warning {
+          font-size: 16px;
+          line-height: 24px;
+          color: var(--green-20);
```
kkosiorowska, line 167: These properties are already in the `simple_text`  class. I think in this class we can only set the color.
```suggestion
          color: var(--green-20);
```

ui/components/Shared/SharedSlideUpMenuPanel.tsx
```diff
@@ -31,8 +31,6 @@ export default function SharedSlideUpMenuPanel({
           padding-top: 25px;
           margin-top: -25px;
           z-index: 1;
-          border-bottom: 1px solid var(--green-120);
```
kkosiorowska, line 34: hmm this is the same component for displaying the accounts list. Are we sure we don't want that border bottom for the filters? I remember talking to Vlad about this. Just to make sure. @VladUXUI , could you confirm?

Screenshot 2023-01-16 at 17 45 18
Screenshot 2023-01-16 at 17 45 35

kkosiorowska (commented):
Look great ✨ Only tiny comments.

ui/components/Shared/SharedSlideUpMenuPanel.tsx
VladUXUI: Thanks for bringing up @kkosiorowska . Let's go with no border for now, i'll make it a bit more clear what the thinking is here when doing the UI audit

kkosiorowska (approved):
LGTM ✅

PR #2879, opened 2023-01-17
Merge multiple requests for NFTs

jagodarybacka:
Requests for multiple collections that are happening quickly can be merged into one.

Latest build: [extension-builds-2879](https://github.com/tallyhowallet/extension/suites/10419272621/artifacts/514952385) (as of Tue, 17 Jan 2023 15:47:28 GMT).

### DIFF
--- a/background/lib/nfts_update.ts
+++ b/background/lib/nfts_update.ts
@@ -34,7 +34,8 @@ function groupChainsByAddress(accounts: AddressOnNetwork[]) {
 
 export function getNFTs(
   accounts: AddressOnNetwork[],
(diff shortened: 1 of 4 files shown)


Shadowfiend:
One thing that isn't clear to me here: why not bulk fetch even if there is more than one page?

jagodarybacka:
I think this PR is stale and can be closed, we are nowhere near hitting our quota on SH so we don't need to merge requests right now.

jagodarybacka:
Stale and no longer a priority, let's close it

PR #2880, opened 2023-01-17
Add custom networks via EIP-3085

0xDaedalus:
Closes https://github.com/tallyhowallet/extension/issues/2762

This PR Adds the ability for the extension to persist information about a network added via `wallet_addEthereumChain` that can later be used to connect to that network.

### To Test
- [ ] Set `SUPPORT_CUSTOM_NETWORKS=true`
- [ ] Navigate to https://chainlist.org/?search=fantom
- [ ] Connect, and click `Add to Wallet` to add Fantom to the wallet (there will be no visual confirmation -- yet)
- [ ] Open indexedDB - you should see the information for the added `Fantom` network inside of the `network` table as well as the corresponding RPCUrls inside of the `rpcUrls` table.

Latest build: [extension-builds-2880](https://github.com/tallyhowallet/extension/suites/10467167727/artifacts/518386010) (as of Thu, 19 Jan 2023 15:08:17 GMT).


background/services/chain/index.ts
```diff
@@ -348,8 +351,13 @@ export default class ChainService extends BaseService<Events> {
     await this.db.initializeBaseAssets()
   }
 
+  async initializeRPCs(): Promise<void> {
+    await this.db.initializeRPCs()
```
hyphenized, line 355: We can move these `this.db.initialize` calls to a new `initialize` method in the db and not have to worry about call order, plus it would allow us to remove the shallow methods from the service.

hyphenized (changes requested):
This is looking great! Left some minor comments

background/services/chain/db.ts
```diff
@@ -183,14 +195,63 @@ export class ChainDatabase extends Dexie {
...
   }
 
+  async addEVMNetwork(
+    chainName: string,
+    chainID: string,
+    decimals: number,
+    symbol: string,
+    assetName: string,
+    rpcUrls: string[]
+  ): Promise<void> {
```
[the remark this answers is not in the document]
0xDaedalus: Unfortunately based on my testing it is quite rare to have icon urls sent in via `wallet_addEthereumChain` (e.g. chainList does not populate icons for any of the chains I've tried to add for testing)

background/services/chain/index.ts
0xDaedalus: love it

hyphenized (approved):
Awesome work, let's 🚢 it!

PR #2884, opened 2023-01-18
Set removed NFTs collections as stale

jagodarybacka:
Set removed NFTs collections as stale so they can be refetched if needed 
### Testing
- add account
- check NFTs page, let collections load
- remove account and add it again
- check if collections are loading correctly

Latest build: [extension-builds-2884](https://github.com/tallyhowallet/extension/suites/10436610272/artifacts/516229618) (as of Wed, 18 Jan 2023 10:26:08 GMT).


kkosiorowska (approved):
Works good ✅

PR #2885, opened 2023-01-18
Calculate total NFTs floor price including assets other than ETH

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2775

### What
- The total floor price should sum all prices we know about, if they are using different currencies for the floor price let's calculate its USD and ETH values and add them to the total sum.
- Let's sort NFTs by price regardless of which currency is used
- Minor UI improvements: showing price tag on badges list, fix for radio button cursor

### Testing
1. Activate Avalanche and BSC chains for wallets that have some NFTs on these chains
2. Check if total floor price in the header of NFTs page includes prices of collections with floor prices in AVAX or BNB
3. Sort collections by price - should sort comparing USD value of currencies

Latest build: [extension-builds-2885](https://github.com/tallyhowallet/extension/suites/10463880816/artifacts/518165647) (as of Thu, 19 Jan 2023 12:44:15 GMT).


0xDaedalus (changes requested):
Have yet to take this for a ride but code is looking good.

I do think there is a gap in testing here.  At the very least we should write some unit tests for `getTotalFloorPrice` and `enrichCollectionWithUSDFloorPrice`.  A component test that asserts the correct prices are rendering would be nice to have as well.

jagodarybacka:
@0xDaedalus yeah I agree, but I think it shouldn't block manual testing - if your manual tests on this PR will be fine can we merge it and I can add unit tests tomorrow?

0xDaedalus:
Sorry but going to have to insist that we block this until tests are in - we can create a separate QA build if needed but I'm really going to start doubling down on requiring tests for PRs as we really should not be treating them as a separate type of dev work.

jagodarybacka:
@0xDaedalus utils' tests added, for UI components there is a separate task already created https://github.com/tallyhowallet/extension/issues/2777

0xDaedalus (approved):
Fantastic!

PR #2898, opened 2023-01-20
Save chain selected by dapp in the database correctly

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2894

### What

Uniswap had a problem with switching chains because the dapp was first requesting to switch the chain and then it was checking if the chain was really switched with `eth_chainId`. We were saving selected chains for dapps incorrectly in the database - with origin equal to extension URL not dapp URL. Now extension should be responding with correct chainID and it should allow Uniswap to accept switched chain.

Latest build: [extension-builds-2898](https://github.com/tallyhowallet/extension/suites/10488088350/artifacts/519969178) (as of Fri, 20 Jan 2023 13:29:58 GMT).


0xDaedalus (approved):
Fantastic find!

PR #2899, opened 2023-01-23
Report a spam for an ability

### LINKED ISSUE #2864: Report spam for ability
- [ ] update abilities removal window

Screenshot 2023-01-13 at 14 25 29

- [ ]  open slideup on click
- [ ] add reasons as radio buttons list

Screenshot 2023-01-13 at 14 28 19

Screenshot 2023-01-13 at 14 28 13

kkosiorowska:
Closes #2864

## Testing Env
```
SUPPORT_ABILITIES=true
```


background/services/abilities/index.ts
```diff
@@ -200,4 +203,13 @@ export default class AbilitiesService extends BaseService<Events> {
...
     }
   }
+
+  // eslint-disable-next-line class-methods-use-this
+  async reportSpam(
+    address: NormalizedEVMAddress,
+    abilitySlug: string,
+    reason: string
+  ): Promise<void> {
+    await createSpamReport(address, abilitySlug, reason)
```
jagodarybacka, line 213: So is it not doing anything to our database or service? I see it is being removed from react as the next step but maybe we should check if spam was actually reported before removing this item?

background/services/abilities/index.ts
```diff
@@ -204,12 +206,17 @@ export default class AbilitiesService extends BaseService<Events> {
...
     abilitySlug: string,
+    abilityId: string,
     reason: string
   ): Promise<void> {
-    await createSpamReport(address, abilitySlug, reason)
+    const result = await createSpamReport(address, abilitySlug, reason)
+    if (result) {
+      this.emitter.emit("removeAbility", { address, abilityId })
+    } else {
+      await this.emitter.emit("reportSpamFailure")
```
0xDaedalus, line 219: I don't think we want an `if/else` here.  We should always remove the ability regardless of if the spam report was successful or not.  In fact I'd argue that there is close to no value to alerting the user that the spam report was unsuccessful since there is nothing they can do about a failure anyway.  Our primary concern here should be getting he spam the user doesn't want out of their face - regardless of if the report was successful or not.

background/main.ts
```diff
@@ -1516,6 +1520,26 @@ export default class Main extends BaseService<never> {
...
       this.store.dispatch(addAbilities(newAbilities))
     })
+    this.abilitiesService.emitter.on(
+      "removeAbility",
+      async ({ address, abilityId }) => {
+        this.store.dispatch(removeAbility({ address, abilityId }))
+      }
+    )
+    this.abilitiesService.emitter.on("reportSpamFailure", () => {
+      this.store.dispatch(setSnackbarMessage("Couldn't report spam"))
```
jagodarybacka, line 1530: It won't be translated 😞

jagodarybacka (dismissed):
I think these are not blockers but up to you if you want to fix them here or in the next PR.

- If I only have one ability then styles are broken (padding at the bottom):
image

kkosiorowska:
> * If I only have one ability then styles are broken (padding at the bottom):

I corrected it in this PR.

kkosiorowska:
> So is it not doing anything to our database or service? I see it is being removed from react as the next step but maybe we should check if spam was actually reported before removing this item?

After thinking about it, I agree with Alex.  The main goal of the user is to remove the ability. We shouldn't block it when reporting spam fails. I reverted to the previous solution.

background/services/abilities/index.ts
0xDaedalus: I think we need to `markAsRemoved` in here as well or else `getActiveAbilities` will return abilities previously marked as spam.

This also suggests to me that we're using redux as the persistent source of truth here rather than as a temporary one - we probably don't want to do that - although this can be addressed in a followup issue.

kkosiorowska: Please check now. I created a separate [issue](https://github.com/tallyhowallet/extension/issues/2934) for the next steps.

ui/_locales/en/messages.json
```diff
@@ -777,7 +777,24 @@
     "timeCloses": "Closes in 1 month",
     "viewWebsiteBtn": "Visit website",
     "markBtn": "Mark as Completed",
-    "deleteBtn": "Delete"
+    "deleteBtn": "Delete",
+    "deleteSlideUpMenu": {
+      "title": "Delete ability?",
+      "desc": "If you delete an ability you won't be able to see it anymore",
+      "spamPrompt": "Is this ability a spam?",
```
0xDaedalus, line 784:
```suggestion
      "spamPrompt": "Is this ability spam?",
```

0xDaedalus (approved):
wspaniały 🎉 🎉

PR #2902, opened 2023-01-23
Don't display prices for untrusted assets.

0xDaedalus:
Closes https://github.com/tallyhowallet/extension/issues/2904

To avoid erroneously showing prices of untrusted assets (and thus making them appear to be legitimate) we should not show the user prices for those assets.

Before:
image

After:
image

### To Test
- [ ] Import the test wallet
- [ ] Switch to Polygon
- [ ] The scam asset `AAVE` should be marked as untrusted and should not display a price

Latest build: [extension-builds-2902](https://github.com/tallyhowallet/extension/suites/10533600698/artifacts/523529251) (as of Mon, 23 Jan 2023 18:26:35 GMT).


hyphenized (approved):
Works nice! 🚀

PR #2907, opened 2023-01-23
Properly use isEnabled to gate abilities polling based on feature flag.

### LINKED ISSUE #2908: Polling for Abilities while the feature is disabled
The feature flag is not being checked properly here:

https://github.com/tallyhowallet/extension/blob/f307fefba52a569fc86d82cb4287be80424ae7fa/background/services/abilities/index.ts#L147

image

0xDaedalus:
Closes #2908 

The previous implementation does not allow for runtime switches.  The pattern here matches how we check for feature flags in all other instances.

Latest build: [extension-builds-2907](https://github.com/tallyhowallet/extension/suites/10536438251/artifacts/523740197) (as of Mon, 23 Jan 2023 20:53:32 GMT).

### DIFF
--- a/background/services/abilities/index.ts
+++ b/background/services/abilities/index.ts
@@ -8,7 +8,7 @@ import {
 } from "./daylight"
 import { AbilitiesDatabase, getOrCreateDB } from "./db"
 import ChainService from "../chain"
-import { FeatureFlags } from "../../features"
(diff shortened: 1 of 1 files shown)


hyphenized (approved):
🚢 🚢 🚢

PR #2926, opened 2023-01-26
Filter abilities

### LINKED ISSUE #2861: Create Filter Abilities selector
- [ ] add filters icon to the header
Screenshot 2023-01-13 at 13 47 05
- [ ]  open slideup on click
- [ ]  add show abilities radio buttons list
- [ ]  save filters in redux slice

### LINKED ISSUE #2864: Report spam for ability

kkosiorowska:
Closes #2861

This PR adds filters for abilities. Several improvements have been made to avoid duplicating code from existing nfts filters. Changes:

- Make `FilterListItem` component for nfts more useable. Change to `SharedToggleItem`.
- Create a dark theme for to  `SharedTooltip`
- Create const UI value for ability type color - `ABILITY_TYPE_COLOR`
- Move ability types from services to separate file
- Fetch more abilities, including all expired (closed) abilities.
- Allow users to filter abilities. Filtering should also affect notifications (green dot + number on banner). Filter by ability state, type and account address. 
- Add tests for filtering abilities

## UI

https://user-images.githubusercontent.com/23117945/214929562-10458bbf-19c2-4fbb-a94b-e23cc4ff6747.mov

https://user-images.githubusercontent.com/23117945/214929653-619321a5-65fe-48aa-8e91-1b23b56879f4.mov

## To Test
Check that all filter works properly.
- [ ] filter by accounts 
- [ ] filter by ability state 
- [ ] filter by ability type 

## Testing Env
```
SUPPORT_ABILITIES=true
```

Latest build: [extension-builds-2926](https://github.com/tallyhowallet/extension/suites/10667657982/artifacts/533401431) (as of Mon, 30 Jan 2023 17:35:27 GMT).


background/abilities.ts
```diff
@@ -0,0 +1,65 @@
+import { NormalizedEVMAddress } from "./types"
```
0xDaedalus, line 1: I'm not sure if the stuff in here feels like it should have its own file.  We're moving away from using `types.ts` files in the repository and this kind of just feels like a renamed `types.ts` file with a few constants in it.  Also I don't know that it deserves to be at the top level of the `background` folder.

background/redux-slices/abilities.ts
```diff
@@ -14,8 +25,17 @@ type AbilitiesState = {
   hideDescription: boolean
 }
 
+const typesFilter: FilterType[] = ABILITY_TYPES.map((type) => ({
```
0xDaedalus, line 28: I think a different approach here could be having a `Set` and filtering based on if a given type is included in the set or not inside of a selector.

This would be a bit cleaner (don't need to worry about the `isEnabled` property) and would let us check if a given ability should be filtered out or not in O(1)

background/redux-slices/abilities.ts
```diff
@@ -14,8 +25,17 @@ type AbilitiesState = {
...
+  type,
+  isEnabled: true,
+}))
+
 const initialState: AbilitiesState = {
-  filter: "incomplete",
+  filters: {
+    state: "open",
+    types: typesFilter,
+    accounts: [],
```
0xDaedalus, line 37: `Set` here as well

background/redux-slices/abilities.ts
```diff
@@ -14,8 +25,17 @@ type AbilitiesState = {
...
 }
 
+const typesFilter: FilterType[] = ABILITY_TYPES.map((type) => ({
+  type,
+  isEnabled: true,
+}))
+
 const initialState: AbilitiesState = {
-  filter: "incomplete",
+  filters: {
```
0xDaedalus, line 34: Conceptually I would argue that we only have _one_ filter - and by changing the properties of `state`, `types`, etc... we are changing the properties of that single filter

background/redux-slices/abilities.ts
```diff
@@ -55,6 +75,45 @@ const abilitiesSlice = createSlice({
...
+      { payload: filter }: { payload: FilterType }
+    ) => {
+      const idx = immerState.filters.types.findIndex(
+        ({ type }) => type === filter.type
+      )
+      immerState.filters.types[idx] = filter
+    },
+    updateFilterAccount: (
+      immerState,
+      { payload: filter }: { payload: FilterAccount }
```
0xDaedalus, line 95: I think that we could get some increased clarity here if we named this `Account` rather than `FilterAccount`.  Same with `FilterType`.

background/redux-slices/utils/account-filter-utils.ts
```diff
@@ -0,0 +1,39 @@
+export type FilterAccount = {
+  id: string
+  isEnabled: boolean
+  name: string
+  thumbnailURL?: string
```
0xDaedalus, line 5: Why are we naming and giving thumbnails to filters?

background/redux-slices/selectors/abilitiesSelectors.ts
```diff
@@ -1,42 +1,60 @@
...
+} from "../utils/account-filter-utils"
+import { selectAccountTotals } from "./accountsSelectors"
 
-const selectAbilities = (state: RootState) => state.abilities.abilities
+const selectAbilities = createSelector(
+  (state: RootState) => state.abilities,
+  (abilitiesSlice) => abilitiesSlice.abilities
+)
+
+export const selectHideDescription = createSelector(
```
0xDaedalus, line 16:
```suggestion
export const selectDescriptionHidden = createSelector(
```

background/redux-slices/utils/abilities.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const filterByAddress = (
+  address: string,
+  accounts: FilterAccount[]
+): boolean => {
+  return !!accounts.find((filter) => filter.id === address)?.isEnabled
+}
+// eslint-disable-next-line import/prefer-default-export
+export const filterAbility = (
```
0xDaedalus, line 34: Love this approach ❤️

ui/pages/Abilities.tsx
```diff
@@ -1,22 +1,54 @@
...
+            <SharedTooltip
+              horizontalPosition="left"
+              width={36}
+              verticalPosition="bottom"
+              type="dark"
+              IconComponent={() => (
+                <SharedIcon
+                  width={24}
+                  icon="toggle.svg"
+                  ariaLabel={t("filters.title")}
```
0xDaedalus, line 42: After hovering over this icon, clicking in, and then clicking out - the tooltip sticks.

https://user-images.githubusercontent.com/94649004/214967003-d90757f4-23aa-4a8b-be5d-11d24a4b09bf.mov



Not necessary to fix in this PR

background/redux-slices/abilities.ts
```diff
@@ -1,11 +1,22 @@
 import { createSlice } from "@reduxjs/toolkit"
-import { Ability } from "../services/abilities"
+import { Ability, AbilityType, ABILITY_TYPES } from "../abilities"
 import { HexString, NormalizedEVMAddress } from "../types"
 import { setSnackbarMessage } from "./ui"
 import { createBackgroundAsyncThunk } from "./utils"
+import { FilterAccount } from "./utils/account-filter-utils"
+
+export type AbilityState = "open" | "closed" | "expired" | "deleted" | "all"
```
0xDaedalus, line 8: I believe that `closed` should be `completed`

background/redux-slices/utils/abilities.ts
```diff
@@ -0,0 +1,43 @@
+import { Ability, AbilityType } from "../../abilities"
```
0xDaedalus, line 1: We 100% need some tests around filtering abilities

background/services/abilities/index.ts
```diff
@@ -83,37 +47,31 @@ const normalizeDaylightAbilities = (
...
-        completed: false,
-        removedFromUi: false,
-        address: normalizeEVMAddress(address),
-        requirement: normalizeDaylightRequirements(
-          // Just take the 1st requirement for now
-          daylightAbility.requirements[0]
-        ),
-      })
-    }
+    normalizedAbilities.push({
```
0xDaedalus, line 50: @VladUXUI Can you confirm that we want the user to be able to see all ability types?  I thought we had discussed only showing a subset of types that we thought were most likely to be relevant

background/abilities.ts
```diff
@@ -0,0 +1,65 @@
...
+
+// https://docs.daylight.xyz/reference/ability-model#ability-types
+export const ABILITY_TYPES = [
+  "claim",
+  "airdrop",
+  "mint",
+  "access",
+  "product",
+  "event",
+  "article",
```
0xDaedalus, line 31: I am 90% sure we don't want to show articles but will double check with vlad

0xDaedalus (changes requested):
Tested and the filtering works great!  I would however like to suggest an alternative implementation of the filters that I think might simplify quite a bit of the logic around getting and filtering abilities.

I think we can represent `type`, `accounts`, and `status` as sets of strings which we persist in redux (think of the "filter state" as the union of these sets).  If we do this - then our "filters" can simply be selectors acting on this "filter state" - meaning our `selectAbilities` selector can be composed of 3 discrete selectors (one for each kind of filter) and we don't need to worry about all of the messy object spread stuff I introduced inside of `selectFilteredAbilities`.  

We also get the added benefit of not needing to worry about the `isEnabled` property filters currently have since we can infer if a filter is enabled by its presence or absence in a given `Set`, or worrying about using `findIndex` since we could use `Set.add` and `Set.delete` (immer should take care of making those operations immutable for us).

ui/pages/Abilities.tsx
kkosiorowska: Yes, the fix is in this https://github.com/tallyhowallet/extension/pull/2899. And it should be merged after it.

background/abilities.ts
kkosiorowska: I wanted to organize the structure. I checked where other services hold types. Types for other services are also at this level like for nfts. I understand that this is the highest level because we refer to it from the services or lib directory. However, I'm not sure where they should be located exactly. Do you have any idea where to put them?

background/services/abilities/index.ts
VladUXUI: > @VladUXUI Can you confirm that we want the user to be able to see all ability types? I thought we had discussed only showing a subset of types that we thought were most likely to be relevant

We discussed having some toggled off by default, but they should all be there.
The ones that should be active by default are Mint,Airdrop,Vote,Access. The rest should be toggled off and user can toggle them on if he wants too

kkosiorowska: I will change it.

kkosiorowska:
> I think we can represent `type`, `accounts`, and `status` as sets of strings which we persist in redux (think of the "filter state" as the union of these sets). If we do this - then our "filters" can simply be selectors acting on this "filter state" - meaning our `selectAbilities` selector can be composed of 3 discrete selectors (one for each kind of filter) and we don't need to worry about all of the messy object spread stuff I introduced inside of `selectFilteredAbilities`.
> 
> We also get the added benefit of not needing to worry about the `isEnabled` property filters currently have since we can infer if a filter is enabled by its presence or absence in a given `Set`, or worrying about using `findIndex` since we could use `Set.add` and `Set.delete` (immer should take care of making those operations immutable for us).

It sounds great. In this approach, there is a problem with non-serializable values. It is possible that this can be fixed however I prefer to make sure that this is definitely what we want to do.

Screenshot 2023-01-27 at 12 39 18

https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

kkosiorowska:
> I think we can represent `type`, `accounts`, and `status` as sets of strings which we persist in redux (think of the "filter state" as the union of these sets). If we do this - then our "filters" can simply be selectors acting on this "filter state" - meaning our `selectAbilities` selector can be composed of 3 discrete selectors (one for each kind of filter) and we don't need to worry about all of the messy object spread stuff I introduced inside of `selectFilteredAbilities`.
> 
> We also get the added benefit of not needing to worry about the `isEnabled` property filters currently have since we can infer if a filter is enabled by its presence or absence in a given `Set`, or worrying about using `findIndex` since we could use `Set.add` and `Set.delete` (immer should take care of making those operations immutable for us).

It sounds great but in this approach, there is a problem with non-serializable values.  Redux best practices don’t recommend this. 

https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

0xDaedalus:
> Redux best practices don’t recommend this.

Ah - fantastic point and one that completely slipped by mind!  I still think that even if we use Array<string> we can get the benefits of:
- Getting rid of the the `isEnabled` property
- Clearer filter composition via multiple selectors
- Not needing the `FilterAccount` type, and simplifying the `FilterType` type into a string enum.

kkosiorowska:
Ready for another look ✅

0xDaedalus (approved):
📈 🎉 💯

PR #2927, opened 2023-01-26
Hide untrusted assets with hide dust option

jagodarybacka:
### What
To make the spam tokens less visible it let's allow to hide them with `Hide asset balances under $2` settings option

https://user-images.githubusercontent.com/20949277/214821734-b2f3d649-f65c-481c-87eb-22468ca0ff56.mov

Latest build: [extension-builds-2927](https://github.com/tallyhowallet/extension/suites/10595543874/artifacts/528090668) (as of Thu, 26 Jan 2023 11:14:13 GMT).


kkosiorowska (approved):
Works well! ✅ It's a hotfix. It would be good to create a separate issue for the second step, implement trust/hide.

jagodarybacka:
@kkosiorowska https://github.com/tallyhowallet/extension/issues/1021 we already have this one, will need some touchup as it's so old

PR #2933, opened 2023-01-27
Prevent price point best effort matching for untrusted assets

hyphenized:
Do not fall back to best effort matches when retrieving price points for untrusted assets

### Before
image

### After
image

## To Test
- [ ] Load an account, check balances in the main and the overview page
- [ ] Smart contract assets that are discovered from token transfers should not display price data unless there's information for that  exact contract address

Latest build: [extension-builds-2933](https://github.com/tallyhowallet/extension/suites/10657658900/artifacts/532691947) (as of Mon, 30 Jan 2023 10:12:37 GMT).

### DIFF
--- a/background/redux-slices/assets.ts
+++ b/background/redux-slices/assets.ts
@@ -223,6 +223,11 @@ export const selectAssetPricePoint = createSelector(
           asset.homeNetwork.chainID === assetToFind.homeNetwork.chainID &&
           hasRecentPriceData(asset)
       )
+
+      /* Don't do anything else if this is an untrusted asset and there's no exact match */
+      if ((assetToFind.metadata?.tokenLists.length ?? 0) < 1) {
+        return undefined
+      }
     }
 
     /* Otherwise, find a best-effort match by looking for assets with the same symbol  */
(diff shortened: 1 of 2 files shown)


jagodarybacka (commented):
That's better 😄 I can't reproduce this bug today 😡 I think it will fix the total balance issue, but am I right that without "hide dust" we will be showing untrusted assets with hidden prices but their prices will be added to the total balance?

0xDaedalus:
I think a stronger solution here might be to not fetch the asset prices in the first place - that prevents this from turning into a whack-a-mole problem where we are constantly hunting for where in the UI we want to hide untrusted asset prices.  If we adopt the mindset that if an asset is untrusted we don't want the application to have _any_ price data about them - I think that should simplify this problem and address the concern in @jagodarybacka's comment above.

jagodarybacka:
Minor fix added for base assets that were treated as untrusted, let's see if everything will be ok now as this is already an improvement and we can iterate later with a more advanced solution.

kkosiorowska (approved):
Works good ✅

PR #2947, opened 2023-01-31
Prevent fetching prices for custom assets

hyphenized:
Revised the indexing service balance alarm so that it does not requests CoinGecko token prices for custom assets.

Custom assets price points might still be cached in Redux so we might need an additional migration to remove them and revert the selector changes introduced in #2933

To Test
----
- [ ] Reinstall the wallet
- [ ] Load an account with custom assets, check the background network requests
- [ ] There should be no request to CoinGecko for custom assets contract addresses

Latest build: [extension-builds-2947](https://github.com/tallyhowallet/extension/suites/10760897294/artifacts/540296223) (as of Fri, 03 Feb 2023 12:20:31 GMT).


background/lib/prices.ts
```diff
@@ -98,6 +98,10 @@ export async function getTokenPrices(
 ): Promise<{
   [contractAddress: string]: UnitPricePoint<FungibleAsset>
 }> {
+  if (tokenAddresses.length < 1) {
```
0xDaedalus, line 101: 📈

background/services/indexing/db.ts
```diff
@@ -222,6 +222,13 @@ export class IndexingDatabase extends Dexie {
         .toCollection()
         .modify(normalizeAssetAddress)
     })
+
+    this.version(5).stores({
```
0xDaedalus, line 226: I think we need a comment here about what exactly the delta is - quite difficult to see just my looking at the code

background/services/indexing/db.ts
```diff
@@ -280,16 +287,18 @@ export class IndexingDatabase extends Dexie {
   }
 
   async getCustomAssetsByNetwork(
-    network: Network
+    networks: EVMNetwork | EVMNetwork[]
```
0xDaedalus, line 290: Lets just make this `EVMNetwork[]` - we can pass in an array of 1 network when we need to only query 1 - and avoid the `Array.isArray` stuff on line 292

background/services/indexing/index.ts
```diff
@@ -177,6 +176,8 @@ export default class IndexingService extends BaseService<Events> {
     const tokenListLoad = this.fetchAndCacheTokenLists()
 
     this.chainService.emitter.once("serviceStarted").then(async () => {
+      this.handlePriceAlarm()
```
0xDaedalus, line 179: 📈

background/services/indexing/index.ts
```diff
@@ -784,9 +823,17 @@ export default class IndexingService extends BaseService<Events> {
...
 
-  private async handleBalanceAlarm(onlyActiveAccounts = false): Promise<void> {
-    // no need to block here, as the first fetch blocks the entire service init
-    this.fetchAndCacheTokenLists()
+  private async handleBalanceAlarm({
+    onlyActiveAccounts = false,
+    fetchTokenLists = true,
+  }: {
+    onlyActiveAccounts?: boolean
+    fetchTokenLists?: boolean
```
0xDaedalus, line 831: 📈

background/services/indexing/tests/index.integration.test.ts
```diff
@@ -12,6 +13,19 @@ import ChainService from "../../chain"
 import PreferenceService from "../../preferences"
 import { getOrCreateDb as getIndexingDB } from "../db"
 
+type MethodSpy<T extends (...args: unknown[]) => unknown> = jest.SpyInstance<
+  ReturnType<T>,
+  Parameters<T>
+>
+
+const getMethodSpy = <T extends (...args: unknown[]) => unknown>(
```
0xDaedalus, line 21: I did not realize how much I loved this until this PR - makes me think we should deprecate the `ExternalizedService` when stubbing methods in favor of this.

background/services/indexing/tests/index.integration.test.ts
```diff
@@ -42,6 +68,8 @@ describe("IndexingService", () => {
     })
 
     sandbox.stub(chainService, "supportedNetworks").value([ETHEREUM, OPTIMISM])
+    // eslint-disable-next-line @typescript-eslint/dot-notation
+    chainService["trackedNetworks"] = [ETHEREUM, OPTIMISM]
```
0xDaedalus, line 72: Why this pattern over stubbing?

0xDaedalus (approved):
Fantastic PR and superb tests 🎉  - will let you review comments as none are blocking and decide if you'd like to address or not.

Shadowfiend (commented):
Couple of small suggestions, plus one more: if you're updating the db, consider going ahead and adding the `trusted` property to the db that is mentioned here:

https://github.com/tallyhowallet/extension/blob/d1305379c3cae2dfbea79ff5dfab13201f9876e7/background/services/indexing/index.ts#L357-362

Obv will take more time than this which seems to already be working, but sets up the next bits for marking assets trusted more clearly, and lets the decision on whether to handle pricing assets happen based on trustedness instead of baking in an easy to forget attachment to all custom assets.

background/services/indexing/tests/index.integration.test.ts
hyphenized: Good catch, overlooked this

background/services/indexing/index.ts
```diff
@@ -669,32 +670,56 @@ export default class IndexingService extends BaseService<Events> {
...
-      (asset) =>
-        asset.symbol === "ETH" ||
-        trackedNetworks
-          .map((n) => n.chainID)
-          .includes(asset.homeNetwork.chainID)
+    const customAssets = new Set(
+      (await this.db.getCustomAssetsByNetwork(trackedNetworks)).map(
+        ({ contractAddress, homeNetwork: { chainID } }) =>
+          `${chainID}:${contractAddress}`
+      )
```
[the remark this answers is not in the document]
hyphenized: Thanks, was testing if using this pattern was actually more readable than using `some` because of the `has(...)` vs `some(a => a == b && a == c)` check

background/services/indexing/index.ts
```diff
@@ -784,9 +823,17 @@ export default class IndexingService extends BaseService<Events> {
     }
   }
 
-  private async handleBalanceAlarm(onlyActiveAccounts = false): Promise<void> {
-    // no need to block here, as the first fetch blocks the entire service init
-    this.fetchAndCacheTokenLists()
+  private async handleBalanceAlarm({
+    onlyActiveAccounts = false,
+    fetchTokenLists = true,
+  }: {
```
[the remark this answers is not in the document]
hyphenized: I fully agree, this was my intent with `handlePriceAlarm` too. Will create a new PR to address this issue.

background/services/indexing/tests/index.integration.test.ts
```diff
@@ -156,9 +185,12 @@ describe("IndexingService", () => {
...
     it("should update cache once token lists load", async () => {
-      sandbox
-        .stub(chainService, "supportedNetworks")
-        .value([ETHEREUM, OPTIMISM])
+      const spy = getMethodSpy<IndexingService["fetchAndCacheTokenLists"]>(
+        indexingService,
+        "fetchAndCacheTokenLists"
+      )
+
+      const cacheSpy = jest.spyOn(indexingService, "cacheAssetsForNetwork")
```
jagodarybacka, line 193: why you are not using `getMethodSpy()` here?

background/services/indexing/tests/index.integration.test.ts
```diff
@@ -223,5 +265,60 @@ describe("IndexingService", () => {
         indexingService.getCachedAssets(ETHEREUM).map((assets) => assets.symbol)
       ).toEqual(["ETH", customAsset.symbol, "TEST"])
     })
+
+    it("should not retrieve token prices for custom assets", async () => {
+      const indexingDb = await getIndexingDB()
+
+      const smartContractAsset: SmartContractFungibleAsset = {
```
jagodarybacka, line 272: we have a factory for assets `createSmartContractAsset()`

background/services/indexing/tests/index.integration.test.ts
hyphenized: There's no need since this is a public method, but this is a good indication that this helper should be renamed

jagodarybacka (approved):
There are no weird asset prices fetched but I had issues with fetching prices (mostly on Polygon) can anyone else manually test it? @kkosiorowska or @0xDaedalus? 😊

PR #2951, opened 2023-01-31
Fix issues for abilities page

### LINKED ISSUE #2946: Daylight: No filter between "Open" + "Expired" abilities
### Discord Discussion Link

_No response_

Chrome

I was trying to filter out "open" abilities from "expired" abilities. 

### What did not work?

All of the "open" abilities are also showing up as "expired" abilities. All of the other filters work fine. 

### Version

None

### Relevant log output

_No response_

kkosiorowska:
Closes #2946 

This PR fixes issues for the abilities page. Changes: 

* fix the issue of filtering expired abilities
* display correct time details label

## UI
Before

After

## To Test
- [ ] try to filter out "open" abilities from "expired" abilities

## Testing Env
```
SUPPORT_ABILITIES=true
```

### DIFF
--- a/background/redux-slices/utils/abilities-utils.ts
+++ b/background/redux-slices/utils/abilities-utils.ts
@@ -1,19 +1,27 @@
 import { Ability, AbilityType } from "../../abilities"
 import { Filter, State } from "../abilities"
 
+const isDeleted = (ability: Ability): boolean => ability.removedFromUi
+
+const isExpired = (ability: Ability): boolean =>
+  ability.closeAt ? new Date(ability.closeAt) < new Date() : false
+
 export const filterByState = (ability: Ability, state: State): boolean => {
   switch (state) {
     case "open":
-      return ability.completed === false && !ability.removedFromUi
-    case "completed":
-      return ability.completed === true && !ability.removedFromUi
-    case "expired":
       return (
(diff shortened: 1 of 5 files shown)


jagodarybacka (approved):
Scope of this PR works fine ✅ 

I find it confusing that I don't see abilities of a specific type on the list unless I have them checked in the "notify me" section. If I have "all" selected in filters I would expect to see them all 🤔 

Btw if you will load the wallet with the feature flag off, switch the flag and reload it - all accounts are by default switched off in the filters panel: 
image

background/redux-slices/abilities.ts
```diff
@@ -56,6 +56,12 @@ const abilitiesSlice = createSlice({
         immerState.abilities[address][ability.abilityId] = ability
       })
     },
+    deleteAbilitiesForAccount: (
```
0xDaedalus, line 59: Will this ever be called separately from `deleteAccount`?  If not - rather than adding a separate action would it make sense to add `delete immerState.abilities[address]` to the logic in `deleteAccount`?

kkosiorowska: `deleteAccount`  is used for us to change the state of the account filter. If we change the account filter to disabled, we don't want to additionally delete abilities. We want them to stop displaying but they shouldn't be deleted.

PR #2952, opened 2023-01-31
Reduce size of NFT redux slice

jagodarybacka:
### What

Importing `vitalik.eth` address that has insane amounts of NFTs results in bloating redux size a lot.
This account has on the Ethereum network:
Collections

Redux slice size before optimizations:
nftsUpdate size  kb  2242 3203125

The easiest thing to decrease the size was to remove the `network` object and replace it with`chainID` string as we can get the network object when we will need it.

After: 
nftsUpdate size  kb  1654 9443359375

Latest build: [extension-builds-2952](https://github.com/tallyhowallet/extension/suites/10713999385/artifacts/536822164) (as of Wed, 01 Feb 2023 14:18:46 GMT).


jagodarybacka:
Ideas of what else can be removed or replaced will be appreciated 😊

0xDaedalus (dismissed):
This is a great step forward - I fear it may not be enough to address our concerns about accounts with large amounts of NFTs but a reduction in size by 50% is a win regardless.

Can we add a test here?  I think a single test that asserts that adding NFTs to the state correctly omits the network and replaces it with the chainID should be enough.

background/redux-slices/tests/nfts_update.unit.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  totalNftCount: 10,
+  floorPrice: {
+    value: 1000000000000000000n,
+    token: {
+      name: "Ether",
+      symbol: "ETH",
+      decimals: 18,
+    },
+  },
+}
```
kkosiorowska, line 21: Non-blocking comment
I think we could move this to the `factories` file.

kkosiorowska (approved):
It looks good. Tests have been added ✅

PR #2981, opened 2023-02-03
 Change visibility for Abilities

### LINKED ISSUE #2978: Abilities should not be shown for read-only accounts
Abilities should not be shown for read-only accounts.

image

kkosiorowska:
Closes #2978

This PR changes visibility for abilities. Abilities should not be shown for read-only accounts.

## UI
Screenshot 2023-02-06 at 08 52 11

Screenshot 2023-02-03 at 14 05 39

## To Test
- [ ] check if abilities are not shown for read-only accounts
- [ ] check if abilities are fetched only for the correct accounts

## Testing Env
```
SUPPORT_ABILITIES=true
```

Latest build: [extension-builds-2981](https://github.com/tallyhowallet/extension/suites/10806567227/artifacts/543588141) (as of Mon, 06 Feb 2023 15:30:44 GMT).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -43,6 +43,7 @@ import { Eligible } from "./services/doggo/types"
 
 import rootReducer from "./redux-slices"
 import {
+  AccountType,
   deleteAccount,
   loadAccount,
   updateAccountBalance,
@@ -601,7 +602,7 @@ export default class Main extends BaseService<never> {
   ): Promise<void> {
     this.store.dispatch(deleteAccount(address))
 
-    if (signer.type !== "read-only" && lastAddressInAccount) {
+    if (signer.type !== AccountType.ReadOnly && lastAddressInAccount) {
       await this.preferenceService.deleteAccountSignerSettings(signer)
     }
 
@@ -620,7 +621,7 @@ export default class Main extends BaseService<never> {
     // remove abilities
     if (
       isEnabled(FeatureFlags.SUPPORT_ABILITIES) &&
-      signer.type !== "read-only"
+      signer.type !== AccountType.ReadOnly
     ) {
       await this.abilitiesService.deleteAbilitiesForAccount(address)
     }
--- a/background/main.ts
+++ b/background/main.ts
@@ -617,7 +617,10 @@ export default class Main extends BaseService<never> {
       await this.nftsService.removeNFTsForAddress(address)
     }
     // remove abilities
-    if (isEnabled(FeatureFlags.SUPPORT_ABILITIES)) {
+    if (
+      isEnabled(FeatureFlags.SUPPORT_ABILITIES) &&
+      signer.type !== "read-only"
+    ) {
       await this.abilitiesService.deleteAbilitiesForAccount(address)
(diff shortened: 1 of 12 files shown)


background/main.ts
```diff
@@ -617,7 +618,10 @@ export default class Main extends BaseService<never> {
       await this.nftsService.removeNFTsForAddress(address)
     }
     // remove abilities
-    if (isEnabled(FeatureFlags.SUPPORT_ABILITIES)) {
+    if (
+      isEnabled(FeatureFlags.SUPPORT_ABILITIES) &&
+      signer.type !== "read-only"
```
jagodarybacka, line 623: Let's use existing [type](https://github.com/tallyhowallet/extension/blob/visibility-for-abilities/background/redux-slices/accounts.ts#L25)

ui/_locales/en/messages.json
```diff
@@ -823,7 +823,9 @@
       },
       "abilityStateTitle": "Show",
       "abilitiesTypesTitle": "Notify me about abilities type",
-      "accountsTitle": "Show/hide accounts"
+      "accountsTitle": "Show/hide accounts",
+      "accountsReadOnlyInfo": "Read-only accounts can’t see abilities. If you can, we recommend you import the address you want o see abilities for.",
```
jagodarybacka, line 827:
```suggestion
      "accountsReadOnlyInfo": "Read-only accounts can’t see abilities. If you can, we recommend you import the address you want to see abilities for.",
```
That "if you can" part sounds weird to me. I would say it should be more like a CTA "You should/can import the address you want to..."

background/main.ts
```diff
@@ -1532,6 +1536,9 @@ export default class Main extends BaseService<never> {
   }
 
   connectAbilitiesService(): void {
+    this.abilitiesService.emitter.on("initAbilities", (address) => {
+      this.store.dispatch(initAbilities(address))
```
jagodarybacka, line 1540: So we are listening on abilities service, then we are checking in redux if an account is read-only, if not then we are going back to abilities service to `pollForAbilities()`, am I right? 
Can't we use other services to know if this account is read-only or not instead of relying on the redux state?

ui/_locales/en/messages.json
kkosiorowska: @VladUXUI What do you think?

background/main.ts
kkosiorowska: I spent some time thinking about it. I haven't been able to find a better way. From the service level, I wasn't able to check the type of account. This solution seems to me the least invasive.

jagodarybacka (approved):
Works as expected ✅ would love to revisit this problem with going back and forth between services and redux in the future

ui/_locales/en/messages.json
VladUXUI: I would leave if you can. Because we don't support private key import or other things. So blocking users from seeing a feature, then telling them to do an action that they can't do, just leads to disappointment and messages on discord.
This way we set the expectations a little bit

PR #2986, opened 2023-02-05
Refactor Indexing service balances alarm handler

hyphenized:
Moved the concern of loading account balances to a separate method such that the balance alarm handler no longer depends on parameters only relevant for loading balances.

## To Test
- [ ] Import a wallet
- [ ] Check that balances are loading normally

Latest build: [extension-builds-2986](https://github.com/tallyhowallet/extension/suites/11091724138/artifacts/564628039) (as of Mon, 20 Feb 2023 16:10:51 GMT).


background/services/indexing/index.ts
```diff
@@ -844,33 +824,37 @@ export default class IndexingService extends BaseService<Events> {
...
+
+        const provider = this.chainService.providerForNetworkOrThrow(network)
+
+        const loadBaseAccountBalance =
+          this.chainService.getLatestBaseAccountBalance(addressOnNetwork)
+
+        const assetsToCheck = provider.supportsAlchemy
+          ? activeAssetsToTrack
+          : // This doesn't pass assetsToTrack as it assumes they've already been cached
+            this.getCachedAssets(network).filter(isSmartContractFungibleAsset)
```
jagodarybacka, line 843: This is really confusing without any context - why does supporting alchemy has something to do with getting cached assets or not? A comment would be helpful 😊

jagodarybacka (commented):
How should we test it? just by importing a wallet and checking if assets are loaded properly?

background/services/indexing/index.ts
hyphenized: When the provider supports alchemy we use [alchemy_getTokenBalances](https://docs.alchemy.com/reference/alchemy-gettokenbalances) to query which assets to track so it's safe to use assets we already stored as tracked in the db. We use the network's cached assets otherwise because we don't know exactly what tokens the user is holding therefore, we query balances for every token we've seen.

jagodarybacka (approved):
Seems like it works 💪

PR #2983, opened 2023-02-05
Fix eth_accounts handling for dApps without permissions

hyphenized:
Certain dApps send requests for `eth_accounts` before connection, this causes the wallet to prompt the user for permissions before interacting with the dApp and, in some cases, to generate multiple prompts, rendering the browser unusable.

## To Test
- [ ] Head to https://app.yieldprotocol.com/pool
- [ ] You should not see the connect to dApp popup without interacting with the dApp

Latest build: [extension-builds-2983](https://github.com/tallyhowallet/extension/suites/10784251627/artifacts/542002439) (as of Sun, 05 Feb 2023 03:19:26 GMT).

PR #2996, opened 2023-02-07
Custom RPCs - Request user confirmation before adding a custom network

### LINKED ISSUE #2957: Implement "Add Network" confirmation prompt.
When a dapp requests to add a network via `wallet_addEthereumChain` - users should be presented with the option to either reject or accept adding that network.  This also serves as visual confirmation to the user that the network was correctly added.

image

https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=15016%3A305984&t=EsspD3H7SCF5MIuk-1

hyphenized:
Closes #2957 

Implements a popup to prompt the user for confirmation before adding a new network and changes `wallet_switchEthereumChain` and `wallet_addEthereumChain` to only work with connected dApps.

image

Pending tasks:
- Add network icons with fallbacks (chainlist & additional color coded / name only)
- Handle adding networks with the same chain as builtin networks
- Handle duplicate requests
- Handle requests for networks with the same chain Id as already stored networks

## Testing Env

```
SUPPORT_CUSTOM_NETWORKS=true
SUPPORT_CUSTOM_RPCS=true
```

## To Test
- [ ] Head over to [chainlist.org](https://chainlist.org), add a few custom networks
- [ ] You should see a confirmation prompt for each network you try to add
- [ ] Clicking on reject or closing the window should cancel the request
- [ ] Clicking on add should save the new network to the wallet

Latest build: [extension-builds-2996](https://github.com/tallyhowallet/extension/suites/11141100660/artifacts/568275569) (as of Wed, 22 Feb 2023 18:00:55 GMT).


ui/pages/AddNewEVMChain.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+  function handleUserResponse(success: boolean): void {
+    dispatch(dispatchAddNetworkUserResponse([requestId, success])).then(() =>
+      window.close()
+    )
+  }
+
+  return (
+    <div className="standard_width">
+      <form action="#">
```
jagodarybacka, line 50: Do we need this `form`?

jagodarybacka (commented):
Didn't hav time to test it manually, initial thoughts:

provider-bridge/index.ts
```diff
@@ -20,7 +20,7 @@ export function connectProviderBridge(): void {
       // if dapp wants to connect let's grab its details
       if (
         event.data.request.method === "eth_requestAccounts" ||
-        event.data.request.method === "eth_accounts"
```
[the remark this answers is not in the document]
hyphenized: This adds the favicon and site title to the params sent to the background script, since we're no longer treating `eth_accounts` (#2983) calls as an indicator that a user wants to connect to a dApp we don't need to pass these anymore

ui/pages/AddNewEVMChain.tsx
hyphenized: We do but I think we also need to move the confirmation handler to the `onSubmit` so the semantic value this form element provides is not lost

jagodarybacka: yeah I was wondering if we have no onSubmit then this element adds little value 😄

background/services/internal-ethereum-provider/index.ts
```diff
@@ -88,7 +88,7 @@ export type ValidatedAddEthereumChainParameter = {
   rpcUrls: string[]
 }
 
-const validateAddEthereumChainParameter = ({
+export const validateAddEthereumChainParameter = ({
```
0xDaedalus, line 91: Good call moving the validation logic out of here.  We can probably go one step further and move the definition into provider-bridge as well

0xDaedalus (changes requested):
Lets go ahead and at the very least add tests for:
- New methods in the provider-bridge service
- Logic inside the new `wallet_addEthereumChain` block

jagodarybacka (changes requested):
Hmmm weird but styles are broken for me 🤔 The rest is working as expected

image

background/services/provider-bridge/tests/index.unit.test.ts
```diff
@@ -104,5 +108,67 @@ describe("ProviderBridgeService", () => {
       expect(stub.called).toBe(false)
       expect(response).toBe(EIP1193_ERROR_CODES.unauthorized)
     })
+
+    it("should wait for user confirmation before calling wallet_AddEtherumChain", async () => {
```
jagodarybacka, line 112:
```suggestion
    it("should wait for user confirmation before calling wallet_addEthereumChain", async () => {
```

jagodarybacka (commented):
Is this ready for another review? I see the styles are still broken 😛

ui/pages/NewCustomNetworkRequest.tsx
```diff
@@ -0,0 +1,256 @@
...
+        .new_chain_logo_wrapper {
+          width: 50px;
+          height: 50px;
+          border-radius: 8px;
+          overflow: hidden;
+        }
+
+        .new_chain_logo {
+          width: 50px;
+          height: 50px;
```
jagodarybacka, line 142: These icons should have the same size: 
image

```suggestion
          width: 56px;
          height: 56px;
```

jagodarybacka (commented):
Looks good, small things left:

cc @VladUXUI should these buttons be fixed to the bottom of the screen? 
image

VladUXUI:
> Looks good, small things left:
> 
> cc @VladUXUI should these buttons be fixed to the bottom of the screen? image

Yeah, this is looking a bit off.
Background seems a bit darker, and buttons are not aligned with the rest of the UI
image

0xDaedalus (approved):
Lets 🚢

PR #2998, opened 2023-02-07
Hide swaps button for networks not supporting swaps

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2959

### What

Hide all links to swaps page on networks not supporting swaps
image
image
 

Latest build: [extension-builds-2998](https://github.com/tallyhowallet/extension/suites/10825993130/artifacts/545049634) (as of Tue, 07 Feb 2023 11:05:17 GMT).


0xDaedalus:
This seems to be a duplicate of https://github.com/tallyhowallet/extension/pull/2962

Happy to hand it over - but want to emphasize that the desired functionality here is for the swap button to be `disabled` rather than removed - which probably looks something like no pointer on mouseover and greying out of the icon.

jagodarybacka:
> swap button to be disabled

@0xDaedalus Vlad left a [comment here](https://github.com/tallyhowallet/extension/issues/2959#issuecomment-1419212402) that it should be hidden

0xDaedalus:
Interesting - that is the complete opposite of what we discussed in Element - but c'est la vie lets 🚢

0xDaedalus (approved):
🚢

PR #3000, opened 2023-02-07
Add rarity rankings to NFTs

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2991
### What
- Show rarity ranks on NFT preview

Other fixes:
- Fix color of slideup's background
- Allow tooltip to show content with "auto" width

image

Note: in the tooltip we show `rank / # of NFTs in collection` which may be different from what OpenSea shows for some collections - if some ranks are shared by many NFT in the collection. As we don't have rarity data on the collection level we can't show this information.

### Testing env

```
SUPPORT_NFT_TAB=true
```

Latest build: [extension-builds-3000](https://github.com/tallyhowallet/extension/suites/10830021187/artifacts/545316928) (as of Tue, 07 Feb 2023 14:10:00 GMT).

### DIFF
--- a/background/lib/simple-hash_update.ts
+++ b/background/lib/simple-hash_update.ts
@@ -39,6 +39,11 @@ type SimpleHashNFTModel = {
     }[]
   }
   owners: { owner_address: string; last_acquired_date: string }[]
+  rarity?: {
+    rank: number | null
+    score: number | null
+    unique_attributes: number | null
+  }
   extra_metadata: {
     attributes?: [{ trait_type?: string | null; value?: string | null }]
   }
@@ -170,6 +175,7 @@ function simpleHashNFTModelToNFT(
     external_url: nftURL = "",
     collection: { collection_id: collectionID },
     extra_metadata: metadata,
+    rarity,
   } = original
 
   const thumbnailURL =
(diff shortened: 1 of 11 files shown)


background/lib/simple-hash_update.ts
```diff
@@ -39,6 +39,11 @@ type SimpleHashNFTModel = {
     }[]
   }
   owners: { owner_address: string; last_acquired_date: string }[]
+  rarity?: {
+    rank: number | null
+    score: number | null
+    unique_attributes: number | null
+  }
```
0xDaedalus, line 46: This could probably be its own named type since we use it in a few places - non blocking of course.

ui/components/NFTS_update/NFTPreview.tsx
```diff
@@ -129,16 +134,41 @@ export default function NFTPreview(props: NFTWithCollection): ReactElement {
...
+                size="small"
+                iconSmall="send"
+                iconPosition="left"
+              >
+                {t("preview.send")}
+              </SharedButton>
+            )}
+          </div>
+
+          {rarityRank !== null && (
```
0xDaedalus, line 151: Seems like we could write a fairly straightforward component test here to

a) ensure that the rank preview does not render if there is no rarity rank
and
b) ensure we're localizing & pulling the rarity rank correctly.

We don't have too many of these tests - so don't want to block on this - but something to think about as a followup and/or for future component related PRs

0xDaedalus (approved):

ui/components/NFTS_update/NFTPreview.tsx
jagodarybacka: there are no UI tests for NFTs yet (but we have a task for it!) but I would love to create them when we will have some space for them 😅

PR #3002, opened 2023-02-07
The Account filter for abilities

kkosiorowska:
This PR fixes some issues with changing the account filter. A couple of  thoughts:
- The event of adding a new account should make enable it in the filter.  
- Read-only accounts shouldn't be added to the filter. 

## To Test

1. Adding accounts
- [ ] Add an account by importing the recovery phrase. Check the abilities filter. The account should be enabled.
- [ ] Add an account by connecting to the ledger. Check the abilities filter. The account should be enabled.
- [ ]  Add read-only account. The account shouldn't be added to the filter.
- [ ] Create a new account. Check the abilities filter. The account should be enabled.

2. Removing accounts
- [ ] Remove imported account. The account should be removed from the filter.
- [ ] Remove ledger account. The account should be removed from the filter.
- [ ] Remove internal account. The account should be removed from the filter.

```
    const state = JSON.parse(stateString, null, 2)
    console.log(state.abilities.filter)
})
```

## Testing Env
```
SUPPORT_ABILITIES=true
```

Latest build: [extension-builds-3002](https://github.com/tallyhowallet/extension/suites/10851278481/artifacts/546900926) (as of Wed, 08 Feb 2023 10:41:23 GMT).


jagodarybacka (approved):
Works fine ✅

PR #3006, opened 2023-02-08
Flip RPC request params

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/2985
### What

Check if wallet address is on the expected position in params list for `eth_call` and `personal_sign` methods.  If it isn't then flip params order.

This should allow dapps that are not following JSON-RPC standard to be able to sign messages anyway.

### Testing
Check if you can connect to dapp and sign message on:
- https://www.premint.xyz/
- https://mintkudos.xyz/
- https://login.xyz/
- https://tallyho.org/web3pledge/

Latest build: [extension-builds-3006](https://github.com/tallyhowallet/extension/suites/10852469412/artifacts/546985369) (as of Wed, 08 Feb 2023 11:39:07 GMT).

### DIFF
--- a/background/services/provider-bridge/index.ts
+++ b/background/services/provider-bridge/index.ts
@@ -25,7 +25,11 @@ import {
 import showExtensionPopup from "./show-popup"
 import { HexString } from "../../types"
 import { WEBSITE_ORIGIN } from "../../constants/website"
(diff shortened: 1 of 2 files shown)


0xDaedalus (approved):
All tested sites work great - 📈

PR #3008, opened 2023-02-08
No change filter state after reloading

kkosiorowska:
This PR makes the filter state for abilities is not updated after reloading.  Accounts added after the flag change should be enabled.

## To Test

- [ ] Run the application when `SUPPORT_ABILITIES=false`. Add new accounts. Enable the feature flag. Go to the abilities page and check the filter. The accounts should be enabled.
- [ ] Change the state of the filter. Make some accounts disabled. Reload the script and check the state again. It should be the same as before reloading.

Latest build: [extension-builds-3008](https://github.com/tallyhowallet/extension/suites/10857046169/artifacts/547299870) (as of Wed, 08 Feb 2023 15:02:41 GMT).

### DIFF
--- a/background/redux-slices/abilities.ts
+++ b/background/redux-slices/abilities.ts
@@ -49,6 +49,15 @@ const initialState: AbilitiesState = {
   hideDescription: false,
 }
 
+const isInitialState = (state: AbilitiesState): boolean =>
+  state.filter.state === initialState.filter.state &&
+  state.filter.types.every(
+    (type, idx) => type === initialState.filter.types[idx]
+  ) &&
+  state.filter.accounts.length === 0 &&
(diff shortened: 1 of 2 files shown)


background/redux-slices/abilities.ts
```diff
@@ -49,6 +49,15 @@ const initialState: AbilitiesState = {
   hideDescription: false,
 }
 
+const isInitialState = (state: AbilitiesState): boolean =>
+  state.filter.state === initialState.filter.state &&
```
jagodarybacka, line 53: I wonder if comparing `JSON.stringify(state) === JSON.stringify(initialState)` wouldn't be enough :thinking:

jagodarybacka (approved):
Works as expected ✅

PR #3013, opened 2023-02-10
Refactor deleting transferred NFTs reducer

jagodarybacka:
### What

Cover edge cases where:
- transferred NFTs weren't loaded to redux cache
- there were multiple NFTs transfers in a row
- transferred NFT was the last one in the collection

### Testing

Test NFTs transfers using steps from the "Transfers" section: https://github.com/tallyhowallet/extension/pull/2843

Latest build: [extension-builds-3013](https://github.com/tallyhowallet/extension/suites/11018524645/artifacts/559156421) (as of Thu, 16 Feb 2023 10:14:02 GMT).


jagodarybacka:
Please don't merge yet - I want to add tests for this reducer - but review and test and see if you can think about another way the collection may stay in the redux despite containing no NFTs.

background/redux-slices/tests/nfts_update.unit.test.ts
```diff
@@ -67,11 +105,278 @@ describe("NFTs redux slice", () => {
...
+
+      it("should handle transfer if collection has many owned NFTs but none cached", () => {
+        state = initializeCollections([{ ...COLLECTION_MOCK, nftCount: 10 }])
+        deleteTransferred(state, [TRANSFER_MOCK])
+
+        const collection = extractCollection(state)
+        expect(collection).toBeDefined()
+        expect(collection?.nfts.length).toEqual(0)
+        expect(collection?.nftCount).toEqual(9)
+      })
```
kkosiorowska, line 139: It seems to me that we should focus on unit testing.  I mean that I test a specific function and compare it with the expected state.

```suggestion
      it("should handle transfer if collection has many owned NFTs but none cached", () => {
        const newState = reducer(state, deleteTransferredNFTs([TRANSFER_MOCK]))

        const collection = extractCollection(newState)
        expect(collection).toBeDefined()
        expect(collection?.nfts.length).toEqual(0)
        expect(collection?.nftCount).toEqual(9)
      })
```

background/redux-slices/tests/nfts_update.unit.test.ts
```diff
@@ -67,11 +105,278 @@ describe("NFTs redux slice", () => {
...
+        })
+        deleteTransferred(state, [TRANSFER_MOCK])
+
+        const collection = extractCollection(state)
+
+        expect(collection).toBeDefined()
+        expect(collection?.nftCount).toEqual(1)
+        expect(collection?.nfts.length).toEqual(1)
+        expect(collection?.nfts[0].id).toBe(secondID)
+      })
```
kkosiorowska, line 208: In this case, we use `initializeCollections`, `updateNFTsInCollection`, and `deleteTransferred`. Finding the issue can be difficult. Maybe we should break these tests into smaller parts.

kkosiorowska (commented):
I tested this and looks great 🤩 Just a couple of thoughts about tests.

background/redux-slices/nfts_update.ts
```diff
@@ -242,20 +300,7 @@ const NFTsSlice = createSlice({
...
-      } = payload
-
-      const collectionToUpdate =
-        immerState.nfts[network.chainID][normalizeEVMAddress(address)][
-          collectionID
-        ]
-
-      collectionToUpdate.nfts = parseNFTs(nfts)
-      collectionToUpdate.hasNextPage = hasNextPage
+      updateNFTsInCollection(immerState, payload)
```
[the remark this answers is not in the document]
jagodarybacka: this is just to untangle it from the redux so we can test it as a regular function, idk if this is the best way to test reducers but pretty convenient. As these f..unctions are kinda big I was even thinking about moving them out of this file for better readability.

background/redux-slices/tests/nfts_update.unit.test.ts
jagodarybacka: yeah, so it is more like an integration test at this point but as we don't have the best test coverage I think it is fine if this is testing more than one function but the whole flow of saving collections and then removing them. Maybe we should move these to `_.integration.test.ts` to make it more clear?

background/redux-slices/tests/nfts_update.unit.test.ts
jagodarybacka: As above - let's treat it as an integration test. I was hesitant to make it a truly "unit" test and write redux state manually as other functions don't have test coverage yet.

background/redux-slices/nfts_update.ts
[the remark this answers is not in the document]
kkosiorowska: There is an example on [the redux page](https://redux.js.org/usage/writing-tests#reducers).  We should probably do it this way.

background/redux-slices/tests/nfts_update.unit.test.ts
kkosiorowska: I agree that it looks more like integration testing. However, it seems to me that we should have unit test coverage of these functions. They have some logic in them and it might be worth it. 

@hyphenized @0xDaedalus  What do you guys think about it? It would be good to establish one approach for the future.

background/redux-slices/nfts_update.ts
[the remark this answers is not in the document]
jagodarybacka: done

background/redux-slices/tests/nfts_update.unit.test.ts
jagodarybacka: Ok, I've extracted it to the integration test file and I'm using `reducer` to dispatch actions. 
I'm still not sure about unit testing `deleteTransferredNFTs` - writing manually this huge state would be more error-prone than making it go through the whole flow IMO. Other reducers like `updateNFTsCollections` can be unit tested for sure but idk if we want to include it all in this PR.

kkosiorowska: I understand what you mean. In this case, however, we trust that the `updateNFTsCollections` method works correctly. Anyway, I think it might be worth adding tests for `updateNFTsCollections`. Yes, you are right it is better to break it up into several PRs. So let's not block it 😀

kkosiorowska (approved):
LGTM ✅

PR #3015, opened 2023-02-10
Retry up to five times if an addresses abilities are still pending

0xDaedalus:
This is a difficult one to test - so will probably need to rely on making sure code & test are accurate.

The way to test this would be:
- [ ] Import an address into the wallet that daylight has never seen - and has a large amount of abilities (probably not possible)
- [ ] If there are a large amount of abilities associated with the address - the response will resolve with N abilities and a status of "pending"
- [ ] Upon seeing the pending status - the app should fire more requests until it either sees a `synced` status or has reached 5 retries.

Latest build: [extension-builds-3015](https://github.com/tallyhowallet/extension/suites/10908416756/artifacts/551098481) (as of Fri, 10 Feb 2023 16:48:02 GMT).

### DIFF
--- a/background/lib/daylight.ts
+++ b/background/lib/daylight.ts
@@ -3,6 +3,7 @@ import { AbilityType } from "../abilities"
 import logger from "./logger"
 
 const DAYLIGHT_BASE_URL = "https://api.daylight.xyz/v1"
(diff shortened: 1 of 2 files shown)


hyphenized (approved):
Looks good :shipit:, great test!

PR #3018, opened 2023-02-11
Improve logging

hyphenized:
This PR updates our logger implementation to accept setting a minimum log level with the goal of reducing the noise generated in the output of CI workflows. It also sets a manual mock for `SerialFallbackProvider`, fixes a couple of flaky test cases and adds a new utility `it.flaky` to debug flaky tests. 

## To Test
- [ ] Check that exporting logs works as expected, UI pages that use the logger now emit entries with `UI` as the label
- [ ] Check the workflow output, you should see a cleaner output
- [ ] Exported logs should include records for the last hour since the last logged entry

Latest build: [extension-builds-3018](https://github.com/tallyhowallet/extension/suites/10918550341/artifacts/551909772) (as of Sat, 11 Feb 2023 07:49:53 GMT).


background/services/chain/__mocks__/serial-fallback-provider.ts
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable class-methods-use-this */
+import { BigNumber } from "ethers"
+import { Block, FeeData } from "@ethersproject/abstract-provider"
+import { makeEthersBlock, makeEthersFeeData } from "../../../tests/factories"
+import type SerialFallbackProvider from "../serial-fallback-provider"
+
+export default class MockSerialFallbackProvider
```
kkosiorowska, line 7: Do we need two classes of `MockSerialFallbackProvider`? The second class exists in the `factories.ts` file.

hyphenized: Nice catch, we do not!

kkosiorowska (approved):
Looks really nice! 😀I tested this and it also fixes an issue with missing logs.

PR #3021, opened 2023-02-13
Show abilities for read-only accounts

kkosiorowska:
This PR shouldn't be merged. It is used only for tests.

Latest build: [extension-builds-3021](https://github.com/tahowallet/extension/suites/11516882762/artifacts/595528758) (as of Mon, 13 Mar 2023 08:15:18 GMT).


VladUXUI:
Can this be a feature flag? so we can turn it on an off for testing but still close this PR

Shadowfiend:
This was mostly used for QA of the Daylight integration pre-launch, after some discussion it sounds like we're comfortable with doing this one-off locally if needed for testing. It's not currently in a state to be immediately feature-flagged, and it's not worth putting in the cycles at this point to get it that clean. Closing!

PR #3026, opened 2023-02-13
Handle multiple banners on Wallet page

### LINKED ISSUE #3023: Handle multiple banners on Wallet page
Handle a situation where there are two banners visible at the same time - default wallet and analytics notifications 
- both should be the same width
- we shouldn't show them at the same time, after analytics is dismissed we should show the default wallet one

image

kkosiorowska:
Close #3023 

Changes:

- set the correct width
- fix the width of icons (checkmark and X)
- display only one banner

Screenshot 2023-02-13 at 16 40 48

Latest build: [extension-builds-3026](https://github.com/tallyhowallet/extension/suites/10948840013/artifacts/554047633) (as of Mon, 13 Feb 2023 15:46:30 GMT).


0xDaedalus (approved):
Works perfectly 👍

PR #3028, opened 2023-02-14
[Plumbing] - Add workspace level .eslintrc.js files

hyphenized:
This PR fixes issues with the extraneous dependencies rule and packages for shared dependencies included at the top level package.json, it also sets exceptions for the same rule for tests and configuration files.

## To Test
- [ ] Attempt to import a package from the top level package.json (e.g. "styled-jsx/css")
- [ ] Check there are no linting warnings


.eslintrc.js
```diff
@@ -101,6 +111,7 @@ module.exports = {
     "**/validate/*.js",
     "**/local-chain/**",
     "!.github",
+    ".eslintrc.js",
```
Shadowfiend, line 114: How come we're ignoring the lint JS file?

Shadowfiend (commented):
Out of curiosity, have these warnings been happening all along?

.eslintrc.js
hyphenized: I added it to the ignore list because it was having issues with the ts parser and excluding this file from the parser with an override generates errors with the Typescript rules

hyphenized:
These warnings have been going [for a while](https://github.com/tallyhowallet/extension/actions/runs/4170467540/jobs/7219461455), one workaround was to have the dependency in both `package.json` files. For test files I think we had to set exceptions.

Shadowfiend:
Sorry for not following up here. I'm kind of torn tbh. That warning kind of seems correct? The goal of that rule is that the package that is referenced should be pinned in its containing project—right now `ui` doesn't pin styled-jsx, the root package does, but the root package declares it specifically because it's used as part of the webpack build.

I wish we could declare the dependency in both and have the version synchronized without having to update it in both package.json files, because it feels right to explicitly declare the styled-jsx dependency in `ui` so that when you look for “which version of styled-jsx is being referenced here” it's in the closest package. The root declaration is really for the build rather than for e.g. type checking. One of the annoying things that happens right now off and on is trying to guess whether a dependency is being defined at root or in the subpackage… Tough nut, this one 🤔

hyphenized:
@Shadowfiend Thanks for clearing this up, I thought we were using workspace level dependencies as shared dependencies across all packages but since that's not the case then this is no longer necessary

PR #3035, opened 2023-02-15
Break asset name on single asset page

jagodarybacka:
Resolves https://github.com/tallyhowallet/extension/issues/3034

(couldn't find a good example but now it should be breaking the word)
image

Latest build: [extension-builds-3035](https://github.com/tallyhowallet/extension/suites/10998123811/artifacts/557629421) (as of Wed, 15 Feb 2023 14:20:49 GMT).


hyphenized (approved):
Looks good!

PR #3036, opened 2023-02-15
Use `erc20` argument when fetching token balances via Alchemy.

0xDaedalus:
image
image

Due to the recent polygon hard-fork it is no longer possible to batch more than 100 requests together on polygon.  Because alchemy relies on batch calling under the hood when processing `alchemy_getTokenBalances` requests sent with a list of token contracts to check - we need to move away from the pattern of passing in a discrete lists of contracts to check.  

Luckily, Alchemy exposes the `erc20` request parameter which returns all erc20 balances for a given address - which we can use not only on polygon but on all chains alchemy supports - so lets do that!

### To Test
- [ ] Import a new wallet
- [ ] You should see token balances in the wallet - likely with both "trusted" and "untrusted" tokens.

Latest build: [extension-builds-3036](https://github.com/tallyhowallet/extension/suites/11007404854/artifacts/558317221) (as of Wed, 15 Feb 2023 21:13:48 GMT).

### DIFF
--- a/background/lib/alchemy.ts
+++ b/background/lib/alchemy.ts
@@ -150,14 +150,11 @@ export async function getAssetTransfers(
  */
 export async function getTokenBalances(
   provider: SerialFallbackProvider,
-  { address, network }: AddressOnNetwork,
-  tokens?: HexString[]
+  { address, network }: AddressOnNetwork
 ): Promise<SmartContractAmount[]> {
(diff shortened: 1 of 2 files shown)


hyphenized (changes requested):
Great work on this fix!

hyphenized (approved):

PR #3038, opened 2023-02-16
Create a global modal

### LINKED ISSUE #3030: Pop-up
# Brief
We need a full page pop-up to let users know about an awesome new thing

# Content
- Video/Gif (TBD)
- Headline (TBD)
- Body text (TBD)
- CTA (TBD)

# Special considerations
- The video/gif will need to be swapped out quite last minute. Maybe we use an outside link with dummy content and change it when we have final content. This way we don't need to submit a new release
- This will need to go live on the 21st or 22nd the latest. This probably means that we need to adjust our release schedule a little bit?

# Figma file
https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=16973%3A314509&t=z4WSTGlWoehlteK2-4

kkosiorowska:
Closes #3030 

## UI
Screenshot 2023-02-17 at 13 47 47

## To Test
- [ ] Check if the global modal is visible.

Latest build: [extension-builds-3038](https://github.com/tallyhowallet/extension/suites/11048787545/artifacts/561427305) (as of Fri, 17 Feb 2023 14:57:32 GMT).

### DIFF
--- a/background/redux-slices/migrations/index.ts
+++ b/background/redux-slices/migrations/index.ts
@@ -21,6 +21,7 @@ import to21 from "./to-21"
 import to22 from "./to-22"
 import to23 from "./to-23"
 import to24 from "./to-24"
+import to25 from "./to-25"
 
 /**
  * The version of persisted Redux state the extension is expecting. Any previous
@@ -60,6 +61,7 @@ const allMigrations: { [targetVersion: string]: Migration } = {
   22: to22,
   23: to23,
   24: to24,
+  25: to25,
 }
 
 /**
--- a/background/redux-slices/migrations/to-25.ts
+++ b/background/redux-slices/migrations/to-25.ts
@@ -1,0 +1,6 @@
+// Missing migration due to incorrect version update from 23 to 25.
+export default (
+  prevState: Record<string, unknown>
+): Record<string, unknown> => {
+  return prevState
+}


0xDaedalus:
Updated with new Copy: 
image

background/redux-slices/migrations/to-26.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export default (prevState: Record<string, unknown>): NewState => {
+  const typedPrevState = prevState as OldState
+
+  return {
+    ...prevState,
+    ui: {
+      ...typedPrevState.ui,
+      showGlobalModal: true,
```
jagodarybacka, line 23: Not a blocker but to be improved later - tbh I'm not sure if doing it based on the redux state is the best way, every time we will want to show this modal to users on the new release automatically we will have to do the migration right?

I think making it based on the local storage (like we keep the notification banners closed by saving their id in the local storage) for example would make it easier to reuse

kkosiorowska: Great suggestion!

ui/pages/Popup.tsx
```diff
@@ -159,6 +160,7 @@ export function Main(): ReactElement {
       <div className="top_menu_wrap_decoy">
         <TopMenu />
       </div>
+      <GlobalModal id="meet_taho" />
```
jagodarybacka, line 163: Wohoo 🎉 it is so much simpler now!

jagodarybacka (approved):
LGTM ✨

PR #3050, opened 2023-02-17
Add base asset even if balance is zero

jagodarybacka:
Resolves #https://github.com/tallyhowallet/extension/issues/2910

### What
For accounts that are not in a "loading" state we were not adding base assets if their balance was zero. 
Accounts could be past their loading state if there were an ENS name loaded faster than balances. 
That way, some accounts have an empty assets list but we should always show at least base assets.

Latest build: [extension-builds-3050](https://github.com/tallyhowallet/extension/suites/11048041063/artifacts/561369975) (as of Fri, 17 Feb 2023 14:23:30 GMT).


background/redux-slices/tests/accounts.unit.test.ts
```diff
@@ -0,0 +1,184 @@
...
+  address: ADDRESS_MOCK,
+  assetAmount: {
+    asset: ETH,
+    amount: 1n,
+  },
+  network: ETHEREUM,
+  retrievedAt: 1,
+  dataSource: "local",
+}
+
```
kkosiorowska, line 33: Non-blocking comment
You can use `createAccountBalance` from factories.

kkosiorowska (approved):
LGTM ✅

PR #3055, opened 2023-02-17
Add rename gif to public images

0xDaedalus:
Rather than loading the rename animation from our website (which could result in long load times for users with poor internet connections) we can simply add it to the bundle to have it load instantly upon extension open.

https://user-images.githubusercontent.com/94649004/219789120-84b97364-7d4d-4ffe-8a3f-bcc604ee2a0c.mov

Latest build: [extension-builds-3055](https://github.com/tallyhowallet/extension/suites/11055583178/artifacts/561958893) (as of Fri, 17 Feb 2023 20:50:58 GMT).


hyphenized (approved):
✅

PR #3057, opened 2023-02-18
Turn on enhanced NFT Support

0xDaedalus:
Latest build: [extension-builds-3057](https://github.com/tahowallet/extension/suites/11166866409/artifacts/570200229) (as of Thu, 23 Feb 2023 18:32:42 GMT).


jagodarybacka (changes requested):
`SUPPORT_NFT_SEND` should be `false` as sending was not in the scope and is not implemented yet

hyphenized (approved):
Activate the Taho

PR #3059, opened 2023-02-20
Make claim type enabled by default

kkosiorowska:
Claim type for abilities should be enabled by default.

Screenshot 2023-02-20 at 16 15 54

## To Test
- [ ] Install the extension. Import a new account and check the filter for abilities.  Claim type should be enabled by default.
- [ ] Make migration. The claim type should be enabled for existing users.

Latest build: [extension-builds-3059](https://github.com/tallyhowallet/extension/suites/11107052947/artifacts/565763395) (as of Tue, 21 Feb 2023 10:59:43 GMT).

### DIFF
--- a/background/abilities.ts
+++ b/background/abilities.ts
@@ -25,11 +25,11 @@ export const ABILITY_TYPES_ENABLED = [
   "airdrop",
   "vote",
   "access",
+  "claim",
 ] as const
 // https://docs.daylight.xyz/reference/ability-model#ability-types
 export const ABILITY_TYPES = [
   ...ABILITY_TYPES_ENABLED,
-  "claim",
   "product",
   "event",
   "article",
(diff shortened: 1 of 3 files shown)


jagodarybacka (commented):
As this feature is already live - do we want to force "claim" to be turned on for users that already have wallet installed?

kkosiorowska:
> As this feature is already live - do we want to force "claim" to be turned on for users that already have wallet installed?

I think not. Only new users should have an updated default state because this change can be strange for them. What do you think? @jagodarybacka

jagodarybacka:
I'm not sure, I imagine some users haven't seen the filters panel at all after they checked this feature and noticed they have no abilities and it would be nice for them to be surprised with a "claim" one. But we don't want to spam them with notifications (or do we? because that's basically the whole point of daylight 😂) 
cc @VladUXUI

kkosiorowska:
There was an issue with fetching abilities. For this reason, let's force "claim" to be turned on for existing users. These changes shouldn't have a bad impact on them.

VladUXUI:
yeah, Claim should definitely be on. That was an oversight on my end

jagodarybacka (approved):
Works fine ✅

PR #3060, opened 2023-02-20
Block Sign button when transaction is not ready

### LINKED ISSUE #2940: Block Sign button when transaction is not ready to be signed
In the new Siging flow we can observe the same effects that were causing problems in the old singing flow.
Sign button should be blocked when transaction is not yet ready - during first load, when gas settings are changed and transaction is recalculating gas, when there is a warning that alerts user that transaction won't be successful.
You pay

kkosiorowska:
Closes #2940 

This PR blocks the Sign button. To see the old solution set the flag to `USE_UPDATED_SIGNING_UI=false`. 

The locking logic was separated into a separate component `TransactionButton`. There is no need to block the button the first time the signing frame appears because we use the  `SignTransactionLoader` when the transaction is undefined.  Button locking doesn't react to visual changes in the component.

 Changes made to the button blocking:
- Both buttons are blocked when the transaction isn't ready. Using value from `selectIsTransactionLoaded` 
- The sign button is blocked when the window focus change. (moved from the old signing flow)
- The sign button for signing a Ledger frame is blocked when there is a warning that alerts 

## To Test
The sign button should be blocked :
- [ ] when gas settings are changed and the transaction is recalculating gas
- [ ] when a warning alerts the user that the transaction won't be successful.
- [ ] when the window focus change

Latest build: [extension-builds-3060](https://github.com/tahowallet/extension/suites/11276677301/artifacts/578234203) (as of Wed, 01 Mar 2023 10:45:50 GMT).


Shadowfiend:
Whoops, good catch here, I had a plan that I apparently never implemented 🙉 Because of the way the new signing flow is structured, the signing frame depends on the signer (Ledger, keychain, potentially others). I suggest extracting the concept of “a button that is disabled for a random timeout based on transaction loaded state” to a component (e.g. `function ButtonWhenTransactionReady(props: SharedButtonProps)` or so), pulling all this logic into there, and using that from `SignerBaseFrame` and `SignerLedgerFrame`.

ui/components/Signing/Signer/TransactionButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  // Runs on updates
+  useEffect(() => {
+    clearDelaySignButtonTimeout()
+
+    if (document.hasFocus()) {
+      delaySignButtonTimeout.current = window.setTimeout(() => {
+        setIsOnDelayToSign(false)
+        // Random delay between 0.5 and 2 seconds
+      }, Math.floor(Math.random() * (5 - 1) + 1) * 500)
```
jagodarybacka, line 66: Why isn't it just 4 😛 
```suggestion
      }, Math.floor(Math.random() * 4 + 1) * 500)
```

jagodarybacka (approved):
Seems to work fine, same as prev version 💪 I wish we will make some UX improvements at some point as disabling buttons is very visible and distracting 🤔 
Some comments, non-blocking but nice to have:

PR #3081, opened 2023-02-24
Token Balances - Fix erc20 asset loading for Alchemy supported networks

hyphenized:
Updated Alchemy token balances JTD validator to accept pageKey strings returned from the API, this string is now properly sent during balance loading until we fetch all token balances or get an invalid response. This fixes an issue when loading erc20 balances for accounts with > 100 tokens.

## To Test
- [ ] Import a read only account with a lot of erc20 tokens (e.g. `mhluongo.eth`)
- [ ] Tokens should appear normally in the balances list


kkosiorowska (approved):
Looks good ✅

background/lib/alchemy.ts
```diff
@@ -150,33 +151,59 @@ export async function getAssetTransfers(
...
+    response = await fetchAndValidate(addressOnNetwork.address, currentPageKey)
+
+    if (!response) {
+      break
+    }
+
+    balances.push(...response.tokenBalances)
+
+    currentPageKey = response.pageKey
+  } while (currentPageKey)
```
0xDaedalus, line 197: I would like to push back pretty strongly here and request for us to implement a follow-up to this merge with a recursive solution here rather than using `do ... while`.

The fact that we needed to drop an eslint-disable in here is a good hint that this does not align with our current code style guidelines - but even without an `await` I think we should prefer to use modern js patterns like generators or recursion (although that is not exactly a "modern" language feature) to `while` loops.  In general I would like to avoid `while` loops whenever we can - especially when we're doing anything async inside the loop.

PR #3084, opened 2023-02-24
Don't show add network popup if SUPPORT_CUSTOM_NETWORKS is not enabled

0xDaedalus:
### To Test
- [ ] Go to chainlist.org
- [ ] Attempt to add a chain
- [ ] You should not see a tally popup asking you to confirm adding the chain

And Separately
- [ ] Checkout v0.25.0
- [ ] Import a wallet
- [ ] Add a network via chainlist
- [ ] Upgrade to this branch
- [ ] Ensure that you don't see the added network in the menu protocol list

Latest build: [extension-builds-3084](https://github.com/tahowallet/extension/suites/11191886046/artifacts/572044288) (as of Fri, 24 Feb 2023 19:50:40 GMT).

### DIFF
--- a/background/services/provider-bridge/index.ts
+++ b/background/services/provider-bridge/index.ts
@@ -36,6 +36,7 @@ import {
 } from "./utils"
 import { toHexChainID } from "../../networks"
 import { TALLY_INTERNAL_ORIGIN } from "../internal-ethereum-provider/constants"
+import { FeatureFlags, isEnabled } from "../../features"
 
 type Events = ServiceLifecycleEvents & {
   requestPermission: PermissionRequest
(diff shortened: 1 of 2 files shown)


background/services/provider-bridge/index.ts
```diff
@@ -515,6 +516,15 @@ export default class ProviderBridgeService extends BaseService<Events> {
           )
 
         case "wallet_addEthereumChain": {
+          if (!isEnabled(FeatureFlags.SUPPORT_CUSTOM_NETWORKS)) {
+            // Attempt to switch to a chain if its one of the natively supported ones - otherwise fail
+            return await this.internalEthereumProviderService.routeSafeRPCRequest(
+              method,
+              params,
```
hyphenized, line 523: Good thinking, these params haven't been validated though which I believe is something the long name service expects

0xDaedalus: I think this is fine actually - reason being if the feature flag is off the IEP service will only attempt to switch to the network before failing so we don't actually care about any of the parameters besides chainID.

hyphenized (approved):
Looks good!

PR #3088, opened 2023-02-27
Disappear an insufficient funds warning

### LINKED ISSUE #2921: Insufficient funds warning is not disappearing
### Discord Discussion Link

_No response_

Chrome

https://user-images.githubusercontent.com/20949277/214549920-df406d09-2cf4-4225-b591-d18d6cffa095.mov

### What did not work?

Insufficient funds warning is not disappearing even after gas settings are changed and there is enough base asset to pay for gas.

### Version

v0.18.9

### Relevant log output

_No response_

kkosiorowska:
Closes #2921 

This PR disappears insufficient funds warning after gas settings are changed and there is enough base asset to pay for gas.

## UI

https://user-images.githubusercontent.com/23117945/221586264-4e84d0bd-9c5d-4a71-8694-568026dd566c.mov

### DIFF
--- a/background/services/enrichment/transactions.ts
+++ b/background/services/enrichment/transactions.ts
@@ -244,7 +244,16 @@ export default async function resolveTransactionAnnotation(
   // If the wallet doesn't have enough base asset to cover gas, push a warning
   if (gasFee + (transaction.value ?? 0n) > baseAssetBalance) {
     txAnnotation.warnings ??= []
-    txAnnotation.warnings.push("insufficient-funds")
+    if (!txAnnotation.warnings.includes("insufficient-funds")) {
+      txAnnotation.warnings.push("insufficient-funds")
+    }
+  } else if (txAnnotation.warnings) {
+    const idx = txAnnotation.warnings.findIndex(
+      (warning) => warning === "insufficient-funds"
+    )
+    if (idx >= 0) {
+      delete txAnnotation.warnings[idx]
+    }
   }
 
   // If the transaction has been mined, get the block and set the timestamp


background/services/enrichment/transactions.ts
```diff
@@ -244,7 +244,16 @@ export default async function resolveTransactionAnnotation(
...
+    if (!txAnnotation.warnings.includes("insufficient-funds")) {
+      txAnnotation.warnings.push("insufficient-funds")
+    }
+  } else if (txAnnotation.warnings) {
+    const idx = txAnnotation.warnings.findIndex(
+      (warning) => warning === "insufficient-funds"
+    )
+    if (idx >= 0) {
+      delete txAnnotation.warnings[idx]
+    }
```
hyphenized, line 256: I think we should use a filter here, deleting an item from the array this way also leaves `undefined` in the item's place which could cause bugs further down the road

background/services/enrichment/transactions.ts
```diff
@@ -244,7 +244,16 @@ export default async function resolveTransactionAnnotation(
   // If the wallet doesn't have enough base asset to cover gas, push a warning
   if (gasFee + (transaction.value ?? 0n) > baseAssetBalance) {
     txAnnotation.warnings ??= []
-    txAnnotation.warnings.push("insufficient-funds")
+    if (!txAnnotation.warnings.includes("insufficient-funds")) {
+      txAnnotation.warnings.push("insufficient-funds")
+    }
+  } else if (txAnnotation.warnings) {
```
hyphenized, line 250: We can move the `txAnnotation.warnings ??= []` outside the parent conditional block and change this to
```suggestion
  } else {
```
This way we can treat this handler as "either we have enough to cover gas costs or not" instead of "if we don't have enough for gas but have warnings..."

hyphenized (changes requested):
Looks great! Works fine on my end, but noticed a small delay in the video 👀. Left some code style comments

jagodarybacka (changes requested):
I think that spinner is not looking good, it moves the content below it. 
If we want a loader for insufficient funds I would suggest using skeleton loader and style it to be exactly the same size as the warning box. Idealy I think loader should only be visible if there was a warning before, if there was no warning then we shouldn't show loader and just show warning immediately we know there is a need. That way we can avoid unnecessary moving UI.

kkosiorowska:
> If we want a loader for insufficient funds I would suggest using skeleton loader and style it to be exactly the same size as the warning box. Idealy I think loader should only be visible if there was a warning before, if there was no warning then we shouldn't show loader and just show warning immediately we know there is a need. That way we can avoid unnecessary moving UI.

We don't want to use exactly a loader for insufficient funds. The addition of the spinner here was intended to solve the delay issue Jorge mentioned. But yes it should be more fluent. I made some improvements and used `SharedSkeletonLoader`. @VladUXUI  please tell what you think.


https://user-images.githubusercontent.com/23117945/222073980-18942d28-82d7-4a40-896c-5007e988fded.mov

jagodarybacka (commented):
Better but not perfect - @VladUXUI what do you think about moving the insufficient funds warning below other UI so it is not moving other elements?
image

VladUXUI:
> 

makes a lot of sense to move the error under! very nice

VladUXUI:
Ghost loader should not be used for content that is going to disappear. Ghost loader are used when content is taking longer to load, and it gives the user the feeling that it's moving faster than it actually is.

VladUXUI:
I think we should put the loading icon in the Sign button, and not above/bellow the network fee. As sign is actually blocked until the amount is checked right?

VladUXUI:
CleanShot 2023-03-06 at 10 24 37

Maybe something like this?
ignore the design of the loader gif itself

kkosiorowska:
> I think we should put the loading icon in the Sign button, and not above/bellow the network fee. As sign is actually blocked until the amount is checked right?

The sign button is blocked when transaction data is not yet ready. There is a problem with the display/disappearance of warnings here. Because we have to wait until the transaction data is ready to display the warning.

kkosiorowska:
> Maybe something like this? ignore the design of the loader gif itself

I think we're able to do something similar. I will try to do it 😀

kkosiorowska:
Improvements have been made. Ready for another look. While doing the task, the idea came up to unlock the reject button.  This will be done in a separate ticket.

https://user-images.githubusercontent.com/23117945/223435001-e0fa0019-8b28-4f57-a5c2-d6bb3d4d8213.mov

jagodarybacka (approved):
Looks good, works as intended with signing transactions, personal signs seem to be not broken 😉 
🚀

PR #3101, opened 2023-03-01
Logging - Log UI errors caught by main error boundary

hyphenized:
This should help with debugging redux and state errors that aren't necessarily caused by background actions

## To Test
- [ ] Add a selector that will access an invalid nested property to the Wallet page, e.g . ```ts useBackgroundSelector((state) => (state as any).something.that.will.throw)```
- [ ] Create a production build
- [ ] Open the wallet page
- [ ] Export logs, you should see the Wallet page in the stack trace

Latest build: [extension-builds-3101](https://github.com/tahowallet/extension/suites/11291295174/artifacts/579279456) (as of Wed, 01 Mar 2023 21:22:40 GMT).


kkosiorowska (approved):
Works well 🚀

PR #3114, opened 2023-03-04
Signing - Prevent signing button from staying locked when attempting to sign data

hyphenized:
Fixes a bug that would cause both reject and sign buttons to remain locked when attempting to sign a message

## To Test
- [x] Head to https://login.xyz, try to sign in with your wallet

Latest build: [extension-builds-3114](https://github.com/tahowallet/extension/suites/11352040150/artifacts/583581270) (as of Sat, 04 Mar 2023 21:05:20 GMT).

### DIFF
--- a/ui/components/Signing/Signer/TransactionButton.tsx
+++ b/ui/components/Signing/Signer/TransactionButton.tsx
@@ -1,10 +1,12 @@
 import React, { ReactElement, useEffect, useRef, useState } from "react"
+import { selectSigningData } from "@tallyho/tally-background/redux-slices/signing"
 import { selectIsTransactionLoaded } from "@tallyho/tally-background/redux-slices/selectors/transactionConstructionSelectors"
 import { useBackgroundSelector, useDebounce } from "../../../hooks"
 import SharedButton, {
   Props as SharedButtonProps,
 } from "../../Shared/SharedButton"
 
+// TODO: Rename this to signing button
 export default function TransactionButton({
   type,
   size,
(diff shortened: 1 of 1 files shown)


ui/components/Signing/Signer/TransactionButton.tsx
```diff
@@ -13,9 +15,12 @@ export default function TransactionButton({
...
 }: SharedButtonProps & { reactOnWindowFocus?: boolean }): ReactElement {
-  const isTransactionDataReady = useBackgroundSelector(
-    selectIsTransactionLoaded
-  )
+  const hasTransactionLoaded = useBackgroundSelector(selectIsTransactionLoaded)
+
+  const hasSignDataRequest = useBackgroundSelector(selectSigningData)
+
+  const isTransactionDataReady = hasTransactionLoaded || hasSignDataRequest
+
```
Shadowfiend, line 23: So looking more closely, I'm trying to remember—are we able to eliminate this logic entirely? IIRC the new signing flow doesn't render the signing button at all until a transaction is known loaded, which means if we're rendering this component all we have to worry about is the delay.

That also seems like a reasonable expectation in terms of keeping the responsibilities of this component tightly scoped to handling the timeout.

Thoughts?

(PS: that should absolutely not block the hotfix.)

hyphenized: These new checks came as part of a feature to [block the sign transaction button while we resolve transaction updates](https://github.com/tahowallet/extension/pull/3060). Gas estimations for example, will dispatch a `updateTransaction` event which changes the transaction state to `pending` through a redux thunk. I think we should keep the transaction state and the document blur/focus handling outside of this button but I'd prefer to do that in a different PR as to not block this.

Shadowfiend: Unfortunately moving readiness out of here has its own implications (risking uncoordinated changes across different signers due to the fact that the sign button is rendered by each signer individually; see my note on the original PR on that). So I think if we need to depend on transaction readiness we should probably unify it even if we don't look it up directly here--feels like it's doable as part of the signature details, which can then be fed to a parameter on this component (maybe a `signingDataReady` field on the details that the signers can forward to this component).  

Doing it that way would also let us put the whole transaction details section in a loading state during an edit... Which might be reasonable? I'm too far down the rabbit hole here but it does seem confusing to eg edit gas or approval allowance and then have the sign button disable temporarily as the *only* indicator that we're processing that edit...

Shadowfiend (approved):
Ok got this tested and confirmed. Also confirmed signing buttons for transactions continue to work as expected.

In fact, I don't see the time-limited disable that we were trying to restore as behavior on `main` or in this branch, but that's a much smaller issue here. Let's rock and roll.

Issue #3128, opened 2023-03-08
Abilities page contains abilities from unimported accounts


jagodarybacka:
Current: If you have an ability on one of your ledger accounts but you decide to not import that account then ability will be imported anyway for this address.

Expected: abilities are fetched only for Ledger accounts that were selected by user during onboarding

Should be fixed together with: https://github.com/tahowallet/extension/issues/3100

PR #3129, opened 2023-03-08
Omit abilities from unknown accounts

jagodarybacka:
Ref #3128

### What
If an account is not imported let's hide abilities for that account. Hotfix for the problem of fetching accounts from the ledger before they are selected to be imported.

### Testing
1. You have to have any abilities on one of your Ledger accounts
2. import Ledger, let address with abilities load on the list but don't select it, select another one
3. Go to abilities page - it should be loading as usual and there should be no ability from the account you have ignored in prev step.

Latest build: [extension-builds-3129](https://github.com/tahowallet/extension/suites/11428154896/artifacts/589009381) (as of Wed, 08 Mar 2023 13:18:42 GMT).


kkosiorowska (approved):
LGTM ✅

PR #3131, opened 2023-03-09
Daylight - Only poll ablities for imported addresses

### LINKED ISSUE #3100: Further optimize daylight requests when connecting a ledger
Ideally we do not want to request abilities for addresses tied to a ledger until those addresses are actually imported into the wallet.   Current state is that the extension requests abilities before ledger addresses are imported (see below comment & video)

_Originally posted by @kkosiorowska in https://github.com/tahowallet/extension/pull/3095#pullrequestreview-1319275833_

hyphenized:
Prevents abilities polling for all ledger addresses discovered instead of only selected addresses when adding new accounts

Closes #3100

Refs:
- https://github.com/tahowallet/extension/pull/3095#pullrequestreview-1319275833

## To Test
- [x] Onboard a ledger, before reaching the address selection screen, monitor the network requests
- [x] There should not be a request for every derived address
- [x] Everything else should work the same

Latest build: [extension-builds-3131](https://github.com/tahowallet/extension/suites/11447290997/artifacts/590431383) (as of Thu, 09 Mar 2023 07:49:11 GMT).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -654,6 +654,8 @@ export default class Main extends BaseService<never> {
               network,
             }
             await this.chainService.addAccountToTrack(addressNetwork)
+            this.abilitiesService.getNewAccountAbilities(address)
+
             this.store.dispatch(loadAccount(addressNetwork))
           })
         )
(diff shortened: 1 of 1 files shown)


jagodarybacka (approved):
Works fine, all types of accounts checked ✅

jagodarybacka:
@kkosiorowska please take a look before merging 😉

kkosiorowska (approved):
LGTM ✅

PR #3137, opened 2023-03-10
Token Balances - Prevent discovered tokens from shadowing known assets balances

hyphenized:
Temporary fix until we refactor the account balances map in the redux slice. Since balances are currently indexed by symbol, a discovered token balance could overwrite a different token balance.

## To Test
- [ ] Start off from the main branch
- [ ] Load any wallet that has two tokens with the same name, bnbspider2.crypto has two different WETH tokens in Polygon
- [ ] You should not see the balance for [this token](https://polygonscan.com/token/0x7ceb23fd6bc0add59e62ac25578270cff1b9f619)
- [ ] Switch to this branch, reload the extension, you should see the missing balance after a few seconds

Latest build: [extension-builds-3137](https://github.com/tahowallet/extension/suites/11578430471/artifacts/599887114) (as of Wed, 15 Mar 2023 13:17:40 GMT).


background/main.ts
```diff
@@ -976,6 +977,26 @@ export default class Main extends BaseService<never> {
...
+            // Only exclude if we have one balance that could be overwritten
+            balances.some(
+              ({ assetAmount: { asset: otherAsset } }) =>
+                otherAsset.symbol === asset.symbol &&
+                isSmartContractFungibleAsset(otherAsset) &&
+                !sameEVMAddress(
+                  asset.contractAddress,
+                  otherAsset.contractAddress
+                )
+            )
```
kkosiorowska, line 995: Non-blocking comment
I'm thinking whether it might be worth moving this to a separate function. 🤔

kkosiorowska (dismissed):
I have tested it and it looks to be working properly.  In my case, I had to wait a little longer to see WETH.

jagodarybacka (commented):
I would say - let's do it properly rather than as a hotfix. So we should be able to display both and mark duplicates. 
cc @VladUXUI please link a design/issue (how we should show duplicated tokens) when it will be ready

Shadowfiend:
Just to confirm my understanding here is correct: we're talking about potentially having e.g. a balance of 3 WETH, and we auto-discover a token that is definitely not the real WETH that shows an apparent balance of, say, 30 WETH, and it overwrites the real WETH?

If so, we definitely need to release this as a hotfix.

background/main.ts
```diff
@@ -976,6 +977,26 @@ export default class Main extends BaseService<never> {
...
 
         balances.forEach((balance) => {
+          const { asset } = balance.assetAmount
+          // Do not update balances for untrusted assets
+          // FIXME: Balances should not be indexed by symbol in redux
+          if (
+            isSmartContractFungibleAsset(asset) &&
+            !asset.metadata?.tokenLists?.length &&
+            // Only exclude if we have one balance that could be overwritten
+            balances.some(
```
0xDaedalus, line 987: This might get a bit costly for accounts with lots of untrusted tokens - I wonder if we can avoid trending towards O(n^2) by constructing a set of symbols that have more than 1 balance associated with them - and then using that as a first check?

kkosiorowska: In my case, the delay in seeing the WETH balance is considerable.

background/main.ts
```diff
@@ -975,7 +975,33 @@ export default class Main extends BaseService<never> {
...
+          .sort((balance, otherBalance) => {
+            const asset = balance.assetAmount.asset as SmartContractAsset
+            const other = otherBalance.assetAmount.asset as SmartContractAsset
+
+            return (
+              (other.metadata?.tokenLists?.length ?? 0) -
+              (asset.metadata?.tokenLists?.length ?? 0)
+            )
+          })
+          .forEach((balance) => sortedBalances.unshift(balance))
```
jagodarybacka, line 1002: Isn't it just reversing the sorted array?

PR #3140, opened 2023-03-10
Add  basic tests for abilities

kkosiorowska:
This PR adds tests for AbilitiesService. PR tests the performance of refreshing abilities and adding them.

Latest build: [extension-builds-3140](https://github.com/tahowallet/extension/suites/11630118352/artifacts/603559665) (as of Fri, 17 Mar 2023 10:57:49 GMT).


setupJest.ts
```diff
@@ -44,3 +44,23 @@ Object.defineProperty(window, "fetch", {
...
+    clear() {
+      store = {}
+    },
+    removeItem(key: string) {
+      delete store[key]
+    },
+  }
+})()
+
+Object.defineProperty(window, "localStorage", { value: localStorageMock })
```
hyphenized, line 66: We can remove this, our version of Jest already [comes with support for localStorage](https://github.com/jsdom/jsdom/blob/master/Changelog.md#11120)

background/services/abilities/tests/index.intedration.test.ts
```diff
@@ -0,0 +1,141 @@
+import sinon from "sinon"
```
0xDaedalus, line 1: Slight type on the filename here - it should be index.`integration`.test.ts

background/services/abilities/tests/index.intedration.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      const lastFetchTime = +new Date(2023, 0, 1, 8)
+      localStorage.setItem("LAST_ABILITY_FETCH_TIME", lastFetchTime.toString())
+      const stub = sandbox
+        .stub(chainService, "getAccountsToTrack")
+        .callsFake(async () => [])
+
+      await abilitiesService.refreshAbilities()
+      expect(stub.called).toBe(true)
+    })
+    it("should not refresh abilities", async () => {
```
0xDaedalus, line 52: Let's add the condition under which this test passes to the test description (as well as the one above)

0xDaedalus (changes requested):
Code looks good - just some description and meta-level stuff to fix here.

I only commented on one of the `it` blocks but I do think that there's a real strong case for structuring test names like `should X when Y` or `should X if Y`.  Particularly when there are conditions when opposite behavior can be expected (like emitting / not emitting, refreshing / not refreshing, etc..)

0xDaedalus (approved):

PR #3145, opened 2023-03-13
Add possibility to sort abilities

kkosiorowska:
This PR adds a new filter for abilities. Abilities are sorted by default by using property `magic`. This means that the most interesting them should display first. Because of this, many users miss out on new activities. A [fix](https://github.com/tahowallet/extension/pull/3142) has also been prepared to make new incoming abilities display before old ones. However, an additional filter has been created to make it easier for users to browse abilities.

**UI**

Screenshot 2023-03-13 at 15 29 19

**TBD**
- Naming of filters
- Sorting by date of opening and closing abilities

### DIFF
--- a/background/abilities.ts
+++ b/background/abilities.ts
@@ -49,6 +49,7 @@ export type Ability = {
   imageUrl?: string
   openAt?: string
   closeAt?: string
+  createdAt: string
   completed: boolean
   removedFromUi: boolean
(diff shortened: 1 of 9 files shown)


kkosiorowska:
The approach to updating abilities will be changed. Let's fix all issues in one PR. https://github.com/tahowallet/extension/pull/3375

PR #3146, opened 2023-03-13
E2E Tests - Intercept posthog requests in e2e tests

hyphenized:
Prevents sending off `NEW_INSTALL` and other analytics events to posthog while executing e2e tests

## To Test
- [ ] Run e2e tests locally, go to posthog dashboard and confirm posthog isn't receiving events

Latest build: [extension-builds-3146](https://github.com/tahowallet/extension/suites/11530117254/artifacts/596450202) (as of Mon, 13 Mar 2023 17:22:51 GMT).


puppycodes:
Another way to approach this also might be to set a specific filter on the Posthog side and push an event so the e2e test is fully completed with Posthog api ping and all. By pushing a certain prop and filtering by it we can exclude it from normal analytics but also check its working.

https://posthog.com/tutorials/filter-internal-users

0xDaedalus (approved):

PR #3153, opened 2023-03-14
Delete accounts for a given network when removing that network.

0xDaedalus:
Since we rely on fetching existing accounts to set up initial state - if we have accounts on corresponding to a network without the corresponding network - our state setup fails on application load.  To prevent this - we should delete accounts for a given network along with the rpcUrls, base assets, and networks when removing the network.

### To Test
- [x] Flip on SUPPORT_CUSTOM_NETWORKS
- [x] Add a network
- [x] Remove that network
- [x] Reload the extension
- [x] You should not see a wall of errors in the console relating to the deleted network in the console.

Latest build: [extension-builds-3153](https://github.com/tahowallet/extension/suites/11549582999/artifacts/597842979) (as of Tue, 14 Mar 2023 12:16:27 GMT).


background/services/chain/db.ts
```diff
@@ -256,12 +256,21 @@ export class ChainDatabase extends Dexie {
...
           this.baseAssets.where({ chainID }).delete(),
           this.rpcUrls.where({ chainID }).delete(),
         ])
+
+        // @TODO - Deleting accounts inside the Promise.all does not seem
+        // to work, figure out why this is happening and parallelize if possible.
+        const accountsToTrack = await this.accountsToTrack
+          .toCollection()
+          .filter((account) => account.network.chainID === chainID)
+        return accountsToTrack.delete()
```
jagodarybacka, line 272: do we need to return anything from this function? Also n NFTs database I was not using `.toCollection` and did `.delete()` right after the filter, not sure if this has any significant effect, probably no 🤔

jagodarybacka (approved):
Works fine ✅ If needed let's refactor later

PR #3154, opened 2023-03-14
Switch to a given network if adding a network that is already added.

0xDaedalus:
Since a given dapp does not necessarily have a sense of what networks are supported within the wallet (and since some dapps rely exclusively on wallet_addEthereumChain) we should support the functionality of switching to a network when receiving a wallet_addEthereumChain message with a chainID that we already support.

### To Test
- [x] Navigate to chainlist.org
- [x] Add the Ethereum Network to wallet
- [x] You should not see a popup
- [x] Add the Avalanche Network to wallet
- [x] You should not see a popup
- [x] Open the wallet popover from the Taho icon—Avalanche Network should be selected.
- [x] In Application/Storage/IndexedDB/internal-ethereum-provider/currentNetwork - you should see the network of `https://chainlist.org` listed as `Avalanche`

Latest build: [extension-builds-3154](https://github.com/tahowallet/extension/suites/14040246805/artifacts/784146217) (as of Tue, 04 Jul 2023 02:33:28 GMT).


0xDaedalus:
Converting back to draft pending some product discussion

hyphenized:
I know this is a draft PR but this should fix the issue with the flaky test, the `wait for next tick` assertion broke because it relied too much on the structure of the code (and the # of macro tasks).
https://gist.github.com/hyphenized/4ce1e44767c9f320d9470e008242b0cd

background/services/provider-bridge/tests/index.unit.test.ts
```diff
@@ -172,5 +179,58 @@ describe("ProviderBridgeService", () => {
...
+          blockExplorerUrls: ["https://etherscan.io"],
+        },
+        "0xd8da6bf26964af9d7eed9e03e53415d37aa96045",
+        "some site",
+        "favicon.png",
+      ]
+
+      const { enablingPermission } = BASE_DATA
+
+      jest.spyOn(featureFlags, "isEnabled").mockImplementation(() => true)
```
hyphenized, line 200: I don't think this is causing any issues but we don't need to stub feature flags here anymore. n/b

hyphenized (commented):
Works well, can we remove the "Network already added screen"?

Shadowfiend:
That screen will only appear once #3481 lands if the user adds an existing network, right? In this flow it never appears because we skip displaying the popup altogether.

With that said, are there any remaining blockers here or are you good to merge?

Shadowfiend:
Yes, this PR changes that :) The `return` for this case is moved above the `showExtensionPopup` call in https://github.com/tahowallet/extension/pull/3154/files#diff-cebbcecfbd654e43999eb66e8653ca07d814ec9972cffb29a2791bb1b3730740 .

hyphenized:
Ah I see, I thought we were unintentionally discarding a feature 😅

hyphenized (approved):
👑 👑 👑

PR #3157, opened 2023-03-15
Does not show price impact when data is not ready

### LINKED ISSUE #3156: In-wallet swaps - no price information & no price impact warning
### Discord Discussion Link

https://discord.com/channels/808358975287722045/1037670242421321769/1085140901564784650

Chrome

Swap MATIC to FTM on Polygon

### What did not work?

Price information can't be calculated -> I believe a user should not be able to proceed with the swap:

Screen Shot 2023-03-15 at 09 13 27

However, the user can proceed. There are moments where a price impact is briefly displayed, but this is for ~ 1 second and then it's showing 'no price information again'.

Screen Shot 2023-03-15 at 09 18 16

Proceeding gives no warning as well (changed it to a 1 matic swap)
Screen Shot 2023-03-15 at 09 20 17

And tx is executed: https://polygonscan.com/tx/0xf800576a2d1b205a86f21684057eac75d20ac01b0966bfc35b5e2962bae01c8a

See below recording at 00:18

https://user-images.githubusercontent.com/71384922/225248135-6002e068-04e5-48ec-b902-a28d019b37aa.mov

Logs:
[logs_v0_28_0__2023-03-15T09_14_08+01_00.txt](https://github.com/tahowallet/extension/files/10977378/logs_v0_28_0__2023-03-15T09_14_08%2B01_00.txt)

### Version

v0.28.0

### Relevant log output

_No response_

kkosiorowska:
Closes #3156 
This PR solves the problem of showing the impact of price when data isn't ready. Previously, when loading the data old price impact showed for 1 second. The changes involve showing the price detail when the data is loaded and up to date.

In the task noticed:
> Price information can't be calculated -> I believe a user should not be able to proceed with the swap:
> 225247416-934d1c0b-ffcc-4b21-bc47-132cf5de6984

However, I don't think we want that. Let me know what you think.

## UI
Before
https://user-images.githubusercontent.com/23117945/225283719-887f22e2-fc0d-48e2-999e-be4c0d94d2d9.mov

After
https://user-images.githubusercontent.com/23117945/225283623-8024cb87-d972-4aa3-bf44-4361f84e05a0.mov

### DIFF
--- a/ui/components/Shared/PriceDetails.tsx
+++ b/ui/components/Shared/PriceDetails.tsx
@@ -36,40 +36,44 @@ export default function PriceDetails(props: PriceDetailsProps): ReactElement {
 
   return (
     <div className="simple_text content_wrap">
-      {!isLoading && amountMainCurrency === undefined ? (
-        t("noAssetPrice")
-      ) : (
+      {!isLoading && (
         <>
-          {amountMainCurrency === "0.00" && "<"}
-          {mainCurrencySign}
-          {amountMainCurrency || "0.00"}
-          {!!priceImpact && priceImpact > 1 && (
-            <span
-              data-testid="price_impact_percent"
-              className="price_impact_percent"
-            >
-              ({formatPriceImpact(priceImpact)}%
-              <SharedTooltip
-                width={180}
-                height={27}
-                horizontalPosition="left"
-                IconComponent={() => (
-                  <SharedIcon
-                    width={16}
-                    icon="icons/m/info.svg"
-                    color={`var(--${getPriceImpactColor(priceImpact)})`}
-                    customStyles="margin-left: -5px;"
-                  />
-                )}
-              >
-                <div>
-                  {t("priceImpactTooltip.firstLine")}
-                  <br />
-                  {t("priceImpactTooltip.secondLine")}
-                </div>
-              </SharedTooltip>
-              )
-            </span>
+          {amountMainCurrency === undefined ? (
+            t("noAssetPrice")
+          ) : (
+            <>
+              {amountMainCurrency === "0.00" && "<"}
+              {mainCurrencySign}
+              {amountMainCurrency || "0.00"}
+              {!!priceImpact && priceImpact > 1 && (
+                <span
+                  data-testid="price_impact_percent"
+                  className="price_impact_percent"
+                >
+                  ({formatPriceImpact(priceImpact)}%
+                  <SharedTooltip
+                    width={180}
+                    height={27}
+                    horizontalPosition="left"
+                    IconComponent={() => (
+                      <SharedIcon
+                        width={16}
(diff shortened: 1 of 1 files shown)


kkosiorowska:
Please @VladUXUI  to confirm the visual change.

jagodarybacka (dismissed):
Works fine ✅ 
I'm wondering if it would look a little bit better if we could hide the price behind a loader when swap is loading - while loading the content there is stale anyway and will be replaced after swap is loaded, right?

Quick demo

https://user-images.githubusercontent.com/20949277/225309375-3275bd78-6bcf-424d-bd65-589d8342f7b0.mov

Naxsun:
Reading through this now again:

_However, I don't think we want that. The price is not calculated because we do not have information from the API about the price point. Let me know what you think._

I don't think we should allow users to proceed with a swap if there's no price information from the api. 
- No price information -> no price impact warning

I think we'll have a lot of users that expect an in-wallet swap will just give them an exchange rate in line with market prices. 

Without a price impact warning we're relying on users to do the comparison between market prices and in-wallet swap quotes themselves, which isn't the way to go imo.

kkosiorowska:
> I don't think we should allow users to proceed with a swap if there's no price information from the api.
> 
> * No price information -> no price impact warning
> 
> I think we'll have a lot of users that expect an in-wallet swap will just give them an exchange rate in line with market prices.
> 
> Without a price impact warning we're relying on users to do the comparison between market prices and in-wallet swap quotes themselves, which isn't the way to go imo.

hmm I think this is a good point of view. I am afraid that by blocking this many users may be negatively affected. What I mean is that we can block a lot of pairs this way and it can make it difficult to use the wallet. Another thing is that this data comes from the API and is not dependent on us. We don't control it. 

Let us know what you think and if you see another solution for this. @jagodarybacka @hyphenized @0xDaedalus

VladUXUI:
This is looking good!
Blocking swaps isn't something we should do, we don't have enough understanding of how that will affect users.
We have a swap "redesign" coming soon we can touch more on that then

VladUXUI:
For now, what we can do is add a warning that we don't have price, like the screen bellow.
image

[Figma](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=60%3A2615&t=hT8ftt0WvRC9Tsyy-4)

VladUXUI:
> Works fine ✅ I'm wondering if it would look a little bit better if we could hide the price behind a loader when swap is loading - while loading the content there is stale anyway and will be replaced after swap is loaded, right?
> 
> Quick demo

I can't QA the PR atm, and hard to tell from video if current iteration "feels" good.
There might be a bit too many things appearing and disappearing with this version. Maybe the dog appearing is enough?

Let's stick to how Karolina has done it and we can revisit this later as well.

PR #3160, opened 2023-03-15
Handle case where there are no cached assets for given network

jagodarybacka:
### What

Fix attempting to filter on `undefined` - while adding a new network with Chainlist there are no cached assets yet

### Testing

- [ ] flip `SUPPORT_CUSTOM_NETWORKS`
- [ ] open background page
- [ ] add new network with chainlist
- [ ] you shouldn't see an error in the console about the line changed in this PR

Latest build: [extension-builds-3160](https://github.com/tahowallet/extension/suites/11578827244/artifacts/599913023) (as of Wed, 15 Mar 2023 13:32:41 GMT).

### DIFF
--- a/background/services/indexing/index.ts
+++ b/background/services/indexing/index.ts
@@ -447,8 +447,7 @@ export default class IndexingService extends BaseService<Events> {
             ({ smartContract: { contractAddress } }) => contractAddress
           )
         )
-        const cachedAssets =
-          this.cachedAssets[addressOnNetwork.network.chainID] ?? []
+        const cachedAssets = this.getCachedAssets(addressOnNetwork.network)
 
         const otherActiveAssets = cachedAssets
           .filter(isSmartContractFungibleAsset)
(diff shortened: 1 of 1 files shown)


background/services/indexing/index.ts
```diff
@@ -447,7 +447,8 @@ export default class IndexingService extends BaseService<Events> {
             ({ smartContract: { contractAddress } }) => contractAddress
           )
         )
-        const cachedAssets = this.cachedAssets[addressOnNetwork.network.chainID]
+        const cachedAssets =
+          this.cachedAssets[addressOnNetwork.network.chainID] ?? []
```
hyphenized, line 451:
```suggestion
        const cachedAssets = this.getCachedAssets(addressOnNetwork.network)
```

hyphenized (changes requested):
Good catch! There's a class method we can use for this, let's change the other instances as well

hyphenized (approved):
Looks good

Issue #3164, opened 2023-03-16
Swaps become unusable if coingecko API rate limit is hit.


0xDaedalus:
https://user-images.githubusercontent.com/94649004/225567091-eee05e5c-d5c1-40cd-99e3-996b8c0a111d.mov

### To Recreate
- [ ] Quickly preview swapping ~10 asset pairs
- [ ] At some point - you will see `429` http errors come back from coingecko
- [ ] Once this happens the Swap screen will be stuck on `Fetching Price`.  Exiting out of the screen and going back will not resolve the issue - and switching the buy asset will not be possible either.

PR #3165, opened 2023-03-16
Disable 429 Retry when querying CoinGecko API.

0xDaedalus:
Closes https://github.com/tahowallet/extension/issues/3164

This change was introduced to resolve an issue where 429 responses from CoinGecko would block swap functionality in the application.  This was happening because the fetchJson function we were using automatically waits and retries requests that come back with a 429 status code.  Due to the high likelihood of us being rate limited when hitting the coingecko api - this is actually undesirable functinality since we want to fail fast in the case that coingecko is unavailable.

By passing in a throttleCallback to fetchJson that always resolves to false  - we can rely on the (already existing) code around the function call to handle cases where we get a 429 or a bad response.  This results in a snappier user experience when querying prices - and does not break any existing functionality - all that is changing is that we now handle 429's how we would handle 400's and 500's - instead of special-casing them.

https://user-images.githubusercontent.com/94649004/225576313-ed20e9d2-cd81-4a51-80f4-9ef31f420b24.mov

### To Test
- [ ] Rapidly swap buy assets or sell asset amounts in the swap page
- [ ] The swap totals should load with `No Price Information` rather than hanging forever

Latest build: [extension-builds-3165](https://github.com/tahowallet/extension/suites/11602728521/artifacts/601621193) (as of Thu, 16 Mar 2023 10:48:43 GMT).

### DIFF
--- a/background/lib/prices.ts
+++ b/background/lib/prices.ts
@@ -42,7 +42,11 @@ export async function getPrices(
   const url = `${COINGECKO_API_ROOT}/simple/price?ids=${coinIds}&include_last_updated_at=true&vs_currencies=${currencySymbols}`
 
   try {
-    const json = await fetchJson(url)
+    const json = await fetchJson({
+      url,
+      // Prevent throttling
+      throttleCallback: async () => false,
+    })
     // TODO fix loss of precision from json
     // TODO: TESTME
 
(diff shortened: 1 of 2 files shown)


kkosiorowska (approved):
Works well for me 🚀

Issue #3144, opened 2023-03-13
Spike for import account with JSON 


jagodarybacka:
Let's understand the lift of doing JSON import

jagodarybacka:
We don't have to document it here if we don't want to, it's just a task to make it visible as an actionable item

Shadowfiend:
Unless we have specific concerns, we should document in public, yes.

jagodarybacka:
Spike has been finished, findings in the [PR](https://github.com/tahowallet/extension/pull/3177)

PR #3177, opened 2023-03-17
Private key JSON import

jagodarybacka:
### What
This PR introduces importing accounts with JSON files (keystore files). In this PR there is a new tab next to the plain text private key import tab, for importing private keys with JSON files.

### Testing

1. To test this feature you need a JSON keystore file. If you don't have one you can generate a new wallet on [MyEtherWallet](https://www.myetherwallet.com/wallet/create/software?type=keystore) website. 
2. Install the wallet with the build from this PR
3. Select private key option on onboarding
4. Put your file and its password, submit and wait a few seconds for the import
5. Your account should be visible in the extension

### Conclusions for the [spike](https://github.com/tahowallet/extension/issues/3144) 

Findings:
- There was a problem finding even one user-facing wallet that allows export account as a JSON file - this format is no longer popular and I was able to identify only one interface that allowed me to create such a file. I couldn't find any software that allows exporting existing wallets as JSON.
- Some wallets are considering JSON format legacy and unsafe. MyEtherWallet even supporting JSON wallet creation on the website is not allowing to import it to their mobile app and they have a big warning about it not being safe to do.
- Once I had the JSON file the implementation of this feature was easy and straight-forward

I see not much harm in having an import feature, I think it will be nice to have but much less used than plain text private keys. For sure we shouldn't bother implementing JSON export feature. 

### Demo

https://user-images.githubusercontent.com/20949277/228903879-685f991d-2c46-4990-a663-f596090853b0.mov

Latest build: [extension-builds-3177](https://github.com/tahowallet/extension/suites/12015664577/artifacts/631221776) (as of Tue, 04 Apr 2023 10:40:41 GMT).


0xDaedalus (approved):
Bravo 🎉 📈

PR #3184, opened 2023-03-24
Switch wallet to newly added network

### LINKED ISSUE #3163: Switch wallet to newly added network after adding a network via `wallet_addEthereumChain`
After adding a network for the first time - the in-wallet network should switch to the given network.

kkosiorowska:
Closes #3163 

This PR updates the network switching logic. When a new network is added, the wallet will be switched to it automatically. 

https://user-images.githubusercontent.com/23117945/227507265-2089b03b-80ab-4e67-a342-4f1a87b28dbf.mov

## To Test
- [ ] add a new custom network
- [ ] check if the wallet has been switched to it

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -75,6 +75,7 @@ import {
   setAccountsSignerSettings,
   toggleCollectAnalytics,
   setShowAnalyticsNotification,
+  setSelectedNetwork,
 } from "./redux-slices/ui"
 import {
   estimatedFeesPerGas,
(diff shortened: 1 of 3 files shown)


jagodarybacka (changes requested):
When user adds one of our built-in networks with chainlist then wallet is not switching to that network, seems like a bad UX - even if we are not adding a new network we should at list select it.

One more thing I'm missing here (but not exactly in the scope of this PR) is some sort of success message or CTA to open a wallet cc @VladUXUI

jagodarybacka (approved):
Works great 👍

PR #3185, opened 2023-03-24
New copy for `Activity`

### LINKED ISSUE #2422: New copy for `Activity` tab when connected to networks alchemy does not support
The won't be accurate since we likely won't have the ability to add historical transactions over time.

image

kkosiorowska:
Closes #2422 

This PR adds a new copy for `Activity`.
## UI
Before
Screenshot 2023-03-24 at 12 41 57

After
Screenshot 2023-03-24 at 12 38 17

### DIFF
--- a/ui/_locales/en/messages.json
+++ b/ui/_locales/en/messages.json
@@ -722,7 +722,7 @@
   },
   "wallet": {
     "activities": {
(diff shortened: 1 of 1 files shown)


hyphenized:
This should only display for networks that are not supported by alchemy 👀

kkosiorowska:
> Will it show all new transactions for this address, or all new transactions made through this Taho instance?
> 
> Copy makes me assume it's -> all new transactions for this address

I'm not quite sure but I think that all new transactions for this address.

0xDaedalus (approved):

PR #3195, opened 2023-03-27
Token Discovery - Remap redux asset balances

hyphenized:
Prevents tokens with the same symbol from overriding existing balances. Changes the way combined amounts data is computed so that balances displayed in overview for e.g. USDC on Mainnet and USDC on Polygon get bundled together only if they both are sourced from a token list (or are network base assets).

## To Test
- [x] Load any wallet that has two or more tokens with the same name, `bnbspider2.crypto`
- [x] You should not be able to see multiple tokens that share the same symbol
- [x] Switch to this branch, reload the extension, you should see all the missing tokens

Latest build: [extension-builds-3195](https://github.com/tahowallet/extension/suites/13697806847/artifacts/757300183) (as of Mon, 19 Jun 2023 08:23:00 GMT).


background/redux-slices/accounts.ts
```diff
@@ -178,13 +181,25 @@ function updateCombinedData(immerState: AccountState) {
...
     }>((acc, combinedAssetAmount) => {
-      const assetSymbol = combinedAssetAmount.asset.symbol
+      const { asset } = combinedAssetAmount
+      /**
+       * Asset amounts can be aggregated if the asset is a base network asset
+       * or comes from a token list, e.g. ETH on Optimism, Mainnet
+       */
+      const canBeAggregated =
+        isNetworkBaseAsset(asset) ||
+        (asset.metadata?.tokenLists?.length ?? 0) > 0
```
jagodarybacka, line 193: Can we add a util function that is checking if an asset has token lists? I feel like we are manually checking it in many places already

background/redux-slices/utils/asset-utils.ts
```diff
@@ -43,15 +44,18 @@ export type AssetDecimalAmount = {
   localizedDecimalAmount: string
 }
 
-function hasChainID(asset: AnyAsset): asset is NetworkBaseAsset {
+/**
+ * All network base assets have a chainID property
+ */
+export function isNetworkBaseAsset(asset: AnyAsset): asset is NetworkBaseAsset {
```
jagodarybacka, line 50: I'm not sure about this change, as we already have a second function that is named `isBuiltInNetworkBaseAsset` and is checking much more things I'm afraid someone will make a mistake and will think that `isNetworkBaseAsset` does the same thing.

I assume that this change is because on custom networks we only have `chainID` as an indicator that this is a base asset?  Can we think about a name that will make it clear which function should be used when? `isAnyNetworkBaseAsset`, `isCustomNetworkBaseAsset`? Lmk what you think 🤔

background/redux-slices/migrations/to-27.ts
```diff
@@ -0,0 +1,52 @@
...
+            [other: string]: unknown
+          }
+        }
+      }
+    }
+    [sliceKey: string]: unknown
+  }
+}
+
+type NewState = {
```
jagodarybacka, line 19: So this is basically the same type from a typescript's perspective?

background/redux-slices/migrations/to-27.ts
```diff
@@ -0,0 +1,52 @@
...
+export default (prevState: Record<string, unknown>): NewState => {
+  const typedPrevState = prevState as PrevState
+
+  const {
+    account: { accountsData },
+  } = typedPrevState
+
+  Object.keys(accountsData.evm).forEach((chainID) =>
+    Object.keys(accountsData.evm[chainID]).forEach((address) => {
+      // Clear all accounts cached balances
```
jagodarybacka, line 46: So redux slice will be populated again, right after the migration?

background/redux-slices/utils/asset-utils.ts
hyphenized: You're right. Though since I could not find an instance of `isBuiltInNetworkBaseAsset` were we are still making use of those extra checks, I think we can replace it with this more generic alternative.

background/main.ts
```diff
@@ -976,62 +976,34 @@ export default class Main extends BaseService<never> {
...
+              balance.network
+            )
+          })
+          .forEach((balance) => {
+            // TODO support multi-network assets
+            const balanceHasAnAlreadyTrackedAsset = assetsToTrack.some(
+              (tracked) =>
+                tracked.symbol === balance.assetAmount.asset.symbol &&
+                isSmartContractFungibleAsset(balance.assetAmount.asset) &&
+                normalizeEVMAddress(tracked.contractAddress) ===
```
jagodarybacka, line 996: we can use `sameEVMAddress` here

jagodarybacka (approved):
I'm not sure if [bnbspider holds](https://etherscan.io/tokenholdings?a=0x208e94d5661a73360d9387d3ca169e5c130090cd) duplicated WETH anymore.

Overall seems fine, assets are reloaded but they are present after a few minutes which is a little bit annoying and I'm afraid people can complain about it. Please @0xDaedalus take a look before merging.

jagodarybacka:
@hyphenized Is this PR still valid and we want to merge it at some point or not really?

hyphenized:
> @hyphenized Is this PR still valid and we want to merge it at some point or not really?

AFAIK This addressed the issue of a collision between two assets named `WETH`, although most of these duplicate assets are likely spam, there should be no problem to merge this as long as we display those as unverified. 

Paging @Shadowfiend for visibility

background/redux-slices/accounts.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3195/commits/4d871de1145302bae1f2473899037bcc6a99f2c2

background/main.ts
```diff
@@ -976,62 +976,34 @@ export default class Main extends BaseService<never> {
...
+            return !isBuiltInNetworkBaseAsset(
+              balance.assetAmount.asset,
+              balance.network
+            )
+          })
+          .forEach((balance) => {
+            // TODO support multi-network assets
+            const balanceHasAnAlreadyTrackedAsset = assetsToTrack.some(
+              (tracked) =>
+                tracked.symbol === balance.assetAmount.asset.symbol &&
```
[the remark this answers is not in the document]
kkosiorowska: There were some changes and the `isSameAsset` function was used, which normalises the addresses.

background/redux-slices/utils/asset-utils.ts
kkosiorowska: I wonder about this. The names are slightly confusing. But however, functions `isBuiltInNetworkBaseAsset` and `isNetworkBaseAsset` take different parameters.  This reduces the occurrence of mistakes. Therefore, it seems to me that it is not a big threat.

kkosiorowska (approved):
It seems to be working well.  :rocket:  Let's merge it.

Hidden tokens appeared on lists of unverified assets

Screenshot 2023-06-19 at 10 18 07

jagodarybacka:
btw I recall we were meant to add another warning for these duplicated tokens, right? It looks very weird to have 3 AAVE tokens and none of them is the real one 🙈 cc @VladUXUI

PR #3199, opened 2023-03-28
Replace deprecated `set-output` command

michalinacienciala:
As explained in
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/, the `set-output` GH Action command will become deprecated after 31st May 2023. There's a new method for setting up outputs (described [here](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter)), which we will now use.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Read .nvmrc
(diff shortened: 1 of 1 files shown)


hyphenized (approved):
Looks good, thanks for updating this!

PR #3201, opened 2023-03-28
Run e2e tests also for PRs which modify feature flag

michalinacienciala:
We want to run `e2e-tests` job not only after merges to `main` and for PRs that have the `e2e` substring in the feature branch name, but also for PRs that modify the feature flags. We do that by adding a new job (`detect-if-flag-changed`) that checks if the '.env.defaults' file (which stores the feature flags configuration) is modified by the PR. The job uses https://github.com/marketplace/actions/paths-changes-filter action to produce a true/false output concluding if the file was modified. If value is `true`, the `e2e-tests` job will be executed (other criteria for execution of the job remain unchanged.

TODO:

- [x] test if the job triggers when there's a change of flag

Latest build: [extension-builds-3201](https://github.com/tahowallet/extension/suites/11866163051/artifacts/620463292) (as of Tue, 28 Mar 2023 16:15:07 GMT).


michalinacienciala:
One concern I have is whether running test for all pushes to the PR branches that modify feature flags won't cause the reaching of the usage limits in the third-party services. @hyphenized, do you think this may be a problem?

hyphenized:
> One concern I have is whether running test for all pushes to the PR branches that modify feature flags won't cause the reaching of the usage limits in the third-party services. @hyphenized, do you think this may be a problem?

Yes, it could cause some false positives but I don't think it's something we should worry about at the moment

michalinacienciala:
In https://github.com/tahowallet/extension/actions/runs/4545184929/jobs/8012252227 I verified that the `e2e-tests` job triggers when there's a change in the '.env.defaults' file (the job failed, but that's an unrelated failure).
Ready for review.

hyphenized (approved):
Looks good, will leave open for the team to give it a 👀

jagodarybacka (approved):
LGTM 📈

michalinacienciala:
@chlolands, you may be interested in this :)

Issue #3189, opened 2023-03-27
Gas: For custom networks


VladUXUI:
For networks that don't support gas control.
Let's use a disabled state + text 
"This network doesn’t support changing Gas settings"

Preview
<img width="653" alt="image" src="https://user-images.githubusercontent.com/80846886/227950998-7650b458-afb4-4815-bb31-810007975e5f.png">

Figma link
https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=17871-336104&t=AYo0RDWV3sYTA1Bh-4

PR #3203, opened 2023-03-28
Disable network fees for custom networks.

0xDaedalus:
Closes https://github.com/tahowallet/extension/issues/3189

Screen Shot 2023-03-28 at 11 08 09 AM

### To Test
- Add the Evmos chain to the wallet
- Attempt to send EVMOS to an address
- Notice that gas settings are greyed out with the correct tooltip.

Latest build: [extension-builds-3203](https://github.com/tahowallet/extension/suites/12040401263/artifacts/633008979) (as of Wed, 05 Apr 2023 08:40:51 GMT).


jagodarybacka (dismissed):
Shouldn't it be the same styles as regular network settings button but without the border? Font looks smaller and the $ value is not white 🤔

hyphenized (dismissed):
We should also remove the $ value as it is not based on the network asset price

ui/components/NetworkFees/FeeSettingsButton.tsx
```diff
@@ -8,6 +14,37 @@ interface FeeSettingsButtonProps {
...
+      <div>
+        <SharedTooltip
+          width={175}
+          height={30}
+          type="dark"
+          horizontalShift={90}
+          horizontalPosition="center"
+          verticalPosition="bottom"
+          IconComponent={() => (
+            <div className="disabled_settings">
```
jagodarybacka, line 34: Styles in `styled-jsx` are not going through the components to the child components so `.disabled_settings` styles are not applied here 😞

kkosiorowska (approved):
The comments were implemented. Let's merge it.✅

Screenshot 2023-04-05 at 10 36 28

Issue #3191, opened 2023-03-27
Custom networks - Swaps disabled state


mr-michael:
On ALL custom networks, the swaps option appears disabled and presents an error message on hover.

The state should be disabled everywhere:
- Main swap button on wallet page
- Small swap icon on each asset
- Single asset page

![image](https://user-images.githubusercontent.com/80846886/227992628-e3cfa5f5-7d90-4721-9dee-3f8bdae0fdf8.png)

https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=17871-336224&t=AYo0RDWV3sYTA1Bh-4

mr-michael:
@VladUXUI thoughts on this page?

Presumably 'swaps' button should be same 'disabled' color as and same message as on main 'wallet' tab?

<img width="369" alt="Screenshot 2023-03-27 at 11 36 23 AM" src="https://user-images.githubusercontent.com/7005061/227991190-889af817-31ef-407c-846e-f78d50800c82.png">

VladUXUI:
Right, very good catch

Here is a preview

![image](https://user-images.githubusercontent.com/80846886/227992410-d5cd9501-917f-4523-b097-a3cf9fb9d80f.png)

The link is the same in figma. Also added disabled for asset card

PR #3205, opened 2023-03-28
Add tooltips indicating that swaps are disabled on custom networks

0xDaedalus:
Closes https://github.com/tahowallet/extension/issues/3191

https://user-images.githubusercontent.com/94649004/228350805-14f99e70-f788-4dbf-ba53-08870ed4faca.mov

### To Test
- [x] Add a custom network with a balance on it.
- [x] Check to make sure a swaps disabled tooltip shows up in the 3 places discussed in #3191

## Testing Env

```
SUPPORT_CUSTOM_NETWORKS=true
SUPPORT_CUSTOM_RPCS=true
```

Latest build: [extension-builds-3205](https://github.com/tahowallet/extension/suites/12366561860/artifacts/656810172) (as of Thu, 20 Apr 2023 11:15:51 GMT).


ui/components/Shared/SharedIconRouterLink.tsx
```diff
@@ -5,10 +5,27 @@ type Props = {
...
+      // @TODO Make accessible
+      // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events
+      <div
+        onClick={(e) => {
+          e.preventDefault()
+          e.stopPropagation()
+        }}
+        className="icon_wrapper"
+      >
+        <i className={`asset_icon ${iconClass}`} />
```
hyphenized, line 25: This icon still changes color on hover

hyphenized (dismissed):
Looks good, two n/b issues:
- The swap button in the asset screen looks shorter when disabled, this might be due to the tooltip affecting layout
- The asset list item action still changes color when hovered, there are global styles on a parent component causing this

jagodarybacka (commented):
image
Well Rootstock is not a custom network, also this tooltip is a little bit too large IMO, maybe it would be nice to change copy or font size? 

Also spacing between icons is different when swaps are disabled

0xDaedalus:
Will leave it up to @VladUXUI to comment on the copy, fixed the font size but was unable to fix the jitter issue.

VladUXUI:
> Will leave it up to @VladUXUI to comment on the copy, fixed the font size but was unable to fix the jitter issue.

It's a good point. Let's go with "swaps aren't supported on this network" as it covers all networks that don't support swaps

kkosiorowska:
I'll take a look at it again 👀

kkosiorowska (approved):
The copy has been updated. Let's not block QA. Works as expected. ✅

The following issues will be resolved in separate PRs.

> The swap button in the asset screen looks shorter when disabled, this might be due to the tooltip affecting layout

> Also spacing between icons is different when swaps are disabled

PR #3233, opened 2023-04-03
Token Discovery - Allow marking assets as trusted

hyphenized:
Allows marking an asset as trusted through the asset warning modal

Trust modal

## To Test
- [ ] Import an account with non tokenlist assets ( e.g. bnbspider2.crypto )
- [ ] Apply [this patch](https://gist.github.com/hyphenized/d1fa04289bf678f4f7a3ef2e285babb7)
- [ ] Attempt to hide untrusted assets

Latest build: [extension-builds-3233](https://github.com/tahowallet/extension/suites/12352492017/artifacts/655743015) (as of Wed, 19 Apr 2023 20:44:52 GMT).


background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -120,12 +120,15 @@ const computeCombinedAssetAmountsData = (
           ? true
           : assetAmount.mainCurrencyAmount > userValueDustThreshold
       const isPresent = assetAmount.decimalAmount > 0
-      const isTrusted = !!(assetAmount.asset?.metadata?.tokenLists.length ?? 0)
+      // TODO: Review after adding "add custom token"
+      const isCustomAsset = !!(
+        assetAmount.asset?.metadata?.tokenLists.length ?? 0
```
0xDaedalus, line 125: I am almost 100% sure we are gonna be adding custom assets that don't have tokenlists - e.g. adding an asset on a local hardhat network.

background/assets.ts
```diff
@@ -39,6 +39,10 @@ export type AssetMetadata = {
   logoURL?: string
   websiteURL?: string
   tokenLists: TokenListCitation[]
+  /**
+   * Set by the user on assets discovered through transaction annotations
+   */
+  trusted?: boolean
```
0xDaedalus, line 45: I think if you took this out of metadata and just put it on `SmartContractFungibleAsset` you'd save a lot of pain related to needing to pass in the empty `{ metadata: { tokenLists: [] }}` everywhere.

0xDaedalus: If you _do_ want to keep it in metadata - I would suggest making the `tokenLists` property optional - as we will very soon have tokens without tokenlists.

background/redux-slices/assets.ts
```diff
@@ -124,6 +144,25 @@ const selectPairedAssetSymbol = (
...
+  ) => {
+    await main.setAssetTrustStatus(asset, trusted)
+    // Update assets slice
+    await dispatch(updateAssetMetadata([asset, { trusted }]))
+    // Update accounts slice cached data about this asset
+    await dispatch(
+      updateAssetCache({
+        ...asset,
+        metadata: { ...asset.metadata, trusted },
+      })
```
0xDaedalus, line 161: Unfortunate that we have to update state in 2 places here but it is what it is.  These kind of places are good impetuses for us to move to React 18 to get batched updating.

background/redux-slices/accounts.ts
```diff
@@ -394,6 +395,33 @@ const accountSlice = createSlice({
         ens: { ...baseAccountData.ens, avatarURL: avatar },
       }
     },
+    /**
+     * Updates cached SmartContracts metadata
+     */
+    updateAssetCache: (
```
0xDaedalus, line 401: This method seems sufficiently complicated to have some tests written for it.

background/redux-slices/assets.ts
```diff
@@ -108,10 +111,27 @@ const assetsSlice = createSlice({
         }
       }
     },
+    updateAssetMetadata: (
```
0xDaedalus, line 114: Let's add a happy-path test here as well.

background/redux-slices/selectors/accountsSelectors.ts
Shadowfiend: There's a TODO in the indexing service that suggests skipping the concept of a custom asset altogether and instead differentiating between baseline trusted (essentially whether it came from a token list) and explicitly trusted (added by the user or marked as trusted by the user).

0xDaedalus: Why do we need to treat them differently in the UI once they are trusted?

hyphenized: > Why do we need to treat them differently in the UI once they are trusted?

If we're adding controls in the UI to manage trusted/untrusted tokens then we'll likely want to avoid displaying the same controls for manually added assets

0xDaedalus: > If we're adding controls in the UI to manage trusted/untrusted tokens then we'll likely want to avoid displaying the same controls for manually added assets

My understanding here differs but I could be wrong cc @VladUXUI.  Just because a token is on one of our default tokenlists does not necessarily mean that the token is relevant to all users.  I can also see users being confused along the lines of "Why can I hide token X but not token Y, I want to hide token Y!".

VladUXUI: If an asset has been marked as trusted, it becomes exactly like the others.
I think the confusion comes from "hide asset" from the asset details page. But that button will be there for all assets, not just the ones that are imported from tx or manually added.

I hope i understood the issue correctly

background/assets.ts
Shadowfiend: Not unlike name resolution, it wouldn't be unreasonable to assume that a token the user adds explicitly is from a special internal token list—and that the only tokens without token lists are tokens we infer from token discovery.

I don't think that necessarily changes the thinking on where to put this property, though.

background/redux-slices/assets.ts
Shadowfiend: Can we dispatch twice and only await the second one to get better behavior here? The updates should execute serially, so awaiting the second should implicitly await the first.

background/assets.ts
Shadowfiend: … Really now that I say that, it feels like there's really no reason to distinguish between a “trusted” asset and an asset the user has manually added, right? Trusted assets could just have been implemented by adding a user-specific token list (which, notably, a future iteration could make exportable and shareable, etc etc) and setting up all of the plumbing for user-added assets with a limited initial interface, exactly like we did for the address book in name resolution. It would even have avoided the need for a migration.

Ugh, kicking myself for not realizing this when I wrote that TODO months ago 😞 Want to be clear we're probably too far in to change the approach now, just thinking aloud.

background/redux-slices/assets.ts
Shadowfiend: Actually, is there a reason that `setAssetTrustStatus` wouldn't run both of these dispatches internally? That kind of coordination seems directly in `Main`'s wheelhouse, and should allow it all to happen in the same tick, I think...

background/assets.ts
```diff
@@ -38,7 +38,11 @@ export type AssetMetadata = {
   coinGeckoID?: string
   logoURL?: string
   websiteURL?: string
-  tokenLists: TokenListCitation[]
+  tokenLists?: TokenListCitation[]
```
Shadowfiend, line 41: Looks like we made this optional but are still passing some explicit empty arrays down below, is that intentional?

hyphenized: Not at all, thanks for pointing that out. Removed the existing one as well for consistency.

background/assets.ts
0xDaedalus: I can certainly see a scenario where a token is on a tokenlist but untrusted (for example if it got hacked or if the tokenlist its on got hacked).  I have the reverse interpretation here where the fact that we treated assets on tokenlists as "trusted" is just a byproduct of the fact that at one point we relied entirely on tokenlists to fetch token metadata.

0xDaedalus (approved):

background/assets.ts
Shadowfiend: Yeah I see that thought. N token lists with attributes of interest as a core primitive to me makes more sense than N attributes on every token, but I haven't thought about it hard enough to have a strong reason for that intuition.

PR #3236, opened 2023-04-03
Add custom chain analytics event

0xDaedalus:
### To Test
- [ ] Add a custom chain to the wallet
- [ ] Observe the `Custom Chain Added` event in posthog.

Latest build: [extension-builds-3236](https://github.com/tahowallet/extension/suites/12046398882/artifacts/633415473) (as of Wed, 05 Apr 2023 13:17:48 GMT).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1654,6 +1654,7 @@ export default class Main extends BaseService<never> {
                 This event is fired when a custom chain is added to the wallet.
                 `,
           chainInfo: chainInfo.chainName,
+          chainId: chainInfo.chainId,
         }
       )
     })
--- a/background/main.ts
+++ b/background/main.ts
@@ -304,10 +304,7 @@ export default class Main extends BaseService<never> {
       chainService
     )
 
-    const analyticsService = AnalyticsService.create(
-      chainService,
-      preferenceService
-    )
+    const analyticsService = AnalyticsService.create(preferenceService)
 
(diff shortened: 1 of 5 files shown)


background/main.ts
```diff
@@ -1735,11 +1759,20 @@ export default class Main extends BaseService<never> {
...
+        const networkNameAtClose =
+          this.store.getState().ui.selectedAccount.network.name
         this.analyticsService.sendAnalyticsEvent(AnalyticsEvent.UI_SHOWN, {
           openTime: new Date(openTime).toISOString(),
           closeTime: new Date().toISOString(),
           openLength: (Date.now() - openTime) / 1e3,
+          networkName:
+            originalNetworkName === networkNameAtClose
+              ? originalNetworkName
+              : "switched networks",
```
hyphenized, line 1775: This'll tell us users switched chains after the popup closed, but it won't catch network changes while the popup is open 🤔

hyphenized (dismissed):
Looks good, left one suggestion

background/main.ts
0xDaedalus: This should be fine - we don't need precise data about what chains our users are using - a directional idea is more than enough.

kkosiorowska (dismissed):
I have tested the solution works great. ✅ Let's fix the conflicts and merge it.

kkosiorowska (approved):

PR #3247, opened 2023-04-04
Add Hardhat Fork Functionality

0xDaedalus:
Necessary environment variables:
```
USE_MAINNET_FORK
MAINNET_FORK_URL
MAINNET_FORK_CHAIN_ID
```

Latest build: [extension-builds-3247](https://github.com/tahowallet/extension/suites/13664299936/artifacts/754727567) (as of Fri, 16 Jun 2023 18:01:08 GMT).


package.json
```diff
@@ -56,7 +56,8 @@
     "background",
     "provider-bridge",
     "provider-bridge-shared",
-    "window-provider"
+    "window-provider",
+    "ci"
```
Shadowfiend, line 60: Do we need this listed as a workspace?

package.json
```diff
@@ -110,6 +111,7 @@
     "eslint-plugin-react": "^7.24.0",
     "eslint-plugin-react-hooks": "^4.2.0",
     "fork-ts-checker-webpack-plugin": "^6.3.2",
+    "hardhat": "^2.14.0",
```
Shadowfiend, line 114: This dependency should probably be in the `ci` package, no? It's unused elsewhere afaict.

Shadowfiend (commented):
I would love to try moving the hardhat dependency into the `ci` package and removing that package from `workspaces` given how much `yarn.lock` garbage it pulls in.

What I expect would happen:
- Drop `hardhat` from top-level `package.json` and drop `ci` from the `workspaces` list there as well.
- Add `hardhat` to `ci/package.json`.
- `git checkout main -- yarn.lock` to revert yarn.lock to the latest main version.
- `yarn install`
- `cd ci; yarn install`
- Then add `ci/yarn.lock`.

If any of this runs into issues I think we can merge as-is rather than wasting too much more time on it though.

Shadowfiend (dismissed):
All righty let's rock and roll 🎸

Shadowfiend:
I guess also not clear to me is how much this differs from what's in `dev-utils/local-chain`… It feels like not at all… But we can reconcile later.

Shadowfiend (approved):
Ok but this time though let's do it lol.

PR #3248, opened 2023-04-04
Allow export of private keys and mnemonics

jagodarybacka:
### What

Let's allow exporting secrets of already imported wallets:
- for HD wallets imported with mnemonics let's allow exporting mnemonics
- for wallets imported with private keys and JSON files let's allow exporting private keys

Not in scope:
- let's not allow exporting private keys of addresses from HD wallet - it would require a change in the `hd-keyring` package so let's drop it for now

### Testing

- make sure unit tests are passing, these changes are to be used on the frontend side in the next PRs

Latest build: [extension-builds-3248](https://github.com/tahowallet/extension/suites/12040382191/artifacts/633008217) (as of Wed, 05 Apr 2023 08:40:22 GMT).


kkosiorowska (approved):
Looks great! ✨

PR #3250, opened 2023-04-04
Token Discovery - Add untrusted assets visibility toggler to settings

### LINKED ISSUE #3213: Separate untrusted tokens option from “Hide asset balances under $2”
“Hide asset balances under $2” currently hides assets <$2 **AND** untrusted assets.

This issue is to separate these out into two options:

- “Hide asset balances under $2”
- “Hide untrusted assets”

The “Hide untrusted assets” should appear immediately under Hide asset balances under $2” on settings

hyphenized:
Adds a toggler to show/hide untrusted assets to the settings page

Closes #3213 

## To Test
- [x] Set a few tokens as trusted
- [x] Go to the settings page and toggle "Hide untrusted assets"
- [x] Check that the wallet asset list does not display the rest of untrusted assets, these should be under hidden assets now
- [x] Verify that trusted assets that are dust do not display if hide dust is enabled
- [x] Assets marked as hidden/trusted should not display the unverified asset warning icon
- Interact with an unverified asset, mark it as trusted/untrusted
  - [x] Wallet should no longer display warning sign in the asset list
  - [x] Wallet should no longer display unverified asset status in the single asset page
  - [ ] Single asset page should display Hide asset / Show asset
  - [ ] Clicking on either of these options should change the asset trusted status
- For unverified assets: 
  - [ ] In the single asset page, clicking on an unverified asset should bring up the warning slide up panel
  - [ ] With `hide untrusted assets` toggled on, the dialog should display `Hide asset` and `Trust asset`
  - [ ] With `hide untrusted assets` toggled off, the dialog should display `Close` and `Trust asset`
- For tokenlist assets: 
	- [x] In the single asset page, you should be able to hide it and send it to the hidden assets list
	- [x] Clicking on "Show" should send a token-list asset back to the main asset list
- [ ] Discovered assets should display transaction source when available

## Testing Env

```
SUPPORT_ASSET_TRUST=true
```

Latest build: [extension-builds-3250](https://github.com/tahowallet/extension/suites/12855255665/artifacts/692999684) (as of Fri, 12 May 2023 12:17:34 GMT).


background/redux-slices/ui.ts
```diff
@@ -15,19 +15,22 @@ export const defaultSettings = {
...
   showingActivityDetailID: string | null
   initializationLoadingTimeExpired: boolean
+  // FIXME: Move these settings to preferences service db
   settings: {
     hideDust: boolean
     defaultWallet: boolean
     showTestNetworks: boolean
     collectAnalytics: boolean
     showAnalyticsNotification: boolean
+    showUntrustedAssets: boolean
```
kkosiorowska, line 33: Shouldn't we do a migration if we are changing the state?

hyphenized: We don't need one because this settings object is guaranteed to exist and we're only checking truthiness for this property value, so `undefined` is fine.

jagodarybacka: As we were always doing migrations with changes like this I think it may be better to continue that habit, even if this is not dangerous to skip this one as this is a boolean, next time someone will skip adding an object which will be a problem 😜 Or we can decide to stop doing unnecessary migrations if everyone agrees with that

Shadowfiend:
Quick Q here: I'm puzzled as to how this interacts with the show/hide hidden assets collapsing in the asset list. Is a hidden asset different from an untrusted asset? How does the user find out the distinction? Or is this work meant to supersede that work?

kkosiorowska (commented):
Currently, after separating hidden assets, the toggler is not doing anything. 

https://user-images.githubusercontent.com/23117945/232706215-39387617-6721-4be7-8306-49fc9b822378.mov

hyphenized:
> Quick Q here: I'm puzzled as to how this interacts with the show/hide hidden assets collapsing in the asset list. Is a hidden asset different from an untrusted asset? How does the user find out the distinction? Or is this work meant to supersede that work?

Hidden assets may be asset balances that are dust, assets that have not been marked as trusted or a combination of the former (assets market as trusted but which balances are considered dust). Does that makes sense? cc @VladUXUI

@kkosiorowska I've fixed the toggler, please give it another look when you can.

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -156,24 +199,14 @@ const computeCombinedAssetAmountsData = (
...
       if (
-        isForciblyDisplayed ||
-        (isTrusted && (hideDust ? isNotDust && isPresent : isPresent))
+        isAssetAmountVisible(assetAmount, {
+          hideDust,
+          hideUntrusted,
+        })
       ) {
         acc.combinedAssetAmounts.push(assetAmount)
-      } else if (isPresent) {
```
kkosiorowska, line 176: I would like to confirm if we want to start showing assets with 0$ balance under the "hidden" section? @VladUXUI 

Screenshot 2023-04-19 at 13 08 51

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -264,7 +314,8 @@ export const selectCurrentAccountBalances = createSelector(
       assets,
       mainCurrencySymbol,
       currentNetwork,
-      hideDust
+      hideDust,
+      hideUntrustedAssets
```
kkosiorowska, line 318: Non-blocking comment

I wonder if it's better to make it one object with settings for assets and pass it here. What I mean is that I would like to avoid increasing the number of parameters for the function.

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -71,12 +72,54 @@ const shouldForciblyDisplayAsset = (
...
+
+  let visible = true
+
+  if (hideDust) {
+    visible = visible && isNotDust
+  }
+
+  if (isSmartContractAmount && hideUntrusted) {
+    visible = visible && isTrusted
+  }
```
kkosiorowska, line 111: 🚀

kkosiorowska (commented):
It looks really good ⚡️Let's still wait for Vlad's response.

background/redux-slices/selectors/accountsSelectors.ts
hyphenized: This is actually a bug, these should appear as "< 0,01" since they are dust balances(not zero). I've added some logic to hide these for now in [`e956630` (#3250)](https://github.com/tahowallet/extension/pull/3250/commits/e956630fb2f7b1b71aa1c37e2bd82d7c00f731ac)

image

background/redux-slices/selectors/accountsSelectors.ts
hyphenized: Agree, we don't need the to pass the current network as a parameter anymore either

VladUXUI:
Couple of things here on the UX side (although uncertain what to actually comment on)
1 - If i marked an asset as trusted, it shouldn't have the yellow Attention icon anymore
2 - Trust/untrust is missing from the token details page 
image

jagodarybacka (changes requested):
- [x] once I trust asset I can no longer untrust it, Hide asset is doing nothing
image

- [x] on single asset page I can see Asset not verified for tokens from token lists 
image

0xDaedalus (changes requested):
I'd like to requests 2 e2e tests here, lets test the following

1. For a token that is initially trusted:
- Set token to untrusted
- Token displays as untrusted in Wallet
- Set it back to trusted
- Token displays as trusted in Wallet

2. For a token that is initially untrusted:
- Set token to trusted
- Token displays as trusted in Wallet
- Set it back to untrusted
- Token displays as untrusted in Wallet

hyphenized:
> * [ ]  once I trust asset I can no longer untrust it, Hide asset is doing nothing
>   image

Thanks for the review, fixed!
> * [ ]  on single asset page I can see Asset not verified for tokens from token lists
>   image

Not sure if this is the correct behaviour, could you confirm @VladUXUI ?

Shadowfiend:
> on single asset page I can see Asset not verified for tokens from token lists
> Not sure if this is the correct behaviour

As I absorb this feature into my brain—why would we see Asset not verified for a token list asset? We have baseline trust for token list assets, right? And Asset not verified is a textual reminder that an asset is not trusted, correct?

VladUXUI:
Token list assets should't have any verified or unverified behaviour/status. As these are already verified and trusted by us

ui/components/Wallet/AssetWarningSlideUp.tsx
```diff
@@ -140,16 +161,47 @@ export default function AssetWarningSlideUp(
             />
           </div>
         </li>
+        {asset.metadata?.discoveryTx && (
+          <li>
+            <div className="left">{t("discoveryTx")}</div>
+            <div className="right">
```
jagodarybacka, line 167: Can we use `SharedAddress` component here?

ui/pages/SingleAsset.tsx
```diff
@@ -10,20 +10,123 @@ import { sameEVMAddress } from "@tallyho/tally-background/lib/utils"
...
 import SharedAssetIcon from "../components/Shared/SharedAssetIcon"
 import SharedButton from "../components/Shared/SharedButton"
 import WalletActivityList from "../components/Wallet/WalletActivityList"
 import SharedBackButton from "../components/Shared/SharedBackButton"
 import SharedTooltip from "../components/Shared/SharedTooltip"
 import { scanWebsite } from "../utils/constants"
+import SharedIcon from "../components/Shared/SharedIcon"
+import AssetWarningSlideUp from "../components/Wallet/AssetWarningSlideUp"
+
+function TrustToggler({
```
jagodarybacka, line 34: `TrustToggler` deserves to be in its own file I think 😉

jagodarybacka (changes requested):
Some ideas and improvements

background/redux-slices/tests/accounts.unit.test.ts
```diff
@@ -274,3 +277,196 @@ describe("Accounts redux slice", () => {
...
+              hideDust: true,
+              hideUntrusted: false,
+            }
+          )
+        ).toBeTruthy()
+
+        expect(
+          isAssetAmountVisible(
+            createCompleteAssetAmount(
+              createSmartContractAsset({ metadata: {} }),
```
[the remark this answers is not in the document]
hyphenized: No metadata means no token lists so this is an unverified, untrusted asset

ui/components/Wallet/AssetWarningSlideUp.tsx
hyphenized: We could but `SharedAddress` has some logic specific to name resolution.

kkosiorowska:
There has been a change in logic for trusted/untrusted assets. The current status of this feature can be found in https://github.com/tahowallet/extension/pull/3394. Let's close it.

PR #3253, opened 2023-04-05
Export private keys from HD wallet addresses

jagodarybacka:
### What

Let's allow exporting private keys from HD wallet addresses.

### Testing

- [x] add HD wallet with the mnemonic
- [x] export private key for account from HD wallet
- [x] export the same private key with metamask and compare them

image

Latest build: [extension-builds-3253](https://github.com/tahowallet/extension/suites/12348519892/artifacts/655435412) (as of Wed, 19 Apr 2023 17:38:25 GMT).


background/services/keyring/tests/index.unit.test.ts
```diff
@@ -141,6 +141,13 @@ describe("Keyring Service", () => {
 
       expect(mnemonic).toBe(HD_WALLET_MOCK.mnemonic)
     })
+    it("should be able to export private key from HD wallet addresses", async () => {
```
0xDaedalus, line 144: One nit here is that I think it might be prudent to add a `describe` block here for `exportPrivateKey` and add this test to it as well as a `should require the wallet to be unlocked` test - only because the fact that the wallet needs to be unlocked is _so_ important here.

0xDaedalus:
I think you can `yarn add tahowallet/extension#export-pk-from-hd-wallet` if you want to get tests passing and not be blocked the the hd-keyring PR.  We'll just need to make sure we change it back before merging `keyring-with-pk` to `main`

kkosiorowska:
I will take a look at that as well 👀

background/services/keyring/tests/index.unit.test.ts
```diff
@@ -245,4 +231,53 @@ describe("Keyring Service", () => {
...
+      const errorMessage = "KeyringService must be unlocked."
+      const exportMnemonic = async () => {
+        await keyringService.exportMnemonic(HD_WALLET_MOCK.addresses[0])
+      }
+      const exportPrivateKey = async () => {
+        await keyringService.exportPrivateKey(PK_WALLET_MOCK.address)
+      }
+
+      expect(exportMnemonic()).rejects.toThrowError(errorMessage)
+      expect(exportPrivateKey()).rejects.toThrowError(errorMessage)
```
kkosiorowska, line 280: Non-blocking comment
There is no difference. However, I have seen that we already use `toThrow`.

```suggestion
      expect(exportMnemonic()).rejects.toThrow(errorMessage)
      expect(exportPrivateKey()).rejects.toThrow(errorMessage)
```

kkosiorowska (approved):
LGTM ✅

PR #3255, opened 2023-04-05
Export private key form

jagodarybacka:
### What
- Create a new component that will wrap a given text or other React component and will blur it unless user if hovering their mouse over it.
- Add Secret Text component to private key export slideup, add copy button

image

### Testing

- [ ] import account with private key
- [ ] export private key from the menu - check pk visible on the screen and test "copy" button

Latest build: [extension-builds-3255](https://github.com/tahowallet/extension/suites/12064461721/artifacts/634749987) (as of Thu, 06 Apr 2023 06:25:22 GMT).


kkosiorowska (approved):
Works great! ✅

PR #3257, opened 2023-04-06
Unlock screen for the account backup 

kkosiorowska:
This PR adds a wallet-unlocking step. The screen should appear whenever the user wants to open the slide-up menu for `Show Private Key`.

https://user-images.githubusercontent.com/23117945/230336203-550e0798-10b1-419c-bcd1-0e42ad81f59f.mov

### To Test
- [x] Check if the screen shows when the wallet is locked
- [x] Check if the screen shows when the wallet is unlocked
- [x] Check if you are able to unlock the screen with the click of a button

Latest build: [extension-builds-3257](https://github.com/tahowallet/extension/suites/12068023649/artifacts/634995263) (as of Thu, 06 Apr 2023 09:40:41 GMT).

### DIFF
--- a/ui/components/AccountItem/ShowPrivateKey.tsx
+++ b/ui/components/AccountItem/ShowPrivateKey.tsx
@@ -176,7 +176,7 @@ export default function ShowPrivateKey({
             flex-direction: column;
             align-items: center;
             justify-content: space-between;
-            height: 90%;
+            height: 91%;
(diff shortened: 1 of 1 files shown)


jagodarybacka (approved):
Overall looks good ✅ 
I don't like that the snackbar with "signing locked" is showing up but we can improve it in the next PR. Let's merge it together and improve later.

PR #3258, opened 2023-04-06
Update link to the Discord channel for bug reporting

michalinacienciala:
We've been linking the `#feedback` Discord channel on the New GH Issue landing page. The channel no longer exists and the `#bug-report` channel is now a designated place for bug reporting in Discord.
Updating name of the channel and the invite link. The invite link does not have the expiry date and allows for unlimited use.

Latest build: [extension-builds-3258](https://github.com/tahowallet/extension/suites/12092805225/artifacts/636786921) (as of Fri, 07 Apr 2023 11:57:13 GMT).

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG.yml
+++ b/.github/ISSUE_TEMPLATE/BUG.yml
@@ -7,7 +7,7 @@ body:
   - type: markdown
     attributes:
       value: >-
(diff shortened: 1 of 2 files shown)


kkosiorowska (commented):
It looks like we should still replace the link in the `BUG.yml` file.

michalinacienciala:
> It looks like we should still replace the link in the `BUG.yml` file.

Thanks for pointing it out, I did not notice it's used there as well. Updated!

kkosiorowska (approved):
Now works great :white_check_mark:

PR #3261, opened 2023-04-06
Custom Token Import - Allow manually importing assets

hyphenized:
Allows manually adding tokens from the asset list and the settings page

image

## Testing Env

```
SUPPORT_CUSTOM_NETWORKS=true
SUPPORT_CUSTOM_RPCS=true
```

## To test
- [ ] Add a custom network
- [ ] Scroll down the asset list, click on add custom asset
- [ ] Input a erc20 token address, check that balance and symbol appear
- [ ] Attempt to "add asset", check that a toast displays and the asset now appears in the asset list after a balance refresh
- [ ] Input the same erc20 token address, verify a warning shows up with "asset already added"
- [ ] Input an account address or an invalid address, you should see an error "invalid contract address"

Latest build: [extension-builds-3261](https://github.com/tahowallet/extension/suites/12160852252/artifacts/641691473) (as of Tue, 11 Apr 2023 15:24:07 GMT).

### DIFF
--- a/background/lib/logger.ts
+++ b/background/lib/logger.ts
@@ -170,6 +170,11 @@ class Logger {
     this.logEvent(LogLevel.error, input)
   }
 
+  buildError(...input: unknown[]): Error {
+    this.error(...input)
+    return new Error(input.join(" "))
+  }
+
   logEvent(level: Exclude<LogLevel, LogLevel.off>, input: unknown[]): void {
     if (logLevels[level] < this.#level) return
 
--- a/background/main.ts
+++ b/background/main.ts
@@ -175,7 +175,6 @@ import {
 import { AddChainRequestData } from "./services/provider-bridge"
 import { AnalyticsEvent, isOneTimeAnalyticsEvent } from "./lib/posthog"
 import { isBuiltInNetworkBaseAsset } from "./redux-slices/utils/asset-utils"
-import { fromFixedPoint } from "./lib/fixed-point"
 
 // This sanitizer runs on store and action data before serializing for remote
 // redux devtools. The goal is to end up with an object that is directly
@@ -1756,12 +1755,7 @@ export default class Main extends BaseService<never> {
   }> {
     const { network } = addressOnNetwork
 
-    const balance = await this.chainService.assetData.getTokenBalance(
-      addressOnNetwork,
-      contractAddress
-    )
-
-    const existingAsset = this.indexingService
+    const cachedAsset = this.indexingService
       .getCachedAssets(network)
       .find(
         (asset): asset is SmartContractFungibleAsset =>
@@ -1769,36 +1763,13 @@ export default class Main extends BaseService<never> {
           sameEVMAddress(contractAddress, asset.contractAddress)
       )
 
-    if (existingAsset) {
-      return {
-        asset: existingAsset,
-        // FIXME: REMOVE FIXED PRECISION
-        balance: fromFixedPoint(balance.amount, existingAsset.decimals, 2),
-        exists: true,
-      }
-    }
+    const result = await this.chainService.queryTokenDetails(
+      contractAddress,
+      addressOnNetwork,
+      cachedAsset
+    )
 
-    const asset = await this.chainService.assetData
-      .getTokenMetadata({
-        contractAddress,
-        homeNetwork: network,
-      })
-      .catch(() => undefined)
-
-    if (!asset) {
-      throw logger.buildError(
-        "Unable to retrieve metadata for custom asset",
-        contractAddress,
-        "on chain:",
-        network.chainID
-      )
-    }
-
-    return {
-      asset,
-      // FIXME: REMOVE FIXED PRECISION
-      balance: fromFixedPoint(balance.amount, asset.decimals, 2),
-    }
+    return { ...result, exists: !!cachedAsset }
   }
 
   async importTokenViaContractAddress(
--- a/background/main.ts
+++ b/background/main.ts
@@ -12,6 +12,7 @@ import {
   getEthereumNetwork,
   isProbablyEVMAddress,
   normalizeEVMAddress,
+  sameEVMAddress,
   wait,
 } from "./lib/utils"
 
@@ -146,7 +147,11 @@ import { getActivityDetails } from "./redux-slices/utils/activities-utils"
 import { getRelevantTransactionAddresses } from "./services/enrichment/utils"
 import { AccountSignerWithId } from "./signing"
 import { AnalyticsPreferences } from "./services/preferences/types"
-import { isSmartContractFungibleAsset, SmartContractAsset } from "./assets"
+import {
+  isSmartContractFungibleAsset,
+  SmartContractAsset,
+  SmartContractFungibleAsset,
+} from "./assets"
 import { FeatureFlags, isEnabled } from "./features"
 import { NFTCollection } from "./nfts"
 import {
@@ -170,6 +175,7 @@ import {
 import { AddChainRequestData } from "./services/provider-bridge"
 import { AnalyticsEvent, isOneTimeAnalyticsEvent } from "./lib/posthog"
 import { isBuiltInNetworkBaseAsset } from "./redux-slices/utils/asset-utils"
+import { fromFixedPoint } from "./lib/fixed-point"
 
 // This sanitizer runs on store and action data before serializing for remote
 // redux devtools. The goal is to end up with an object that is directly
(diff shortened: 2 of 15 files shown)


VladUXUI:

background/lib/logger.ts
```diff
@@ -170,6 +170,11 @@ class Logger {
     this.logEvent(LogLevel.error, input)
   }
 
+  buildError(...input: unknown[]): Error {
+    this.error(...input)
+    return new Error(input.join(" "))
+  }
+
```
kkosiorowska, line 177: Should we always use this function now?

background/main.ts
```diff
@@ -1740,6 +1746,61 @@ export default class Main extends BaseService<never> {
...
+      addressOnNetwork,
+      contractAddress
+    )
+
+    const existingAsset = this.indexingService
+      .getCachedAssets(network)
+      .find(
+        (asset): asset is SmartContractFungibleAsset =>
+          isSmartContractFungibleAsset(asset) &&
+          sameEVMAddress(contractAddress, asset.contractAddress)
```
kkosiorowska, line 1769: Non-blocking comment
I have a feeling that maybe we could create a separate function for this condition. Something like `isMatchingFungibleAsset`.

ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -0,0 +1,346 @@
...
+          <SharedInput
+            label={t("input.contractAddress.label")}
+            errorMessage={error ? t("error.invalidToken") : ""}
+            onChange={handleContractChange}
+          />
+          <div className="tooltip_wrap">
+            <SharedTooltip
+              width={219}
+              horizontalShift={100}
+              customStyles={{ "--tooltip-icon-color": "var(--green-20)" }}
```
kkosiorowska, line 168: Non-blocking comment
It is for the tooltip component. Maybe it is simpler to add a color option in the props in that case?

ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -0,0 +1,346 @@
...
+                t={t}
+                i18nKey="input.tooltip"
+                components={{
+                  url: (
+                    <a
+                      href={HELPDESK_CUSTOM_TOKENS_LINK}
+                      rel="noopener noreferrer"
+                    >
+                      Help Center
+                    </a>
```
kkosiorowska, line 180: Why not use SharedLink?

ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -0,0 +1,346 @@
...
+                size={40}
+                logoURL={assetData?.asset.metadata?.logoURL}
+                symbol={assetData.asset.symbol}
+              />
+            ) : (
+              <PlaceholderIcon />
+            )}
+            <div className="token_details">
+              <div className="balance">
+                <strong>{assetData?.balance ?? "Balance"}</strong>
```
kkosiorowska, line 200: Do we not want to translate `Balance` and `Name`?

kkosiorowska (commented):
I took a quick look. Some minor thoughts.

background/main.ts
```diff
@@ -1740,6 +1746,61 @@ export default class Main extends BaseService<never> {
     return this.chainService.removeCustomChain(chainID)
   }
 
+  async queryCustomTokenDetails(
```
0xDaedalus, line 1749: This feels like it should be a method in `chainService` - I know we unfortunately already have some complicated methods in `main.ts` but in general I think we've moved away from putting application logic in this file and have mostly been using it as a connector between the ui thread & background thread (either through event listeners & emitters or direct method calls via background async thunks).

ui/_locales/en/messages.json
```diff
@@ -540,6 +541,17 @@
...
+    "addCustomAssetSettings": {
+      "title": "Add custom asset",
+      "input.contractAddress.label": "Contract address",
+      "input.tooltip": " Visit your preferred block explorer and find the asset you want to import.<br>Copy and paste the contract address here.<br><br>For more information visit <url>Help Center</url>",
+      "warning.alreadyExists.title": "Asset already exists, you will see it once your balance is over 0",
+      "warning.alreadyExists.desc": "If you still have issues seeing an asset, turn off “hide balance 2$” from Settings",
+      "error.invalidToken": "Invalid contract address",
+      "submit": "Add asset",
+      "snackbar.success": "Asset added successfully",
+      "footer.hint": "How to add a custom asset?"
```
0xDaedalus, line 553: This should probably either be

`How to add a custom asset`
or
`How do I add a custom asset?` if we want to keep the `?`

cc @VladUXUI

ui/_locales/en/messages.json
```diff
@@ -741,7 +753,9 @@
       "contractCreation": "(Contract creation)",
       "transactionTo": "To:",
       "transactionFrom": "From:",
-      "loadingActivities": "Digging deeper..."
+      "loadingActivities": "Digging deeper...",
+      "addCustomAssetPrompt": "Can't find an asset?",
+      "addCustomAssetAction": "add custom asset"
```
0xDaedalus, line 758:
```suggestion
      "addCustomAssetAction": "Add custom asset"
```

ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -0,0 +1,346 @@
...
+        strokeLinecap="round"
+      />
+      <path
+        d="M6.0625 15.1273L7.5916 13.0885C7.77416 12.8451 8.13 12.8198 8.34515 13.035L9.8125 14.5023"
+        stroke="currentColor"
+        strokeWidth="1.5"
+        strokeLinecap="round"
+      />
+      <circle cx="9.5" cy="8.5" r="1.5" fill="currentColor" />
+    </svg>
```
0xDaedalus, line 51: Rather than defining this inline - lets place a `.svg` file in the `ui/public/images` folder and then import it here.

ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -0,0 +1,346 @@
...
+    if (!isProbablyEVMAddress(contractAddress)) {
+      setState({ loading: false, assetData: null, error: true })
+      return
+    }
+
+    const details = (await dispatch(
+      checkTokenContractDetails({
+        contractAddress: normalizeEVMAddress(contractAddress),
+      })
+    )) as unknown as AssetData
```
0xDaedalus, line 110: As a rule I think we should avoid this kind of casting - especially since we know that `checkTokenContractDetails` can return `null` in some cases.  Even though we handle this case a few lines by showing an error if `details === null` we don't want our typings to diverge from what is actually going on.

0xDaedalus (changes requested):

background/lib/logger.ts
jagodarybacka: I don't think so, we should carefully pick errors that are meaningful for debugging before we put them into logs

hyphenized: If it helps reduce repetition. Something to note is that this only works well for string errors and, on the frontend side an uncaught error will also be logged by the error boundary at `ErrorFallback`.

background/main.ts
hyphenized: We could but we're comparing a contract address against an asset, which is a good fit for sticking to `sameEVMAddress` here I think.

ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -0,0 +1,346 @@
...
+)
+
+export default function SettingsAddCustomAsset(): ReactElement {
+  const { t } = useTranslation("translation", {
+    keyPrefix: "settings.addCustomAssetSettings",
+  })
+
+  const history = useHistory()
+
+  type AssetData = AsyncThunkFulfillmentType<typeof checkTokenContractDetails>
```
jagodarybacka, line 73: does this type need to be inside the component? most often types are defined above the react components using them

ui/pages/Settings/SettingsAddCustomAsset.tsx
hyphenized: Yes it would be simpler, but the goal is to allow customizing more than just the tooltip icon color by overriding css variables. In that scenario, it wouldn't be very practical to have one property per style override.

ui/pages/Settings/SettingsAddCustomAsset.tsx
hyphenized: Whoops 🤦  good catch!

ui/pages/Settings/SettingsAddCustomAsset.tsx
jagodarybacka: If i remember correctly dispatch is not returning correct types so the best we can do is just be careful and make the casts here the same as in the background actions 🤔

background/main.ts
hyphenized: I've moved most of the logic to the chain service, which should make this easier to test as well.

ui/pages/Settings/SettingsAddCustomAsset.tsx
hyphenized: I agree, however I consider this is an exception given the type definition for `AssetData` is the "correct" thunk return type.

ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+  const [{ loading, error, assetData }, setState] = useSetState<{
```
jagodarybacka, line 75: Isn't it achieving the same outcome as the `useReducer`?

PR #3263, opened 2023-04-07
Separate assets in the asset list 

### LINKED ISSUE #3224: Separate hidden assets from non-hidden assets in the asset list
image

- [ ] The `Hide Asset Balances under $2` should no longer hide untrusted assets.

kkosiorowska:
Closes #3224 

This PR separates hidden assets from non-hidden assets in the asset list. The `Hide Asset Balances under $2` should no longer hide untrusted assets. All hidden assets should be separated and displayed together.

## UI
Before
Screenshot 2023-04-10 at 11 34 28

After
https://user-images.githubusercontent.com/23117945/231146605-98e9b23b-0b8c-4099-8d30-e39198d27c44.mov

Screenshot 2023-04-12 at 11 20 59
Screenshot 2023-04-12 at 11 19 14

## To Test
- [x] untrusted assets should be separated
- [x] assets under $2 should be separated
- [x] `Hide Asset Balances under $2` should no longer hide untrusted assets

Latest build: [extension-builds-3263](https://github.com/tahowallet/extension/suites/12179823482/artifacts/643115140) (as of Wed, 12 Apr 2023 09:29:23 GMT).

### DIFF
--- a/background/redux-slices/migrations/index.ts
+++ b/background/redux-slices/migrations/index.ts
@@ -23,12 +23,13 @@ import to23 from "./to-23"
 import to24 from "./to-24"
 import to25 from "./to-25"
 import to26 from "./to-26"
+import to27 from "./to-27"
 
 /**
  * The version of persisted Redux state the extension is expecting. Any previous
  * state without this version, or with a lower version, ought to be migrated.
  */
-export const REDUX_STATE_VERSION = 26
+export const REDUX_STATE_VERSION = 27
 
 /**
  * Common type for all migration functions.
@@ -64,6 +65,7 @@ const allMigrations: { [targetVersion: string]: Migration } = {
   24: to24,
   25: to25,
   26: to26,
+  27: to27,
 }
 
 /**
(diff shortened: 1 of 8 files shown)


jagodarybacka (changes requested):
Code looks fine but I'm not sure about UX:
- I feel like there should be only one "digging deeper..."
image
- I see "show hidden..." when they are already shown and "hide..." when they are hidden, I feel like the label of the button should correspond to the action that will be performed after clicking on it
- we started showing assets with 0$ balance under the "hidden" section
image

- if we would like to make the UX better I feel like after clicking on the button that will show hidden assets it should auto scroll a little bit so user can see that something actually was displayed below the visible part of the screen

https://user-images.githubusercontent.com/20949277/230915782-f5f597a5-99d3-421b-a324-447cf69b9b02.mov

VladUXUI:
Agree with all the points above that Jagoda made

kkosiorowska:
Take another look, please. @VladUXUI @jagodarybacka

jagodarybacka:
One small thing that annoys me - every time I open the wallet hidden assets height animation is running, would it be possible to skip it if it is already rendered with the hidden assets section open?

https://user-images.githubusercontent.com/20949277/231195498-7fbf4615-b63b-4b23-80cd-e81fde02ba13.mov

jagodarybacka (approved):
Looks good ✅

PR #3267, opened 2023-04-11
Custom Token Import - Add network selector

### LINKED ISSUE #3262: Add network selector to custom token import

_Originally posted by @VladUXUI in https://github.com/tahowallet/extension/issues/3261#issuecomment-1500072361_

hyphenized:
Closes #3262 

Adds a slide up network selector to custom token import

image

## Testing Env

```
SUPPORT_CUSTOM_NETWORKS=true
SUPPORT_CUSTOM_RPCS=true
```

## To test
- [ ] Head over to the settings page to add a custom asset
- [ ] Change the network
- [ ] Input a token contract address

Latest build: [extension-builds-3267](https://github.com/tahowallet/extension/suites/12208407512/artifacts/645197431) (as of Thu, 13 Apr 2023 09:53:57 GMT).


ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -141,8 +163,83 @@ export default function SettingsAddCustomAsset(): ReactElement {
...
+        }
+      `}</style>
+      <SharedSlideUpMenu
+        isOpen={isNetworkSelectOpen}
+        isScrollable
+        customStyles={{ display: "flex", flexDirection: "column" }}
+        close={() => setNetworkSelectOpen(false)}
+      >
+        <div className="network_select">
+          <div className="network_select_title">Select Network</div>
```
kkosiorowska, line 189: Missing translation.

kkosiorowska (commented):
I've looked into this and have a few small thoughts.

- When I select a network and close the wallet the network does not save when I open it again. I noticed this when I closed my wallet for a while to copy the address token. It is a small thing and a minor point.

- It's not quite within the scope of this PR but I noticed it when testing. Styles look bad when the network has a long name.

  Screenshot 2023-04-11 at 15 19 23

- After we add several networks, the header with the close button disappears.
  Screenshot 2023-04-11 at 15 28 10

kkosiorowska (approved):
LGTM :white_check_mark:

Shadowfiend:
My 2c: ideally we would resolve their balance during the adding process and warn them if they've got <$2 balances hidden that the token won't appear by default.

VladUXUI:
Ideally any asset should follow the <$2 rule. So definitely hide them if that toggle is on.
But yeah, good to let people know.

We had this screen designed 
image

PR #3268, opened 2023-04-11
CI - Fix new line handling in build env block detection

### LINKED ISSUE #3262: Add network selector to custom token import

hyphenized:
Closes #3262 

Fixes missing escaping of new line sequences '\n'

Before: 
image

## Testing Env

```
SUPPORT_CUSTOM_NETWORKS=true
SUPPORT_CUSTOM_RPCS=true
```

## To test
- [ ] Check logs for build workflow
- [ ] You should see multiple env vars added correctly

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Dev build
         if: github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')
         run: |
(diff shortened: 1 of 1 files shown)


Shadowfiend (approved):
It worked! Lovely!

Checked the generated code, flags are on as expected, checked the generated code on main and they are off, so it's not just an accident that they're on in this PR 😁 

Let's ship it!

PR #3271, opened 2023-04-12
Custom networks - Fixes/Polishing

hyphenized:
Fixes the following issues found during QA:

- Removing a network does not remove it from in memory tracked networks
- Removing a network does not remove stored permissions that reference such network
- Invalid database index on internal ethereum provider db
- Block explorer is not persisted with custom network data
- dApp disconnects after adding a network
- Entering the asset page for a custom network asset crashes the page

## To Test
- [x] Add a network, remove it, try to add it again
- [x] Add a network, dApp should not disconnect
- [x] Add a custom asset, enter the activity list for the asset from the asset list

Latest build: [extension-builds-3271](https://github.com/tahowallet/extension/suites/12326907158/artifacts/653841586) (as of Tue, 18 Apr 2023 22:53:24 GMT).


kkosiorowska (commented):
> * Add a network, dApp should not disconnect

Unfortunately, when I have connected to uniswap and then added another network I was disconnected from the dapp

> * Add a custom asset, enter the activity list for the asset from the asset list

Do you know of any test custom asset that currently does not yet exist? I would like to test if the adding actually works.

0xDaedalus:
> Do you know of any test custom asset that currently does not yet exist? I would like to test if the adding actually works.

If you go to a given chain's block explorer - it will generally have a `Top ERC-20 Tokens` or `Top ERC-20 Tokens my Marketcap` section you can browse through.  This is what it looks like on moonscan:

image

You should be able to grab any token you want from there

kkosiorowska (commented):
In my case, it takes quite a long time to display the new token. The user gets a message that the resource has been added, but it does not appear in the list for a long while. It is a bit confusing and I wonder if we can improve it somehow ? 🤔

https://user-images.githubusercontent.com/23117945/232697953-989c2b36-bd92-40ae-8f00-c62ba351352f.mov

background/networks.ts
```diff
@@ -62,6 +62,10 @@ export type SmartContract = NetworkSpecific & {
 export type EVMNetwork = Network & {
   chainID: string
   family: "EVM"
+  /**
+   * Provided for custom networks
+   */
+  blockExplorerURL?: string
```
0xDaedalus, line 68: 📈

0xDaedalus (approved):
Woo! lets gooo 🚢

0xDaedalus (approved):

PR #3274, opened 2023-04-14
Fix a snackbar position for tabbed onboarding

### LINKED ISSUE #3151: Fix toast positioning within tab pages
E.g. when we toggle default wallet on, the toast should be under the button and not centered

image
image

kkosiorowska:
Closes #3151 

This PR fixes a snackbar position for tabbed onboarding.  When we toggle default wallet on, the snackbar should be under the button and not centered.

### UI

https://user-images.githubusercontent.com/23117945/231990342-2ea58c25-1745-420f-adbf-bb3c293fdfbe.mov

Latest build: [extension-builds-3274](https://github.com/tahowallet/extension/suites/12281069291/artifacts/650518670) (as of Mon, 17 Apr 2023 10:12:25 GMT).

### DIFF
--- a/ui/components/Snackbar/Snackbar.tsx
+++ b/ui/components/Snackbar/Snackbar.tsx
@@ -1,11 +1,21 @@
-import React, { ReactElement, useCallback, useEffect, useRef } from "react"
+import React, {
+  ReactElement,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from "react"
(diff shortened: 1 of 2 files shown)


jagodarybacka (changes requested):
If user has already added account then we have a single panel (while adding second account) and snackbar would look better centered:
image

jagodarybacka (approved):
Looks good ✅

PR #3275, opened 2023-04-14
WIP linear build

Shadowfiend:
Latest build: [extension-builds-3275](https://github.com/tahowallet/extension/suites/12249334851/artifacts/648236505) (as of Fri, 14 Apr 2023 21:02:35 GMT).


Shadowfiend:
Turns out secrets are elided in job outputs, so you can't pass secrets from one job to another, making this approach less useful than originally anticipated... Will ponder more, but closing for now.

PR #3277, opened 2023-04-17
Update Alchemy `getTokenMeta` validator

0xDaedalus:
Alchemy will sometimes return token metadata with a `null` value for the token's decimals.

### To Test
- [x] Import vlitalik.eth
- [x] Ensure you don't see the following message in the console:

```
Alchemy token metadata response didn't validate, did the API change? 
{decimals: null, logo: null, name: 'VOS.AI Invitation', symbol: 'VOS.AI'}
```

Latest build: [extension-builds-3277](https://github.com/tahowallet/extension/suites/12294212981/artifacts/651443453) (as of Mon, 17 Apr 2023 19:07:41 GMT).


background/lib/alchemy.ts
```diff
@@ -261,7 +261,7 @@ export async function getTokenMetadata(
     throw new Error("Alchemy token metadata response didn't validate.")
   }
   return {
-    decimals: json.decimals,
+    decimals: json.decimals ?? 0,
```
hyphenized, line 264: Why not 18? 🤔

0xDaedalus: Good question - its because a token without a `decimals` property won't be able to use that `decimals` property to convert from units of `ether` to a "human readable" number.  If the token's decimals are hard-coded elsewhere in the contract then there's really nothing we can do about it - so I think it makes sense to assume `0`.

Shadowfiend:
Just a very small question here before I test + merge: is this also testable via adding the symbol in question as a custom token? (With the flag flipped ofc.)

Shadowfiend (approved):
Verified the error in question happened before and didn't after this change. Also checked that the assets in question made it into the indexing DB with the right default `decimals`. Let's :shipit:

PR #3281, opened 2023-04-18
Bump Dexie to 3.0.4

Shadowfiend:
Though the underlying fix isn't to something we directly call, better safe than sorry on the data layer.

Should be an easy merge here, supersedes #2850 .


Shadowfiend:
Closing in favor of #3415 .

Issue #3276, opened 2023-04-17
Can't correctly import account using recovery phrase when it was earlier imported using private key and removed


michalinacienciala:
Steps to reproduce:
1. After unpacking the extension import at least one account (with whatever method).
2. Add some other account using private key. Added under `Private key` category.
3. Remove that account. Removed
4. Add that account using recovery phrase. Added under `Read Only` account (NOOK).
5. Remove that account. Removed.
6. Add that account using recovery phrase. Not added at all (NOOK)

If after that the account is added using private key, it's added correctly. 

Observed on: [extension-builds-3266](https://github.com/tahowallet/extension/suites/12238675333/artifacts/647467018)

jagodarybacka:
Fixed in #3282

PR #3282, opened 2023-04-18
Fix unhiding removed accounts

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3276

### What
If the account is added again, it should no longer be hidden, whether it was added with a private key or mnemonic.

### Testing

Follow the steps described in the linked task or:

- add account with private key (first you have to export this private key in Metamask from the HD wallet with mnemonic you know)
- remove that account
- add HD wallet with mnemonic you used to export private key

You should be able to add the same address with both private key and mnemonic

Latest build: [extension-builds-3282](https://github.com/tahowallet/extension/suites/12315887155/artifacts/653029617) (as of Tue, 18 Apr 2023 14:45:36 GMT).

### DIFF
--- a/background/services/keyring/index.ts
+++ b/background/services/keyring/index.ts
@@ -408,6 +408,7 @@ export default class KeyringService extends BaseService<Events> {
 
     if (!address) return null
 
+    this.#hiddenAccounts[address] = false
     await this.persistKeyrings()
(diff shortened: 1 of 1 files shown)


hyphenized (approved):
Tested, works well

PR #3283, opened 2023-04-19
Block the signing button when a limit is changing

### LINKED ISSUE #3135: Infinite approval even after manually specifying a lower limit
I recently made a deposit on Convex, which requested an infinite approval of a Curve LP token. I edited to approval to 100 of that same token, and signed... only the realize the wallet had reverted to an infinite approval :(

kkosiorowska:
Closes #3135 

This PR makes sure that the user sets the correct limit. The problem is that the user has the option to sign the transaction without approving the limit. It may appear to him that he has changed the limit, however, the limit will be set correctly when he clicks the confirm button. The changes ensure that the limit is set correctly.

### UI
https://user-images.githubusercontent.com/23117945/233965993-23a88e83-a5a8-449b-ae5a-18bfdc4e2e58.mov

### To Test
- [x] check if the sign button is disabled when the user is updating the limit field (check for ledger account also)

Latest build: [extension-builds-3283](https://github.com/tahowallet/extension/suites/12442603973/artifacts/662365206) (as of Mon, 24 Apr 2023 14:05:13 GMT).


jagodarybacka:
I'll review 👀

ui/utils/signing.ts
```diff
@@ -0,0 +1,9 @@
+import React from "react"
+
+export type SignerFrameContextValue = {
+  shouldBlockedSigning: boolean
+  toggle: (value: boolean) => void
```
jagodarybacka, line 5: I would specify what is this toggling in the function name

jagodarybacka (changes requested):
Works fine, checked other types of transactions too and they are all ok. 

I wonder if we can somehow avoid using context here, as this is adding another layer of state management. But on the other hand, this button locking mechanism is purely on the frontend side so I'm ok leaving it here. Let's improve naming and we are good to go I think.

Shadowfiend:
I suggest we use redux state instead of a context here. We can make it fairly generic to support any situation where an edit action is happening in the signing flow, e.g. an `additionalSigningStatus` redux field that can be set to `"editing" | undefined`.

A slightly bigger issue IMO: the current approach doesn't really tell the user why the sign button is disabled—say they start editing, update, and forget to hit save. Then they see they can't sign. How do they know that all they have to do is click save?

kkosiorowska:
> I suggest we use redux state instead of a context here. We can make it fairly generic to support any situation where an edit action is happening in the signing flow, e.g. an `additionalSigningStatus` redux field that can be set to `"editing" | undefined`.

Let's do it this way.

> A slightly bigger issue IMO: the current approach doesn't really tell the user why the sign button is disabled—say they start editing, update, and forget to hit save. Then they see they can't sign. How do they know that all they have to do is click save?

@VladUXUI Do you have any suggestions to solve this problem?

VladUXUI:
Of the top of my head, we should add “you have unsaved changed” either as a notification or as a tooltip on hover over disabled button.
or just autosave the amount after 3/4 seconds of pause in writing 

I would go with tooltip tbh, it has the least amount of questions and unknowns

ui/components/Signing/Signer/TransactionButton.tsx
```diff
@@ -86,22 +89,40 @@ export default function TransactionButton({
     }
   }, [focusChangeNonce])
 
+  const renderButton = () => (
```
jagodarybacka, line 92: Idk if we are getting much cleaner code making it a function

jagodarybacka (approved):
Works fine

PR #3284, opened 2023-04-19
Submit swap button should be disabled if there is a setup error

### LINKED ISSUE #3230: Submit swap button should be disabled in case of setup error
User should not be able to click "Get final quote" if there is error like "insufficient balance"
image

kkosiorowska:
Closes #3230 

This PR fixes the issue with submit swap button. The user should not be able to click button if there is an error like "insufficient balance". 

### UI
Screenshot 2023-04-19 at 15 55 08

### To Test
- [ ] Check if submit swap button is disabled in case of setup error


jagodarybacka:
I'll review 👀

jagodarybacka (approved):
Works fine ✅

Issue #3269, opened 2023-04-11
Input Fields: Ui fix


VladUXUI:
Input fields labels currently have a solid background. This is not working that well on different background colours.
if possible it would be good to have a transparent background, but so that the line isn't visible

Current
![image](https://user-images.githubusercontent.com/80846886/231195716-20fd6563-7a7b-478b-945e-30632b4f75cc.png)

Ideal
![image](https://user-images.githubusercontent.com/80846886/231195761-93808828-f8e6-4384-9bff-436ba4a4c277.png)

hakeemullahjan:
@VladUXUI please can you assign this to me?

PR #3291, opened 2023-04-20
Make input label's background transparent

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3269

### What
Try to make transparent background that will ignore input border by using blur effect.

WIP - Animation is not perfect as blur catches border color during the transition.

image

https://user-images.githubusercontent.com/20949277/233369054-187cae70-a046-43ce-88d1-dbf40176278b.mov

Latest build: [extension-builds-3291](https://github.com/tahowallet/extension/suites/12368619258/artifacts/656954723) (as of Thu, 20 Apr 2023 12:47:03 GMT).

### DIFF
--- a/ui/components/Keyring/KeyringUnlock.tsx
+++ b/ui/components/Keyring/KeyringUnlock.tsx
@@ -95,7 +95,6 @@ export default function KeyringUnlock({
                 setErrorMessage("")
               }}
               errorMessage={errorMessage}
-              focusedLabelBackgroundColor="var(--green-95)"
             />
           </div>
           <div>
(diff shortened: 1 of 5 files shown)


Shadowfiend:
So, this is imperfect but I feel like it's still workable as-is. Unless you're morally opposed @jagodarybacka I would love to let Vlad look at it and potentially land it as-is?

jagodarybacka:
Well Vlad can take a look but to me it is not ready yet, but I'll let you decide if it looks better or worse than with the background.

VladUXUI:
but judging by the video, i'm not fan of the blinking orange on transition
i would say close the pr and try to improve another time.

But this is an awesome step in the right direction :D, thank you @jagodarybacka

jagodarybacka:
Yeah I believe it was a right direction but more polishing will be needed, let's drop it for now

Shadowfiend:
Thanks folks!

PR #3292, opened 2023-04-20
Fix the UI issues  when swaps are disabled

kkosiorowska:
This PR solves ui issues found in PR solves ui issues found in https://github.com/tahowallet/extension/pull/3205#pullrequestreview-1393765519

### UI 
https://user-images.githubusercontent.com/23117945/233371373-b8116ac3-c908-4336-a5a2-174564b65d7d.mov

### DIFF
--- a/ui/components/Wallet/WalletAccountBalanceControl.tsx
+++ b/ui/components/Wallet/WalletAccountBalanceControl.tsx
@@ -62,15 +62,21 @@ function ActionButtons(props: ActionButtonsProps): ReactElement {
               </SharedSquareButton>
             </div>
           ) : (
-            <SharedTooltip
-              type="dark"
-              width={180}
-              horizontalPosition="center"
-              verticalPosition="bottom"
-              verticalShift={-30}
-              horizontalShift={30}
-              IconComponent={() => (
-                <div className="button_wrap">
+            <div className="button_wrap">
+              <SharedTooltip
+                type="dark"
+                width={180}
+                horizontalPosition="center"
+                verticalPosition="bottom"
+                verticalShift={-30}
+                horizontalShift={22}
+                customStyles={{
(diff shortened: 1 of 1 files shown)


jagodarybacka:
I'll review 👀

jagodarybacka (commented):
I know I'm nitpicking but if we are improving it then let's improve it for good 😉 
I would say (@VladUXUI please correct me if I'm wrong):
- all 3 tooltips with this message should have the same fontsize
- if we will decide on a smaller font size (which looks better IMO) then the line height looks like it's too large
-   let's adjust margins between tooltips and their elements (tooltip touching button looks worse IMO but the one on the item list is too far away)
image
image
image

VladUXUI:
> I know I'm nitpicking but if we are improving it then let's improve it for good 😉 I would say (@VladUXUI please correct me if I'm wrong):

You are not nitpicking and you are not wrong!
Agree here that we should standardise everywhere. Tooltip dark should have only one design, if i did anything different in figma, slap me silly about it :D

here are the details for it.
background = green-120
Font = green 20
Font size = 14px
Font Padding = 16 px
Font alignment = centre
Margin = 8px (all around, left/right/top/bottom)

Let me know if that helps to clarify things

kkosiorowska:
I standardized the spacing as much as I could. Personally, I think that if we want to change the margins for tooltip we would have to do it in a separate issue. The SharedTooltip component has `horizontalShift` and `verticalShift` properties. Changing the margin can have a negative impact on some tooltips I think.

Screenshot 2023-04-21 at 12 09 49

Screenshot 2023-04-21 at 12 09 42

Screenshot 2023-04-21 at 12 09 56

jagodarybacka (approved):
Alright they are lookin much better ✅

PR #3300, opened 2023-04-21
Change expect timeout on CI to handle loading NFTs

jagodarybacka:
### What
We have a problem with the E2E test that was checking NFTs balance after a wallet was loaded.
Seems like this request may take longer than 30s but CI wasn't waiting long enough. Let's try making `expect` timeout 60s. Tried 5 times on this PR and it seems fine.

Latest build: [extension-builds-3300](https://github.com/tahowallet/extension/suites/12437097546/artifacts/661990529) (as of Mon, 24 Apr 2023 10:28:04 GMT).


kkosiorowska (approved):
LGTM ✅

PR #3302, opened 2023-04-21
Custom Networks - Add delete network confirmation modal

hyphenized:
Adds a confirmation modal when removing networks 

image

## Testing Env

```
SUPPORT_CUSTOM_NETWORKS=true
SUPPORT_CUSTOM_RPCS=true
```

## To Test

- [x] Attempt to remove a custom network
- [x] Closing or cancelling the modal should not delete the network
- [x] After confirming the network should be removed from the wallet

Latest build: [extension-builds-3302](https://github.com/tahowallet/extension/suites/12461669678/artifacts/663857786) (as of Tue, 25 Apr 2023 07:54:29 GMT).


kkosiorowska:
I'll review 👀

ui/pages/Settings/SettingsCustomNetworks.tsx
```diff
@@ -29,8 +33,96 @@ export default function SettingsCustomNetworks(): ReactElement {
...
+  const [showConfirmDelete, setShowConfirmDelete] = useState(false)
+  const [networkToDelete, setNetworkToDelete] = useState<EVMNetwork | null>(
+    null
+  )
+
+  const handleModalConfirm = () => {
+    if (networkToDelete) {
+      dispatch(removeCustomChain(networkToDelete.chainID))
+      setShowConfirmDelete(false)
+      setNetworkToDelete(null)
```
kkosiorowska, line 45: Non-blocking comment
It could be packaged into a separate cleaning function.

kkosiorowska (changes requested):
Looks really good ⚡️Some comments:

- [x] When the network name is long the text jumps to the next line. It doesn't look good in my opinion. Maybe we could improve it somehow?
- [x] Also when the text jumps to the next lines the buttons are truncated.

Screenshot 2023-04-24 at 09 22 00

I have noticed one more thing. It is not within the scope of this PR. I have added the EVMOS network. After removing the network, I see that in the portfolio I still have a preview of the EVMOS token. Shouldn't it disappear when the network is removed?

https://user-images.githubusercontent.com/23117945/233927986-094fa936-0ef7-4b0c-8a7b-b7a88ca6ea54.mov

Shadowfiend (commented):
Need to do a quick test drive but this looks good. 

I had trouble bringing up the add custom RPC page but that's unrelated to this change, will pursue separately.

kkosiorowska:
@Shadowfiend is OOO. Let's not block the main tests.

PR #3303, opened 2023-04-23
project setup minor changes

GermaVinsmoke:
- Added `.yarn/*` in `.gitignore` because it creates `.yarn` folder inside the project and it's dependent on the local device
- `.tool-versions` is for those who are using `asdf` instead of `nvm`
- `"start": "run-p -n -l devtools \"dev {@}\" --"`
   - `-n` - to show task name before the running task (shows which script command is called to run this thing)
   - `-l` - to show task name on each line of the output
- `"@redux-devtools/cli"` because it's needed in `start` script

Latest build: [extension-builds-3303](https://github.com/tahowallet/extension/suites/12638966661/artifacts/685745455) (as of Mon, 08 May 2023 20:19:31 GMT).


.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 16.10.0
```
Shadowfiend, line 1: No clear reason to add this given asdf supports `.nvmrc` directly, and `.tool-versions` doesn't seem to be a formal or informal standard used widely. Maybe I'm missing context here, though?

I don't believe anyone on the core team currently uses asdf specifically, and if that's the case it seems safer to avoid any indication that we maintain compatibility of any sort there.

package.json
```diff
@@ -80,6 +80,7 @@
     "@babel/preset-typescript": "^7.15.0",
     "@babel/register": "^7.14.5",
     "@playwright/test": "^1.31",
+    "@redux-devtools/cli": "^2.0.0",
```
Shadowfiend, line 83: We already have this dependency in the `background` package, but pinned to a 1.0 version. Seems like yarn is hoisting the dependency for us here. 

I think it's right to have the dependency listed explicitly here (and dropped from `background`, probably, as that package doesn't use the CLI tools), but it feels safer to do that without changing the version, and then separately bump the version so we can specifically look at what changes are introduced. 

Lastly, it looks like `yarn.lock` probably needs to be updated? Haven't double checked that though.

Shadowfiend (changes requested):
Thank you for the contribution! Took a quick look and left some notes. Excited for the additional `run-p` flags and the added clarity they'll bring 💪

.tool-versions
GermaVinsmoke: Alright, I thought `asdf` was getting more users so maybe it'll be a good addition.
I checked on the internet, `asdf` provides legacy support for files like `.nvmrc`.
This can be removed.

package.json
GermaVinsmoke: So, I'll have to replace this version with the `^`1.0.7` right?
About `yarn.lock`, I'm thinking what to do because it's updating the whole `yarn.lock` file for some weird reason. Even without touching `package.json` file, it's updating the whole `yarn.lock`.

GermaVinsmoke:
@Shadowfiend  Most of the problems were coming because of wrong version of **yarn**. Switched to classic Yarn v1 and everything worked fine now.

.tool-versions
GermaVinsmoke: Removed this file in new commit

.gitignore
```diff
@@ -10,3 +10,4 @@ app/extension-scripts/api/playground/
 .vscode
 playwright-report
 test-results/
+.yarn/*
```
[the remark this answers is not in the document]
GermaVinsmoke: Was because of v3 so removed this also from `.gitignore`

package.json
GermaVinsmoke: Switched to `^1.0.7`. `yarn.lock` was already having the definition for `@redux-devtools/cli: ^1.0.7`, so there was no update in the file.

package.json
```diff
@@ -80,6 +80,7 @@
     "@babel/preset-typescript": "^7.15.0",
     "@babel/register": "^7.14.5",
     "@playwright/test": "^1.31",
+    "@redux-devtools/cli": "^1.0.7",
```
Shadowfiend, line 83: Can you do me one more favor and drop this dependency from `background/package.json`? Otherwise we'll start having to track in two places when we need to version bump, and I don't think `background` uses that package directly.

Shadowfiend (approved):
Aight lgtm 👍

Shadowfiend:
@GermaVinsmoke was looking at why this didn't already merge—looks like there are some unsigned commits in the branch. I'm going to rebase the commits and sign off on the relevant pieces and merge that way so we can get ourselves out of the current failure mode.

Shadowfiend:
Closing in favor of #3379, which I'm going to instamerge.

PR #3304, opened 2023-04-24
Add workflow commenting release PRs with tests list

michalinacienciala:
The PR introduces a GH Actions workflow which adds a comment with tests list to the PRs with the release candidates (PRs that want to merge `release-*` branches to the `main` branch). The test list should be specified in the `./.github/workflows/test-list/release-test-list.md` file. The comment will be added only once, right after the PR gets created.

TODO:
- [ ] add the file with test list (under `./.github/workflows/test-list/release-test-list.md`)
- [ ] temporarily modify the workflow to test it on different workflow triggers

### DIFF
--- a/.github/workflows/test-list.yml
+++ b/.github/workflows/test-list.yml
@@ -1,0 +1,25 @@
+# This workflow adds a comment with tests list to the PRs with the release
+# candidates (PRs that want to merge `release-*` branches to `main`). The test
+# list is specified in the `./.github/workflows/test-list/release-test-list.md`
+# file. The comment is added only once, right after the PR gets created.
+
+name: Add test list to release PRs
+
+on:
+  pull_request:
+    types:
+      - opened
+    branches:
+      - main
+
+jobs:
+  add-release-test-list:
+    runs-on: ubuntu-latest
+    if: startsWith(github.head_ref, 'release-')
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v3
+      - uses: mshick/add-pr-comment@v2
+        with:
+          message-path: ./.github/workflows/test-list/release-test-list.md


michalinacienciala:
The above comment was created during testing of the workflow. It's a sample of what will be added to the release PRs.

.github/workflows/test-list/release-test-list.md
```diff
@@ -0,0 +1,255 @@
+# Release checklist
+
+This release checklist should be performed before release will be published.
+
+- dev tests should be performed by the dev team
+- if something is not working please create an issue and link it here. Checkbox should be checked only if everything was fine
+- in case of serious issues or doubts you should ask the team first
+
+**Don't forget to approve the PR when checklist will be finished ✅**
```
michalinacienciala, line 9:
```suggestion
**Don't forget to approve the PR when checklist is finished ✅**
```

.github/workflows/test-list/release-test-list.md
```diff
@@ -0,0 +1,255 @@
...
+   - [ ] connect with dapp without MM available using Taho
+   - [ ] connect with dapp with MM available using Taho
+   </details>
+
+<details>
+<summary>⚙️ Settings</summary>
+
+- [ ] check `hide balance under $2` option
+- [ ] check bug reports - export logs
+</details>
```
michalinacienciala: What do you think about adding some checks for other options that are not covered on the list?
For example:
```
- [ ] check enabling test networks - enable/disable, send funds on test network
- [ ] check connected websites - disconnect single / all
- [ ] check links - Need help, Discord, Twitter, GitHub
- [ ] check analytics - on/off, deleting data, privacy policy link
```

michalinacienciala:
@jagodarybacka, let's review each other's changes before merging. I left two small comments to the checklist.

.github/workflows/test-list/release-test-list.md
jagodarybacka: I would be careful with adding new things to this list - I would totally want much more things to be tested here but we want it test most important things.

> check enabling test networks

we had tests about test networks before on the checklist but as test networks are not that important to majority of users and they won't bring us revenue we decided to drop it

> check connected websites

this makes sense and is quick to test, will add

> check links 

links are changing very rarely, no need to test them often

> check analytics

mixed feelings - this is important but will be hard to manually test if data was actually deleted. But we should test it either with E2E or make the unit tests coverage better

jagodarybacka (approved):
I think it's ready 🚀

.github/workflows/test-list/release-test-list.md
```diff
@@ -0,0 +1,232 @@
...
+   - [ ] check balance
+   - [ ] check NFTs
+   - [ ] check abilities
+   - [ ] check activities
+   - [ ] check overview page
+2. Add read-only account with UNS
+   - [ ] check assets
+   - [ ] check balance
+   - [ ] check NFTs
+   - [ ] check abilities
```
kkosiorowska, line 36: I've only noticed it now. Abilities are not available to read-only accounts. It's a bit confusing and I think it's worth adding more precise information about what we want to check here.

jagodarybacka: @kkosiorowska ah sorry my bad, could you remove that and make a PR? Feel free to suggest other tweaks to the checklist. I still want it to be as short and concise as possible, we can assume our release checklist tester will know how to test these points. We can also do a Notion/Coda page with a description of how to test each point - not a need right now but in the future I guess that will be nice to have.

kkosiorowska: Sure, I will do it later today.

PR #3307, opened 2023-04-24
Display a warning message when custom asset balance is dust

hyphenized:
Shows a warning message when the user has enabled "Hide assets under 2$" and attempts to add a token which balance is below the threshold.

image

## To Test
- [x] Apply [this patch](https://gist.github.com/hyphenized/3c513e7399b2b7d82c1142a9919ac107)
- [x] Reinstall the extension, import `bnbspider2.crypto` as a readonly account
- [x] On settings, enable "Hide assets under 2$"
- [ ] Go to the add custom token screen and attempt to add:
		- `0xdbdb4d16eda451d0503b854cf79d55697f90c8df` You should see the dust balance warning
		- `0x8965349fb649a33a30cbfda057d8ec2c48abe2a2` You should not see the warning as there's no price data
		- `0x6c28AeF8977c9B773996d0e8376d2EE379446F2f` You should not see the warning since the $ value exceeds the threshold
- [x] With the hide dust setting turned off it should not display the warning for all previous test cases

Latest build: [extension-builds-3307](https://github.com/tahowallet/extension/suites/12466162180/artifacts/664180075) (as of Tue, 25 Apr 2023 11:12:04 GMT).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -5,6 +5,7 @@ import { configureStore, isPlain, Middleware } from "@reduxjs/toolkit"
 import { devToolsEnhancer } from "@redux-devtools/remote"
 import { PermissionRequest } from "@tallyho/provider-bridge-shared"
 import { debounce } from "lodash"
+import { utils } from "ethers"
 
 import {
   decodeJSON,
@@ -120,7 +121,7 @@ import {
   setDeviceConnectionStatus,
   setUsbDeviceCount,
 } from "./redux-slices/ledger"
-import { OPTIMISM } from "./constants"
+import { OPTIMISM, USD } from "./constants"
 import { clearApprovalInProgress, clearSwapQuote } from "./redux-slices/0x-swap"
 import {
   AccountSigner,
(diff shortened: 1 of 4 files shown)


background/main.ts
```diff
@@ -1845,6 +1849,7 @@ export default class Main extends BaseService<never> {
   ): Promise<{
     asset: SmartContractFungibleAsset
     amount: bigint
+    mainCurrencyAmount: number | undefined
```
kkosiorowska, line 1852: I remember that we talked about the convention that we should use the following version.
```suggestion
    mainCurrencyAmount?: number
```

kkosiorowska (commented):
I haven't tested it yet but minor thoughts.

kkosiorowska (commented):
It seems that everything works as it should. I worked on part of this task, so it would be good for someone else to look at it as well. @jagodarybacka @0xDaedalus 

> 0x6c28AeF8977c9B773996d0e8376d2EE379446F2f You should not see the warning since the $ value exceeds the threshold

The price is currently under $2. Not able to test this point.

jagodarybacka:
I had multiple problems: 
- AVAX `0x8965349fb649a33a30cbfda057d8ec2c48abe2a2` is not recognized as a contract
- tried the same path on `berrry.eth` with [SLP](https://etherscan.io/token/0x0c6f06b32e6ae0c110861b8607e67da594781961) - not found
- with other tokens from PR description - I can't see the warning because prices are not loading 😞

kkosiorowska:
@jagodarybacka  Please check again it should work now.

ui/_locales/en/messages.json
```diff
@@ -551,6 +551,7 @@
       "networkSelect.title": "Select Network",
       "warning.alreadyExists.title": "Asset already exists, you will see it once your balance is over 0",
       "warning.alreadyExists.desc": "If you still have issues seeing an asset, turn off “hide balance 2$” from Settings",
+      "warning.dust.title": "Asset balance is under <$2, you wil see it once your balance is over $2",
```
jagodarybacka, line 554: `under` and `<` seems redundant

jagodarybacka (approved):
Really hard to test without these patches but it seems to work fine

PR #3309, opened 2023-04-25
Refactor keyring redux slice to remove `importing` field

jagodarybacka:
Implementation of changes requested in https://github.com/tahowallet/extension/pull/3089#discussion_r1173088819

### What

Let's remove the `importing` field from the redux keyring slice and handle waiting for new signer import with promises.

### Testing

Everything should work as before:
- [x] import account with mnemonic
- [x] import private key
- [x] import private key with JSON
- [x] generate new wallet
- [x] migration from `main`

Latest build: [extension-builds-3309](https://github.com/tahowallet/extension/suites/12522917658/artifacts/668449636) (as of Thu, 27 Apr 2023 11:38:39 GMT).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1668,38 +1668,8 @@ export default class Main extends BaseService<never> {
     return this.keyringService.exportPrivateKey(address)
   }
 
-  async importSigner(
-    signerRaw: SignerRawWithType
-  ): Promise<{ success: boolean; errorMessage?: string }> {
-    let address = null
-
-    try {
-      address = await this.keyringService.importSigner(signerRaw)
-    } catch (error) {
-      logger.error("Internal signer import failed:", error)
-
-      return {
-        success: false,
-        errorMessage: "Unexpected error during account import.",
-      }
-    }
-
-    if (!address) {
-      return {
-        success: false,
-        errorMessage:
-          "Failed to import new account. Address may already be imported.",
-      }
-    }
-
-    this.store.dispatch(
-      setNewSelectedAccount({
-        address,
-        network: this.store.getState().ui.selectedAccount.network,
-      })
-    )
-
-    return { success: true }
+  async importSigner(signerRaw: SignerRawWithType): Promise<string | null> {
+    return this.keyringService.importSigner(signerRaw)
   }
 
   async getActivityDetails(txHash: string): Promise<ActivityDetail[]> {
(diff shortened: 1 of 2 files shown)


background/main.ts
```diff
@@ -1668,8 +1668,36 @@ export default class Main extends BaseService<never> {
     return this.keyringService.exportPrivateKey(address)
   }
 
-  async importSigner(signerRaw: SignerRawWithType): Promise<HexString | null> {
-    return this.keyringService.importSigner(signerRaw)
+  async importSigner(
+    signerRaw: SignerRawWithType
+  ): Promise<{ success: boolean; errorMessage?: string }> {
```
Shadowfiend, line 1673: How about we make this an alternate?

```suggestion
  ): Promise<{ address: true } | { success: false; errorMessage: string }> {
```

Basically communicate "you will only see an error message in case of failure".

background/main.ts
```diff
@@ -1668,8 +1668,36 @@ export default class Main extends BaseService<never> {
...
+    let address = null
+
+    try {
+      address = await this.keyringService.importSigner(signerRaw)
+    } catch (error) {
+      return {
+        success: false,
+        errorMessage: `Unexpected error during account import. Error: ${error}`,
```
Shadowfiend, line 1681: I'm thinking we shouldn't bubble an error out unguarded to the UI here to avoid leaking potentially sensitive info now or in the future. Logs should catch the real error.

background/redux-slices/keyrings.ts
```diff
@@ -47,26 +44,8 @@ export const emitter = new Emittery<Events>()
...
-    // Set the selected account as the first address of the last added keyring,
-    // which will correspond to the last imported keyring, AKA this one. Note that
-    // this does rely on the KeyringService's behavior of pushing new keyrings to
-    // the end of the keyring list.
-    dispatch(
-      setNewSelectedAccount({
-        address,
-        network: ui.selectedAccount.network,
-      })
-    )
```
Shadowfiend, line 69: Fwiw I do feel like the logic in `main.importSigner` mostly belongs here rather than main--main should always do as little wiring as possible to connect to the services IMO.

Shadowfiend (commented):
Left some small comments but overall looks good. Can't test right now 😔 so not hitting the approval.

hyphenized (approved):
Tested, works well! Will leave open in case you want to address the comments

background/main.ts
jagodarybacka: I wanted to make this type this way but I have a problem making typescript happy about returning something more complicated than a simple object here 🤔 Something weird is going on with types for `createBackgroundAsyncThunk`, I will take a look later if I will have enough time.

hyphenized (approved):
Looks good, let's fix the types for async thunks in a separate ticket.

Issue #3273, opened 2023-04-13
Tokens that are added manually should be visible in the asset list even if they have a balance of zero.


0xDaedalus:
They should, however, still be hidden if the user elects to hide assets with balances under $2

kkosiorowska:
We have decided to close this task because it will add more confusion to user and not enough value.

PR #3313, opened 2023-04-26
Make custom tokens visible in the asset list even if they have a balance of zero

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3273

## What
Currently, when a user adds a token manually, it will not show up in the list of assets when the balance is zero. We should display an asset on the list even when its balance is empty.

## UI
Screenshot 2023-04-26 at 14 20 43
Screenshot 2023-04-26 at 14 20 59

## To Test
- [ ] add a custom token whose balance is zero (`0x841fad6eae12c286d1fd18d1d525dffa75c7effe` for Fandom network)
- [ ] the asset should appear in the list of assets
- [ ] hide dust in the settings, the token should be displayed on the list of hidden assets
- [ ] make sure that we do not show other assets that have a balance of zero

Latest build: [extension-builds-3313](https://github.com/tahowallet/extension/suites/13604126632/artifacts/749906098) (as of Wed, 14 Jun 2023 15:42:00 GMT).


background/redux-slices/utils/asset-utils.ts
```diff
@@ -352,3 +352,10 @@ export function isUntrustedAsset(asset: AnyAsset | undefined): boolean {
   }
   return false
 }
+
+export function isCustomAsset(asset: AnyAsset | undefined): boolean {
```
jagodarybacka, line 356: I would assume the asset will be always present when passed to this function - expecting undefined may be more convenient but may also hide some bugs
```suggestion
export function isCustomAsset(asset: AnyAsset): boolean {
```

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -163,17 +164,17 @@ const computeCombinedAssetAmountsData = (
...
 
       // Hide dust, untrusted assets and missing amounts.
       if (
         isForciblyDisplayed ||
+        (isCustom && !hideDust) ||
         (isTrusted && (hideDust ? isNotDust && isPresent : isPresent))
       ) {
         acc.combinedAssetAmounts.push(assetAmount)
-      } else if (isPresent) {
+      } else if (!isCustom ? isPresent : true) {
```
jagodarybacka, line 177:
```suggestion
      } else if (isCustom || isPresent) {
```

background/redux-slices/accounts.ts
```diff
@@ -310,10 +311,15 @@ const accountSlice = createSlice({
...
-            existingAccountData.balances[updatedAssetSymbol] === undefined &&
-            !isBuiltInNetworkBaseAsset(asset, network) // add base asset even if balance is 0
+            existingAccountData.balances[updatedAssetSymbol] === undefined
           ) {
-            return
+            // add base asset even if balance is 0 or is a custom asset
+            if (
+              !isCustomAsset(asset) &&
+              !isBuiltInNetworkBaseAsset(asset, network)
+            ) {
```
jagodarybacka, line 320:
```suggestion
            if (
              isCustomAsset(asset) ||
              isBuiltInNetworkBaseAsset(asset, network)
            ) {
```

background/redux-slices/accounts.ts
```diff
@@ -310,10 +311,15 @@ const accountSlice = createSlice({
         if (existingAccountData !== "loading") {
           if (
             updatedAccountBalance.assetAmount.amount === 0n &&
-            existingAccountData.balances[updatedAssetSymbol] === undefined &&
-            !isBuiltInNetworkBaseAsset(asset, network) // add base asset even if balance is 0
+            existingAccountData.balances[updatedAssetSymbol] === undefined
           ) {
-            return
+            // add base asset even if balance is 0 or is a custom asset
+            if (
```
jagodarybacka, line 317: Can we combine these two ifs?

jagodarybacka (changes requested):
Some improvements ⬇️ 

Also - on other networks trusted and untrusted assets switched sections:

https://user-images.githubusercontent.com/20949277/235885224-7d7fb89c-31e9-46b9-a572-823f882765c2.mov

background/redux-slices/accounts.ts
kkosiorowska: It isn't what we expect. Take a look at the photo below. 
**Correct condition**
If the balance is zero and it isn't a custom asset and it isn't a base network asset we shouldn't add it to balances (go to return).

Screenshot 2023-05-08 at 10 14 19

kkosiorowska:
Let's wait with this PR until we merge this [one](https://github.com/tahowallet/extension/pull/3250).

hyphenized:
> Let's wait with this PR until we merge this [one](https://github.com/tahowallet/extension/pull/3250).

Good catch! The conditionals for token visibility have already been refactored there but we can also cherry pick some of that work here if we need to push this out faster

0xDaedalus:
Lets add a couple of unit tests here as well - we should just need two:  One for `computeCombinedAssetAmountsData` and one for `updateAccountBalance`

jagodarybacka:
@kkosiorowska Is this PR still valid and we should merge it at some point or not?

kkosiorowska:
> @kkosiorowska Is this PR still valid and we should merge it at some point or not?

@jagodarybacka PR was blocked by Verified/Unverified tokens. It needs to be updated and tested again. I should be looking into it this week.

kkosiorowska:
> I don't think this is a good idea fwiw. If i sell my asset, how do i hide the dust from it?

We are able to force a newly added asset to be displayed (when the balance is 0). In the case where we sell an asset and its balance is close to zero we are able to hide this.

Shadowfiend:
> > I don't think this is a good idea fwiw. If i sell my asset, how do i hide the dust from it?
> 
> We are able to force a newly added asset to be displayed (when the balance is 0). In the case where we sell an asset and its balance is close to zero we are able to hide it.

I think if we've implemented, in the custom token add page, the message that the asset will be added but hidden due to having 0 balance + hide dust being enabled, then we should avoid getting to this level of contortion in the code, because it's going to have the potential to get real nasty. e.g. when do we consider an asset to be ”new“ and subject to this? Do we have to wait for balance to go over 0 and then drop back under it? Are we instead saying the asset is visible at exactly 0 but not at fractionally over 0? Etc.

background/redux-slices/accounts.ts
```diff
@@ -312,7 +313,11 @@ const accountSlice = createSlice({
           if (
             updatedAccountBalance.assetAmount.amount === 0n &&
             existingAccountData.balances[updatedAssetSymbol] === undefined &&
-            !isBuiltInNetworkBaseAsset(asset, network) // add base asset even if balance is 0
+            // add base asset even if balance is 0 or is a custom asset
+            !(
+              isUntrustedAsset(asset) ||
```
hyphenized, line 318: If the asset already exists it will likely be part of a token list assets and not a custom asset, in that case this won't trigger right?

kkosiorowska:
> EDIT to say: if it's not clear, I'm basically for killing this PR completely, BUT I think it's ultimately a call for Vlad/Haley to make 😁

It seems to me that Vlad has noticed an important issue.  I don't think this PR adds much value for the user. I'm okay with closing it.

@VladUXUI  Please let us know what you think.

VladUXUI:
Yeah, let's close. I think this will add more confusion to user and not enough value. Seeing 0 token assets isn't useful for many people. Also as a user i will rarely add a 0 balance asset just for the fun of it

Issue #3326, opened 2023-04-27
Styling regression - extra spacing on txn signing


mr-michael:
<img width="396" alt="Screen Shot 2023-04-26 at 9 40 41 PM" src="https://user-images.githubusercontent.com/7005061/234739501-66d80d67-b29a-4637-a2cc-631a74f44d7c.png">

<img width="380" alt="Screen Shot 2023-04-26 at 9 42 18 PM" src="https://user-images.githubusercontent.com/7005061/234739503-eb52e631-c193-4729-829b-9bb7eb60d2ae.png">

PR #3337, opened 2023-04-28
Fix style regression issue for transaction warning

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3326

Please check this [PR](https://github.com/tahowallet/extension/pull/3088). When the network has the ability to change fees, we should leave extra space for a possible warning that will appear. Otherwise, there should be no extra space. 

For internal swaps, we have an additional tab - "Raw data". When we make an external swap there is a problem with additional space for example for the Optimism network. In this case, we should leave extra space for a possible warning. It is not the best solution but it is a quick fix. 

## UI
Before
Screenshot 2023-04-28 at 11 13 49

After
Screenshot 2023-04-28 at 11 05 10

Screenshot 2023-04-28 at 11 07 32

Screenshot 2023-04-28 at 13 57 49

https://user-images.githubusercontent.com/23117945/235107275-8b4f5fa2-bfc6-4635-95aa-e85349eae4a3.mov

**Extreme case**

https://user-images.githubusercontent.com/23117945/235142515-606be4de-0c6d-442f-b79c-5448ab8872d2.mov

## To Test
- [x] check styles on the internal swap page - when the network has a fee setting (it should have extra space)
- [x]  check styles on the internal swap page - when the network does not have a fee setting (it should not have extra space)
- [x] check styles on the external swap page for custom networks (it should not have extra space) 
- [x] check styles on the external swap page for the Optimism network (it should have extra space)

Latest build: [extension-builds-3337](https://github.com/tahowallet/extension/suites/12547180032/artifacts/670314053) (as of Fri, 28 Apr 2023 09:18:40 GMT).

### DIFF
--- a/background/constants/networks.ts
+++ b/background/constants/networks.ts
@@ -260,3 +260,7 @@ export const CHAIN_ID_TO_OPENSEA_CHAIN = {
   [AVALANCHE.chainID]: "avalanche",
   [BINANCE_SMART_CHAIN.chainID]: "bsc",
 }
+
+export const NETWORKS_WITH_FEE_SETTINGS = new Set(
+  [ETHEREUM, POLYGON, ARBITRUM_ONE, AVALANCHE].map((network) => network.chainID)
+)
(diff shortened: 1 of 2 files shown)


0xDaedalus (approved):
📈

PR #3348, opened 2023-05-01
Custom Networks - Add balances lookup fallback for networks without multicall support

### LINKED ISSUE #3336: Can't add USDC on zkSync era
### Discord Discussion Link

n/a

Chrome

Adding USDC as a token on zkSync era

### What did not work?

Wallet: Version 0.31.0135a1b4 from https://github.com/tahowallet/extension/pull/3147

Wallet address: https://explorer.zksync.io/address/0xf0ee42aa7a347a2d1e9dbdf5cfc42b66843ab33d

Tried to add USDC, token contract:  0x3355df6D4c9C3035724Fd0e3914dE96A5a83aaf4

Result in Taho:
Screen Shot 2023-04-28 at 08 50 36

Result in MM:
Screen Shot 2023-04-28 at 08 53 34

Screen Shot 2023-04-28 at 08 58 01

Same applies to WETH: 0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91

### Version

v0.31.0

### Relevant log output

_No response_

hyphenized:
Closes #3336

Loads token balances through multiple requests on networks where the multicall contract might not be available

## To Test
- [ ] Comment out [this line](https://github.com/tallycash/extension/blob/35391a5891ee21ba41984d3106663c826518cb11/background/lib/multicall.ts#L70)
- [ ] Import this address as read only `0xf0ee42aa7a347a2d1e9dbdf5cfc42b66843ab33d`
- [ ] Import this token `0x3355df6D4c9C3035724Fd0e3914dE96A5a83aaf4` on zkSync era
- [ ] Importing the token should work and there should be no errors logged in the background page

Latest build: [extension-builds-3348](https://github.com/tahowallet/extension/suites/12696304699/artifacts/681316461) (as of Fri, 05 May 2023 09:40:23 GMT).

### DIFF
--- a/background/lib/erc20.ts
+++ b/background/lib/erc20.ts
@@ -75,7 +75,7 @@ export async function getBalance(
   provider: BaseProvider,
   tokenAddress: string,
   account: string
-): Promise<BigInt> {
+): Promise<bigint> {
   const token = new ethers.Contract(tokenAddress, ERC20_ABI, provider)
 
   return BigInt((await token.balanceOf(account)).toString())
--- a/background/lib/erc20.ts
+++ b/background/lib/erc20.ts
@@ -7,6 +7,7 @@ import {
   FunctionFragment,
   TransactionDescription,
 } from "ethers/lib/utils"
+import { poll, PollOptions } from "@ethersproject/web"
 import { SmartContractAmount, SmartContractFungibleAsset } from "../assets"
 import { EVMLog, SmartContract } from "../networks"
 import { HexString } from "../types"
@@ -16,8 +17,10 @@ import {
   CHAIN_SPECIFIC_MULTICALL_CONTRACT_ADDRESSES,
   MULTICALL_ABI,
   MULTICALL_CONTRACT_ADDRESS,
+  networkSupportsMultiCall,
 } from "./multicall"
 import logger from "./logger"
+import { isFulfilledPromise } from "./utils/type-guards"
 
 export const ERC20_FUNCTIONS = {
   allowance: FunctionFragment.from(
@@ -72,7 +75,7 @@ export const ERC2612_INTERFACE = new ethers.utils.Interface(ERC2612_ABI)
  * Get an account's balance from an ERC20-compliant contract.
  */
 export async function getBalance(
-  provider: BaseProvider,
+  provider: Provider,
   tokenAddress: string,
   account: string
 ): Promise<bigint> {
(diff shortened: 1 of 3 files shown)


background/lib/erc20.ts
```diff
@@ -193,41 +196,82 @@ export const getTokenBalances = async (
...
+    ceiling: 500,
+    timeout: 5000,
+    retryLimit: 2,
+  }
+
+  // Attempt to fetch all balances through multiple RPC requests
+  // There's a risk of going to get rate limited depending on the
+  // number of tokens queried for the network
+
+  const balances = await Promise.allSettled(
```
0xDaedalus, line 249: Have you performance tested this?  I would assume if you import an account with over ~100 tokens this would absolutely blow up on a low-resource machine

background/lib/multicall.ts
```diff
@@ -10,13 +10,70 @@ export interface AggregateContractResponse {
   }>
 }
 
+const NETWORKS_SUPPORTING_MULTICALL_BY_CHAIN_ID = [
```
0xDaedalus, line 13: I think this is going to result in a net increase of headaches down the line.  Multicall is at this point basically a public good being deployed to more and more chains every day (through the official channel or not) and this is going to be a pain to maintain.  

I think by merging this PR we're trading one class of bug for another - IMO its better to have functionality that is clearly breaking (token won't load at import) that we can address by updating a contract address than to have functionality that is ambiguously breaking (token balances aren't loading, app feels sluggish) that we first have to diagnose and then still make an update (in this case - to chain ID).

background/lib/erc20.ts
```diff
@@ -193,41 +196,82 @@ export const getTokenBalances = async (
...
+    tokenAddresses.map(
+      async (contractAddress): Promise<SmartContractAmount> => {
+        // Poll will keep retrying until:
```
0xDaedalus, line 252: If this takes more than 1 minute to resolve - we'll get redundant stacking requests from the balance alarm

background/lib/multicall.ts
Shadowfiend: Worth noting that multicall is *something we can deploy* if we're talking about an EVM chain. We could even prompt for it in the UI if we wanted.

jagodarybacka (commented):
Which network should be used to import `0x3355df6D4c9C3035724Fd0e3914dE96A5a83aaf4`?
It seems to fail for every network I try 🤔

background/lib/erc20.ts
hyphenized: I have not however this should not be an issue for custom networks as most of the tokens will likely have been imported manually by the users.

background/lib/erc20.ts
hyphenized: The polling timeout set for every request should prevent this from taking one minute

background/lib/multicall.ts
0xDaedalus: I'm a bit fearful that the functionality introduced in this PR falls into the "one step forward, two steps back" category but will leave the decision on wether we want to merge or not in your hands.

Keep in mind that with this implementation:
- We'll have to wait for chrome extension releases to get token querying working on a new chain.
- We'll need to think of a way to get this working on hardhat forks or local networks with custom chain ids (might impact @michalinacienciala's e2e CI work)
- The core team will be the lynchpin when token functionality is needed on new networks, rather than any member of the community (or core team) deploying a contract on a given chain.

Do you think the pros in this PR outweigh the cons listed above?

background/lib/erc20.ts
jagodarybacka: Can we explicitly fetch it just for custom tokens?

background/lib/multicall.ts
jagodarybacka: If we have so many cons then we should do thorough tests before merging including testing on slower machines and on Windows too considering the worst scenarios we can think of

Shadowfiend:
Ok had a deep look here. I think there's value to this as a research path, but I don't think we have the bandwidth to do the diligence needed to feel good about this long term. I'm going to close the PR, but leave a placeholder issue linking back to it for future investigation.

PR #3355, opened 2023-05-02
Catch Enter keypress on Unlock screen

jagodarybacka:
### What
To prevent Enter keypress to mess with other components listening for that key let's catch it on the Unlock screen so user can type password and press Enter to confirm.

### Testing
Confirm you can unlock by pressing Enter on:

- [x] full screen unlock screen - tabbed onboarding
- [x] popup full screen - for example while adding or removing accounts
- [x] slideup - export mnemonic
- [x] slideup - export private key

image

Latest build: [extension-builds-3355](https://github.com/tahowallet/extension/suites/12615522781/artifacts/675227855) (as of Tue, 02 May 2023 10:41:45 GMT).


kkosiorowska (approved):
LGTM :white_check_mark:

PR #3358, opened 2023-05-03
Fix issue with getting abilities

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3363

## What

- Update the parameters for getting the abilities because there were changes to some parameters. Removal of unnecessary.
- Fetch open, completed and expired abilities to update the cache. This is required to compare the state of the data from the API with the state in the cache. There is no way to get information from the Daylight API about the removed abilities. 
- Removal of an out-of-date link -  `https://docs.daylight.xyz/reference/retrieve-wallets-abilities`

## UI
https://user-images.githubusercontent.com/23117945/236416129-ad2e4cc7-6c4c-4807-8aaf-46f0574a3495.mov

## To Test
Testing is difficult because we need an account with old cache abilities that no longer exist in the Daylight API. You can use the prepared patch file. The first refresh action adds abilities to the cache. The second allows you to call the `removeAbilities` method.

- [x] Use the patch file
- [x] Install new extension
- [x] Test ability should appear as open
- [x] Wait a minute and then open the waller again
- [x] Ability should be moved to deleted

```
git apply <<- EOF
diff --git a/background/services/abilities/index.ts b/background/services/abilities/index.ts
index b19283d1..b33b1301 100644
--- a/background/services/abilities/index.ts
+++ b/background/services/abilities/index.ts
@@ -12,7 +12,7 @@ import ChainService from "../chain"
 import { normalizeEVMAddress } from "../../lib/utils"
 import { Ability, AbilityRequirement } from "../../abilities"
 import LedgerService from "../ledger"
-import { HOUR } from "../../constants"
+import { MINUTE } from "../../constants"
 
 const normalizeDaylightRequirements = (
   requirement: DaylightAbilityRequirement
@@ -137,6 +137,25 @@ export default class AbilitiesService extends BaseService<Events> {
 
     const newAbilities: Ability[] = []
 
+    const shouldRemoveAbility = localStorage.getItem("shouldRemoveAbility")
+    if (shouldRemoveAbility === null) {
+      normalizedAbilities.push({
+        type: "mint",
+        title: "Test Ability",
+        description: null,
+        abilityId: "test-abilityId",
+        linkUrl: "",
+        completed: false,
+        removedFromUi: false,
+        address: address as NormalizedEVMAddress,
+        requirement: {
+          type: "hold",
+          address: "",
+        },
+        slug: "test-slug",
+      })
+    }
+
     await Promise.all(
       normalizedAbilities.map(async (ability) => {
         const isNewAbility = await this.db.addNewAbility(ability)
@@ -149,12 +168,15 @@ export default class AbilitiesService extends BaseService<Events> {
     if (newAbilities.length) {
       this.emitter.emit("newAbilities", newAbilities)
     }
-    /**
-     * We are not able to get information about the removed abilities from the Daylight API.
-     * To update the cache we have to compare the data with the received abilities.
-     * The ability can be open completed or expired. Therefore, we need to get the abilities for these 3 types.
-     */
-    await this.removeAbilities(address, normalizedAbilities)
+
+    if (shouldRemoveAbility === "true") {
+      /**
+       * We are not able to get information about the removed abilities from the Daylight API.
+       * To update the cache we have to compare the data with the received abilities.
+       * The ability can be open completed or expired. Therefore, we need to get the abilities for these 3 types.
+       */
+      await this.removeAbilities(address, normalizedAbilities)
+    }
   }
 
   async markAbilityAsCompleted(
@@ -182,7 +204,7 @@ export default class AbilitiesService extends BaseService<Events> {
   async refreshAbilities(): Promise<void> {
     const lastFetchTime = localStorage.getItem(this.ABILITY_TIME_KEY)
 
-    if (lastFetchTime && Number(lastFetchTime) + HOUR > Date.now()) {
+    if (lastFetchTime && Number(lastFetchTime) + MINUTE > Date.now()) {
       return
     }
     localStorage.setItem(this.ABILITY_TIME_KEY, Date.now().toString())
@@ -194,6 +216,13 @@ export default class AbilitiesService extends BaseService<Events> {
     for (const address of addresses) {
       this.emitter.emit("initAbilities", address as NormalizedEVMAddress)
     }
+
+    const counter = localStorage.getItem("counter")
+    if (counter === "1") {
+      localStorage.setItem("shouldRemoveAbility", "true")
+    }
+    const newValue = counter ? +counter : 0 + 1
+    localStorage.setItem("counter", newValue.toString())
   }
 
   async reportAndRemoveAbility(


EOF
```

Latest build: [extension-builds-3358](https://github.com/tahowallet/extension/suites/12776968774/artifacts/687252540) (as of Tue, 09 May 2023 15:12:15 GMT).


hyphenized:
Reviewing 👀

hyphenized (approved):
Works well! Left 2 minor comments

background/services/abilities/index.ts
```diff
@@ -113,6 +113,23 @@ export default class AbilitiesService extends BaseService<Events> {
...
+    address: HexString,
+    abilities: Ability[]
+  ): Promise<void> {
+    const cachedAbilities = await this.db.getAllAbilities()
+    const diffAbilities = cachedAbilities.filter(
+      (cachedAbility) =>
+        !abilities.find(
+          ({ abilityId }) => cachedAbility.abilityId === abilityId
+        )
+    )
```
jagodarybacka, line 126: I doubt it will be a real problem as users probably won't own that much abilities but can we make this algorithm a little bit better by mapping `abilities` ids instead of nesting another loop?

Something like:
```suggestion
    const ids = new Set(abilities.map(a => a.abilityId))
    const diffAbilities = cachedAbilities.filter(
      (cachedAbility) =>
        !ids.has(cachedAbility.abilityId)
    )
```

background/services/abilities/index.ts
```diff
@@ -113,6 +113,23 @@ export default class AbilitiesService extends BaseService<Events> {
...
+
+    diffAbilities.forEach(({ abilityId }) =>
+      this.markAbilityAsRemoved(normalizeEVMAddress(address), abilityId)
+    )
+  }
```
jagodarybacka, line 131: Does it make sense to make it a bulk update on database instead of marking them one by one?

kkosiorowska:
The approach to updating abilities will be changed. Let's fix all issues in one PR. https://github.com/tahowallet/extension/pull/3375

PR #3354, opened 2023-05-02
Fix intro copy

0xDaedalus:
Closes https://github.com/tahowallet/extension/issues/3315

image

Latest build: [extension-builds-3354](https://github.com/tahowallet/extension/suites/12606909335/artifacts/674566777) (as of Tue, 02 May 2023 01:17:42 GMT).

PR #3339, opened 2023-04-28
Add zksync era specific multicall contract address

0xDaedalus:
Contract scan link: https://explorer.zksync.io/address/0x47898B2C52C957663aE9AB46922dCec150a2272c#contract

### To Test
- [ ] Add zkSync Era to the wallet.
- [ ] Attempt to add an ERC-20 token (e.g. [Liquity].(https://explorer.zksync.io/address/0x503234F203fC7Eb888EEC8513210612a43Cf6115))
- [ ] Adding the token should not fail.

PR #3335, opened 2023-04-27
Add (beta) label to custom network menu options

0xDaedalus:
image

image

image

Latest build: [extension-builds-3335](https://github.com/tahowallet/extension/suites/12538186689/artifacts/669589536) (as of Thu, 27 Apr 2023 22:27:38 GMT).

PR #3332, opened 2023-04-27
Redirect to the activity page after submitting a swap

kkosiorowska:

PR #3361, opened 2023-05-03
v0.32.0

jagodarybacka:
## What's Changed
* Custom Networks - Copy changes by @hyphenized in https://github.com/tahowallet/extension/pull/3328
* Add (beta) label to custom network menu options by @0xDaedalus in https://github.com/tahowallet/extension/pull/3335
* Fix style regression issue for transaction warning by @kkosiorowska in https://github.com/tahowallet/extension/pull/3337
* Add zksync era specific multicall contract address by @0xDaedalus in https://github.com/tahowallet/extension/pull/3339
* Add workflow commenting release PRs with tests list by @michalinacienciala in https://github.com/tahowallet/extension/pull/3304
* Fix intro copy by @0xDaedalus in https://github.com/tahowallet/extension/pull/3354
* Change custom network copy to "EVM-compatible blockchain" by @0xDaedalus in https://github.com/tahowallet/extension/pull/3351
* Round corners on custom network icons by @0xDaedalus in https://github.com/tahowallet/extension/pull/3352
* Redirect to the activity page after submitting a swap by @kkosiorowska in https://github.com/tahowallet/extension/pull/3332
* v0.31.1 by @hyphenized in https://github.com/tahowallet/extension/pull/3329
* Custom Networks - QA Build by @hyphenized in https://github.com/tahowallet/extension/pull/3147
* Impersonate MM on SpaceID by @jagodarybacka in https://github.com/tahowallet/extension/pull/3359
* Fix top network selector styles when network name is long by @jagodarybacka in https://github.com/tahowallet/extension/pull/3357

**Full Changelog**: https://github.com/tahowallet/extension/compare/v0.31.1...v0.32.0
Latest build: [extension-builds-3361](https://github.com/tahowallet/extension/suites/12651203301/artifacts/677889151) (as of Wed, 03 May 2023 16:05:22 GMT).

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG.yml
+++ b/.github/ISSUE_TEMPLATE/BUG.yml
@@ -51,6 +51,7 @@ body:
       label: Version
       description: What version of the extension are you running?
       options:
+        - v0.32.0
         - v0.31.1
         - v0.31.0
         - v0.30.1
--- a/manifest/manifest.json
+++ b/manifest/manifest.json
@@ -1,6 +1,6 @@
 {
   "name": "Taho",
-  "version": "0.31.1",
+  "version": "0.32.0",
   "description": "The community owned and operated Web3 wallet.",
   "homepage_url": "https://taho.xyz",
   "author": "https://taho.xyz",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tallyho/tally-extension",
   "private": true,
-  "version": "0.31.1",
+  "version": "0.32.0",
   "description": "Taho, the community owned and operated Web3 wallet.",
   "main": "index.js",
   "repository": "git@github.com:thesis/tally-extension.git",


SJQuality (approved):
QA Approved

beemeeupnow:
After adding a chain via chainlist the "back" arrow at the top-left of this page is not working:

stuck_on_custom_networks

beemeeupnow:
It would be nice if we detected when the user types small or specific amounts and have display behave accordingly.
(instead of showing 0, show 0.001 if that is what the user typed)

Also would be good to have a visual display of the gas unit. "~1.1 gwei" could be accompanied with the amount of gas unit.

For example, sending 0.001 xDAI appears like this:
small_amount_display

beemeeupnow:
Something I noticed on the activity view:
placement of the 'x' here feels awkward, maybe if it was raise a bit so that the padding above it matched the padding on the right it wouldn't feel that wa
activity_awkward_x

I was able to spot it and figured that it was likely to close the pane, but it didn't have a good feeling. (there was a fleeting second where I wondered if this would do some other action related to hiding this activity from the list)

beemeeupnow:
Performed some exploratory testing with Gnosis chain mainnet, Celo alfajores testnet and Canto.

Not sure where the problem is but here's an odd behavior I found when using a fresh address:
* Add Canto via chainlist
* Change selected network in wallet UI to Ethereum
* Go to https://canto.io
* Connect wallet with top-right button on the page
* Experience a loop where the page repeatedly refreshes, asking you to connect your wallet

Works with another wallet extension.
After testing with other extension and trying Taho again the page no longer loops automatically, but the app does not stay connected to the wallet.

Shadowfiend:
> placement of the 'x' here feels awkward

The transaction from/to should be below the X. Guessing we're in a situation without a scanner link, and the X is being floated right, so without the header link the content below is getting pulled up. Def something we can fix.

Screenshot 2023-05-09 at 11 21 32

jagodarybacka:
> After adding a chain via chainlist the "back" arrow at the top-left of this page is not working

Already reported https://github.com/tahowallet/extension/issues/3365

> could be accompanied with the amount of gas unit

So we have two parts here:
- amount send/swapped is not showing with decimals - what precision should we use?
- we are showing gas price in gwei but user doesn't know the amount of gas units that the transaction will consume

cc @VladUXUI what you think?

> placement of the 'x' here feels awkward

Added issue https://github.com/tahowallet/extension/issues/3373

> odd behavior I found with Canto

Added issue https://github.com/tahowallet/extension/issues/3374

VladUXUI:
> * amount send/swapped is not showing with decimals - what precision should we use?

i think 2 decimals should be enough? The way i wanted this to work from the beginning is that we show 2 decimals after the last 0, so 0.00000012 or 0.0012 or 0.12 and so on. This way we cover a wider range of asset prices. Assets that have higher price are more likely to have more decimals when swapping/trading.

For now, let's go with two decimals, hard to really choose just 1 rule for them all


> * we are showing gas price in gwei but user doesn't know the amount of gas units that the transaction will consume
Gas units is something that nobody understands really, so not a fan of adding it. Also it's a bigger and scarier number. Home come we don't have USD equivalent here?

michalinacienciala:
> For now, let's go with two decimals
By "two decimals" you mean just the two decimals after the `.` or this idea of "we show 2 decimals after the last 0"?

> hard to really choose just 1 rule for them all

Yep, although I think we should do some refactoring of how we handle displaying the amounts at some point, as currently we are pretty inconsistent...
I've ticketed this some time ago: https://github.com/tahowallet/extension/issues/3175.

beemeeupnow:
> >  i think 2 decimals should be enough? The way i wanted this to work from the beginning is that we show 2 decimals after the last 0, so 0.00000012 or 0.0012 or 0.12 and so on. This way we cover a wider range of asset prices. Assets that have higher price are more likely to have more decimals when swapping/trading.

I really like this, especially for scenarios when the value is <0 it makes a lot of sense since small values become more significant and it still allows dynamic granularity.

> >   Gas units is something that nobody understands really, so not a fan of adding it. Also it's a bigger and scarier number. Home come we don't have USD equivalent here?

I mean approximate amount of the gas asset, which would help for cases where we don't have USD equivalent.

Maybe that's another issue I encountered while testing Gnosis chain is not having USD values show up.

VladUXUI:
I don't know how complicated "two decimals after last 0 is". Need to have a discussion around it.
But i'm good to go with only two decimals after "." for now.

>I mean approximate amount of the gas asset, which would help for cases where we don't have USD equivalent. (following same approximation as above)

Yeah, i think this is treating the symptom and not the issue. The issue is that USD values don't show up, that is because of coingeko atm. But we are working on a more general fix using oracle pricing that should land next sprint hopefully

PR #3369, opened 2023-05-08
Impersonate MM on app.0xsplits.xyz

hyphenized:
## To Test 
- [ ] Set wallet as default, then try to connect to https://app.0xsplits.xyz/accounts/


kkosiorowska (approved):
:white_check_mark:

PR #2796, opened 2022-12-16
Fix issue with swapping on Pancakeswap with Ledger

kkosiorowska:
Closes #2767 

This PR allows a swap on Pancakeswap with Ledger when the user has changed the Ledger settings.

## Steps to Recreate
- go to [pancakeswap](https://pancakeswap.finance/swap) to make swap with Ledger account
- try to do a swap when bling signing is enabled on the settings
- disable blind signing on the settings
- try to do a swap again

Latest build: [extension-builds-2796](https://github.com/tallyhowallet/extension/suites/9909132501/artifacts/478991299) (as of Fri, 16 Dec 2022 13:29:41 GMT).

PR #3380, opened 2023-05-12
Failing to Prepare: Add a SigningService hook for pre-notifying signers of a request

### LINKED ISSUE #3031: Investigate possible Ledger failure blocking signing
Related PR: https://github.com/tallyhowallet/extension/pull/2796/files

The fix it implements possibly breaks if a Ledger transport is no longer valid.  We should investigate if this blocks signing with non-Ledger accounts as well as it throws an exception before anything else gets to happening

Shadowfiend:
> “By failing to prepare, you are preparing to fail.”
>
> ― Benjamin Franklin

Some signers can require various types of setup when a new signing request is about to come in. Two quick examples are hardware wallets, which may need to explicitly refresh the connection state to make sure that the wallet is still connected and ready, or remote signers like WalletConnect (currently not supported), which might need to establish a remote connection to be able to make a request.

For the case where a Ledger did not have arbitrary data signing enabled specifically, 8dc160b landed a fix where a full connection refresh would be kicked off, fixing a specific bug. However, this led to an issue where arbitrary data signing information would be checked in certain cases when a Ledger was disconnected, throwing an error early in the signing process and preventing advancing to the signing stage even for non-Ledger signers.

This commit reworks that fix, instead providing a new method on the SigningService, which is responsible for coordinating between signers. This method, `prepareForSigningRequest`, is called immediately prior to passing a signing request (data, typed data, or transaction signing) to the UI, giving the signers an opportunity to do any setup they might need.

The method can be used in blocking fashion in cases where the UI will be presented quickly and the signer should already be prepared when the UI is presented; this is used for transaction signing and data signing. For typed data signing, the data is enriched before presenting the UI, and this will generally take much longer than the current implementation of signer preparation, so the call to `prepareForSigningRequest` is left un-awaited.

Currently, the signing service's implementation simply always asks the Ledger service to refresh its connection information in preparation for the request. Future refinement could pass the signing address and check whether the Ledger service is responsible for that address before refreshing connection information, but the current implementation seems to have pretty low overhead.

Fixes #3031.

## Testing

- [x] Repeat the test at https://github.com/tahowallet/extension/pull/2796#issue-1500202808 to ensure we have not introduced a regression for that fix.
- [x] After the above, disconnect the Ledger and try to swap one more time. The signing screen should come up, reflecting the disconnected state of the Ledger. Without this branch, the signing screen should be locked in a loading state.

Latest build: [extension-builds-3380](https://github.com/tahowallet/extension/suites/12845908941/artifacts/692276466) (as of Fri, 12 May 2023 03:12:50 GMT).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1195,17 +1195,8 @@ export default class Main extends BaseService<never> {
     this.internalEthereumProviderService.emitter.on(
       "transactionSignatureRequest",
       async ({ payload, resolver, rejecter }) => {
-        /**
-         * There is a case in which the user changes the settings on the ledger after connection.
-         * For example, it sets disabled blind signing. Before the transaction signature request
-         * ledger should be connected again to refresh the state. Without reconnection,
-         * the user doesn't receive an error message on how to fix it.
-         */
-        const isArbitraryDataSigningEnabled =
-          await this.ledgerService.isArbitraryDataSigningEnabled()
-        if (!isArbitraryDataSigningEnabled) {
-          this.connectLedger()
-        }
+        await this.signingService.prepareForSigningRequest()
+
         this.store.dispatch(
           clearTransactionState(TransactionConstructionStatus.Pending)
         )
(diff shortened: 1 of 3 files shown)


kkosiorowska (approved):
Great job, I have tested and it looks like everything works well 🚀.

PR #3381, opened 2023-05-12
Ledger-en-dary: Categorize accounts by signer id instead of just keyring id

Shadowfiend:
Previously, the only type of id being taken into account to differentiate between different signers was the keyring id. This meant that two accounts from two different Ledgers appeared under the same entry in the account list, "Ledger 1". Additionally, it meant if you renamed the Ledger category, it was for all Ledgers, not just for one.

The underlying change to fix this is fairly minor: `keyringId` is now generalized to `signerId`, which is derived separately for each account signer type. The derivation is set up such that adding a signer type will require specifying how to differentiate between signers of that type. Any signers that prefer not to differentiate can return `null`, or another value.

NOTE: this change will probably interact with #3089 !

## Testing

The easiest way to test is to add two Ledgers to an existing account. Without this branch, you should see something similar to the below:

image

With this branch, you should see this:

image

Also confirm:
- [x] Both Ledgers are editable separately.
- [x] Editing an entry before this branch remains applicable after switching to this branch, but with that renamed entry associated only with the first Ledger.

Latest build: [extension-builds-3381](https://github.com/tahowallet/extension/suites/12845936179/artifacts/692278164) (as of Fri, 12 May 2023 03:14:58 GMT).

### DIFF
--- a/background/redux-slices/selectors/accountsSelectors.ts
+++ b/background/redux-slices/selectors/accountsSelectors.ts
@@ -45,6 +45,7 @@ import {
   ReadOnlyAccountSigner,
   SignerType,
 } from "../../services/signing"
+import { assertUnreachable } from "../../lib/utils/type-guards"
 
 // TODO What actual precision do we want here? Probably more than 2
 // TODO decimals? Maybe it's configurable?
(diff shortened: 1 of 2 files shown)


hyphenized (commented):
This looks good to me but I'd prefer that we worked these on top of #3089 so we can avoid having to rework that PR based on these changes.

jagodarybacka:
We can also merge this and prepare another PR to https://github.com/tahowallet/extension/pull/3089 even if it won't be merged right now

Shadowfiend:
Yes definitely the latter. We'll get some thoughts from the LA folks on updates to that branch, but halting forward motion in any way that isn't directly related to the signer service there is asking a lot.

mhluongo:
This will seriously improve my life, thank you :pray:

Issue #3341, opened 2023-05-01
In-wallet swap - missing price info on swap UI, yet available on sign tx screen


Naxsun:
### Discord Discussion Link

na

### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

Swap MATIC to FTM on Polygon

### What did not work?

There is no price information & price impact available on the swaps UI screen. However, there is price information on the swaps - tx confirmation screen.

Reproduced this a couple of times.
![image](https://user-images.githubusercontent.com/71384922/235445294-e1683987-08ef-49fc-831f-792fbf4fd593.png)

![image](https://user-images.githubusercontent.com/71384922/235445301-3211864c-73ae-4f27-b7a9-9695bbe56cb6.png)

[logs_v0_31_0__2023-05-01T13_23_24+02_00.txt](https://github.com/tahowallet/extension/files/11365100/logs_v0_31_0__2023-05-01T13_23_24%2B02_00.txt)

### Version

v0.31.0

PR #3383, opened 2023-05-15
Change `tokenToEthRate` threshold for fetching prices

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3341

### What

It is possible to get prices for assets with a low token to ETH rate, so let's change the threshold to 0.1 to get more accurate price info.

Note: this shouldn't cause much more traffic to Coingecko especially when we will merge the price oracle PR

### Testing

- [ ] setup pair mentioned in the issue (MATIC-FTM on Polygon) - there should be a price displayed for Fantom

### DIFF
--- a/background/redux-slices/utils/0x-swap-utils.ts
+++ b/background/redux-slices/utils/0x-swap-utils.ts
@@ -107,7 +107,7 @@ export async function getAssetAmount(
 }
 
 /**
- * If the tokenToEthRate of a is less than 1
+ * If the tokenToEthRate of a is less than 0.1
  * we will probably not get information about the price of the asset.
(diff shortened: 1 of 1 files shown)


kkosiorowska (approved):
LGTM ✅

Shadowfiend:
Will let you merge @kkosiorowska since you reviewed here.

> Note: this shouldn't cause much more traffic to Coingecko especially when we will merge the price oracle PR

Assuming that PR doesn't land right now, how much of a difference are we expecting?

jagodarybacka:
From what I was testing - I can't see a difference using our testing accounts but that may have some effect for accounts that own more illiquid tokens or while user will try to swap with low liquidity assets. Edge case I would say.

PR #3385, opened 2023-05-15
RFB 5: Control-compatible addresses

Shadowfiend:
The implementation section still needs a little more love, but the core of the ideas is there.

Latest build: [extension-builds-3385](https://github.com/tahowallet/extension/suites/13097193642/artifacts/710933010) (as of Tue, 23 May 2023 18:55:12 GMT).

### DIFF
--- a/rfb/rfb-5-control-compatile-addresses.md
+++ b/rfb/rfb-5-control-compatile-addresses.md
@@ -243,7 +243,29 @@ name of the account, because Taho only considered a name to apply to a single
 network. Exploring this problem space led to the development of control
 compatibility as a concept.
 
-// TODO
+For general name resolution, all names should be saved or resolved as on a
+particular network---for example, if a user renames an account while on
+Ethereum mainnet in the UI, the name should be persisted as the name for that
+account on Ethereum. If the ENS resolver on Ethereum mainnet returns an address
+for the name, that address should be considered to be valid on Ethereum
+mainnet. All name resolution requests should thus return an `AddressOnNetwork`,
+as above, that corresponds to the network the address's name was resolved on.
+
+Once that name percolates into the UI, the UI should use control compatibility
+to determine whether to display it in a given context. For example, a keyring
+account's name might be present no matter what network the user is viewing. A
+Ledger account's name, on the other hand, might only display if it's on a
+transaction-compatible network.
+
+Similarly, when entering a recipient address in the Send page, the name should
+only be considered resolved if a resolved name is returned that is
+control-compatible to the requested network. This might be adjusted by giving
+the user warning feedback that the address is known (and allowing the name to
+be used) but not guaranteed to be controlled by the same person.
+
+Note that the generalized problem of displaying accounts that the user controls
+on one network but not another in the Taho UI is not solved, and is a separate
+issue for Taho's design team to tackle.
 
 ## Related Links
 
--- a/rfb/rfb-5-control-compatile-addresses.md
+++ b/rfb/rfb-5-control-compatile-addresses.md
@@ -9,7 +9,7 @@ ongoing development can deprioritize certain thinking that maintains the
 underlying flexibility needed to rerpesent these differences.
 
 Currently, the code base assumes that a given address is always controlled by
-the same underlying key, keyring, or hardware device, even if it is on a
+the same underlying key, keyring, or external device, even if it is on a
 different network. For regular keys, in practice, this is a fair assumption: if
 Bob has a key that can sign for address `0x0abc` with a derivation path of
 `44'/60'/0'/0/0`, the likelihood that Bob will have a different key that can
@@ -18,11 +18,12 @@ low.
 
 This assumption falls apart when it comes to accounts that are not derived or
 controlled directly by in-memory private keys, however. For example, if Bob is
-viewing the Ethereum network and has a key for `0x0abc` at `44'/60'/0'/0/0` can
-sign a transaction for that address, they can interact with everything as
-expected--dApps, etc. If Bob switches to the Rootstock network, on the other
-hand, and they control that key using a Ledger, Ledger itself might require a
-different app to sign a transaction for that key on Roostock.
+viewing the Ethereum network and has a key stored on a Ledger for `0x0abc` at
+`44'/60'/0'/0/0` can sign a transaction for that address, they can interact
+with everything as expected--dApps, etc. If Bob switches to the Rootstock
+network, on the other hand, and they control that key using a Ledger, Ledger
+itself might require a different app to sign a transaction for that key on
+Roostock.
 
 This is the simplest scenario. A more complex scenario appears if the address
 in question is actually a smart contract or account abstraction-based wallet.
@@ -77,7 +78,7 @@ and correspond to the same user on multiple networks or not.
 
 To allow the above, we can introduce two concepts for any given pair of networks:
 
-- Two networks are _transaction-compatible_ for a given hardware wallet if sign
+- Two networks are _transaction-compatible_ for a given external wallet if sign
   a transaction for both networks from the same key material without requiring
   any on-wallet manipulation[^2].
 - An _address_ is considered _control-compatible_ across the pair of networks
@@ -93,8 +94,11 @@ control-compatible across 2 networks:
 - The address was derived from a hardware wallet that controls private key
   material for the address AND the networks are transaction-compatible[^2].
 - The address has been resolved via a name service lookup on both networks and
-  the owner of the records is the same on both networks. [^3] This can change
-  over time, so if the wallet leverages these
+  the owner of the records is the same on both networks.[^3] This can change
+  over time, so if we choose for the wallet to leverage this condition, it
+  should monitor ownership for changes and adjust its understanding of control
+  compatibility based on that information. _Most likely this is not worth
+  implementing at this time._
 
 In addition to these 3 possible paths, the 2 networks must have the same
 address format; e.g., an address cannot be control-compatible between Ethereum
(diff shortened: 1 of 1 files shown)


mhluongo:
Basic idea seems sound. A table with examples to demonstrate *control-compatibility* would be nice.

We might also need to start directly mapping addresses to control mechanisms throughout the UI. Eg knowing that a BTC address and an ETH address are controlled by the same hardware wallet would be useful if we're ever showing cross-chain transactions

Shadowfiend:
> A table with examples to demonstrate control-compatibility would be nice.

Good call. 

> We might also need to start directly mapping addresses to control mechanisms throughout the UI.

We have all the tools to do that (indeed, this is basically what #3381 finishes implementing), but the issues in play on this RFB are broader I think.

Shadowfiend:
Pushed some clarifications + a table, still need to add more details for implementation.

rfb/rfb-5-control-compatile-addresses.md
```diff
@@ -0,0 +1,155 @@
...
+sign for address `0x0abc` but at derivation path `44'/137'/0'/0/0` is quite
+low.
+
+This assumption falls apart when it comes to accounts that are not derived or
+controlled directly by in-memory private keys, however. For example, if Bob is
+viewing the Ethereum network and has a key for `0x0abc` at `44'/60'/0'/0/0` can
+sign a transaction for that address, they can interact with everything as
+expected--dApps, etc. If Bob switches to the Rootstock network, on the other
+hand, and they control that key using a Ledger, Ledger itself might require a
+different app to sign a transaction for that key on Roostock.
```
pdyraga, line 25: This section is confusing to me. 

This is what I believe it says:
- Bob has an in-memory key for `0x0abc`
- Bob also has a key for `0x0abc` on Ledger.

We say that Bob has to switch to another Ledger application to sign with that address on Rootstock and to another Ledger application to sign with that address on Ethereum. This is correct but it would be exactly the same if this account was only on Ledger. Why is the in-memory key important here?

EDIT: Okay, I think I understand - we are not talking about an in-memory key here. Let me suggest a small rewrite to make it more clear: 
```suggestion
This assumption falls apart when it comes to accounts that are not derived or
controlled directly by in-memory private keys, however. For example, if Bob is
viewing the Ethereum network and has a key for `0x0abc` at `44'/60'/0'/0/0`
controlled by Ledger, they can sign transactions and interact with everything as
expected--dApps, etc. If Bob switches to the Rootstock network, on the other
hand, Ledger itself might require a different app to sign a transaction for that
key on Rootstock.
```

rfb/rfb-5-control-compatile-addresses.md
```diff
@@ -0,0 +1,155 @@
...
+These are the cases where the wallet should assume an address is
+control-compatible across 2 networks:
+
+- The wallet controls private key material for the address (e.g. the address is
+  derived from an internal keyring).
+- The address was derived from a hardware wallet that controls private key
+  material for the address AND the networks are transaction-compatible[^2].
+- The address has been resolved via a name service lookup on both networks and
+  the owner of the records is the same on both networks. [^3] This can change
+  over time, so if the wallet leverages these
```
pdyraga, line 97: then what? 😉

rfb/rfb-5-control-compatile-addresses.md
```diff
@@ -0,0 +1,155 @@
...
+base data type should be an `AddressOnNetwork` or `NameOnNetwork`. This should
+be the case both on the way into the UI (e.g. when displaying autocompletes or
+transaction information) AND on the way out of the UI (e.g. when requesting
+signature or otherwise sending information into the services). The network
+information may not always be needed or used, but it should be present so that
+fidelity is not lost.
+
+The intent is to allow the UI to manage addresses as tied one-to-one to
+networks, and to allow flexibility behind the scenes for addresses that might
+be known to be controlled by the same address on multiple networks.
```
pdyraga, line 151: From this RFB:

> (...) In particular, the goal is for an address to have enough information carried with it that the UI and internal services can inform the user **whether an address is known to exist and correspond to the same user on multiple networks or not.**

If we just pass `AddressOnNetwork` or `NameOnNetwork`, is it enough to decide about the control compatibility? Shouldn't the source be passed on as well?

rfb/rfb-5-control-compatile-addresses.md
Shadowfiend: Lolll let me look at that 😆

rfb/rfb-5-control-compatile-addresses.md
Shadowfiend: I think that's unclear wording on my part. Using `*OnNetwork` gives us the base information we need to later determine control compatibility. If we find out about an address but don't know which network it's for, we'll never be able to derive control compatibility information. If we have the address and network, we can check whether Taho controls private keys to it, or has been connected to a hardware wallet that does, etc.

pdyraga: Ack, that makes sense. It's my lack of understanding of the extension architecture and code. I was under the impression there is a need for an API in the extension - in a pure function mode - that receives (address, network, source)` tuples and decides about compatibility. It looks like the goal is to always pass the network along the name and/or address and the wallet can resolve the source and decide.

rfb/rfb-5-control-compatile-addresses.md
```diff
@@ -0,0 +1,272 @@
...
+
+Once that name percolates into the UI, the UI should use control compatibility
+to determine whether to display it in a given context. For example, a keyring
+account's name might be present no matter what network the user is viewing. A
+Ledger account's name, on the other hand, might only display if it's on a
+transaction-compatible network.
+
+Similarly, when entering a recipient address in the Send page, the name should
+only be considered resolved if a resolved name is returned that is
+control-compatible to the requested network. This might be adjusted by giving
```
jagodarybacka, line 262: Does it mean that we will be able to use ENS for Sending on networks other than Ethereum or not? Assuming the account user wants to send funds to is not an imported signer but some other address

rfb/rfb-5-control-compatile-addresses.md
```diff
@@ -0,0 +1,272 @@
...
+ongoing development can deprioritize certain thinking that maintains the
+underlying flexibility needed to rerpesent these differences.
+
+Currently, the code base assumes that a given address is always controlled by
+the same underlying key, keyring, or external device, even if it is on a
+different network. For regular keys, in practice, this is a fair assumption: if
+Bob has a key that can sign for address `0x0abc` with a derivation path of
+`44'/60'/0'/0/0`, the likelihood that Bob will have a different key that can
+sign for address `0x0abc` but at derivation path `44'/137'/0'/0/0` is quite
+low.
```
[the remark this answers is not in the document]
Shadowfiend: The probability I'm noting here is that mnemonic A at a given derivation path will produce the same private key (and thus control the same address) as mnemonic B at a different derivation path. Does that answer your question? I can clarify this section if so.

rfb/rfb-5-control-compatile-addresses.md
Shadowfiend: If it's not an address we control we should either disallow it or make it clear to the user that there is no guarantee they're actually sending to the same person, and the funds could get lost.

rfb/rfb-5-control-compatile-addresses.md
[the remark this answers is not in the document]
jagodarybacka: aaaah ok, that makes sense 👌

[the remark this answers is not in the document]
Shadowfiend: Awesome, I'll reword a little to clarify this.

rfb/rfb-5-control-compatile-addresses.md
```diff
@@ -0,0 +1,272 @@
...
+| New `HDKeyring`        | RSK     | Yes                 | Private key        |
+| New `HDKeyring`        | BTC     | No                  | Address format[^4] |
+| Private key import     | Polygon | Yes                 | Private key        |
+| Private key import     | RSK     | Yes                 | Private key        |
+| Private key import     | BTC     | No                  | Address format     |
+| Ledger walllet         | Polygon | Yes                 | Tx compatible      |
+| Ledger walllet         | RSK     | Yes                 | Tx compatible      |
+| Ledger walllet         | BTC     | No                  | Address format     |
+| WalletConnect Safe[^5] | Polygon | No                  | Tx incompatible    |
+| WalletConnect Safe     | RSK     | No                  | Tx incompatible    |
```
pdyraga, line 122: Nitpicking but I find the Reason for those two cases a bit confusing.

If, for the given mechanism, we have Transaction compatibility than I believe we also have Control compatibility.
Not having Transaction compatibility does not implicate not having Control compatibility. In my opinion and at least in theory.

This is not the current state on Ledger I believe but I can think of a situation when Ledger asks to switch to Arbitrum app (no Transaction compatibility because I had to switch apps) while the same mechanism is used for signing on both networks (has Control compatibility).

rfb/rfb-5-control-compatile-addresses.md
```diff
@@ -0,0 +1,272 @@
...
+name) is control-compatible on the specified `otherNetwork`. To check control
+compatibility, the signing service should check:
+
+- If the two networks have incompatible address formats, return `false`.
+- If there is no known signer for that address, return `false`.
+- If there is a known signer (internal, Ledger, or otherwise), the backing service
+  for that signer should have a new implemented method, `isTransactionCompatible`,
+  defined below; return the result of that method.
+
+Additionally
```
pdyraga, line 202: Yes sir?

pdyraga (approved):
Just some final nitpickin'. Overall, it looks good to me.

rfb/rfb-5-control-compatile-addresses.md
Shadowfiend: For the love of…

rfb/rfb-5-control-compatile-addresses.md
Shadowfiend: It doesn't imply it directly, but if we don't know we have transaction compatibility we don't *know* that we have control compatibility. A contract multisig is the perfect example. If you look at the cited Wintermute attack, the issue was sending to a contract multisig on a network that would have supported the same multisig, but the multisig wasn't deployed yet. The attack was to deploy the multisig in such a way that the intended recipients did not control the multisig anymore, and seize the funds that way. Ultimately if we don't know for a fact that the connected signer can sign for multiple networks, we can't assume that it will for control compatibility purposes, or we expose the user to attacks.

rfb/rfb-5-control-compatile-addresses.md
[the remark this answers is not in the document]
Shadowfiend: Gave it a shot @ https://github.com/tahowallet/extension/pull/3385/commits/9927e7363f4ee15af15a430b726e002a124c1108

Shadowfiend:
Ok, I think this is ready to merge from my end.

pdyraga (approved):

jagodarybacka (approved):
🚀

PR #3387, opened 2023-05-17
Add header  in transaction details view

### LINKED ISSUE #1167: Transaction Details - Add header
Our transaction details need a bit of love .

Currently we don't have title implemented. So once you click on the transaction in activity you no longer know what transaction it is.
Components are Icon + Tx type + Button(icon) for Etherscan.

Should look something like this
image

[Link to Figma](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?node-id=6%3A0)

kkosiorowska:
Closes #1167 

This PR adds a header to the transaction details view and solves the problem of missing header for the custom network. Changes:
-  Separate component for the activity icon
- Special hook to return activity details view
- Minor improvements

In addition, the problem with missing padding when the network does not have NFTs has been fixed, check the screenshots below.

### UI

**Before**
Screenshot 2023-05-17 at 10 12 35

Screenshot 2023-05-17 at 10 40 32

Screenshot 2023-05-17 at 10 42 32

**After**
Screenshot 2023-05-17 at 10 18 34

Screenshot 2023-05-17 at 13 52 32

Screenshot 2023-05-17 at 10 10 39

### To Test
- [x] Check top padding for the activity list when a network does not have NFTs.
- [x] Check the header for custom network. The link should be working properly.
- [x] check the header for default network. The link should be working properly.

Latest build: [extension-builds-3387](https://github.com/tahowallet/extension/suites/12979887290/artifacts/702340713) (as of Thu, 18 May 2023 07:07:45 GMT).

### DIFF
--- a/ui/components/Shared/SharedActivityIcon.tsx
+++ b/ui/components/Shared/SharedActivityIcon.tsx
@@ -1,8 +1,15 @@
 import React, { ReactElement } from "react"
 import classNames from "classnames"
 
+export type ActivityIconType =
+  | "receive"
+  | "send"
+  | "approve"
+  | "swap"
+  | "contract_interaction"
+
 type SharedActivityIconProps = {
-  type: "receive" | "send" | "approve" | "swap" | "contract_interaction"
+  type: ActivityIconType
   size: number
 }
 
--- a/ui/components/Shared/SharedActivityIcon.tsx
+++ b/ui/components/Shared/SharedActivityIcon.tsx
@@ -1,0 +1,48 @@
+import React, { ReactElement } from "react"
+import classNames from "classnames"
+
+type SharedActivityIconProps = {
+  type: "receive" | "send" | "approve" | "swap" | "contract_interaction"
+  size: number
+}
+
+export default function SharedActivityIcon({
+  type,
+  size,
+}: SharedActivityIconProps): ReactElement {
+  return (
+    <>
+      <div className={classNames("activity_icon", type)} />
+      <style jsx>{`
+        .activity_icon {
+          background: url("./images/activity_contract_interaction@2x.png");
+          background-size: cover;
+          width: ${size}px;
+          height: ${size}px;
+          margin-right: 4px;
+          margin-left: 9px;
+        }
+        .receive {
+          background: url("./images/activity_receive@2x.png");
+          background-size: cover;
+        }
+        .send {
+          background: url("./images/activity_send@2x.png");
+          background-size: cover;
+        }
+        .approve {
+          background: url("./images/activity_approve@2x.png");
+          background-size: cover;
+        }
+        .swap {
+          background: url("./images/activity_swap@2x.png");
+          background-size: cover;
+        }
+        .contract_interaction {
+          background: url("./images/activity_contract_interaction@2x.png");
+          background-size: cover;
+        }
+      `}</style>
+    </>
+  )
+}
--- a/ui/components/Shared/SharedActivityIcon.tsx
+++ b/ui/components/Shared/SharedActivityIcon.tsx
@@ -21,7 +21,7 @@ export default function SharedActivityIcon({
           width: ${size}px;
           height: ${size}px;
           margin-right: 4px;
-          margin-left: 9px;
+          margin-left: 4px;
         }
         .asset-transfer-receive {
           background: url("./images/activity_receive@2x.png");
--- a/ui/components/Shared/SharedActivityIcon.tsx
+++ b/ui/components/Shared/SharedActivityIcon.tsx
@@ -1,12 +1,6 @@
 import React, { ReactElement } from "react"
 import classNames from "classnames"
-
-export type ActivityIconType =
-  | "receive"
-  | "send"
-  | "approve"
-  | "swap"
-  | "contract_interaction"
+import { ActivityIconType } from "../../hooks/activity-hooks"
 
 type SharedActivityIconProps = {
   type: ActivityIconType
@@ -29,23 +23,23 @@ export default function SharedActivityIcon({
           margin-right: 4px;
           margin-left: 9px;
         }
-        .receive {
+        .asset-transfer-receive {
           background: url("./images/activity_receive@2x.png");
           background-size: cover;
         }
-        .send {
+        .asset-transfer-send {
           background: url("./images/activity_send@2x.png");
           background-size: cover;
         }
-        .approve {
+        .asset-approval {
           background: url("./images/activity_approve@2x.png");
           background-size: cover;
         }
-        .swap {
+        .asset-swap {
           background: url("./images/activity_swap@2x.png");
           background-size: cover;
         }
-        .contract_interaction {
+        .contract-interaction {
           background: url("./images/activity_contract_interaction@2x.png");
           background-size: cover;
         }
--- a/ui/components/Wallet/WalletActivityDetails.tsx
+++ b/ui/components/Wallet/WalletActivityDetails.tsx
@@ -216,7 +216,7 @@ export default function WalletActivityDetails(
           }
           .header_title {
             margin-left: 4px;
-            margin-right: 14px;
+            margin-right: 8px;
             font-family: "Segment";
             font-style: normal;
             font-weight: 600;
--- a/ui/components/Wallet/WalletActivityDetails.tsx
+++ b/ui/components/Wallet/WalletActivityDetails.tsx
@@ -113,7 +113,7 @@ function DestinationCard(props: DestinationCardProps): ReactElement {
 
 interface WalletActivityDetailsProps {
   activityItem: Activity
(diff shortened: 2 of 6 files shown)


Shadowfiend:
Quick non-blocking note, but it looks like we might not be properly resolving the base asset symbol in the "Amount" field of transaction details (currently says "ETH" when based on the other stuff I see it should probably say "EVMOS").

ui/components/Shared/SharedActivityIcon.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { ReactElement } from "react"
+import classNames from "classnames"
+
+export type ActivityIconType =
+  | "receive"
+  | "send"
+  | "approve"
+  | "swap"
+  | "contract_interaction"
```
Shadowfiend, line 9: I think we should be able to reuse the existing set here from `TransactionAnnotation["type"]`. This should let the type reflect new possibilities if/as we add them. 

Because of how the component is written below, new types would just default to the contract interaction icon (which feels like correct behavior).

ui/pages/Wallet.tsx
```diff
@@ -123,7 +123,9 @@ export default function Wallet(): ReactElement {
           />
           <div
             className={classNames("panel standard_width", {
-              no_padding: panelNumber === 1,
+              no_padding:
+                panelNumber === 1 &&
+                NETWORKS_SUPPORTING_NFTS.has(selectedNetwork.chainID),
```
Shadowfiend, line 128: It feels like it's probably time (after this PR) to move this to [`useSwitchablePanels`](https://github.com/tahowallet/extension/blob/116b51aa7a2ec3062571b100e6b6fde1e2924ac6/ui/hooks/index.ts#L71) and let the NFT container define its structure separately from the others instead of sprinkling these weird conditionals everywhere.

ui/hooks/activity-hooks.ts
```diff
@@ -0,0 +1,74 @@
...
+export default function useActivityViewDetails(
+  activity: Activity,
+  address: string
+): ActivityViewDetails {
+  let details
+  switch (activity.type) {
+    case "asset-transfer":
+      details = {
+        label: isReceiveActivity(activity, address)
+          ? i18n.t("wallet.activities.tokenReceived")
```
Shadowfiend, line 38: Since we're already in a book context, could we `useTranslation` and shorten these keys?

ui/components/Wallet/WalletActivityListItem.tsx
```diff
@@ -54,75 +44,15 @@ export default function WalletActivityListItem(props: Props): ReactElement {
     }
   }, [bottomRef])
 
-  // TODO Replace this with better conditional rendering.
```
Shadowfiend, line 57: More idle/non-blocking thinking to externalize past knowledge/context:

Ironically my TODO here was that it would be better to have components that specialized based on type, similar to how the signing summaries work, rather than having a data structure that's trying to fit data about all render types within it. 

Now that we're sharing info between components I think this approach is marginally better, but I still don't love that the activity details type has to carry fields that don't apply equally to all activity types. It's one of the open problems for me around the whole `ActivityItem` structure.

ui/hooks/activity-hooks.ts
```diff
@@ -0,0 +1,74 @@
...
+    name?: string
+  }
+  assetLogoURL?: string
+  assetSymbol: string
+  assetValue: string
+}
+
+export default function useActivityViewDetails(
+  activity: Activity,
+  address: string
```
Shadowfiend, line 31: Would love a more descriptive name here. Maybe `activityInitiatorAddress`? This makes it clear that it's used as a reference for who initiated the activity. We could use the same variable name in the activity list item.

Shadowfiend (changes requested):
Left a few small nits and some broader context/future considerations. Have not tested yet, but I don't need to be primary reviewer here if someone else wants to take it to merge.

kkosiorowska:
> Quick non-blocking note, but it looks like we might not be properly resolving the base asset symbol in the "Amount" field of transaction details (currently says "ETH" when based on the other stuff I see it should probably say "EVMOS").

I have tested the solution for default and custom networks. But I forgot to upload one screenshot for EVMOS. It seems that everything is good. I have updated the task description. Thanks for noticing!

Screenshot 2023-05-17 at 13 52 32

ui/components/Shared/SharedActivityIcon.tsx
kkosiorowska: Great suggestion. To use `TransactionAnnotation["type"]` I had to slightly modify the type to extract the icons for send and receive activities. Please take another look.

Shadowfiend: Looks awesome!

ui/hooks/activity-hooks.ts
```diff
@@ -0,0 +1,89 @@
...
+      }
+  }
+  return {
+    recipient: activity.recipient,
+    assetLogoURL: activity.assetLogoUrl,
+    assetSymbol: activity.assetSymbol,
+    assetValue: activity.value,
+    ...details,
+  } as ActivityViewDetails
+}
```
Shadowfiend, line 89: The reason we end up needing `as` here is that the `icon` field isn't constant in the `details` objects. I'd prefer to avoid `as` because the compiler wouldn't warn us if e.g. we set an `icon` that wasn't in `ActivityIconType`—by definition `as` skips actually checking that the type is correct. We could fix this by adding `as const` to all the icon assignments (e.g. `icon: "contract-interaction" as const`) and dropping `as ActivityViewDetails`, but I'd like to suggest instead we flip the structure here and set some base details that can then be `return`ed directly in the cases:

```suggestion
export default function useActivityViewDetails(
  activity: Activity,
  activityInitiatorAddress: string
): ActivityViewDetails {
  const { t } = useTranslation("translation", {
    keyPrefix: "wallet.activities",
  })

  const baseDetails = {
    recipient: activity.recipient,
    assetLogoURL: activity.assetLogoUrl,
    assetSymbol: activity.assetSymbol,
    assetValue: activity.value,
  }

  switch (activity.type) {
    case "asset-transfer":
      return {
        ...baseDetails,
        label: isReceiveActivity(activity, activityInitiatorAddress)
          ? t("tokenReceived")
          : t("tokenSent"),
        icon: isReceiveActivity(activity, activityInitiatorAddress)
          ? "asset-transfer-receive"
          : "asset-transfer-send",
      }
    case "asset-approval":
      return {
        ...baseDetails,
        label: t("tokenApproved"),
        icon: "asset-approval",
        assetValue:
          activity.value === INFINITE_VALUE
            ? t("infiniteApproval")
            : activity.value,
      }
    case "asset-swap":
      return {
        ...baseDetails,
        icon: "asset-swap",
        label: t("tokenSwapped"),
      }
    case "contract-deployment":
    case "contract-interaction":
    default:
      return {
        ...baseDetails,
        icon: "contract-interaction",
        label: t("contractInteraction"),
      }
  }
}
```

What do you think?

Shadowfiend:
Tested this, overall everything looks good. Let me know which of the two paths (`as const` vs the `...baseDetails` rewrite) you want to take in that last comment, and then I think we'll be good to merge 💪

ui/components/Wallet/WalletActivityDetails.tsx
```diff
@@ -205,13 +211,18 @@ export default function WalletActivityDetails(
...
-            width: 304px;
-            margin-bottom: 10px;
+            align-items: center;
+            margin: 18px 0;
           }
-          .header_button {
-            margin-top: 10px;
+          .header_title {
+            margin-left: 4px;
+            margin-right: 14px;
```
VladUXUI, line 219: margin-right: 14px; should be 8px

ui/components/Wallet/WalletActivityListItem.tsx
```diff
@@ -206,34 +145,6 @@ export default function WalletActivityListItem(props: Props): ReactElement {
           button:hover {
             background-color: var(--green-80);
           }
-          .activity_icon {
-            background: url("./images/activity_contract_interaction@2x.png");
-            background-size: cover;
-            width: 14px;
-            height: 14px;
-            margin-right: 4px;
-            margin-left: 9px;
```
VladUXUI, line 215: margin-left: 9px; should be 4px

ui/hooks/activity-hooks.ts
kkosiorowska: I like the idea of using `baseDetails`. Let's improve it!

VladUXUI (commented):
Added some nitpicking at pixels :D otherwise looking awesome!

Shadowfiend (approved):
Test drive checks out, Vlad's notes have been addressed, let's rock and roll.

Issue #3370, opened 2023-05-09
Remove extra spacing on txn signing


kkosiorowska:
Remove extra spaces on the txn signing when there is no warning. Make sure that the warning after changing the network settings appears smoothly. Check out the video and this [PR](https://github.com/tahowallet/extension/pull/3337).

https://user-images.githubusercontent.com/23117945/237054020-fdfd3483-f889-4f02-8ca2-17de1fd6ae07.mov

Issue #3342, opened 2023-05-01
Add a green - yellow - red indicator on the sign tx screen for price impact


Naxsun:
### Discord Discussion Link

n/a

### What would you like to be able to do?

Currently there is no price impact displayed on the swap tx screen. In below example the value 'from' to 'to' is quite a bit different. Giving this a red color would make it stand out more.

![image](https://user-images.githubusercontent.com/71384922/235445423-2f619c12-efd2-4f89-b0ec-452c6bcb9883.png)

Issue #3175, opened 2023-03-17
 Unify display of values of txs/assets/fees etc


michalinacienciala:
There is a number of places where we show a rounded amount/value of a given tx/asset/fee etc.
In some of them we don't show any tooltip and in some we show tooltip with the amount that's the same as in the text we're hovering on.
In some places we show the rounded values without any indicator that the value got approximated and in some we use the `~`, `<` or `...` signs.

It would be good to unify our notation. We don't necessarily want to always display the amounts with the same precision - depending on the context we may want to round to two decimal places, or to six or eight. That's fine. But we should have a way of indicating that the shown amount got approximated. And in most places (although probably not all) where we show 0 or 2 decimal digits, it would be good to show more precise amount in a tooltip on hover.

We should think of a convention we want to use and fix all the places where it's not followed.

Some places we should think about:
1. Total account balance
2. Assets balances (on various views)
3. Values of txs in the Asset history or Activity view
4. Details of transactions
5. Values on `Swap Assets` view
6. NFTs values
7. Portfolio values

As part of this work we could also think whether we want to keep the current behavior regarding use of `,` and `.` in the displayed values. For values in USD we use `.` as a decimal separator (for example `$1,051.63`), whereas for other assets we use `,` as decimal separator (for example `23911,783`). This may be a bit confusing, especially if both notations are used on the same view (like on `Swap Assets` view).

VladUXUI:
Can't really have 1 rule for them all. Some place are worth having 100% accuracy, other places not really. This is due to constraint on space and anything after 3 decimals is usually incalculable by humans or it's value is really low. Some asset with bigger price are different tho. For example most people aren't able to tell the difference between 0.325 eth and 0.3259382.

When we use "~" it means we don't really know the exact price
When we use "<" it means that this is the max amount you will pay (meaning again that we don't know the exact price)
I don't remember us using "..." in asset values or prices. But it probably shouldn't be there?

I do like the hover showing the full amount, but only if it was rounded up/down.

VladUXUI:
Do you have an example of a convention being broken that we might want to look into?

michalinacienciala:
One thing that felt "wrong" to me was display of tx amount when small values are being sent.
For example, when I'm sending 0.005 ETH I see:
<img width="367" alt="image" src="https://github.com/tahowallet/extension/assets/78352137/d759b028-d659-4865-a492-561649106190">
And then on the Activity list:
[...]

Issue #2943, opened 2023-01-30
Show decimal token amounts on swap confirm screen


jagodarybacka:
As user is swapping assets we should be showing amounts with more precision instead of rounding them
![Swap assets](https://user-images.githubusercontent.com/20949277/215525757-d97bac3c-5d82-4422-87e6-60a0548a0803.png)

PR #3392, opened 2023-05-18
Swap Transaction screen redesign

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3342
Resolves https://github.com/tahowallet/extension/issues/2943
Resolves https://github.com/tahowallet/extension/issues/3175
Resolves https://github.com/tahowallet/extension/issues/3370
Resolves https://github.com/tahowallet/extension/issues/2804

### What

Redesign and fixes for Swaps.

Scope:
- removed summary screen, `Review swap` button will take user to Sign Transaction screen
- implemented balance change summary component on the Sign Transaction screen
- added route and DAO fee info to transaction details
- code improvements on Swaps page
- removed unused files and merged transactions Details components
- improved price truncate util

### Testing

- [x] test in-wallet Swaps on various networks and with different tokens
- [x] test error`not enough for network fees`

Test for regressions on other signing screens:
- [x] send 
- [x] contract interactions
- [x] token approval
- [x] message signing

image

Latest build: [extension-builds-3392](https://github.com/tahowallet/extension/suites/13140392467/artifacts/714277697) (as of Thu, 25 May 2023 09:13:14 GMT).


background/lib/utils/index.ts
```diff
@@ -41,19 +41,48 @@ export function normalizeAddressOnNetwork({
   }
 }
 
+/**
+ * Manually truncate number, try to cut as close to `decimalLength` as possible
+ * but look for significant digits in the decimal up to `maxDecimalLength`
+ * @param value floating point number as a string or number
+ * @param decimalLength desired length of decimal part
+ * @param maxDecimalLength max length of decimal part - will try to look
+ *                        for sgnificant digits up to this point
```
Shadowfiend, line 50:
```suggestion
 *                         for significant digits up to this point
```

background/lib/utils/index.ts
```diff
@@ -41,19 +41,48 @@ export function normalizeAddressOnNetwork({
   }
 }
 
+/**
+ * Manually truncate number, try to cut as close to `decimalLength` as possible
+ * but look for significant digits in the decimal up to `maxDecimalLength`
```
Shadowfiend, line 46: I know “significant” implies this, but perhaps we should clarify that we will look past `decimalLength` up to `maxDecimalLength` if the number is <1.

background/lib/tests/utils.unit.test.ts
```diff
@@ -0,0 +1,101 @@
+import { truncateDecimalAmount } from "../utils"
+
+type TruncateDecimalAmountType = [
+  number | string,
+  number,
+  number | undefined,
+  string
+]
```
Shadowfiend, line 8: I know it's going to be more verbose, but I strongly recommend reworking these as an object, without necessarily defining a separate type:

```typescript
type TruncateDecimalAmountType = {
  value: number | string
  decimalLength: number
  maxDecimalLength: number | undefined
  expected: string
}
```

You can then unpack this directly in the tests:

```typescript
    it.each(integersCases)(
       "given integer should return integer",
       ({ value, decimalLength, maxDecimalLength, expected }) => {
```

This'll make the test cases themselves a little easier to understand, I think.

background/lib/utils/index.ts
```diff
@@ -41,19 +41,48 @@ export function normalizeAddressOnNetwork({
...
+    return hasSignificantDigits ? `${integer}.${decimalsTruncated}` : integer
   }
-  return valueString
+
+  // integer = 0 so we are trying to get better decimal precision
+  if (decimalLength < maxDecimalLength && !hasSignificantDigits) {
+    return truncateDecimalAmount(value, decimalLength + 1, maxDecimalLength)
+  }
+
+  return hasSignificantDigits ? `${integer}.${decimalsTruncated}` : integer
```
Shadowfiend, line 85: Can we just clamp this using `maxDecimalLength`, as in:

```typescript
  const firstSignificantDecimalDigit =
    // If the integer component is nonzero...
    integer.length > 1 || integer[0] !== "0" ?
      // All decimal digits are significant.
      0 :
      // Otherwise, the first significant digit is the
      // first nonzero decimal digit.
      decimals.findIndex((digit) => digit !== "0")

  const decimalsTruncated = decimals.substring(
    0,
    Math.max(
      // Take the first significant digit, unless the max
      // allowable decimal length is shorter.
      Math.min(firstSignificantDecimalDigit, maxDecimalLength),
      // Take up to the first allowed significant digit,
      // unless the requested minimum decimal length is
      // longer.
      decimalLength
    )
  )

  return `${integer}.${decimalsTruncated}`
```

?

background/lib/tests/utils.unit.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  [0.00123, 1, undefined, "0"],
+  [0.00123, 2, undefined, "0"],
+  [0.00123, 3, undefined, "0.001"],
+  [0.00123, 4, undefined, "0.0012"],
+
+  [0.00123, 0, 5, "0"],
+  [0.00123, 1, 5, "0.001"],
+  [0.00123, 2, 5, "0.001"],
+  [0.00123, 3, 5, "0.001"],
+  [0.00123, 4, 5, "0.0012"],
```
Shadowfiend, line 46: It looks like we support strings for the `value` but never actually test that path. Might be worth adding a few test cases for those.

Also missing: I don't see a version of this that actually checks the max decimal length? Feels like we need to check `[0.0000123, 4, 5, "0.000001"]` or something like that.

Shadowfiend (changes requested):
For future reference, this would have been much easier to review as 3-4 separate PRs 😬 Dropping a quick review of the `truncateDecimalAmount` changes, which I love, and which I think can be split out to a PR by themselves fwiw.

I'll continue to move through the changes today.

background/redux-slices/0x-swap.ts
```diff
@@ -503,3 +512,24 @@ export const selectInProgressApprovalContract = createSelector(
   (state: { swap: SwapState }) => state.swap.inProgressApprovalContract,
   (approvalInProgress) => approvalInProgress
 )
+
+export const selectSwapBuyAssets = createSelector(
+  (state: { assets: AssetsState }) => state.assets,
```
Shadowfiend, line 517: Non-blocking: makes me twitch to pull state from a different slice inside a slice. I would rather we move this to its own selectors file to make it clear it's gluing stuff from multiple slices together (and it can wield the `RootState` type that way).

background/redux-slices/0x-swap.ts
```diff
@@ -479,6 +485,9 @@ export const executeSwap = createBackgroundAsyncThunk(
         type: "asset-swap",
         fromAssetAmount: sellAssetAmount,
         toAssetAmount: buyAssetAmount,
+        estimatedPriceImpact: quote.estimatedPriceImpact
```
Shadowfiend, line 488: I was puzzled as to why we weren't already using this, but it looks like the feature is only a few months old. Nice!

That said, the way this is implemented on the 0x side (https://github.com/0xProject/0x-api/blob/63fd378c3fd198fd1760cd684763956defccfb55/src/services/swap_service.ts#L99 ) looks like it's using fixed-point math, so what do you think about using `parseToFixedPointNumber` to pull a `FixedPointNumber` here?

ui/components/Signing/SignatureDetails/TransactionSignatureDetails/DetailsPanel.tsx
```diff
@@ -1,11 +1,200 @@
...
+          {t("networkFees.estimatedNetworkFee")}
+        </div>
+        <FeeSettingsButton onClick={() => setNetworkSettingsModalOpen(true)} />
+      </span>
+      {isSwapAnnotation(transactionDetails.annotation) && (
+        <SwapAssetDetails
+          transactionRequest={transactionDetails}
+          annotation={transactionDetails.annotation}
+        />
+      )}
```
Shadowfiend, line 147: Not immediately blocking, but let's instead add a `TransactionAdditionalDetails` component here that can work exactly like `TransactionSignatureSummary`, specializing based on `type`.

This will also eliminate the need for `isSwapAnnotation`, and set us up for additional tidbits of extra info for other transaction types (e.g. imagine seeing the current balance and new balance for asset transfers).

ui/_locales/en/messages.json
```diff
@@ -698,7 +698,11 @@
       "noApprovalTarget": "Attempting to approve transfer without an approval target.",
       "nonContractAsset": "Attempting to approve transfer of a non-contract asset."
     },
-    "getFinalQuote": "Get final quote",
+    "reviewSwap": "Review swap",
+    "transaction": {
+      "header": "Swap assets",
+      "balanceChange": "Balance change"
+    },
```
Shadowfiend, line 705: For these, I think we've namespaced stuff under the [`signTransaction` key](https://github.com/tahowallet/extension/blob/2b115ee5849ef51bd0f3563a04d8a4f565e08db2/ui/_locales/en/messages.json#L635-L663).

ui/components/SignTransaction/SignTransactionPanelCombined.tsx
```diff
@@ -1,11 +1,11 @@
 import React, { ReactElement } from "react"
-import SignTransactionDetailPanel from "./SignTransactionDetailPanel"
```
Shadowfiend, line 2: This component can also go away actually, but I'll take care of that hehe.

ui/components/Signing/SignatureDetails/TransactionSignatureDetails/DetailsPanel.tsx
```diff
@@ -1,11 +1,200 @@
-import React, { ReactElement } from "react"
-import { EnrichedEVMTransactionRequest } from "@tallyho/tally-background/services/enrichment"
-import SignTransactionDetailPanel from "../../../SignTransaction/SignTransactionDetailPanel"
+/* eslint-disable react-hooks/exhaustive-deps */
```
Shadowfiend, line 1: Would love to find a way to kill this, btw. Probably not worth futzing with it right now though 🙈

ui/components/Signing/SignatureDetails/TransactionSignatureDetails/TransactionSignatureSummary/SwapAssetSummary.tsx
```diff
@@ -1,220 +1,34 @@
...
-              </div>
-              <div>{source.proportion * 100}%</div>
-            </div>
-          ))}
-        </div>
+        <SwapQuoteBalanceChange
+          fromAsset={fromAssetAmount}
+          toAsset={toAssetAmount}
+          priceImpact={estimatedPriceImpact}
+        />
```
Shadowfiend, line 21: Beautiful 🤩

background/lib/utils/index.ts
jagodarybacka: hmm, it's not passing tests in this form - leaving zeros in the decimal part but I'll take a second look if you think doing recursion here is too much 🙈

background/redux-slices/0x-swap.ts
jagodarybacka: We are not getting that number with a huge precision and right now we are not doing much besides checking if we should display it and what color to use so to me it seems like it is ok to be as simple as possible

background/lib/tests/utils.unit.test.ts
```diff
@@ -1,70 +1,115 @@
...
+  { value: 0.0001, decimalLength: 1, maxDecimalLength: 1, expected: "0" },
+  { value: 0.0001, decimalLength: 1, maxDecimalLength: 2, expected: "0" },
+  { value: 0.0001, decimalLength: 1, maxDecimalLength: 3, expected: "0" },
+  { value: 0.0001, decimalLength: 1, maxDecimalLength: 4, expected: "0.0001" },
+  { value: 0.0001, decimalLength: 1, maxDecimalLength: 5, expected: "0.0001" },
+
+  { value: "0.00123", decimalLength: 0, expected: "0" },
+  { value: "0.00123", decimalLength: 3, expected: "0.001" },
+  { value: "0.00123", decimalLength: 4, expected: "0.0012" },
+  { value: "0.00123", decimalLength: 0, maxDecimalLength: 5, expected: "0" },
```
Shadowfiend, line 75: I'm surprised by this one. Shouldn't it yield `0.00123`?

background/lib/utils/index.ts
Shadowfiend: This was laziness on my part haha. The recursion isn't the only part, I just think we're ending up with too many conditionals here. This version passes tests and makes it a little clearer where we're making decisions based on max vs decimalLength vs significant digit I think:

```typescript
  const firstSignificantDecimalDigit =
    [...decimals].findIndex((digit) => digit !== "0") + 1

  const decimalTruncationLength =
    integer.length > 1 || integer[0] !== "0"
      ? // For a value >=1, always respect decimalLength.
        decimalLength
      : // For a value <1, use the greater of decimalLength or first
        // significant decimal digit, up to maxDecimalLength.
        Math.min(
          Math.max(decimalLength, firstSignificantDecimalDigit),
          maxDecimalLength
        )

  // If the truncation point includes no significant decimals, don't include
  // the decimal component at all.
  if (decimalTruncationLength < firstSignificantDecimalDigit) {
    return integer
  }

  const decimalsTruncated = decimals.substring(0, decimalTruncationLength)

  return `${integer}.${decimalsTruncated}`
```

Ultimately I'll leave it up to you if that feels clearer—feel free to pull it in or not!

background/lib/tests/utils.unit.test.ts
Shadowfiend: Answering my own question: no, but that's because we return early right now. I think that's a mistake:

```typescript
  if (decimalLength === 0) {
    return integer
  }
```

I think this should be `if (decimalLength === 0 && maxDecimalLength === 0)`.

ui/components/Signing/SignatureDetails/TransactionSignatureDetails/TransactionSignatureSummary/SwapAssetDetails.tsx
```diff
@@ -8,19 +8,21 @@ import { TransactionSignatureSummaryProps } from "./TransactionSignatureSummaryP
 export default function SwapAssetDetails({
   annotation: { sources },
 }: TransactionSignatureSummaryProps<AssetSwap>): ReactElement {
-  const { t } = useTranslation("translation", { keyPrefix: "swap" })
+  const { t } = useTranslation()
```
Shadowfiend, line 11: Probably can use `signTransaction.swap` as the `keyPrefix` here?

background/lib/utils/index.ts
```diff
@@ -41,19 +41,49 @@ export function normalizeAddressOnNetwork({
...
+ * @param decimalLength desired length of decimal part
+ * @param maxDecimalLength max length of decimal part - will try to look
+ *                        for significant digits up to this point
+ * @returns truncated number
+ */
 export function truncateDecimalAmount(
   value: number | string,
-  decimalLength: number
+  decimalLength: number,
+  maxDecimalLength = decimalLength
```
Shadowfiend, line 57: Incidentally we should probably add this parameter to the `truncateDecimalAmount` call in `SignTransactionTransferInfoProvider`!

ui/components/Swap/SwapQuoteBalanceChange.tsx
```diff
@@ -0,0 +1,165 @@
...
+        .balance_token {
+          display: flex;
+          align-items: center;
+        }
+        .sell {
+          color: var(--trading-sell);
+        }
+        .buy {
+          color: var(--trading-buy);
+        }
```
Shadowfiend, line 144: Do we know whether the sell and buy colors here are other common colors in the design system, or if calling them trading-sell and trading-buy is the most general way to reference them?

ui/pages/Swap.tsx
```diff
@@ -126,58 +123,24 @@ export default function Swap(): ReactElement {
...
-      }
-      if (
-        // Explicitly add a network's base asset.
-        isBuiltInNetworkBaseAsset(asset, currentNetwork)
-      ) {
-        return true
-      }
-      return false
-    })
+    const assets = selectSwapBuyAssets(state) as SwappableAsset[]
```
Shadowfiend, line 136: Phew, I feel like it is time to make `selectSwapBuyAssets` return actual `SwappableAsset`s:

```typescript
export const selectSwapBuyAssets = createSelector(
  (state: RootState) => state.assets,
  selectCurrentNetwork,
  (assets, currentNetwork) => {
    return assets.filter(
      (
        asset
      ): asset is SwappableAsset & {
        recentPrices: SingleAssetState["recentPrices"]
      } => {
...
```

Shadowfiend (commented):
Still not quite done, but almost there. Left a couple more notes—I like how much removing the final quote has removed from our swap page state machine 😅

ui/components/Swap/SwapQuoteBalanceChange.tsx
jagodarybacka: image
Green in Figma is called `trading/incoming", pink is not named at all, lmk if you have better name ideas 😉

background/lib/tests/utils.unit.test.ts
jagodarybacka: Hmm, I wrote that `if` that is early returning an integer when the `decimalLength = 0`, because that was in that function before these changes 🤔 
I think it is not the worst thing to do - if you want to get an integer then you don't care about the decimal part. If you care about the decimal part then you should at least set `decimalLength = 1` - because that util will cut decimal part if it will contain only zeros so you will get a nice integer anyway. What do you think about it?

jagodarybacka:
Alright so there is still a problem that goes like this - sometimes we are able to show $ price on the swap screen but on the transaction screen we are losing it. 

It happens because swaps screen is using [checkCurrencyAmount](https://github.com/tahowallet/extension/blob/swap-tx-redesign/background/redux-slices/0x-swap.ts#L360) that will fetch the prices if there are no recent prices saved on the asset. Transaction screen is using [saved recent prices](https://github.com/tahowallet/extension/blob/swap-tx-redesign/ui/components/Swap/SwapQuoteBalanceChange.tsx#L50) which may be empty for some assets.

I think we should ensure that if prices are fetched then they are saved/cached so we can avoid problems like that and just use one method to get $ price where needed instead of having multiple ways to do the same thing which is causing bugs and problems.

But I think it is not blocking to this PR so I'll undraft it and I can continue fixing that in the next PR.

background/lib/utils/index.ts
jagodarybacka: Refactored + removed that exception for `decimalLength=0` case

background/lib/tests/utils.unit.test.ts
Shadowfiend: Appreciate it. Surprisingly complicated for a function that feels so straightforward conceptually!

background/lib/utils/index.ts
Shadowfiend: … Lol. I guess I'm cleaning that up now before I embarrass myself again 😆

ui/components/Swap/SwapQuoteBalanceChange.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const { fromAsset, toAsset, priceImpact } = props
+  const { t } = useTranslation()
+  const assets = useBackgroundSelector(getAssetsState)
+  const mainCurrencySymbol = useBackgroundSelector(selectMainCurrencySymbol)
+  const mainCurrencySign = useBackgroundSelector(selectMainCurrencySign)
+
+  const rawFromAmount = formatUnits(fromAsset.amount, fromAsset.asset.decimals)
+  const fromAmount = truncateDecimalAmount(rawFromAmount, 2, 8)
+  const rawToAmount = formatUnits(toAsset.amount, toAsset.asset.decimals)
+  const toAmount = truncateDecimalAmount(rawToAmount, 2, 8)
```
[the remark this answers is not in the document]
Shadowfiend: I think we do need to continue to use `toLocaleString`—in many ways I hate that we've got `truncateDecimalAmount` at all, because it makes assumptions about number format that aren't localizable.

We do have the tools to unify all of it by taking the number and then passing it to `toLocaleString`, or using [`formatToParts`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts) and dealing with the `fraction` component. That said, I don't think it's a pressing concern at the moment.

ui/components/Swap/SwapQuoteBalanceChange.tsx
Shadowfiend: What about using `-incoming` and `-outgoing`? It looks like there's Trading/Outgoing style but it's a slightly different color, we can check with Vlad on that.

Shadowfiend:
> I think we should ensure that if prices are fetched then they are saved/cached so we can avoid problems like that and just use one method to get $ price where needed instead of having multiple ways to do the same thing which is causing bugs and problems.
>
> But I think it is not blocking to this PR so I'll undraft it and I can continue fixing that in the next PR.

Agree across the board, love the thinking. One way to force-fetch prices that also updates the cache ✅ and pushing to a separate PR ✅

ui/components/Swap/SwapQuoteBalanceChange.tsx
Shadowfiend: Checked with Vlad here—indeed, `-outgoing` is meant to be the pink we're using here, just hasn't made it into the design system file. Let's use those two names unless we have a strong reason to diverge!

background/lib/utils/index.ts
Shadowfiend: The new flow uses `localizedDecimalAmount`—which is unfortunately subject to the problem of not always showing the amount of precision we might prefer.

Bleh. Can't make myself go down that rabbit hole right now 😅

Anyway, I said I'd clean it up and I'm cleaning it up (#3404 😝).

ui/utils/swap.ts
```diff
@@ -283,3 +284,46 @@ export function useSwapQuote(useSwapConfig: {
     requestQuoteUpdate: debouncedRequest,
   }
 }
+
+export function getOwnedSellAssetAmounts(
+  assetAmounts: CompleteAssetAmount[] | undefined,
+  currentNetwork: EVMNetwork
+): CompleteAssetAmount<SwappableAsset>[] {
+  // TODO We're special-casing ETH here in an odd way. Going forward, we should
+  // filter by current chain and better handle network-native base assets
```
Shadowfiend, line 293: I think this TODO no longer applies? ETH isn't special-cased, though we're letting through assets that are not known to be on this chain.

Shadowfiend (commented):
I believe most of my concerns are cleared here, so going to move into testing tomorrow.

Shadowfiend (approved):
This looks awesome 🤩 Approving, but going to give @VladUXUI a chance to do some design QA before landing this. Let's rock and roll 🎸

VladUXUI:
This is working and looking awesome! No changes on my part!

Shadowfiend:
Lfg!

PR #3394, opened 2023-05-19
Verified/Unverified tokens 

### LINKED ISSUE #3213: Separate untrusted tokens option from “Hide asset balances under $2”

kkosiorowska:
Closes #3213 

Currently, the user sees a lot of unverified assets. The user should be able to hide them in the settings options. Unverified assets are a challenge to security. The user should be able to verify the asset and hide it if they have done it by mistake. This PR adds options to handle verified/unverified tokens.

## What

Tasks to be completed in the next PRs: 
- https://github.com/tahowallet/extension/issues/3399
- https://github.com/tahowallet/extension/issues/3400
- https://github.com/tahowallet/extension/issues/3401

[Figma](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?type=design&node-id=19229-386562&t=RniE6V5sKkBkEcUr-0)

### Changes
- Adding a toggle to the settings for untrusted assets (Carrying out the migration for new state)
- Update hidden assets for untrusted assets list (Only unverified assets should appear in the list of assets)
- Adding a banner when a user opens the list of unverified assets (When the user closes it, it should not be displayed again in the list)
- Adding tests to the display logic for trusted/untrusted tokens

### Additionally
- Fixing the problem of too-long token names
- Change of terminology from `scanWebsite` to `blockExplorer`

## UI

https://github.com/tahowallet/extension/assets/23117945/0632dc03-1101-4907-8d52-afe0f322c84a

## To Test

- [x] Check that the migration has been completed correctly without errors
- [x] See how the application behaves when the feature flag is disabled

Toggler for unverified tokens on settings page

- [x] By default, the toggler should be disabled. The list of unverified assets should not be visible. 
- [x] Enable display of unverified tokens. The list of unverified assets should appear on the wallet page.
- [x] The list should be hidden when the account does not have unverified tokens in the selected network. (Even if the toggler is enabled)
- [x] Disable display of unverified tokens. The list should disappear.

Token verification

- [x] Verify the asset and add them to the regular list.
- [x] Open a single page for the asset. Check that the correct untrusted token status messages appear.

## Testing Env

```
SUPPORT_UNVERIFIED_ASSET=true
```

Latest build: [extension-builds-3394](https://github.com/tahowallet/extension/suites/13153872262/artifacts/715278429) (as of Thu, 25 May 2023 18:12:31 GMT).


background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -152,35 +172,31 @@ const computeCombinedAssetAmountsData = (
...
+          acc.combinedAssetAmounts.push(assetAmount)
+        } else if (isUntrustedAsset(assetAmount.asset)) {
+          // FIXME: Assets with very low values should still be displayed as dust e.g. "<0,01"
+          const isPresent = assetAmount.decimalAmount > 0
+
+          // Don't display hidden assets whose calculated decimal amount is zero
+          if (!isPresent) return acc
+
+          acc.untrustedAssetAmounts.push(assetAmount)
+        }
```
Shadowfiend, line 195: We're trying to both categorize and filter here, and we're duplicating work that is mostly related in a way that's not going to be easy to maintain down the line, I think. We've also ended up bifurcating our logic for when to display untrusted assets from the logic for when to display trusted ones, which is going to lead to some weird behaviors.

My suggestion is we:
- Rename `isAssetAmountVisible` to `determineAssetDisplayAndTrust(...): { displayAsset: boolean; trustedAsset: boolean }` (using `Display` here so we use that term across decisions around whether to display an asset or not).
- Call it here as `const { displayAsset, trustedAsset } = determineAssetDisplayAndTrust(assetAmount, { hideDust, showUntrusted })`
- Use `displayAsset` to decide whether to add the asset and `trustedAsset` to decide which list to add it to.

Shadowfiend (commented):
Left some initial comments here, I think *mostly* from existing work. We'll have to discuss what should vs shouldn't be blocking merge.

background/main.ts
```diff
@@ -50,9 +50,14 @@ import {
   loadAccount,
   updateAccountBalance,
   updateAccountName,
+  updateAssetCache,
```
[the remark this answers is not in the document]
kkosiorowska: Let's rename it.

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -73,31 +73,37 @@ const shouldForciblyDisplayAsset = (
...
       : assetAmount.mainCurrencyAmount > userValueDustThreshold
   const isPresent = assetAmount.decimalAmount > 0
-  const isTrusted = !isUntrustedAsset(assetAmount.asset)
 
-  return !!(
-    isForciblyDisplayed ||
-    (isTrusted && (hideDust ? isNotDust && isPresent : isPresent))
-  )
+  const trustedToBeVisible = showUntrusted || isTrusted
+  const enoughBalanceToBeVisible = isPresent && (isNotDust || !hideDust)
```
Shadowfiend, line 93: Can't handle it… I think we should introduce a `const showDust = !hideDust` and use it here. Every time I run into `|| !hideDust` my brain flips 😅

background/redux-slices/selectors/accountsSelectors.ts
```diff
@@ -229,14 +232,16 @@ export const selectAccountAndTimestampedActivities = createSelector(
   getAccountState,
   getAssetsState,
   selectHideDust,
+  selectShowUntrustedAssets,
```
Shadowfiend, line 235: Looking at this, I think we should probably name `showUntrusted` to `showUntrustedAssets` throughout the code? It's def clearer.

background/redux-slices/tests/accounts.unit.test.ts
```diff
@@ -279,148 +279,196 @@ describe("Accounts redux slice", () => {
...
             mainCurrencyAmount: 0,
           }),
           {
             hideDust: true,
+            showUntrusted: false,
           }
         )
-      ).toBeTruthy()
+
+        expect(displayAsset).toBeFalsy()
```
Shadowfiend, line 355: I would love to change all of the `toBeTruthy` and `toBeFalsy` to `toBe(true)` and `toBe(false)`. Our API is very specifically a boolean one, not one that can return undefined or null.

ui/pages/SingleAsset.tsx
```diff
@@ -91,6 +91,7 @@ export default function SingleAsset(): ReactElement {
       localizedDecimalAmount: undefined,
     }
 
+  const isTokenListAsset = !!asset?.metadata?.tokenLists?.length
```
Shadowfiend, line 94: Shouldn't `isUntrusted` be taking this into account?

ui/utils/networks.ts
```diff
@@ -113,3 +117,9 @@ export const getNetworkIcon = (network: EVMNetwork): string => {
 
   return FALLBACK_ICONS_BY_CHAINID[network.chainID] ?? ""
 }
+
+export const getScanWebsiteUrl = (network: EVMNetwork): string | undefined => {
```
Shadowfiend, line 121: We should unify terminology here by making `scanWebsite` be `blockExplorer`.

ui/_locales/en/messages.json
```diff
@@ -576,6 +580,12 @@
         "description": "You can also add a custom RPC, this can be a bit risky as you might add a malicious RPC without knowing, that could result in loss of funds.",
         "addBtn": "Add custom RPC (Advanced)"
       }
+    },
+    "unverifiedAssets": {
+      "tooltip": {
+        "firstPart": "Discover assets that you own but are not on our asset list. Some spam/scam assets will show up, so thread carefully.",
+        "secondPart": "These will show up in the bottom of asset page until you verify them."
```
Shadowfiend, line 587:
```suggestion
        "secondPart": "These will show up on the bottom of the asset page until you verify them."
```

ui/_locales/en/messages.json
```diff
@@ -819,6 +815,24 @@
...
       "description": "They help us improve the wallet.  You can disable anytime",
       "settingsLink": "Change settings"
+    },
+    "unverifiedAssets": {
+      "stateOfHiddenAssets1": "Hide",
+      "stateOfHiddenAssets2": "See",
+      "hiddenAssets": "{{stateOfHiddenAssets}} unverified assets ({{amount}})",
+      "verifyAsset": "Verify asset"
+    },
+    "trustedAssets": {
```
Shadowfiend, line 825: Should probably be `verifiedAssets`. Let's chat on Discord though and see here—there's a world where we need to lean in and talk about “verified”/“unverified” instead of trusted/untrusted across the codebase 🙃

ui/components/Shared/SharedBanner.tsx
```diff
@@ -12,7 +12,7 @@ type BannerProps = {
   customStyles?: string
 }
 
-type CanBeClosedProps =
+export type CanBeClosedProps =
```
Shadowfiend, line 15: Non-blocking, but I feel like this should be a single prop, `canBeClosedWithId?: string | undefined`. If the id exists and is undefined, it can be closed. If it doesn't or it's undefined, it cannot be closed. We should only split them apart if the id may exist separately from the `canBeClosed` flag's state. Otherwise we have to track a separate type, mix it in, export it, and do weird prop spreading—just doesn't feel like a good fit for a component prop.

ui/pages/SingleAsset.tsx
kkosiorowska: Yes `isUntrusted` takes this into account. But we need to check this value when the asset has been verified by the user.  I will move this to a separate function and add a comment to make it clear.

Shadowfiend:
I know we discussed how to handle the removal flow—rather than stacking that on top of this already-large PR, let's:

- Only execute the unverified asset click handler from `CommonAssetListItem` if the unverified list item feature flag is on.
- Only display the unverified asset stuff on the single asset page if the feature flag is on.
- Work to merge this PR without too much additional work (i.e., we can probably punt a lot of review work that isn't about the direct functionality to a follow-up PR).

Once this PR is merged, we can do the removal flow in a separate PR, and work on toggling the feature flag.

Shadowfiend (commented):
Ok, I left a few more thoughts, but I've entered the phase here where we have enough changes that I'm not going to do much more detailed review on anything that isn't new commits. Some things we could do better will get through, but that's fine.

I'm going to move to testing on Thursday morning unless you tell me not to 😁

background/assets.ts
```diff
@@ -44,6 +44,10 @@ export type AssetMetadata = {
    * legitimate asset.
    */
   trusted?: boolean
+  /**
+   * Set by the user to hide assets that have been verified by mistake.
+   */
+  hidden?: boolean
```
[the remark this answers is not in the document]
kkosiorowska: Let's separate this from the main PR. I will remove this part and prepare a separate PR for asset removal.

background/redux-slices/selectors/accountsSelectors.ts
kkosiorowska: Done, `const showDust = !hideDust` was introduced.

background/redux-slices/tests/accounts.unit.test.ts
kkosiorowska: Done.

ui/_locales/en/messages.json
```diff
@@ -576,6 +580,12 @@
         "description": "You can also add a custom RPC, this can be a bit risky as you might add a malicious RPC without knowing, that could result in loss of funds.",
         "addBtn": "Add custom RPC (Advanced)"
       }
+    },
+    "unverifiedAssets": {
+      "tooltip": {
+        "firstPart": "Discover assets that you own but are not on our asset list. Some spam/scam assets will show up, so thread carefully.",
```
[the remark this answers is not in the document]
kkosiorowska: Done.

background/services/indexing/index.ts
```diff
@@ -334,10 +335,10 @@ export default class IndexingService extends BaseService<Events> {
             annotationAddressesOnNetwork
           )
 
-        // An asset has baseline trust if we are already tracking the asset
-        // (e.g. via a previously baseline-trusted interaction or via a token
+        // An asset has baseline verify if we are already tracking the asset
+        // (e.g. via a previously baseline-verified interaction or via a token
         // list) OR the sender is a tracked address.
-        const baselineTrustedAsset =
+        const baselineVerifiedAsset =
```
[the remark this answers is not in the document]
kkosiorowska: Done.

background/main.ts
```diff
@@ -1915,7 +1916,7 @@ export default class Main extends BaseService<never> {
     addressNetwork: AddressOnNetwork
   }): Promise<void> {
     const { metadata = {} } = asset
-    // Manually imported tokens are trusted
+    // Manually imported tokens are verified
```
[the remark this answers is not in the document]
kkosiorowska: Changes: 

- The name was changed from `importAccountCustomToken` to `importCustomToken`.
- The logic for marking assets as verified has been pushed to the `IndexingService`.
- Unnecessary call to `addCustomAsset` removed.

At this moment, once the token has been verified, we update this status in the db.

Screenshot 2023-05-24 at 09 50 15

background/redux-slices/selectors/accountsSelectors.ts
kkosiorowska: Done.

background/services/indexing/index.ts
```diff
@@ -416,12 +411,16 @@ export default class IndexingService extends BaseService<Events> {
...
       async ({ addressNetwork, assetTransfers }) => {
         assetTransfers.forEach((transfer) => {
-          const fungibleAsset = transfer.assetAmount
-            .asset as SmartContractFungibleAsset
-          if (fungibleAsset.contractAddress && fungibleAsset.decimals) {
+          const fungibleAsset = transfer.assetAmount.asset
+          if (
+            isSmartContractFungibleAsset(fungibleAsset) &&
+            fungibleAsset.contractAddress &&
+            fungibleAsset.decimals
```
Shadowfiend, line 418: I think `isSmartContractFungibleAsset` already does this check, no?

background/services/indexing/index.ts
```diff
@@ -573,19 +572,22 @@ export default class IndexingService extends BaseService<Events> {
...
-    isTrusted: boolean
+    metadata: AnyAssetMetadata
   ): Promise<void> {
-    await this.db.updateAssetMetadata(asset, { trusted: isTrusted })
+    await this.db.updateAssetMetadata(asset, metadata)
     await this.cacheAssetsForNetwork(asset.homeNetwork)
   }
 
-  async importAccountCustomToken(
+  async importCustomToken(
```
Shadowfiend, line 583: A TODO for follow up work: to complete this rename we need to drop the `AddressOnNetwork` parameter and make the balance retrieval work across addresses (though only on the asset's network), as adding the asset is not an address-specific action.

ui/components/Wallet/AssetListItem/styles.ts
```diff
@@ -23,11 +23,13 @@ export default `
     justify-content: center;
     margin-left: 16px;
   }
+  .max_width {
```
Shadowfiend, line 26: Is there a reason we don't want that max-width on all items? If flexbox handles it on the verified assets, maybe we can see if there's a way to tweak the flexbox here rather than setting a pixel max.

Shadowfiend (commented):
Moar comments! Almost entirely for follow up PRs I think.

background/assets.ts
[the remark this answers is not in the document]
kkosiorowska: Let's do it in https://github.com/tahowallet/extension/issues/3399

background/main.ts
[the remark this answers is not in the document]
kkosiorowska: Let's fix it in https://github.com/tahowallet/extension/issues/3401

background/services/indexing/index.ts
kkosiorowska: Let's fix it in https://github.com/tahowallet/extension/issues/3401

ui/components/Shared/SharedBanner.tsx
kkosiorowska: Let's fix it in https://github.com/tahowallet/extension/issues/3403

background/redux-slices/utils/asset-utils.ts
```diff
@@ -332,23 +332,37 @@ export function heuristicDesiredDecimalsForUnitPrice(
   )
 }
 
+/**
+ * Check if the asset has a list of tokens.
+ * Assets that do not have it are considered unverified.
+ *
+ */
+export function isUnverifiedAsset(asset: AnyAsset | undefined): boolean {
```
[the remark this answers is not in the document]
kkosiorowska: Done.

ui/components/Wallet/UnverifiedAsset/AssetWarningSlideUp.tsx
```diff
@@ -76,73 +77,85 @@ export default function AssetWarningSlideUp(
...
+                        handleUpdateAssetMetadata({ trusted: true })
+                      }
+                    >
+                      {t("addToAssetList")}
+                    </SharedButton>
+                  )}
+                </>
+              ) : (
+                <SharedButton size="medium" type="secondary" onClick={close}>
+                  {t("close")}
```
Shadowfiend, line 158: Should we pull this from the global translation rather than having a namespaced one? e.g. as we do @ https://github.com/tahowallet/extension/blob/b9ce174531155d523581ee740f8c6ebf8d023eba/ui/components/Signing/Signer/SignerLedger/SignerLedgerFrame.tsx#L29 .

ui/pages/Settings.tsx
```diff
@@ -320,7 +320,7 @@ export default function Settings(): ReactElement {
   const generalList = [
     setAsDefault,
     hideSmallAssetBalance,
-    isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET) && unverifiedAssets,
+    unverifiedAssets,
```
Shadowfiend, line 323: If we're leaving asset removal for a future PR, we should probably still block this on the feature flag, right?

ui/pages/SingleAsset.tsx
```diff
@@ -133,7 +144,9 @@ export default function SingleAsset(): ReactElement {
                 logoURL={asset?.metadata?.logoURL}
                 symbol={asset?.symbol}
               />
-              <span className="asset_name ellipsis">{symbol}</span>
+              <span className="asset_name">
+                {trimWithEllipsis(symbol, MAX_SYMBOL_LENGTH)}
```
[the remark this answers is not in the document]
Shadowfiend: Realized this might have been my question about max-width—if so, I 100% support using the CSS approach to truncate here, I was just wondering if we could leverage flexbox to handle the max width of the container (e.g. by making the left content lower priority than the right content—I don't remember off the top of my head what flexbox calls that) vs setting an explicit pixel `max-width`.

Either way, this is pretty minor and doesn't need to burn much more time right now.

Shadowfiend (commented):
Closed out most of the comments, intending to move more heavily into testing tomorrow!

ui/pages/Settings.tsx
kkosiorowska: The options for adding and deleting an asset are locked under the feature flag. Currently, unverified assets are hidden under the dust toggle so it seems to me that this part should not be hidden. Let's discuss this on discord.

background/main.ts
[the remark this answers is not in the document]
kkosiorowska: The property name changed from `trusted` to `verified`.

ui/pages/SingleAsset.tsx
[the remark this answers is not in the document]
kkosiorowska: We are currently using the trimWithEllipsis function for this in some of these cases.  That is why I chose to use it again and solve this issue. Let's solve this problem in this task https://github.com/tahowallet/extension/issues/3405

background/main.ts
```diff
@@ -1064,6 +1069,13 @@ export default class Main extends BaseService<never> {
 
     this.indexingService.emitter.on("assets", async (assets) => {
       await this.store.dispatch(assetsLoaded(assets))
+
+      if (isSmartContractFungibleAsset(assets[0])) {
+        await this.store.dispatch(
+          updateAssetMetadata([assets[0], assets[0].metadata ?? {}])
```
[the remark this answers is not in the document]
kkosiorowska: I have updated the logic and moved it to `assetsLoaded`. Take a look at it again please.

VladUXUI:
Overall looking and working good!
Here are some smaller notes:

- [ ] This button needs a mouse over to our white

image

- [ ]  Needs to change copy for the modal of an asset that is verified
Asset verified by you
Be aware of scam and spam assets, only interact with assets you trust.

- [ ] Let's not remember state of expanded menu.
This automatically stays open if i switch accounts, and that is not really a good thing.
It's also ok to close everytime a user opens up a modal to verify

ui/components/Wallet/UnverifiedAsset/AssetWarningSlideUp.tsx
kkosiorowska: Done.

kkosiorowska:
> * [ ]  This button needs a mouse over to our white

Screenshot 2023-05-25 at 20 11 23


> * [ ]   Needs to change copy for the modal of an asset that is verified
>   Asset verified by you
>   Be aware of scam and spam assets, only interact with assets you trust.

Screenshot 2023-05-25 at 20 12 23

Screenshot 2023-05-25 at 20 12 48


> * [ ]  Let's not remember state of expanded menu.
>   This automatically stays open if i switch accounts, and that is not really a good thing.
>   It's also ok to close everytime a user opens up a modal to verify


https://github.com/tahowallet/extension/assets/23117945/b2dcb05f-47dd-4643-89ef-4bbda96f5489

background/redux-slices/assets.ts
```diff
@@ -88,6 +94,7 @@ const assetsSlice = createSlice({
           if (duplicates.length === 0) {
             mappedAssets[newAsset.symbol].push({
               ...newAsset,
+              metadata: newAsset.metadata ?? {},
```
Shadowfiend, line 97: Do we need this in this slice? Metadata will always be included in `...newAsset`, and the only remaining component is making sure it's present, but I don't think we strictly need to do that here.

ui/components/Wallet/UnverifiedAsset/AssetVerifyToggler.tsx
```diff
@@ -47,7 +47,7 @@ export default function AssetVerifyToggler({
       <style global jsx>
         {`
           .asset_verify_toggler:hover .icon {
-            color: ${hoverColor};
+            background-color: ${hoverColor};
```
Shadowfiend, line 50: Do we need this even though we have the `hoverColor` specified above?

Shadowfiend (approved):
I think we're good to move on this. A couple of additional observations:
- `updateAssetMetadata` thunk seems like it should only be calling into `main`. `main` should call into `IndexingService` to update the metadata, and the `IndexingService` should re-emit `assets` with the updated metadata as a result, which will then percolate to the UI. The idea is always that if the UI needs to know about a change, chances are another service will also want to know about it. Making this change will remove the need for the `updateMetadata` reducer entirely.
- We should also remove the need for the `IndexingService` to emit `updateAssetReferences`. `IndexingService` shouldn't need to know about asset references, as it's a UI-specific detail—instead, we should probably emit this as part of the `assets` event handler.


I noted to @VladUXUI that the collapsing of the unverified asset list when bringing up the Verify Asset modal is jarring, so we'll hear more from him there.

Since we're behind a feature flag here, I'm going to go ahead and merge so we can keep moving!

VladUXUI:
@Shadowfiend is correct here, it is very jarring, but that might be because we close the expanded menu after we close the modal.

Can we close only if users switches accounts?

ui/components/Wallet/UnverifiedAsset/AssetVerifyToggler.tsx
kkosiorowska: Unfortunately, yes. Without this part of the code, the icon does not change color when hovered over. I think it has to do with the fact that we are nesting elements that react to hover. I will remove the unneeded `hoverColor` in the next PR.

kkosiorowska:
> @Shadowfiend is correct here, it is very jarring, but that might be because we close the expanded menu after we close the modal.
> 
> Can we close only if users switches accounts?

I agree, it's annoying after a long time. I missed it by making improvements. Let's fix this in the next PR.

background/redux-slices/assets.ts
kkosiorowska: You are right because this logic appears here [earlier](https://github.com/tahowallet/extension/blob/main/background/services/indexing/index.ts#L642). I will fix this in the next PR.

Shadowfiend:
Want to be clear: collapsing the list is the only past I found jarring! Rest is 👌 from my perspective, such as it is 😄

PR #3404, opened 2023-05-25
Unsigned Integers: Remove everything in `ui/components/SignTransaction`

Shadowfiend:
This is the first step in eliminating all vestiges of the old transaction signing flow. The feature flag can't be dropped yet, as there are a few other entanglements left.

A few components were still in use:
- Several around Ledger status; these have been moved to SignerLedger/ under the new signing structure.
- The loading screen for signature data was still in use; it's been moved to `Signing/SigningLoading`.
- The network account info top bar, which is static in the signing flow (instead of modifiable in the main wallet) was also in SignTransaction; it's been moved to `SigningNetworkAccountInfoTopBar` under `Signing`.
- The `SignTransactionDetailWarning` component, used to display warnings such as when assets are being transferred directly to a contract address; it's been moved to `TransactionSignatureDetailsWarning` under the `TransactionSignatureDetails` section of `Signing`.

Other than name adjustments, the moved components have no changes.

------

Draft until #3392 lands.

## To Test

Testing is mostly focused on making sure we're still following the new flow in signing, since we're ripping out a lot of code here.

- [ ] Ensure that each type of transaction signing has its summary data displayed correctly.
  - [ ] ETH transfer (can be from the extension).
  - [ ] ERC20 transfer (can be from the extension).
  - [ ] Approval transaction (the Tally test wallet does not have SUSHI approved, so an easy way is to try to swap SUSHI).
  - [ ] Normal contract interaction (a swap with an approved asset will work here).
- [ ] For all of the above, make sure fee changes are applied correctly and, if there are other internal form elements (e.g., the approval transaction amount), that these are also applied correctly.
- [ ] Ensure that each type of data signing has its summary data displayed correctly.
  - [ ] Sign In With Ethereum. Can be checked at https://login.xyz/ under Get Started.
  - [ ] Check regular message signing still looks visually consistent. Can be checked at https://mintkudos.xyz/ on Polygon by hitting Connect Wallet.
- [ ] Ensure Ledger accounts are going through the new Ledger flow by checking that connection status is displayed at the top of the signing screen for data, typed data, and contract interactions.


ui/hooks/signing-hooks.ts
```diff
@@ -40,10 +44,62 @@ export const useAreKeyringsUnlocked = (redirectIfNot: boolean): boolean => {
   return keyringStatus === "unlocked"
 }
 
+// FIXME Remove after USE_UPDATED_SIGNING_UI = true
```
jagodarybacka, line 47: It seems like this function is already unused 🤔

ui/pages/PersonalSign.tsx
```diff
@@ -1,102 +1,17 @@
...
-    }
-
-    return <Signing request={signingDataRequest} />
-  }
-
-  if (
-    typeof signingDataRequest === "undefined" ||
-    typeof signerAccountTotal === "undefined"
-  ) {
+  if (currentAccountSigner === null || signingDataRequest === undefined) {
```
jagodarybacka, line 12: Seems like we are doing this check inside `Signing` component already, right?

jagodarybacka (commented):
Went through the code, will test it manually

jagodarybacka:
Alright, all transactions flows look ok ✅

hyphenized:
I'm also having problems while connecting to https://login.xyz/, https://wagmi.sh/examples/sign-in-with-ethereum works fine though

Shadowfiend:
Yeah I need to have a look here. Not sure what would cause that in this change, have we confirmed that's not happening on main?

jagodarybacka:
@Shadowfiend It is [happening on main](https://github.com/tahowallet/extension/actions/runs/5124093665/jobs/9215539267)

Shadowfiend:
Got it. The fact that e2e tests were passing before also indicates this is likely a change on that site. We should fix I think, but likely don't need to block this PR.

jagodarybacka (approved):
It's good to go - other fixes are planned to be implemented in the next PR, right? Lmk if so and we can 🚢

Shadowfiend:
Sorry, yes—other fixes in followups, let's ship this when ready!

jagodarybacka:
Alright, let's merge it 🚀

Issue #3406, opened 2023-05-25
Missing $ price on swap transaction screen


jagodarybacka:
Sometimes we are able to show $ price on the swap screen but on the transaction screen we are losing it. 

It happens because swaps screen is using [checkCurrencyAmount](https://github.com/tahowallet/extension/blob/swap-tx-redesign/background/redux-slices/0x-swap.ts#L360) that will fetch the prices if there are no recent prices saved on the asset. Transaction screen is using [saved recent prices](https://github.com/tahowallet/extension/blob/swap-tx-redesign/ui/components/Swap/SwapQuoteBalanceChange.tsx#L50) which may be empty for some assets.

We should ensure that if prices are fetched then they are saved/cached so we can avoid problems like that and just use one method to get $ price where needed instead of having multiple ways to do the same thing which is causing bugs and problems.

https://github.com/tahowallet/extension/assets/20949277/7e02ee87-76fb-423c-9cdc-86b54f8c658b

_Originally posted by @jagodarybacka in https://github.com/tahowallet/extension/issues/3392#issuecomment-1561328227_

PR #3408, opened 2023-05-25
Cache new asset price points

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3406

It's working but very ugly approach right now 🤢 will think about something better 

### What
If we are using a function that will trigger fetching new prices for assets then these prices should be cached.
Let's try dispatching `newPricePoint` on demand with these new prices.

Latest build: [extension-builds-3408](https://github.com/tahowallet/extension/suites/13148307583/artifacts/714837068) (as of Thu, 25 May 2023 14:28:21 GMT).

### DIFF
--- a/background/redux-slices/0x-swap.ts
+++ b/background/redux-slices/0x-swap.ts
@@ -362,14 +362,16 @@ export const fetchSwapPrice = createBackgroundAsyncThunk(
           quoteRequest.assets.buyAsset,
           assets,
           quote.buyAmount,
-          quoteRequest.network
+          quoteRequest.network,
+          dispatch
         ),
         sellCurrencyAmount: await checkCurrencyAmount(
           Number(quote.sellTokenToEthRate),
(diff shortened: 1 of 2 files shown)


Shadowfiend:
To approach the “something” better here, wanted to suggest a quick thought: right now, we generate the buy and sell currency amounts alongside the regular lookup of the quote.

Instead of doing that, we could calculate buy and sell currency amounts live in the swap component based on selected data from the redux store. Then, when we generate the quote, we trigger a request to get up-to-date price data to either the indexing service or the asset slice. Then the indexing service/asset slice is in charge of inspecting whether the price data that is currently available is recent enough, and trigger a lookup if not. Once the lookup completes and that data is available, the price info would display (since we'd be using a selector that would trigger immediately).

What this doesn't let us do is show a loading spinner while the price data resolves, so there's more to ponder there.

mhluongo:
Closing as stale

Issue #3399, opened 2023-05-24
Make it possible for the user to delete a verified/unverified asset


kkosiorowska:
The user should have the ability to remove assets that have been verified by mistake. We do not want to remove the asset from the memory, but hide it from the user.

PR #3413, opened 2023-05-26
Remove the verified/unverified assets

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3399

## What

The user should be able to delete the verified/unverified asset. However, the asset should not be removed but hidden from the user. To see a deleted asset again, the user should add it as a custom asset.

## UI

https://github.com/tahowallet/extension/assets/23117945/9b03a6cc-f75e-41ba-b992-f4f7de204842

## To Test

- [x] Remove verified/unverified asset. The asset should disappear. 
- [x] After removal, the metadata should be updated in the db for this token. Check table `customAssets`.
- [x] After removal, check that asset has been removed from the redux state.
- [x] Try adding the asset again. The asset should be added correctly.

## Testing Env

```
SUPPORT_UNVERIFIED_ASSET=true
```

Latest build: [extension-builds-3413](https://github.com/tahowallet/extension/suites/13270975374/artifacts/723896525) (as of Wed, 31 May 2023 13:00:16 GMT).


background/assets.ts
```diff
@@ -44,6 +44,10 @@ export type AssetMetadata = {
    * legitimate asset.
    */
   verified?: boolean
+  /**
+   * Set by the user to remove asset that has been verified by mistake.
+   */
+  removed?: boolean
```
Shadowfiend, line 50: I think this only needs to be visible in the db, as the UI can't really make use of it at the moment.

background/services/indexing/index.ts
```diff
@@ -582,8 +582,9 @@ export default class IndexingService extends BaseService<Events> {
     addressNetwork: AddressOnNetwork
   ): Promise<void> {
     const { metadata = {} } = asset
-    // Manually imported tokens are verified
+    // Manually imported tokens are verified and not removed
     metadata.verified = true
+    metadata.removed = false
```
Shadowfiend, line 587: If I'm not misreading the above, if `asset` has no `metadata` field yet, these two changes won't take effect. Maybe this is why we're sprinkling defaulting of the field in a few other places. This seems like the right place to default it:

```typescript
  asset.metadata = {
    ...(asset.metadata ?? {}),
    verified: true,
  }
```

Shadowfiend (commented):
Dropped a couple of small comments, but my thought one is: I think we should mark an asset as removed in the database, but not percolate any of that out to the UI. The UI should basically see the asset as actually deleted, but the db can track that it once existed and it was removed, to be used only for checking if an unverified, discovered (not added) asset should be added to the asset list.

The flow would be:
- User hits “Don't Show”.
- Indexing service is notified.
- Db is updated to mark the asset as removed with a field.
- Further queries of assets in the db always omit `removed: true`.
- Redux is updated to not have the asset at all (same as before we added the `removed` flag).
- When a new transaction comes in that mentions the asset and we go to add it as unverified in the indexing service, the asset is not added if it already exists in the db as `removed: true`.
- When adding a custom asset, if it already exists as `removed: true`, both `removed` is set to false and `metadata.verified` is set to `true`.

Wdyt?

kkosiorowska:
> Dropped a couple of small comments, but my thought one is: I think we should mark an asset as removed in the database, but not percolate any of that out to the UI. The UI should basically see the asset as actually deleted, but the db can track that it once existed and it was removed, to be used only for checking if an unverified, discovered (not added) asset should be added to the asset list.

It makes a lot of sense let's do it this way.

background/services/indexing/db.ts
```diff
@@ -287,22 +299,24 @@ export class IndexingDatabase extends Dexie {
     )
   }
 
-  async getCustomAssetsByNetworks(
+  async getActiveCustomAssetsByNetworks(
     networks: EVMNetwork[]
   ): Promise<SmartContractFungibleAsset[]> {
```
jagodarybacka, line 304:
```suggestion
  ): Promise<CustomAsset[]> {
```
shouldn't it return custom assets?

jagodarybacka (commented):
minor comments, will test manually in a sec

jagodarybacka (commented):
Can we fix css here? (we can do it later as well)
image

background/redux-slices/assets.ts
```diff
@@ -175,6 +182,39 @@ export const refreshAsset = createBackgroundAsyncThunk(
...
+      asset: SmartContractFungibleAsset
+    },
+    { extra: { main } }
+  ) => {
+    await main.hideAsset(asset)
+  }
+)
+
+/**
+ * Removes the custom asset from the user interface.
```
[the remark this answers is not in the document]
kkosiorowska: I think we can remove it. The removal functions to check whether the assets are the same anyway regardless of whether it is a custom asset.

jagodarybacka (approved):
All good, would love a browser tooltip here if we are using ellipsis but let's not block it
image

PR #3416, opened 2023-05-29
Opening activity details from token verification level

kkosiorowska:
## What
The user should be able to view activity details from the token verification level. But in many cases, activities for unverified tokens are not yet available for some log time. It is important to secure such a case. If the activity is not yet available the preview should not be available but user should be able to copy tx hash. 

## UI

https://github.com/tahowallet/extension/assets/23117945/b3c51fb8-e0f9-4cd4-9c47-d93806faafab

https://github.com/tahowallet/extension/assets/23117945/d2f3db8d-721c-4982-8e55-a1f82b971573

## To Test
- [x] Check that the activity details preview is not available when the activity is not yet loaded. User should be able to copy tx hash.
- [x] Check activity preview is working correctly (Token BANANA for testertesting.eth)

## Testing Env
```
SUPPORT_UNVERIFIED_ASSET=true
```

Latest build: [extension-builds-3416](https://github.com/tahowallet/extension/suites/13219232197/artifacts/720058452) (as of Mon, 29 May 2023 13:47:01 GMT).


Shadowfiend:
Hmm... any idea why we end up with a tx we don't have details for? Fix seems fine for now but it would be even better if we could look these up on demand, though I realize we're not all the way set up for that the way activity details work right now.

kkosiorowska:
> Hmm... any idea why we end up with a tx we don't have details for? Fix seems fine for now but it would be even better if we could look these up on demand, though I realize we're not all the way set up for that the way activity details work right now.

It seems to me that it is a thing of our queue `transactionsToRetrieve`. Transaction hashes are added to it, but they are not retrieved for a long time. I think we would have to look at this separately and maybe improve it.

jagodarybacka:
If the transaction cannot be shown on the slideup, can we allow to copy hash as a text?

kkosiorowska:
> If the transaction cannot be shown on the slideup, can we allow to copy hash as a text?

I wonder if it won't be a bit confusing for the user. What I mean by that is that it doesn't know what to expect when it clicks. Clicking on one address will open the activity details page, and clicking on a second address for other token will copy it. @VladUXUI  do you have any suggestions for this?

jagodarybacka:
the perfect solution would be to just fix these transactions - they are also not showing up on the single asset pages but if we have found the asset in the transaction that mean the transaction should be displayed on that asset's page 😉 That's not blocking for this PR anyway

Shadowfiend:
Separately, and later, I wonder if we can introduce some sort of call to force a transaction house look up for situations like this.

jagodarybacka (approved):
Works fine ✅

PR #3418, opened 2023-05-29
Use fork in E2E tests of transactions

michalinacienciala:
This commit introduces automated E2E tests for asset transactions and changes to
the testing framework needed for the tests to run successfully.

Changes to the framework include:
* The changes from https://github.com/tahowallet/extension/pull/3247 which
added support for the hardhat functionality in the wallet's code.
* Creation of a Hardhat project in `./ci` folder. In the project we configure a
`hardhat` network, which is a network that's a fork of Mainnet. The fork begins
on the block specified in the `blockNumber` parameter in `hardhat.config.ts`
(currently it's set to `17094516`, which is a starting point for our fork-based
E2E tests). Any actions performed on the `hardhat` network will only be applied
locally, they will not affect the real Mainnet network.
* Creation of the helper files and functions in `./e2e-tests/utils`.
* Modification of the GH Actions workflow. A new job (`e2e-tests-fork`) is added
to the `build.yml` workflow. The job uses the build artifact from earlier jobs,
installs the project, installs Hardhat and runs the local `hardhat` network with
the Mainnet fork. Then it runs the Playwright E2E tests designed for fork (the
ones with `-fork` suffix) and upload the test artifacts in case of tests failure.
Also the old `e2e-tests` job gets modified - it's now running only those E2E
tests which are not intended for running on fork (don't contain the `-fork`
suffix.)

The following tests got added:
* User can send base asset (executed on forked Mainnet)
* User can open ERC-20 transfer from asset list and can reject the transfer
  (executed on forked Mainnet)
* User can send ERC-20 from header (executed on forked Mainnet)
* User can send base asset (executed on Goerli testnet)
Most of the tests are run on Mainnet fork (to not waste the funds), but one test
is run on Goerli - this is to verify the display of the transaction on the
activity views (activity history doesn't work currently on fork).

TODO:

- [x] Change the settings of workflow artifacts accessibility (storing of artifacts turned off until we switch to importing accounts using JSON, not seed)
- [x] Change the `TEST_WALLET_RECOVERY_PHRASE` secret to the seed phrase of the `testertesting.eth` wallet (**please don't do that before previous TODOs are checked!**)
- [x] Once secrets are set, do final testing of the workflow
- [x] Clean the testing config from the workflow once tests are done

Latest build: [extension-builds-3418](https://github.com/tahowallet/extension/suites/14283396113/artifacts/802860953) (as of Thu, 13 Jul 2023 18:53:15 GMT).


jagodarybacka (commented):
Not finished yet, first batch of comments

e2e-tests/transactions.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+        await walletPageHelper.onboarding.addAccountFromSeed({
+          phrase: recoveryPhrase,
+          onboardingPage: page,
+        })
+      } else {
+        throw new Error("RECOVERY_PHRASE environment variable is not defined.")
+      }
+
+      await walletPageHelper.setViewportSize()
+      await walletPageHelper.goToStartPage()
```
jagodarybacka, line 28: As we are repeating these lines maybe we can add a util or method on the `walletPageHelper` to onboard with a seed phrase

jagodarybacka (commented):
One more quick comment. Also where do you plan to cut the scope of this PR - all the TODOs in the code are planned to be implemented here or in the next PRs?

ci/README.md
```diff
@@ -0,0 +1,47 @@
...
+   suggested and a default value)  
+   `CHAIN_API_URL=<Insert_your_API_URL_here>` (needed for running Hardhat, may
+   be e.g. Alchemy or Infura API URL)
+
+2. Run the following commands:  
+   `cd ..`
+   `yarn install`  
+   `yarn start` (or `yarn build`)  
+   `cd ci`  
+   `yarn install`
```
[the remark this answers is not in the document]
michalinacienciala: You're right, we don't. Removed in 6d2b9aec865f5407ab70422190b8ee3ea82187f8.

ci/README.md
```diff
@@ -0,0 +1,47 @@
...
+1. Configure required environment variables (either by setting them in `../.env`
+   or directly in the console):  
+   `USE_MAINNET_FORK=true` (needed for building the package)  
+   `MAINNET_FORK_CHAIN_ID=1337` (needed for running Hardhat, `1337` is the
+   suggested and a default value)  
+   `CHAIN_API_URL=<Insert_your_API_URL_here>` (needed for running Hardhat, may
+   be e.g. Alchemy or Infura API URL)
+
+2. Run the following commands:  
+   `cd ..`
```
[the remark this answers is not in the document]
michalinacienciala: Clarified in 6d2b9aec865f5407ab70422190b8ee3ea82187f8.

ci/README.md
```diff
@@ -0,0 +1,47 @@
...
+3. Unpack `./dist/chrome` to Chrome Extensions.
+
+# What wallet features work / don't work on `hardhat` network
+
+When run on `1337`:
+
+:heavy_plus_sign: we can use specific block as a start.  
+:heavy_plus_sign: we can send txs and do swaps (balances updated)  
+:heavy_plus_sign: we can browse NFTs  
+:heavy_minus_sign: we can't browse Activities
```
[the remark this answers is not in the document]
michalinacienciala: Clarified in 3a9d61f0971a7523cab29466d5863b468b1fcb9e.

ci/README.md
```diff
@@ -0,0 +1,47 @@
...
+# What wallet features work / don't work on `hardhat` network
+
+When run on `1337`:
+
+:heavy_plus_sign: we can use specific block as a start.  
+:heavy_plus_sign: we can send txs and do swaps (balances updated)  
+:heavy_plus_sign: we can browse NFTs  
+:heavy_minus_sign: we can't browse Activities  
+:heavy_minus_sign: wallet shows strange assets  
+:heavy_minus_sign: if no cashe exists, loading assets takes long, account
```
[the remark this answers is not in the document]
michalinacienciala: Fixed in 3a9d61f0971a7523cab29466d5863b468b1fcb9e.

ci/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+The E2E tests are located in the `../e2e-tests` directory. Only those with
+`-fork` suffix are meant to be run on forked Mainnet.
+
+To run those tests, go to `..` and run `npx playwright test` .
+
+To run specific test, go to `..` and run `npx playwright test <file_name>`.
+
+To run all tests designed for fork, go to `..` and run
+`find e2e-tests -type f -name "*-fork.*" -exec npx playwright test {} \;`.
```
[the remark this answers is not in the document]
michalinacienciala: Added in 1a3a8905514b4023b584435c596322c7f7bab6cf.

ci/hardhat.config.ts
```diff
@@ -0,0 +1,23 @@
...
+  defaults: "../.env.defaults",
+})
+
+const config: HardhatUserConfig = {
+  networks: {
+    hardhat: {
+      forking: {
+        enabled: true,
+        url: process.env.CHAIN_API_URL || "",
+        blockNumber: 17094516,
```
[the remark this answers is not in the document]
michalinacienciala: Modified a bit (and documented) in e43c36cd85d051a1a4c876580a7ab7ac6a36c1f9.

ci/tsconfig.json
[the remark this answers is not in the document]
michalinacienciala: We can use the one from root, the one I created was obsolete. Removed in 12a05d4901f5f38803bebe238ad350632f7e82b2.

michalinacienciala:
@jagodarybacka , I've addressed your comments. I also commented out the storing of Playwright artifacts in CI and updated the TODOs in the PR description accordingly. Once there will be no more comments to be addressed, we'll be able to proceed with the TODOs.

> Also where do you plan to cut the scope of this PR - all the TODOs in the code are planned to be implemented here or in the next PRs?

Regarding the TODOs in code, I was thinking of addressing thm (apart from the ones in `build.yml`) in a separate PRs (they are outside of the initial suggested testing steps). But if you would like to see some resolved as part of #3418, let me know.

michalinacienciala:
The TODOs which will not be delivered as part of this PR are tracked here: https://github.com/tahowallet/extension/issues/3489.

@Shadowfiend, could you set `FORK_TEST_WALLET_RECOVERY_PHRASE` secret to some dummy value (I wan't to make sure it's not leaked in logs before we set the proper value. Also, please set up ALCHEMY_URL secret (it should have format `https://mainnet.infura.io/v3/<key>`). Once I'll have them, I'll follow with the further steps.

.github/workflows/build.yml
```diff
@@ -140,14 +175,79 @@ jobs:
...
+      - name: Extract extension
+        run: unzip -o chrome.zip -d dist/chrome
+      - name: Cache Hardhat
+        uses: actions/cache@v2
+        with:
+          path: ci/cache
+          key: ${{ runner.os }}-hardhat-${{ hashFiles('ci/cache/*.json') }}
+      - name: Run Hardhat
+        env:
+          CHAIN_API_URL: https://mainnet.infura.io/v3/${{ secrets.ALCHEMY_API_KEY }}
```
michalinacienciala, line 227: @Shadowfiend, do both `ALCHEMY_API_KEY` and `DEV_ALCHEMY_API_KEY` secrets point to a Ethereum Mainnet-based key? I wonder if I should conditionally use `DEV_ALCHEMY_API_KEY` here (when tests are run on a dev build).

michalinacienciala:
> not I believe in secrets.

Sounds intriguing, can you elaborate? :)


> For testing the other one you should just be able to use an `env:` block right? It needn’t be secret.

Well I figured I should be good without the testing. We're switching off the storing of artifacts until we move to the import via JSON, so no risk of leaking the seed phrase there. And GH in its logs does obfuscate all the texts matching the secrets, so we should be good also on this front.
So I think we can already set the final value of the `FORK_TEST_WALLET_RECOVERY_PHRASE` secret.

There is one thing that I still need to figure out - I noticed that the failed e2e tests do not mark the GH Actions step as failed. Not sure yet why is that...

.github/workflows/build.yml
michalinacienciala: I'm getting `Invalid JSON-RPC response received: invalid project id` when starting Hardhat with this config, maybe it's related to the fact that the build was built using `DEV_ALCHEMY_API_KEY`.

.github/workflows/build.yml
```diff
@@ -63,6 +63,39 @@ jobs:
...
+        run: |
+          echo -e ${{ steps.env-block.outputs.result }} > .env
+          echo 'USE_ANALYTICS_SOURCE="BETA"' >> .env
+          yarn build
+        env:
+          ALCHEMY_KEY: ${{ secrets.DEV_ALCHEMY_API_KEY || 'oV1Rtjh61hGa97X2MTqMY9kEUcpxP-6K' }}
+          BLOCKNATIVE_API_KEY: ${{ secrets.DEV_BLOCKNATIVE_API_KEY || 'f60816ff-da02-463f-87a6-67a09c6d53fa' }}
+          DAYLIGHT_API_KEY: ${{ secrets.DAYLIGHT_API_KEY }}
+          COMMIT_SHA: ${{ github.sha }}
+          USE_MAINNET_FORK: true
```
michalinacienciala, line 77: Looks that somehow `USE_MAINNET_FORK=true` gets ignored or overwritten (I'm saying that because `e2e-tests-fork` job skips the fork-based tests and executes the rest). I'll investigate tomorrow.

michalinacienciala: Nvm, fixed that - had to add `USE_MAINNET_FORK: true` `env` to the step running tests.

michalinacienciala:
> > not I believe in secrets.
> 
> Sounds intriguing, can you elaborate? :)

@Shadowfiend, you meant "The alchemy value is already set for prod and not prod (dev) in secrets, I believe", right? Gosh, I only just now realized that :D

> There is one thing that I still need to figure out - I noticed that the failed e2e tests do not mark the GH Actions step as failed. Not sure yet why is that...

That's not a problem anymore, I had to get rid of using `find` for deciding which tests should be run on which environment.

The current outstanding things are those:
- `FORK_TEST_WALLET_RECOVERY_PHRASE` is not set yet as a secret
- there's a problem with running Hardhat (that `Invalid JSON-RPC response received: invalid project id` issue)

e2e-tests/utils/onboarding.ts
```diff
@@ -33,6 +33,30 @@ export default class OnboardingHelper {
...
+      .find((page) => /onboarding/.test(page.url()))
+
+    if (!onboarding) {
+      // Should never happen
+      throw new Error("Onboarding page closed too early")
+    }
+
+    return onboarding
+  }
+
```
michalinacienciala, line 59: This fragment of code got removed in https://github.com/tahowallet/extension/pull/3503/files, but I had problems with building the code after I merged this change to my feature branch from main. I restored it in 39c8f872627f3d1d19041d5211b916ad1c74f391 and after that building and onboarding works for me. But maybe I should have be able to onboard without restoring this code?

Shadowfiend: I think you should be able to just reference the unscoped `getOnboardingPage` as an import rather than invoking it on the `onboarding` helper. It wasn't using anything specific to the onboarding helper and was duplicating the code, thus the removal.

We can also introduce a version on the onboarding helper that just calls through to the top-level helper.

.github/workflows/build.yml
michalinacienciala: Gosh, I was using Infura's URL syntax instead of Alchemy's 🤦 
Works after updating to the correct one.

e2e-tests/utils/onboarding.ts
michalinacienciala: Addressed in [3bf0776](https://github.com/tahowallet/extension/pull/3418/commits/3bf07760d8dec45c7443aa474d115e0a44b07db4).

michalinacienciala:
Leaving as draft until we set the `TEST_WALLET_RECOVERY_PHRASE` secret and remove the testing config.

Shadowfiend:
Wrapping up for now, but this looks like it's well on the way to working. Regular e2e tests failed on some NFT stuff that's been failing them across PRs for the last day or two, a deeper look is warranted but probably no need to block this PR. Will let you take it out of draft when ready!

michalinacienciala:
@Shadowfiend, ready for a final review 🚀

.github/workflows/build.yml
```diff
@@ -63,6 +63,39 @@ jobs:
         with:
           name: extension-builds-${{ github.event.number || github.event.head_commit.id }}
           path: dist/*.zip
+      - name: Dev build on Mainnet fork
+        if: github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')
```
Shadowfiend, line 67: Non-blocking: I would love to move a bunch of this to a structure more like what we have @ https://github.com/thesis/valkyrie/blob/main/.github/workflows/build-and-test.yml#L14-L68 . Basically a step that centralizes all the categories of builds we're trying to choose between, so that the `if` statements are clearer.

e2e-tests/utils/walletPageHelper.ts
```diff
@@ -37,6 +37,132 @@ export default class WalletPageHelper {
...
+  }
+
+  /**
+   *  The function checks elements of the main page that should always be present.
+   */
+  async verifyCommonElements(
+    network: RegExp,
+    testnet: boolean,
+    accountLabel: RegExp
+  ): Promise<void> {
```
Shadowfiend, line 102: Non-blocking: this is opaque enough at the call site that I'd like to suggest we move to an object instead:

```suggestion
  async verifyCommonElements({
    network, testnet, accountLabel
  }: {
    network: RegExp,
    testnet: boolean,
    accountLabel: RegExp
  ): Promise<void> {
```

This would make the invocations much clearer IMO:

```typescript
  walletPageHelper.verifyCommonElements({ network: /.../, testnet: true, accountLabel: /.../ })
```

e2e-tests/utils/transactions.ts
```diff
@@ -0,0 +1,344 @@
...
+    const senderButton = assetActivityItemPopup
+      .getByTestId("tx_participant_wrap")
+      .filter({ hasText: "From:" })
+      .getByRole("button", { name: sendFromAddressShortened })
+    await expect(senderButton).toHaveAttribute(
+      "title",
+      dedent(`
+        Copy to clipboard:
+        ${sendFromAddressFull}
+      `)
```
Shadowfiend, line 256: Non-blocking: this feels like it would be better as a regex rather than pulling in a new dep just for this test, fwiw.

Shadowfiend (approved):
Feels like we could probably have included `e2e` in the branch name so we could actually have seen the tests run, but I'm good with this as-is and we'll see how it does on main!

PR #3424, opened 2023-05-31
Verified/Unverified tokens  - QA Build 

kkosiorowska:
## Checklist

- Token verification
  - [x] Open a single page for an unverified asset. There should be information that the asset is not verified.
  - [x] Operations such as send/ swap should be blocked for unverified assets.
  - [x] Verify the asset. The token should be added to the list of regular assets.
  - [x] Open a single page for this asset. The page should display information that the asset has been verified by the user.
  - [x] After verification of the token, the ability to execute transactions and swaps should be unlocked.

- Removing and adding assets
  - [x] Hide a selected verified/unverified token. The token should be removed from the list.
  - [x] Try adding a removed asset by contract address. The token should be added correctly. An imported token should appear as a verified asset.

- Asset warning 
  - [x] ,,Added from transaction'' should be a link to that transaction or if this is not possible tx hash should be copied.
  - [x] ,,Contract  address''  should be a link to the block explorer for the selected asset.

Latest build: [extension-builds-3424](https://github.com/tahowallet/extension/suites/13466658236/artifacts/738968790) (as of Thu, 08 Jun 2023 14:13:53 GMT).


jagodarybacka:
🔴 Blocking in my opinion:
- [ ] Swaps assets list is showing unverified tokens even when toggle in the settings is off - https://github.com/tahowallet/extension/issues/3438 I think it is worth to think about fixing this bug before flipping the flag

🤔 Questions and possible improvements:
- [ ] this tooltip is marked as WIP text in Figma - do we want to fix it? 
image
- [ ] I don't really like that the hidden assets list is collapsing every time I change the page an go back to Wallet page. If I am browsing assets I own that are unverified it is annoying to constantly toggle the list
- [ ] have we decided that the whitespace on top is fine? 
image
- [ ] Header and X button are not in the same line 
image
- [ ] Figma shows contract address as a text area but I guess we are fine with single-line input, right? 
image
- [ ] margins around Knowledge Center are too big I think 
image
- [ ] if 2 accounts own the same unverified token then which transaction should we show in the "Added from transaction"?
- [ ] if we can show a link to the etherscan for contract address maybe instead of copying a link to the transaction let's allow users to see the transaction on the etherscan?

Shadowfiend:
Agree on blocking. Will let @VladUXUI chime in on the rest.

jagodarybacka:
>  browsing assets that are unverified, you are probably not going to change pages

But that involves selecting an asset -> going to a single asset page -> verifying it -> going back to Wallet -> seeing the hidden assets list collapsed

kkosiorowska:
Questions for @VladUXUI 

> Figma shows contract address as a text area but I guess we are fine with single-line input, right?

> margins around Knowledge Center are too big I think

You have not addressed these points so I understand that this should be fixed. Let me do it.

> Header and X button are not in the same line

I think we've talked about this before, but let us know if it should be improved. This state is also in another component.

Screenshot 2023-06-07 at 12 20 06

> The transaction should open up in Taho, as a tx-details modal not as copy link.
I can't find in my accounts an asset that is discovered in Transactions to test it out

If the activity is available to the user, the transaction view will be opened.  Unfortunately, in many cases, the transactions have not yet been retrieved, which is why there is a copy option.  

You can check this for `testertesting.eth` account, polygon network, and token symbol `Banana`. Please let us know if we are staying on this.

https://github.com/tahowallet/extension/assets/23117945/43610d88-131d-404c-8e1c-3c27efaa2df0

Shadowfiend:
> >  browsing assets that are unverified, you are probably not going to change pages
> 
> 
> 
> But that involves selecting an asset -> going to a single asset page -> verifying it -> going back to Wallet -> seeing the hidden assets list collapsed 

You can just verify an asset directly in the list, no? No need to go to the single asset page unless it's one you're particularly puzzled over, and I'm understanding Vlad to mean that we're expecting that to be a relatively corner case during a mass verification/unverification action.

Shadowfiend:
For the contract address textarea, let's only make the change if it's quick and easy. If we need to reimplement all our input styling and stuff, let's skip it for now and focus on the other things we're trying to wrap up this sprint.

jagodarybacka:
> You can just verify an asset directly in the list

Yup, I'm just complaining because I noticed that it is slightly inconvenient and I would love us to have great UX, but it's not a huge deal 😉

VladUXUI:
> Figma shows contract address as a text area but I guess we are fine with single-line input, right?

This is for the add custom asset. And that is good to stay as is.

> Header and X button are not in the same line

They should indeed be in same line, meaning that the X is not positioned correctly, the title is.


>If the activity is available to the user, the transaction view will be opened. Unfortunately, in many cases, the transactions have not yet been retrieved, which is why there is a copy option.

Ahhh i see, let's go with open in etherscan in that case, and use the same icon as contract instead of copy, as Jagoda mentioned

kkosiorowska:
> > If the activity is available to the user, the transaction view will be opened. Unfortunately, in many cases, the transactions have not yet been retrieved, which is why there is a copy option.
> 
> Ahhh i see, let's go with open in etherscan in that case, and use the same icon as contract instead of copy, as Jagoda mentioned

Let's fix it here [#3454](https://github.com/tahowallet/extension/pull/3457)

kkosiorowska:
> > if 2 accounts own the same unverified token then which transaction should we show in the "Added from transaction"?
>
>The account that you are seeing the unverified token on.

Let's fix it here https://github.com/tahowallet/extension/pull/3458

jagodarybacka (approved):
Alright, I think it's good to go 🚀

PR #3431, opened 2023-06-02
Fix duplicate base assets being stored in redux asset cache

### LINKED ISSUE #3445: Wallet popup lagging due to asset cache growing too large
On every balance update, the wallet is storing duplicate instances of base asset for custom networks, likely due to the check here 

https://github.com/tahowallet/extension/blob/dda5d8d81e859f9129643425556fda2c0c4aeb1e/background/redux-slices/assets.ts#L85

hyphenized:
Closes #3445 

Fixes an issue when caching assets in the redux state that did not correctly handle already existing network base assets.

## To Test
- [x] Do a fresh install on `main`
- [x] Add a custom network, after a few minutes check the assets slice in the redux state, the last entries should be duplicates of the custom network's base asset.
- [x] Switch to this branch and reload the extension, after a few minutes check the assets slice, there should not be any duplicates.

Latest build: [extension-builds-3431](https://github.com/tahowallet/extension/suites/13384491623/artifacts/732526029) (as of Mon, 05 Jun 2023 17:25:44 GMT).

### DIFF
--- a/background/redux-slices/migrations/to-29.ts
+++ b/background/redux-slices/migrations/to-29.ts
@@ -1,0 +1,14 @@
+// This migration ensures the assets collection is cleared of duplicated
+// base assets
+
+export default (
+  prevState: Record<string, unknown>
+): Record<string, unknown> => {
+  const { assets, ...newState } = prevState
+
+  // Clear assets collection; these should be immediately repopulated by the
+  // IndexingService in startService.
+  newState.assets = []
+
+  return newState
+}
--- a/background/redux-slices/migrations/to-29.ts
+++ b/background/redux-slices/migrations/to-29.ts
@@ -1,5 +1,5 @@
 // This migration ensures the assets collection is cleared of duplicated
-// base assets
+// base assets https://github.com/tahowallet/extension/issues/3445
 
 export default (
   prevState: Record<string, unknown>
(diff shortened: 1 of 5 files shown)


background/redux-slices/utils/asset-utils.ts
```diff
@@ -398,5 +398,9 @@ export function isSameAsset(asset1: AnyAsset, asset2: AnyAsset): boolean {
     return false
   }
 
+  if (isNetworkBaseAsset(asset1) && isNetworkBaseAsset(asset2)) {
+    return asset1.chainID === asset2.chainID
+  }
+
```
kkosiorowska, line 404: Can we use the `sameNetworkBaseAsset` function here?

background/redux-slices/migrations/to-29.ts
```diff
@@ -0,0 +1,14 @@
+// This migration ensures the assets collection is cleared of duplicated
+// base assets
```
Shadowfiend, line 2: Let's drop a link to an issue here.

Shadowfiend (approved):
Def going to still be some bugs lurking here, but let's ship this so we stop the bleeding.

Shadowfiend:
We need to add some unit tests for `isSameAsset` while we're in here. Any chance we can slot some of those in this week?

Shadowfiend (approved):
A'iiight all set here, tested it out and did some extra logging, all looks right.

PR #3432, opened 2023-06-03
Ledger Respect: Attempt #2 at fixing Ledger connection issues

### LINKED ISSUE #3419: Blank page on disconnected ledger
Introduced in v0.34.0

If the Ledger goes to sleep/screensaver and a dapp requests a signature, there is an error and the wallet UI shows up blank.

Ledger_sleeping

v0.32.0 (working) for reference:
v0 32 0_check_Ledger

v0.34.0 (broken) and higher:
v0 34 0_blank

_Originally posted by @beemeeupnow in https://github.com/tahowallet/extension/issues/3407#issuecomment-1564524469_

Shadowfiend:
Maybe fixes #3419.

Compare against #3420, and let's land the one that seems to work + be most correct while we debug further.

Latest build: [extension-builds-3432](https://github.com/tahowallet/extension/suites/13347808183/artifacts/729885779) (as of Sat, 03 Jun 2023 12:24:37 GMT).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -1269,14 +1269,13 @@ export default class Main extends BaseService<never> {
         resolver: (result: string | PromiseLike<string>) => void
         rejecter: () => void
       }) => {
-        // Don't await, as the below enrichment is expected to take longer than
-        // signer prep. If that assumption breaks, we should probably await the
-        // two in parallel.
-        this.signingService.prepareForSigningRequest()
+        // Run signer preparation and enrichment in parallel.
+        const [, enrichedSignTypedDataRequest] = await Promise.all([
+          this.signingService.prepareForSigningRequest(),
+          this.enrichmentService.enrichSignTypedDataRequest(payload),
+        ])
 
-        const enrichedsignTypedDataRequest =
-          await this.enrichmentService.enrichSignTypedDataRequest(payload)
-        this.store.dispatch(typedDataRequest(enrichedsignTypedDataRequest))
+        this.store.dispatch(typedDataRequest(enrichedSignTypedDataRequest))
 
         const clear = () => {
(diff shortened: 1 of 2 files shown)


Shadowfiend:
Testing some more, this looks like it does in fact do the trick:

- `onConnection` does not block on a connected Ledger that was unlocked then auto-locked.
- Refreshing the Ledger connection is working correctly.
- Errors are logged correctly that were previously being thrown and breaking the loading of the signing UI.

kkosiorowska (commented):
I have tested the solution and it seems to be working well. I have additionally tested the bug from https://github.com/tahowallet/extension/pull/3380 and it does not cause regression either.

jagodarybacka (approved):
I like this solution more as this is simpler

Non blocking issues I see:

- [deriveAddressOnLedger](https://github.com/tahowallet/extension/blob/38e67acfe9aa1295a6df92529c857a815f594883/background/services/ledger/index.ts#L116-L124) which is throwing error while Ledger is asleep is not guarded with a try-catch here and this method is used in other places as well.

- I see the lag from loading the signing screen is a little bit longer here when Ledger is ready for signing (`Promise.all` fault I guess) so would be nice to make sure the skeleton loader is visible instead of just a blank screen.

https://github.com/tahowallet/extension/assets/20949277/840d96aa-628f-4cdb-a74c-1e1fbb331a07

Shadowfiend:
> I see the lag from loading the signing screen is a little bit longer here when Ledger is ready for signing (Promise.all fault I guess) so would be nice to make sure the skeleton loader is visible instead of just a blank screen.

Yeah wondering if we can move to `allSettled` or something there as well. Def worth poking at.

PR #3434, opened 2023-06-05
Fix UI crash when gas fee data is unavailable

hyphenized:
Prevents a crash when trying to send funds after adding a network under slow network conditions or certain RPCs

## To Test
- [ ] Add a custom network, e.g. hardhat, the first RPC on this network should be unavailable
- [ ] Attempt to send some funds, UI should not crash

Latest build: [extension-builds-3434](https://github.com/tahowallet/extension/suites/13375997039/artifacts/731891139) (as of Mon, 05 Jun 2023 12:10:42 GMT).

### DIFF
--- a/ui/components/NetworkFees/FeeSettingsText.tsx
+++ b/ui/components/NetworkFees/FeeSettingsText.tsx
@@ -137,7 +137,7 @@ export default function FeeSettingsText({
   networkSettings = customNetworkSetting ?? networkSettings
   const baseFeePerGas =
     useBackgroundSelector((state) => {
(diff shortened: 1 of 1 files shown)


kkosiorowska (commented):
I'm not quite sure how to test it. How can I actually add a hardhat network? 🤔

jagodarybacka:
Is it happening on regular custom networks added via Chainlist? I think this is a safe change to merge anyway but would be good to add some background 😉

Shadowfiend (approved):
This is just good practice anyway, so let's go ahead and move forward without doing extensive testing.

Issue #3429, opened 2023-06-02
Use per-network feeRecipient for 0x swaps


kkosiorowska:
Currently `COMMUNITY_MULTISIG_ADDRESS` in https://github.com/tahowallet/extension/blob/main/background/constants/index.ts#L26 is a single hardcoded multisig address. Due to difficulties interacting with these Safes on L2s, we want to deploy new Safes that will receive the funds from the original multisig, and then use those as the primary multisigs for L2s.

To do this, we need to be able to look up the community multisig address _for a specific chain_.

We need to:
- Add a function to do this lookup.
- Use it in `get0xApiBase` to construct the `gatedParameters` (https://github.com/tahowallet/extension/blob/main/background/redux-slices/0x-swap.ts#L118).
- Leave the existing `COMMUNITY_MULTISIG_ADDRESS` for EVM-compatible chains as a fallback while we set up the addresses with the community multisig council.

PR #3443, opened 2023-06-05
Use the multisig community address for a specific chain

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3429

## What
The `gatedParameters` for 0x swaps should be set up for a specific chain. Currently we use `COMMUNITY_MULTISIG_ADDRESS` which is a single hardcoded multisig address. We should look up the community multisig address for a specific chain.

The PR gives us the ability to have the addresses for specific chains. At the moment, the addresses are not yet available. They should be deployed in the coming days. Therefore, by default, we should use the existing address.

https://github.com/tahowallet/extension/assets/23117945/778dce26-b91b-41a5-b77d-0324757e2f91

## To Test
- [ ] Try to swap and check parameters for 0x swaps. By default, the existing `COMMUNITY_MULTISIG_ADDRESS` should be used.

Latest build: [extension-builds-3443](https://github.com/tahowallet/extension/suites/13434138434/artifacts/736437224) (as of Wed, 07 Jun 2023 11:57:09 GMT).


background/constants/index.ts
```diff
@@ -23,9 +23,12 @@ export const MINUTE = 60 * SECOND
 export const HOUR = 60 * MINUTE
 export const DAY = 24 * HOUR
 
-export const COMMUNITY_MULTISIG_ADDRESS =
+export const DEFAULT_COMMUNITY_MULTISIG_ADDRESS =
   "0x99b36fDbC582D113aF36A21EBa06BFEAb7b9bE12"
 
+// TODO add correct addresses
+export const COMMUNITY_MULTISIG_ADDRESS: Record<string, string> = {}
```
Shadowfiend, line 30: I think we can be more explicit in what this contains by expanding the type and adjusting the name a bit:

```suggestion
export const COMMUNITY_MULTISIG_ADDRESS_BY_CHAINID: {
  [chainID: string]: string
} = {}
```

background/redux-slices/0x-swap.ts
```diff
@@ -133,11 +134,23 @@ const get0xApiBase = (network: EVMNetwork) => {
   return `${prefix}${base}`
 }
 
-const gatedParameters = {
-  affiliateAddress: COMMUNITY_MULTISIG_ADDRESS,
-  feeRecipient: COMMUNITY_MULTISIG_ADDRESS,
-  buyTokenPercentageFee: SWAP_FEE,
+const getMultisigAddressByNetwork = (network: EVMNetwork) => {
```
Shadowfiend, line 137:
```suggestion
function getMultisigAddressByNetwork(network: EVMNetwork) {
```

Let's stick to regular function declarations unless we need the variable-assignment version for specific reasons (e.g. `this` binding, etc). Amongst other things, it gives us better stack traces (as functions attach their names to the stack in a way that `const`s don't).

background/redux-slices/0x-swap.ts
```diff
@@ -133,11 +134,23 @@ const get0xApiBase = (network: EVMNetwork) => {
...
+  // Look up the community multisig address for a specific chain
+  return (
+    COMMUNITY_MULTISIG_ADDRESS[network.chainID] ??
+    DEFAULT_COMMUNITY_MULTISIG_ADDRESS
+  )
 }
+
+const getGatedParameters = (network: EVMNetwork) => {
```
Shadowfiend, line 145:
```suggestion
function getGatedParameters(network: EVMNetwork) {
```

Shadowfiend (commented):
Quick nonblocking simplification thought.

Shadowfiend (commented):
Looks good in principle, will let @hyphenized bring this home since he's feature buddy here.

hyphenized:
This looks good to me, will merge later in case we want to address the non blocking comment.

hyphenized (approved):

PR #3446, opened 2023-06-05
Facilitated Connections: Add updated dApp Connections flag, clean up a bunch of other feature flags

Shadowfiend:
I went ahead and rolled in the removal of almost all remaining feature flags. The notable exceptions are NFT stuff (which is being handled in #3421) and `USE_UPDATED_SIGNING_UI`, which I'm slowly going through in separate cycles.

The main goal here was adding `ENABLE_UPDATED_DAPP_CONNECTIONS`.

## Testing

- [x] Custom networks appears and works as expected.
- [x] Analytics are still showing up (including the button to delete analytics data, the button to disable analytics, and actually reporting analytics to Posthog).
- [x] RNS names are resolving.
- [x] Keyring lock/unlock is available in the account list.
- [x] There is no `Forgot Password` button on the signing unlock screen.

These are fairly straightforward changes, so I think superficial testing is sufficient here, and we can do some deeper stuff when release time comes.

Latest build: [extension-builds-3446](https://github.com/tahowallet/extension/suites/13389037962/artifacts/732885317) (as of Mon, 05 Jun 2023 20:55:49 GMT).


.env.defaults
```diff
@@ -4,40 +4,35 @@ BLOCKNATIVE_API_KEY="f60816ff-da02-463f-87a6-67a09c6d53fa"
...
+HIDE_SWAP_REWARDS=true
 SUPPORT_WALLET_CONNECT=false
-SUPPORT_CUSTOM_NETWORKS=true
 SUPPORT_CUSTOM_RPCS=false
-SUPPORT_UNVERIFIED_ASSET=false
\ No newline at end of file
+SUPPORT_UNVERIFIED_ASSET=false
+USE_UPDATED_SIGNING_UI=true
+SUPPORT_MULTIPLE_LANGUAGES=false
+SUPPORT_ARBITRUM_NOVA=false
```
hyphenized, line 31: Should we flip this one to true, and remove it as well? This network does not seem to need some kind of special treatment like Optimism/Polygon do 🤔

hyphenized (commented):
Code looks good, will test in a bit

.env.defaults
Shadowfiend: No clue what the right next step is on this one, but toggling a feature flag is def out of scope for this particular PR. Maybe let's discuss the flip in Discord though.

Shadowfiend: https://discord.com/channels/1079490158639976609/1079519391219191908/1115388946667548703

PR #3447, opened 2023-06-05
Dial-up Connection: Add placeholders for new dApp Connections flow

Shadowfiend:
Visual components that should be hidden in the new dApp Connections flow
are hidden when `ENABLE_UPDATED_DAPP_CONNECTIONS` is enabled, and
placeholder code exists for new components (though currently they don't
add any content).

Along the way, add an `isDisabled` complement to `isEnabled` for feature
flags so that usage is clearer, and update the declaration style (using
`function` instead of `const` for consistency) and docs of `isEnabled`.

Lastly, lightly refactor the way that top menus are handled. Previously,
there was an always-present top menu “decoy” that was used to avoid
animating the top menu due to the way transitions were implemented.
Transitions are now implemented to target the `main` element, which is
always outside of the top menu, eliminating the need for the decoy and
some other CSS styles to boot.

## Testing

- [x] Make sure that the dApp connections popup still shows the option to
      switch to another wallet above the reject/connect buttons.
  - [x] Make sure clicking “Yes, switch wallet” on that screen still works.
- [x] Make sure “Use Taho as your default wallet” still appears on the Settings page.
- [x] Make sure, for a connected dApp, that the lightning bolt still appears between network and account info and pops up info when clicked.
- [x] Make sure the top menu still stays stable when switching pages inside
      the wallet tab (e.g. going to an asset page, swap page, or send page).

Latest build: [extension-builds-3447](https://github.com/tahowallet/extension/suites/13394518219/artifacts/733293045) (as of Tue, 06 Jun 2023 03:00:50 GMT).


hyphenized (commented):
Nice work! That decoy was annoying 😅. Everything looks good except for a couple of small visual glitches, one when switching from the wallet page to settings (the eth icon displays for a few seconds?). The other one I noticed is that when opening the unlock signing page the animation looks a bit odd.

Shadowfiend:
Kk I expect both should be fixed now.

Issue #3414, opened 2023-05-26
Restructure Settings page


Shadowfiend:
Since we now have more items in settings, we can start to group them a bit more.
Here is the proposed grouping category for now.

**General**

- Use Taho as default wallet
- Connected websites
- Analytics
__________________

**Networks & Assets**

- Hide asset balance under $2
- Show unverified assets
- Custom Network
- Custom Assets
- Show testnet networks
___________________

**Help Center**

- Bug report
- Need help?

![image](https://github.com/tahowallet/extension/assets/80846886/af26e39f-e147-4b43-8811-e0f29ab6d1e6)

[Figma file](https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?type=design&node-id=1917-75412&t=3T895ofxYIP5OPfi-4)

PR #3450, opened 2023-06-06
Restructure the settings page

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3414

## What
Currently, the settings page includes more elements. Elements should be grouped for better clarity.

## UI

Before
Screenshot 2023-06-06 at 18 32 08

After
Screenshot 2023-06-06 at 18 31 50

## To Test
- [ ] Go to the settings page and check that items are grouped.

Latest build: [extension-builds-3450](https://github.com/tahowallet/extension/suites/13531705094/artifacts/744119081) (as of Mon, 12 Jun 2023 08:28:19 GMT).


ui/utils/lists.ts
```diff
@@ -16,4 +16,8 @@ export const intersperseWith = <T, K>(
   return result
 }
 
+export const excludeFalsyValues = <T>(items: (T | boolean)[]): T[] => {
+  return items.filter((item): item is Exclude<T | boolean, boolean> => !!item)
+}
```
hyphenized, line 21: Sweet! Just a small nit, we can remove the `boolean` union from the parameter as it is implied that T can be a `boolean`, we should also fix the return type here.

```suggestion
export const excludeFalsyValues = <T>(items: T[]): Exclude<T, boolean>[] => {
  return items.filter((item): item is Exclude<T, boolean> => !!item)
}
```

ui/pages/Settings.tsx
```diff
@@ -414,6 +429,24 @@ export default function Settings(): ReactElement {
...
+            margin-bottom: 24px;
+            padding-bottom: 24px;
+          }
+          .group:last-child {
+            border-bottom: none;
+            padding: 0px;
+            margin: 0px;
+          }
+          .group_title {
+            color: var(--green-20);
```
hyphenized, line 443: Is this the right shade? It looks a bit darker in figma

Current: 
image
Design:
image

ui/_locales/en/messages.json
```diff
@@ -485,18 +485,23 @@
   "readOnly": "Read-only",
   "readOnlyNotice": "Read-only mode",
   "settings": {
+    "group": {
+      "general": "General",
+      "walletOptions": "Networks & Assets",
+      "helpCenter": "Help Center"
```
hyphenized, line 491:
```suggestion
      "walletOptions": "Networks & assets",
      "helpCenter": "Help center"
```

hyphenized (changes requested):
This looks good to go, just a few small nits

ui/pages/Settings.tsx
```diff
@@ -319,36 +320,50 @@ export default function Settings(): ReactElement {
...
+        customNetworks,
+        addCustomAsset,
+        enableTestNetworks,
+      ],
+    },
+    helpCenter: {
+      title: t("settings.group.helpCenter"),
+      items: [bugReport, needHelp],
+    },
+  }).map(({ title, items }) => ({ title, items: excludeFalsyValues(items) }))
```
Shadowfiend, line 349: Rather than subjecting the entire list to feature-flag-related action (which would likely stay well after the feature flags die due to the fact that they're not obviously related), I would prefer to continue using the paradigm we've been using of spreading a potentially-empty array:

```typescript
   ...(isEnabled(FeatureFlags.SUPPORT_MULTIPLE_LANGUAGES) ? [languages] : [])
```

In fact, we could introduce a utility for this, as we do similar things in a few places:

```typescript
  wrapIfEnabled<T>(flag: FeatureFlagType, valueToWrap: T): [T] | [] {
    return isEnabled(flag) ? [valueToWrap] : []
  }

  // ...

  [
    // ...
    ...wrapIfEnabled(FeatureFlags.SUPPORT_MULTIPLE_LANGUAGES, languages),
    ...wrapIfEnabled(FeatureFlags.SUPPORT_ACHIEVEMENTS_BANNER, notificationBanner),
  ]
```

ui/pages/Settings.tsx
Shadowfiend: I just discovered most of our color variables don't match the design system. Will fix in the dApp Connection work, so let's leave this one alone.

VladUXUI, line 443: color: var(--green-20); Let's change to 
color: var(--green-40)

In my design the main content is white and title are 20. But i forgot we have mouse over on most items, so let's change the title to 40 and should have same results

VladUXUI (commented):
Looking good!
Just one little change, non blocking

ui/utils/lists.ts
kkosiorowska: It is no longer used.

hyphenized (approved):
🚢

Issue #3438, opened 2023-06-05
Showing incorrect tokens on "sell" swap assets list


jagodarybacka:
Sell asset list on the swap page is showing assets that shouldn't be there
- we are showing assets with 0 balance, that we don't own anymore
- we are showing unverified assets even if "show unverified assets" in the settings is turned off
![image](https://github.com/tahowallet/extension/assets/20949277/1ffea857-5a34-4996-9833-77dfaed86a71)

PR #3452, opened 2023-06-07
Fix issue for assets list for swapping/sending 

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3438

## What
The user should not be able to send and swap unverified assets. In addition, assets with 0 balance should not appear in the list of assets.

## UI

Before

https://github.com/tahowallet/extension/assets/23117945/44eb61d8-9783-4c55-8a43-b7850cfe457c

After

https://github.com/tahowallet/extension/assets/23117945/c549328e-d03e-4d7f-a488-46c3add42df5

## Testing
- [x] Check the list of assets that can be sent. There should not be unverified assets and assets with a balance of 0.
- [x] Check the list of assets that can be swap. There should not be unverified assets and assets with a balance of 0.
- [x] Verify the asset and see if it appears in the lists above.

## Testing Env

```
SUPPORT_UNVERIFIED_ASSET=true
```

Latest build: [extension-builds-3452](https://github.com/tahowallet/extension/suites/13430489328/artifacts/736164593) (as of Wed, 07 Jun 2023 09:22:45 GMT).

### DIFF
--- a/background/redux-slices/utils/asset-utils.ts
+++ b/background/redux-slices/utils/asset-utils.ts
@@ -16,6 +16,7 @@ import {
   OPTIMISM,
   POLYGON,
 } from "../../constants"
+import { FeatureFlags, isEnabled } from "../../features"
 import { fromFixedPointNumber } from "../../lib/fixed-point"
 import { sameEVMAddress } from "../../lib/utils"
 import { AnyNetwork, NetworkBaseAsset, sameNetwork } from "../../networks"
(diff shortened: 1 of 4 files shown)


background/redux-slices/utils/asset-utils.ts
```diff
@@ -375,6 +376,17 @@ export function isUnverifiedAssetByUser(asset: AnyAsset | undefined): boolean {
   return false
 }
 
+/**
+ * Assets that are untrusted and have not been verified by the user
+ * should not be swapped or sent.
+ */
+export function canBeUsedForTransaction(asset: AnyAsset): boolean {
```
jagodarybacka, line 383: I'm not a fan of this function name but don't have a better idea for now 🤔

jagodarybacka (approved):
Works fine ✅

background/redux-slices/utils/asset-utils.ts
```diff
@@ -375,6 +376,17 @@ export function isUnverifiedAssetByUser(asset: AnyAsset | undefined): boolean {
...
+  if (!isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET)) {
+    return true
+  }
+  return isUntrustedAsset(asset) ? !isUnverifiedAssetByUser(asset) : true
```
Shadowfiend, line 387: My 2c is `isUnverifiedAssetByUser` should (a) check if the asset is untrusted already (ie `isUnverifiedAssetByUser` being true should imply `isUntrustedAsset` being true) and in fact I believe it does and (b) should probably have a converse, `isVerifiedAsset`, that is what this function is basically trying to do (and would save us the extra function).

PR #3455, opened 2023-06-07
Route transactions through Flashbots

### LINKED ISSUE #3349: Flashbots Protect MEV-Share Spike
Flashbots recently released their beta implementation of MEV-share!

Details: MEV kickbacks are currently live. Transactions sent via FB Protect RPC are automatically routed through MEV-share. If a searcher bids to backrun the transaction, the user receives 90% of their bid.

- Integration guide for wallets: https://docs.flashbots.net/flashbots-mev-share/orderflow-providers/integration-guide
- MEV-Share gh repo: https://github.com/flashbots/mev-share
- Background on MEV-Share: https://collective.flashbots.net/t/mev-share-programmably-private-orderflow-to-share-mev-with-users/1264/1

**Scope of this spike**

- [ ] What is level of effort required for us add a "Protect with Flashbots" toggle to Settings? Enabling the toggle would change Ethereum RPC to go through Flashbots. Disabling the toggle would put the user back in the normal RPC round robin.
- [ ] Same as above. But instead the toggle would appear on the Sign txn page (not Settings) _for Ethereum L1 txns only_.

jagodarybacka:
Resolves #3349 

### What

Add a checkbox on the transaction screen and a toggle on the settings page to toggle using Flashbots RPC for sending transactions. 

- in Settings user decides if this feature should be used at all
- if the toggle is turned off then there is no checkbox on the transactions screen and Flashbots RPC is not being used
- if the toggle is turned on then there is a checkbox on the transaction screen if this transaction is applicable to Flashbots supported tx type (`eth_sendRawTransaction`) and if the transaction is on the Ethereum network
- user can uncheck the checkbox on the transaction screen and Flashbots won't be used for that transaction
- for the next transaction the checkbox is checked again - if user doesn't want to use Flashbots at all they should switch it off on the Settings page

#### Assumptions:
- I implemented it keeping in mind that we may want to add support custom RPCs defined by users - Flashbots are treated here as a custom RPC that cannot be edited in the UI
- I assumed that only one custom RPC can be used for a given chain at a given time and that custom RPC will have max priority over other RPCs 
- User will be able to decide which methods should be routed through custom RPC - in the case of Flashbots only `eth_sendRawTransaction` makes sense

### Things to consider

- [x] what kind of privacy setup we want to use? [See docs](https://docs.flashbots.net/flashbots-protect/rpc/mev-share#protect-rpc-reference)
  - default stable
- [x] do we want to add a tooltip to explain a bit more what this toggle does?
  - tooltip added
- [x] do we want to support Flashbots on Goerli?
  - not right now

### Testing

- [x] send transaction on Ethereum while toggle in Settings is off - should use different RPC than flashbots
- [x] send transaction on Ethereum while toggle in Settings is on - checkbox should be visible on the transaction page, should be checked by default, transaction should use flashbots
- [x] send transaction on Ethereum while toggle in Settings is on but uncheck the checkbox on tx screen - tx should go through different RPC
- [x] after sending or rejecting tx with unchecked checkbox setup another tx - checkbox should be checked and tx should go through Flashbots
- [x] test with swaps
- [x] test with dexes
- [x] regression tests for personal signs - signing screen shouldn't include checkbox 
- [x] flashbots shouldn't be used for other JSON-RPC calls than `eth_sendRawTransaction` (we have one chain id call while the provider is added to the service which is fine I think) 
- [x] send transaction on other chains while toggle is on - shouldn't use Flashbots and the checkbox shouldn't be there on the transaction screen
- [x] test migration from `main`

image
image
image

Latest build: [extension-builds-3455](https://github.com/tahowallet/extension/suites/14356778724/artifacts/808514345) (as of Mon, 17 Jul 2023 16:57:39 GMT).


Shadowfiend:
> do we want to add a tooltip to explain a bit more what this toggle does?

My 2c here is yes, something like:

> Sends all transactions through Flashbots to protect against frontrunning and shares MEV back with you. This can leak some information about your transactions to searchers, see [the Flashbots documentation](https://docs.flashbots.net/flashbots-protect/rpc/mev-share#stable-configuration) for more.

I would also consider making the “Use Flashbots RPC” toggle be “Use Flashbots RPC by default” since we allow toggling per-tx.

Shadowfiend:
Haven't had a chance to go deep here quite yet, but left some initial thoughts above. Excited to see this working!

jagodarybacka:
> Re: Goerli: only if it's trivial, and if so we should add Sepolia as well.

Seems like they only support Goerli but the docs are not saying anything really about it, I can't even find info if I have to use the same endpoint or a different one to send to the testnet. Maybe let's drop it for now.
image

Re: filtering methods

yeah not sure if that will be needed for other RPCs but I would like to avoid getting rate limited by Flashbots so I think only routing transactions through them is the best option. We can ignore that when we will decide to implement user-imported RPCs.

Shadowfiend:
It does look like Sepolia is supported from various other parts of their docs. Since the instructions for Goerli are this:

> Add https://rpc-goerli.flashbots.net with a chainID of 5 and currency of ETH.

We could just see if this works equivalently on Sepolia. Agree there's no need to burn time on it rn though.

jagodarybacka:
oh thanks, their search bar wasn't helpful at all 🙈

jagodarybacka:
@VladUXUI Can you take a look at this build and let us know what would you like to change regarding UX?

VladUXUI:
Took a look it this and did some digging in documentation.
As far as i understand, send / approve and other simple interactions don't go through Flashbot mempool, i mean Flashbot automatically send these to the public mempool.
Would it then make sense to only have Flashbots in transactions that need actual protection?

This is so users don't think they will get MEV from send/approve transactions.

VladUXUI:
Also, what is the benefit of having the toggle in each transaction?

background/main.ts
```diff
@@ -1859,6 +1860,19 @@ export default class Main extends BaseService<never> {
     await this.chainService.removeCustomChain(chainID)
   }
 
+  async toggleFlashbotsProvider(shouldUseFlashbots: boolean): Promise<void> {
+    if (shouldUseFlashbots) {
+      const flashbotsProvider = makeFlashbotsProvider()
```
kkosiorowska, line 1865: I wonder if we shouldn't move this to `addCustomProvider`  method.

background/redux-slices/ui.ts
```diff
@@ -308,6 +317,14 @@ export const sendEvent = createBackgroundAsyncThunk(
   }
 )
 
+export const toggleFlashbots = createBackgroundAsyncThunk(
+  "ui/toggleFlashbots",
+  async (value: boolean, { dispatch, extra: { main } }) => {
+    dispatch(toggleUseFlashbots(value))
+    await main.toggleFlashbotsProvider(value)
```
kkosiorowska, line 324: Should the order not be reversed? What I mean is that the update of the UI always occurs after the update in the db.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -113,15 +120,33 @@ function isConnectingWebSocketProvider(provider: JsonRpcProvider): boolean {
...
-    (RPC_METHOD_PROVIDER_ROUTING[Number(chainID)] ?? []).some((m: string) =>
-      method.startsWith(m)
+    (ALCHEMY_RPC_METHOD_PROVIDER_ROUTING[Number(chainID)] ?? []).some(
+      (m: string) => method.startsWith(m)
     )
   )
 }
 
+function customOrDefaultProvider(
+  chainID: string,
```
kkosiorowska, line 133: It is not used anywhere.

.env.defaults
```diff
@@ -34,3 +34,4 @@ SUPPORT_ACHIEVEMENTS_BANNER=false
 SUPPORT_NFT_SEND=false
 USE_MAINNET_FORK=false
 ENABLE_UPDATED_DAPP_CONNECTIONS=false
+SUPPORT_FLASHBOTS_RPC=false
```
kkosiorowska, line 37: The description for the dev panel is missing in `messages.json`. We often forget about it but don't use the dev panel probably at all. We could somehow solve this issue.

kkosiorowska (commented):
I haven't tested some minor comments.

background/main.ts
jagodarybacka: Hmm, I wanted to keep service agnostic about Flashbots

.env.defaults
jagodarybacka: Yeah at this point I would vote to drop that feature as a whole as it is causing a burden without adding much value

background/services/chain/serial-fallback-provider.ts
jagodarybacka: I wanted to make it follow the same arguments set as `alchemyOrDefaultProvider` but maybe it doesn't make that much sense 😂

.env.defaults
Shadowfiend: Let's double check no one's life depends on it in discord but then yeah I'm ready to kill it with fire.

background/constants/networks.ts
```diff
@@ -246,7 +251,7 @@ export const CHAIN_ID_TO_COINGECKO_PLATFORM_ID: {
  * The structure is network specific with an extra `everyChain` option.
  * The methods in this array will be directed towards alchemy on every network.
  */
-export const RPC_METHOD_PROVIDER_ROUTING = {
+export const ALCHEMY_RPC_METHOD_PROVIDER_ROUTING = {
```
Shadowfiend, line 254: Non blocking: is this used outside of chain service? It seems like that's where it belongs...

background/main.ts
Shadowfiend: This isn't really making the provider right? It's returning the descriptor/provider creator for it. There's a naming thing here I'd like to get right, though I don't think we necessarily need to block on it.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -919,43 +967,67 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
...
+    supportedMethods: ["eth_sendRawTransaction"],
+    creator: () => getProviderCreator(FLASHBOTS_RPC_URL),
+  }
+}
+
 export function makeSerialFallbackProvider(
   chainID: string,
-  rpcUrls: string[]
+  rpcUrls: string[],
+  customRpc?: { rpcUrl: string; supportedMethods?: string[] }
```
Shadowfiend, line 992: We should name the type for custom RPCs--perhaps `RpcConfig`? 

Also, why not change `rpcUrls` to `rpcConfigs: RpcConfig[]` or so? There doesn't seem to be much difference between the two.

Shadowfiend (commented):
Couple of thoughts as I finally get to this, sorry for the delay. More to come!

background/redux-slices/ui.ts
Shadowfiend: To be clear, but not blocking: we should update the db, and the service should propagate that update to the UI. Preferences are a little wobbly right now in this regard, but that's the ideal data flow through the system.

VladUXUI:
Hopefully is not too late, but could we change toggle inside transaction to a checkmark instead of a toggle?
Toggles are generally used for more permanent states, and this is a situational one.
Maybe something like this?
image

jagodarybacka:
We don't need a feature flag here I think, right?

Shadowfiend:
I think we probably should have merged sooner and used the feature flag, but since we didn't do that, I think the flag has become unnecessary.

ui/components/Signing/SignatureDetails/TransactionSignatureDetails/DetailsPanel.tsx
```diff
@@ -124,6 +143,27 @@ export default function DetailPanel({
...
+                onChange={toggleFlashbotsRPC}
+                label={t("signTransaction.useFlashbots")}
+                labelPosition="left"
+                customStyles={{ width: "100%" }}
+                customStylesForLabel={{
+                  color: "var(--green-40)",
+                  fontSize: "14px",
+                  marginRight: "auto",
+                }}
+              />
```
kkosiorowska, line 163: While testing the changes for https://github.com/tahowallet/extension/pull/3495 I noticed one thing. The empty space does not have a pointer cursor but changes the state of the checkbox. I wonder if in this case, only the checkmark should be clickable.

https://github.com/tahowallet/extension/assets/23117945/67e8b6b1-dbf2-4df8-9faf-7511ac4d2d5a

jagodarybacka: I like that the checkbox label is clickable tbh, that's the default checkbox behavior iirc, I'll just fix that cursor.

Shadowfiend: Labels should always trigger (for buttons) or focus (for fields) their labeled items—but agree the cursor would be a nice touch.

kkosiorowska:
Migration works correctly ✅

ui/pages/Settings.tsx
```diff
@@ -196,41 +208,17 @@ export default function Settings(): ReactElement {
   }
 
   const unverifiedAssets = {
-    title: "",
+    title: t("settings.showUnverifiedAssets"),
+    tooltip: () => (
```
kkosiorowska, line 212: Non-blocking
After adding the tooltip property and merge we should also use this for `autoLockSettings`

kkosiorowska (changes requested):
I spent some time testing Flashbots. It seems to work fine in almost all cases. Some notes:

- We use Flashbots when the checkbox is disabled. Steps to reproduce:
1. Send transaction using flashbots
2. Change the network and send the transaction from the new network.
3. Back to the Ethereum network and try to send transaction when the checkbox is disabled. 



https://github.com/tahowallet/extension/assets/23117945/975f9420-dcc6-428b-a296-b5112f146edd

- I can't do the swap using Flashbots. I have this problem also when using Uniswap dApp.
https://github.com/tahowallet/extension/assets/23117945/243b70d5-bad2-46d4-be87-1c1d10dda727



https://github.com/tahowallet/extension/assets/23117945/2fe172bc-2dd0-4224-bdee-ece399f34022

beemeeupnow:
> * I can't do the swap using Flashbots. All my transactions are dropped. I have this problem also when using Uniswap dApp.

I did not encounter the issue with Uniswap.

It seems the current build may not have a 0x API key so I could not test the built-in swap.

Shadowfiend:
> It seems the current build may not have a 0x API key so I could not test the built-in swap.

This changed this week and we haven't gotten a chance to add a dev key for the dev builds, sorry folks.

background/main.ts
```diff
@@ -1221,7 +1221,8 @@ export default class Main extends BaseService<never> {
           }
         }
 
-        const rejectAndClear = () => {
+        const rejectAndClear = async () => {
+          await this.autoToggleFlashbotsProvider()
```
Shadowfiend, line 1225: Should we be doing this in `handleAndClear` as well?

background/main.ts
```diff
@@ -1205,8 +1207,6 @@ export default class Main extends BaseService<never> {
             // eslint-disable-next-line @typescript-eslint/no-use-before-define
             rejectAndClear
           )
-
-          this.autoToggleFlashbotsProvider()
```
Shadowfiend: Not blocking: this whole thing makes me think `autoToggleFlashbotsProvider` behavior really shouldn't be in `main`. The flashbots stuff probably needs to be a first-class concept in the chain service itself, maybe as a generic “temporarily use this RPC for this transaction” thing. The RPC would be included as an optional parameter in the `broadcastSignedTransaction` call so that its usage is atomic rather than mutating global state that is subject to race conditions.

kkosiorowska (approved):
Looks like the issue has been solved :white_check_mark: Today I had no more problems with the swaps on Uniswap. But still a few of my transactions have been dropped, but I don't think this is due to these changes.

background/main.ts
kkosiorowska: Let's make it this way. Please create a task for the refactor.  We are ready for merge :rocket:

PR #3458, opened 2023-06-08
Fix issue for discovery transaction hash

kkosiorowska:
## What
Currently, we store the transaction hash for the first discovered transaction in the metadata for this asset. As a result, we are not able to relate the second account to another discovered transaction hash. We should track this for all addresses. In addition, when a user deletes an account, we should also update this in the asset state.

## To be determined

The current migration updates objects from the `customAssets` table. From this level, we are unable to relate the saved discoveryTxHash to the account from which it was added. As a result, we are unable to display the correct discovered transaction hash for the existing account.  This is not the best solution but I am not quite sure how to solve this issue. 

## UI

https://github.com/tahowallet/extension/assets/23117945/a8297672-6427-419b-b518-4397ac51681b

## Testing
- [x] Test the migration of data from the `main` branch. `discoveryTxHash` should be removed from custom assets.

Use the below addressees for the tests. Both addresses have the `Banana` token which should be added through transactions.
- `0xE49FC0028A9c5eeB0aA886B886168653439B928d`
- `0x52813a94a5D02291Eed4e22D8264464af95375D6`

- [x] Check the tx hash for added accounts. They should be different.
- [x] Remove the account and check the state of the db and redux. The asset should be updated. 

## Testing Env

```
SUPPORT_UNVERIFIED_ASSET=true
```

Latest build: [extension-builds-3458](https://github.com/tahowallet/extension/suites/13666652837/artifacts/754888933) (as of Fri, 16 Jun 2023 19:48:50 GMT).


background/services/indexing/index.ts
```diff
@@ -603,6 +603,27 @@ export default class IndexingService extends BaseService<Events> {
     this.emitter.emit("removeAssetData", asset)
   }
 
+  async removeDiscoveryTxHash(address: string): Promise<void> {
+    const assets = (await this.db.geAllCustomAssets()).filter(
+      (asset) =>
+        asset.metadata?.discoveryTxHash &&
+        asset.metadata?.discoveryTxHash[address]
```
jagodarybacka, line 610: 😜 
```suggestion
        asset.metadata?.discoveryTxHash?.[address]
```

jagodarybacka (commented):
Minor comments, will test it manually in a sec

jagodarybacka:
Just flagging - I'm wondering what is a desired behavior here - as "asset verified by you" is showing on all accounts that own a given token regardless if their transaction was the one that led to asset discovery or not. Most of these accounts don't have "Transaction hash" as we don't have the transaction that they got the asset with. 

That being said I think it is fine as it is and we shouldn't go into that rabbit hole 😉

background/assets.ts
```diff
@@ -51,9 +51,11 @@ export type AssetMetadata = {
  */
 export type NetworkSpecificAssetMetadata = AssetMetadata & {
   /**
-   * Set on assets discovered through transaction annotations
+   * Set on assets discovered through transaction annotations per address
    */
-  discoveryTxHash?: HexString
+  discoveryTxHash?: {
+    [address: HexString]: HexString
```
Shadowfiend, line 57: Scenarios I want to make sure we don't have:

(1)
- I have an unverified asset I see on Polygon.
- I see that the asset got picked up from transaction X, but transaction X was on BNB 🙅 

(2)
- I have an unverified asset I see on Polygon.
- I see that the asset got picked up from transaction X, and it was on Polygon.
- I verify it, and then switch to BNB.
- I see the same asset is verified, and when I click on the info, I see it was verified because of transaction X, which is not on BNB 🙅 

In the first path, we should only show the unverified asset as having been discovered on Polygon, and the transaction should always be from that network. In the second path, once we've verified the asset, we should only display the transaction hash in the slideover if it's from the same network.

So, I think this is fine, but we need to make sure that the `address` key here is *only* on the network's `homeNetwork` (and same for the tx hash). If the address or transaction are not from the same network as the asset's `homeNetwork`, I think we shouldn't include it. The other option is we *do* include it, but then we have to use `AddressOnNetwork` (maybe by converting this to a list of `[AddressOnNetwork, HexString]` pairs).

Alternatively, we can use `discoveryTxHash` for the home network, and include a list of other `AddressOnNetwork`/transaction hash pairs for other networks we've seen the asset on as a separate property.

I'm in favor of the *least complicated solution here* right now, which seems like it would be not tracking transaction hashes on networks other than the home network.

hyphenized: I feel like I'm missing something here but I believe neither of these should be possible because the asset is bound to a network?

Shadowfiend: Ah! Realizing Jagoda's note wasn't about different *networks* but about different *accounts* showing the asset. I'm fine with that, so it feels like we're probably good to merge here.

background/main.ts
```diff
@@ -648,6 +648,9 @@ export default class Main extends BaseService<never> {
     await this.signingService.removeAccount(address, signer.type)
 
     this.nameService.removeAccount(address)
+
+    // remove discovery tx hash for custom asset
+    this.indexingService.removeDiscoveryTxHash(address)
```
Shadowfiend, line 653: This is another one where we should really be responding to a signal from `SigningService` that the address is no longer being managed (or from the `ChainService` that it's no longer being tracked).

background/services/indexing/db.ts
```diff
@@ -242,6 +242,17 @@ export class IndexingDatabase extends Dexie {
...
         "&[contractAddress+homeNetwork.name],symbol,contractAddress,homeNetwork.family,homeNetwork.chainID,homeNetwork.name",
     })
+
+    // Fix issue for discovery transaction hash
+    this.version(6).upgrade((tx) => {
+      return tx
+        .table("customAssets")
+        .toCollection()
+        .modify((customAsset: CustomAsset) => {
+          // eslint-disable-next-line no-param-reassign
```
Shadowfiend, line 252: Let's make sure we drop a short explanation anytime we `eslint-disable`.

Shadowfiend (approved):
All righty going to take this fellow across the finish line, looks like all concerns are addressed other than some non-blocking stuff I dropped just now.

PR #3451, opened 2023-06-07
Cable Connection: Implement standard flow for dApp connect popup

Shadowfiend:
The dApp Connect popup now shows the top bar for switching
between Taho being default and not, and includes an explanatory
popover triggered by clicking on an `i` icon.

This PR does not yet implement the “first time” flow, where the
default wallet explanation appears without user interaction;
that is in #3464. It also does not implement automatically closing
the dApp connection popup and bringing up the next wallet if
default connection is turned off; that is in #3462.

## Testing

- [x] Try to connect to a dApp while Taho is marked as default.
      The Taho dApp connection flow should come up.
  - [x] There should be a top bar on the connection modal that
        matches the one in the wallet popup.
  - [x] When clicking the default toggle on/off, opening the wallet
        popup should show the same default state.
  - [x] Toggling default to off (or not Taho) and then closing the
        dApp Connection modal, then reloading the dApp and attempting
        to connect, should trigger the other wallet's dApp Connection
        flow.

Additionally, without updated dApp Connections enabled:

- [x] Try to connect to a dApp while Taho is marked as default.
      The Taho dApp connection flow should come up.
  - [x] There should NOT be a top bar on the connection modal that
        matches the one in the wallet popup.

## Testing Environment

```
ENABLE_UPDATED_DAPP_CONNECTIONS=true
```

Latest build: [extension-builds-3451](https://github.com/tahowallet/extension/suites/13578654320/artifacts/747855360) (as of Tue, 13 Jun 2023 18:45:28 GMT).

PR #3462, opened 2023-06-08
Repeater Connection: Forward requests to new default when default is switched off during dApp connection flow

Shadowfiend:
Adds the ability to pass through to the next provider as soon
as the user toggles default wallet off.

Can be reviewed and merged separately from #3451.

Ex:

https://github.com/tahowallet/extension/assets/8245/9fcb3c77-8242-4461-831b-7988243be4ed

## Testing

- [ ] Make sure Taho is set as default and connect to a dApp, waiting for the dApp Connection popup to appear.
- [ ] Switch the default wallet toggle to off in the top bar of the dApp Connection popup.
  - [ ] The dApp Connection popup should disappear, and the next wallet should pop up its connection flow immediately.

Note that you can get into a state where you've previously requested a MetaMask connection, but did not complete the flow (e.g. by not unlocking MetaMask). In this case MetaMask will auto-decline the connection request from the dApp, so the above flow will appear not to work. Make sure MetaMask is in an unlocked idle state before testing the flow above.

## Testing Env

```
ENABLE_UPDATED_DAPP_CONNECTIONS=true
```

Latest build: [extension-builds-3462](https://github.com/tahowallet/extension/suites/13633228421/artifacts/752242181) (as of Thu, 15 Jun 2023 15:12:02 GMT).

### DIFF
--- a/ui/pages/DAppConnect/DAppConnectPage.tsx
+++ b/ui/pages/DAppConnect/DAppConnectPage.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from "react"
+import React, { ReactElement, useEffect } from "react"
 import { AccountTotal } from "@tallyho/tally-background/redux-slices/selectors"
 import { PermissionRequest } from "@tallyho/provider-bridge-shared"
 import { useTranslation } from "react-i18next"
@@ -7,10 +7,12 @@ import {
   isDisabled,
   isEnabled,
 } from "@tallyho/tally-background/features"
+import { selectDefaultWallet } from "@tallyho/tally-background/redux-slices/ui"
 import SharedButton from "../../components/Shared/SharedButton"
 import SharedAccountItemSummary from "../../components/Shared/SharedAccountItemSummary"
 import RequestingDAppBlock from "./RequestingDApp"
 import SwitchWallet from "./SwitchWallet"
+import { useBackgroundSelector } from "../../hooks"
 
 type DAppConnectPageProps = {
   permission: PermissionRequest
@@ -29,6 +31,19 @@ export default function DAppConnectPage({
 }: DAppConnectPageProps): ReactElement {
   const { t } = useTranslation("translation", { keyPrefix: "dappConnection" })
   const { title, origin, faviconUrl, accountAddress } = permission
+
+  const isDefaultWallet = useBackgroundSelector(selectDefaultWallet)
+
+  useEffect(() => {
+    if (
+      isEnabled(FeatureFlags.ENABLE_UPDATED_DAPP_CONNECTIONS) &&
+      !isDefaultWallet
+    ) {
+      // When default wallet is toggled off, wait for the toggle animation to
+      // complete and close the window out.
+      setTimeout(() => window.close(), 300)
+    }
+  }, [isDefaultWallet])
 
   return (
     <>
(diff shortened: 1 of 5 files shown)


window-provider/index.ts
```diff
@@ -165,6 +169,34 @@ export default class TallyWindowProvider extends EventEmitter {
...
+            // received.
+            .sort(([id], [id2]) => Number(BigInt(id2) - BigInt(id)))
+            .forEach(([id, { sendData, reject, resolve }]) => {
+              window.walletRouter
+                ?.routeToNewDefault(sendData.request)
+                // On success or error, call the original reject/resolve
+                // functions to notify the requestor of the new wallet's
+                // response.
+                .then((...a) => resolve(...a))
+                .catch((...b) => reject(...b))
```
hyphenized, line 194: Minor nit here that this is the same as `.then(resolve, reject)`

window-provider/index.ts
```diff
@@ -165,6 +169,34 @@ export default class TallyWindowProvider extends EventEmitter {
...
+
+              this.requestResolvers.delete(id)
```
hyphenized, line 196: I think it's safe to remove this since we've already cleared the map

ui/pages/DAppConnect/DAppConnectPage.tsx
```diff
@@ -30,6 +32,19 @@ export default function DAppConnectPage({
...
+  const isDefaultWallet = useBackgroundSelector(selectDefaultWallet)
+
+  useEffect(() => {
+    if (
+      isEnabled(FeatureFlags.ENABLE_UPDATED_DAPP_CONNECTIONS) &&
+      !isDefaultWallet
+    ) {
+      // When default wallet is toggled off, wait for the toggle animation to
+      // complete and close the window out.
+      setTimeout(() => window.close(), 300)
```
hyphenized, line 44: I think this is causing an issue when Taho is the only wallet available. On https://pancakeswap.finance/ when Taho is not the default, connecting with the injected option causes the popup to close immediately.

hyphenized (changes requested):
Looks great! Found one bug and left a couple of comments.

window-provider/index.ts
Shadowfiend: I find this clearer (except probably for using `.then(resolve).catch(reject)` instead of spreading), tbh, but don't have a particularly strong opinion. If you feel strongly I'm happy to change it.

ui/pages/DAppConnect/DAppConnectPage.tsx
Shadowfiend: Incidentally this shouldn't trigger in that scenario—it should really only trigger when the user actively toggles it off, i.e. on transition. Let me noodle on this a little.

If this isn't causing that behavior with the feature flag off, let's not let it block the merge since there's a follow-up in the mix.

window-provider/index.ts
Shadowfiend: https://github.com/tahowallet/extension/pull/3462/commits/117fed883489277e9a6ad93584475e3b99515aa1

window-provider/index.ts
Shadowfiend: Switched to `.then(resolve).catch(reject)` (+ prettier) in https://github.com/tahowallet/extension/pull/3462/commits/117fed883489277e9a6ad93584475e3b99515aa1 .

hyphenized (approved):
🚢

PR #3463, opened 2023-06-09
Fix broken E2E test removing wallet

michalinacienciala:
After changes made in https://github.com/tahowallet/extension/pull/3447 the E2E tests started to fail on `await page.locator(".profile_button").nth(1).click()` in `e2e-tests/accounts.spec.ts` file. The element could not be found. It looks that previously some DOM elements were duplicated and we had two elements with class name `profile_button`. Now we have only one, so we can remove the `.nth(1)` locator from the step.


jagodarybacka (approved):
Looks good ✅

Shadowfiend:
Excited that the e2e tests picked this up, and that I caused it 🙃

Shadowfiend:
(And thankful you found it!)

PR #3465, opened 2023-06-12
Update Ledger connection status icons

### LINKED ISSUE #1066: Ledger - Sign-tx screens missing status graphics
Adding this as a UI fix and not a bug. As i realise it might not be a quick and easy task. It is also not blocking.

image

hyphenized:
Closes #1066

Updates and adds ledger connection icons for disconnected / error / unknown states. Assets for the ledger background were not reused as to avoid positioning issues around the different icons used for each state.

## To Test
- [x] Connect a ledger account, onboarding should display the same blank state ledger icon
- [x] Disable blind signing, attempt to sign a transaction, you should see the new "check error" icon / CTA
- [x] Disconnect the ledger, ledger icon should display new disconnected state

Latest build: [extension-builds-3465](https://github.com/tahowallet/extension/suites/13854686465/artifacts/769693087) (as of Mon, 26 Jun 2023 05:52:54 GMT).


Shadowfiend:
Heads up that I think the disconnected icon *also* needs to be updated to look like https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?type=design&node-id=6062-145221&t=aadjjnIsgiBaqRVa-4 .

ui/public/images/connect_ledger_indicator_check.svg
Shadowfiend, line 1: I think this and the Disconnected bit need to be sans text, so we can i18n the text (e.g. using https://css-tricks.com/fitting-text-to-a-container/#aa-magic-number-it-with-viewport-units or https://css-tricks.com/fitting-text-to-a-container/#aa-just-use-svg). Ledger is a brand name so we don't need to worry about it, but for these I think we have to be able to translate the text itself.

kkosiorowska (commented):
> Disable blind signing, attempt to sign a transaction, you should see the new "check error" icon / CTA

I disabled blind signing but I don't see new icons.

Screenshot 2023-06-13 at 15 53 26

Screenshot 2023-06-13 at 15 59 43

 > Disconnect the ledger, ledger icon should display new disconnected state

In this case, I don't see the icon too.

Screenshot 2023-06-13 at 16 01 40

hyphenized:
@kkosiorowska Fixed, was able to replicate this by attempting to swap but the icon appeared extremely small on my end. It's related to svgs shrinking on flexbox

kkosiorowska (commented):
It seems to be working well 🚀 However, I have one comment. The icons are a bit different from those in the Figma. What I mean is that the circle with the icon is bigger than in the Figma.  @VladUXUI  Is this okay?

- [ ] However, this should be improved. I'm talking about the red circle.
Screenshot 2023-06-15 at 09 50 52

Screenshot 2023-06-15 at 09 52 16

- Connected
Screenshot 2023-06-15 at 09 55 27

Screenshot 2023-06-15 at 09 56 42

- Check error
Screenshot 2023-06-15 at 09 39 05

Screenshot 2023-06-15 at 09 56 47

Shadowfiend:
Why did we diverge from the design here? I haven't dug deep but I'm not sure what kind of limitation would prevent us from implementing the design as-is.

VladUXUI:
This is a mistake indeed, but i think it's my fault. I might have sent wrong SVG to Jorge

Shadowfiend (approved):
Just eyeballed it all, looks effin' gorgeous. Amazing, thank you ❤️ 

Feels like we can do a little better with the icon during onboarding as well, but we can leave that for another time.

PR #3472, opened 2023-06-13
Create E2E tests for verified/unverified tokens

michalinacienciala:
This commit introduces E2E test that checks the functiolality of verified/unverified tokens. Following general steps are executed:
1. Import account
2. Enable `Show unverified assets`
3. Hide asset
4. Trust asset
5. Hide trusted asset
A number of checks is performed during each step.

The commit also introduces helper functions and adds `data-testid` attribute to a couple of DOM elements.

TODO:

- [x] As some of the code is similar or identical as in the https://github.com/tahowallet/extension/pull/3418 PR which is yet not merged to `main`, some conflicts may arise and will need to be resolved before this change lands on `main`.
- [x] Also some changes will need to be made once
https://github.com/tahowallet/extension/pull/3195 gets merged to `main`, as this PR solves a bug causing failures in the tests (the failing part of the test was temporarily commented out).
- [x] Add `E2E_TEST_ONLY_WALLET_JSON_BODY` and `E2E_TEST_ONLY_WALLET_JSON_PASSWORD` secrets in GitHub's settings.
- [x] Wait for https://github.com/tahowallet/extension/pull/3559 to be merged and merge main to feature branch (should fix the failing `e2e-tests` job)
- [x] Handle TODOs in the code

Latest build: [extension-builds-3472](https://github.com/tahowallet/extension/suites/15116290312/artifacts/863816729) (as of Tue, 15 Aug 2023 18:12:30 GMT).


michalinacienciala:
The tests were passing locally, but we have a failing check in CI. I'll investigate the failures.

e2e-tests/utils/assets.ts
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    await verifyAssetPopup
+      .locator("li")
+      .filter({
+        hasText: "Contract address",
+      })
+      .locator(".right")
+      .click({ trial: true })
+    // TODO: Click and verify the scan website address
```
michalinacienciala, line 254: I tried implement that, but had some troubles. We may consider merging this PR without it and I'll work on this in a separate PR (as my next task).

e2e-tests/utils/walletPageHelper.ts
```diff
@@ -167,7 +170,19 @@ export default class WalletPageHelper {
...
 
-  async verifyDefaultWalletBanner(): Promise<void> {
+  async waitForAssetsToLoad(timeout?: number): Promise<void> {
+    await expect(this.popup.getByText("Digging deeper")).toHaveCount(0, {
+      timeout: timeout ?? 120000,
+    })
+    await expect(this.popup.locator(".spinner")).toHaveCount(0, {
+      timeout: timeout ?? 120000,
+    })
+    // await expect(this.popup.locator(".token_icon_fallback")).toHaveCount(0, {
```
jagodarybacka, line 180: Can we delete it?

e2e-tests/utils/walletPageHelper.ts
```diff
@@ -185,6 +200,86 @@ export default class WalletPageHelper {
...
+    const numberOfAccounts = await this.popup
+      .getByTestId("wallet_address_item")
+      .count()
+    const accountLabelButton = this.popup
+      .getByTestId("wallet_title")
+      .filter({
+        hasText: accountLabel,
+      })
+      .getByRole("button")
+    await this.popup.getByTestId("wallet_title").click({ trial: true }) // TODO: why trial?
```
jagodarybacka, line 217: what about this comment?

e2e-tests/utils/walletPageHelper.ts
```diff
@@ -185,6 +200,86 @@ export default class WalletPageHelper {
...
+
+  /**
+   * Function closing the Accounts popup.
+   */
+  async closeAccountsPopup(): Promise<void> {
+    const accountsPopup = await this.popup.$(".slide_up_menu:not(.closed)")
+    if (accountsPopup) {
+      const closeButton = await accountsPopup.$(
+        'button[aria-label="Close menu"]'
+      )
```
jagodarybacka, line 272: don't we want to add proper test ids?

e2e-tests/utils/assets.ts
```diff
@@ -0,0 +1,468 @@
...
+    const balance = await activityLeftContainer.getByText(
+      /^(\d|,)+(\.\d{2,4})*$/
+    )
+    await expect(balance.getByText(expectedBalance)).toHaveCount(1, {
+      timeout: 120000,
+    })
+
+    if (assetType === "base" || assetType === "knownERC20") {
+      await expect(
+        activityLeftContainer.getByText(/^\$(\d|,)+\.\d{2}$/)
```
jagodarybacka, line 88: These regex should be named variables

e2e-tests/utils/assets.ts
```diff
@@ -0,0 +1,468 @@
...
+    // human eye.
+    // await expect(
+    //   this.popup.getByRole("button", { name: "Verify asset" }).first()
+    // ).toBeHidden()
+    // await expect(this.popup.getByText("Asset not verified")).not.toBeVisible()
+    // await expect(
+    //   this.popup.getByText(
+    //     "Be aware of scam and spam assets, only interact with assets you trust."
+    //   )
+    // ).not.toBeVisible()
```
jagodarybacka, line 325: can we add some html attribute or check it by CSS class to make it easier to test?

jagodarybacka (commented):
A bunch of possible improvements 🏗️

e2e-tests/utils/walletPageHelper.ts
michalinacienciala: I couldn't do it simpler than this (maybe there's something I didn't think of?).

The problem is that the `slide_up_menu` DOM element for Accounts contains a bunch of identical `slide_up_menu` child elements, all containing their own Close button. All Close buttons except of one are not visible for users. On the DOM side, the difference is that those invisible buttons are wrapped with the `div` element of `slide_up_menu closed` class, whereas the visible button is wrapped by `slide_up_menu with_overflow` `div`.
Adding a test id to the `button` or the `div` element does not solve the issue - the id is added to all the elements and we still have the problem with identifying the one we want to click.

As a workaround we use [ElementHandle](https://playwright.dev/docs/api/class-elementhandle) to locate the right item to click, by specifying that the element shouldn't be wrapped by the DOM item of the `closed` class (I couldn't do that using classic Locator syntax).

And now, going back to your suggestion - switching to use of the test ids everywhere in the function would not work from the above mentioned reasons. And sticking with use of the ElemenHandle to locate the right `slide_up_menu` and using the test id to locate the Close menu would also not work, because the `getByTestId` cannot be called on the ElementHandle objects.


We could add a `data-testid="slide_up_menu_close"` line after [this one](https://github.com/tahowallet/extension/blob/80c90e062169b08051833f39965bd4474cb6f449/ui/components/Shared/SharedSlideUpMenu.tsx#L88C14-L88C14) and then try to do
```
this.popup.getByTestId("slide_up_menu_close").click()
```
in the `closeAccountsPopup()` function, but this does not work. It finds multiple identical matching DOM elements. The only significant difference between the matched elements is that the parent element of the one we want to click has class `slide_up_menu with_overflow`

Anyway, as all of this is not straightforward, I've added a comment to the `closeAccountsPopup()` function, that explains why we use ElementHandle there: b4a61d2651b58ce4fe5449aae16798f8803c9eeb.

e2e-tests/utils/walletPageHelper.ts
michalinacienciala: Oops, this step and a couple of others nearby were not supposed to land in the final code. Removed in 4e740dfb74bf890032a65b859d9cd4df8457e2af.

e2e-tests/utils/walletPageHelper.ts
```diff
@@ -185,6 +200,86 @@ export default class WalletPageHelper {
     ).toHaveText(network)
   }
 
+  /**
+   * Function adding new address to an already imported account
+   */
+  async addAddressToAccount(accountLabel: string): Promise<void> {
```
michalinacienciala, line 203: BTW, due to the fact that during work on this PR  I switched from importing the tests account via recovery phrase to import via JSON, I no longer use this function (and `closeAccountsPopup` as well). However, those function may come handy some day. What is the best practice here? Leave it as it is or maybe comment out or remove completely?

e2e-tests/utils/walletPageHelper.ts
michalinacienciala: Removed in e9d795129ad2b12dc2b986bad459d0d91b245311.

e2e-tests/utils/onboarding.ts
```diff
@@ -26,6 +26,10 @@ const DEFAULT_PASSWORD = "12345678"
 // It owns some NFTs/badges.
 export const account1Address = "0x6f1b1f1feb01235e15a7962f16c389c7f8218ed6"
 export const account1Name = /^e2e\.testertesting\.eth$/
+// The account2 is the he testertesting.eth account. It's used for manual
+// testing, so it's balance may fluctuate. It can be used to test features that
+// don't depend on the constant balance or state of the assets.
+export const account2Name = /^testertesting\.eth$/
```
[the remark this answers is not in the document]
michalinacienciala: Addressed in 95db29f06e775a98d4a7d0711b3111606c6e0fe2.

e2e-tests/regular/token-trust.spec.ts
```diff
@@ -0,0 +1,555 @@
...
+      if (jsonPassword) {
+        await walletPageHelper.onboardWithJSON(
+          "./e2e-tests/utils/JSON.json",
+          jsonPassword
+        )
+      } else {
+        throw new Error(
+          "E2E_TEST_ONLY_WALLET_JSON_PASSWORD environment variable is not defined."
+        )
+      }
```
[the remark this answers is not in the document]
michalinacienciala: Hmmm, but will that make much difference? When calling the method, we will still need to provide the environment variables as arguments. And this means that we'll still need to have the `if`/`else` conditions present in the `token-trust.spec.ts`, right? Or maybe there's something I'm missing here?

ui/_locales/en/messages.json
```diff
@@ -663,7 +663,7 @@
       }
     },
     "unverifiedAssets": {
-      "tooltip": "Discover assets that you own but are not on our asset list. Some spam/scam assets will show up, so tread carefully.<br/><br/>These will show up on the bottom of the asset page until you verify them."
+      "tooltip": "Discover assets that you own but are not on our asset list. Some spam/scam assets may show up, so treat them with caution.<br/><br/>They will show up on the bottom of the asset page until you verify them."
```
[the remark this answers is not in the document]
michalinacienciala: No problem  :)

e2e-tests/regular/token-trust.spec.ts
```diff
@@ -0,0 +1,555 @@
...
+      /**
+       * Ensure there are no unverified assets on the main page.
+       */
+      await assetsHelper.assertNoUnverifiedAssetsOnWalletPage()
+
+      // In order for the next few tests to make sense the wallet address should
+      // have a positive balance of at least one of the listed assets. At the
+      // moment of writing the test, the wallet had positive balance for all
+      // those assets.
+      const untrustedAssets = [
```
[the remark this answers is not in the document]
michalinacienciala: See 89019b4e156ccc33e6874001d380ee834ea139d1.

e2e-tests/fork-based/transactions.spec.ts
```diff
@@ -143,12 +145,12 @@ test.describe("Transactions", () => {
         /**
          * Verify elements on the asset activity screen
          */
-        await transactionsHelper.verifyAssetActivityScreen(
+        await assetsHelper.assertAssetDetailsPage(
           /^Ethereum$/,
-          /^testertesting\.eth$/,
+          account2Name,
```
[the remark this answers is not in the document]
michalinacienciala: > can we use a variable for testertesting.eth string as well?

Did you mean this comment for line 115 (and other similar)? If so, I've applied that in 95db29f06e775a98d4a7d0711b3111606c6e0fe2.

> Also other strings like Ethereum etc - so we can avoid typos or other easy bugs

and

> Network's strings should be reusable variables, same with most common assets like /^MATIC$/, /^WMATIC$/ etc

I feel a bit torn about this...

The good side of moving commonly used labels to some external place is what you said and also possibility to apply changes to the labels quickly, without having to modify them in many places.

But there are also some cons, mainly in cases where we also have the same labels abstracted in the app's code (as this is the case for network names and base assets labels stored in `network.ts`). What I mean is that having similar implementation in tests and in tested code may lead to introducing same bug in both places and not catching it by the tests.

Another issue is that in tests we sometimes want to use RegExp, sometimes string, sometimes value with all letters capitalized, sometimes only with the first letter captalized etc, depending on the context. We'll either need a separate variable for each different version of the label we use or we'll need to use some method that modifies given variable into the formatting we need.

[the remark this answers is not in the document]
michalinacienciala: Not saying definitely NO to this idea though. The pros may outweigh the cons...
What would be the best way to approach this? Having all the constant values used by the E2E tests stored in one helper file? Or storing them in files they are related to (something like what we have currently, where names of testing accounts are in the onboarding helper)? The first solution seems more universal to me, but I'm curious if there some good practice I should follow here?

e2e-tests/utils/assets.ts
michalinacienciala: According to Playwright documentation
> Element is considered visible when it has non-empty bounding box and does not have visibility:hidden computed style. Note that elements of zero size or with display:none are not considered visible.

One thing that we could do is to do something to make the elements' style be computed to `visibility:hidden`. I'm not sure though how easily can we do this...

Another approach would be to use ElemenHandle syntax to check that elements we want to make sure are invisible are children of the `div` element with the `hidden_assets` class, but no `visible` class. But this seems to me unnecessarily complex, given that checking
```
await expect(
      this.popup.getByTestId("hidden_assets_container")
    ).not.toBeVisible()
```
already tells us that the section with unverified assets does not have the `visible` class.

e2e-tests/regular/token-trust.spec.ts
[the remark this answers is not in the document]
hyphenized: If we move these to a helper we can reduce duplication. Since we're doing very similar things with `TEST_WALLET_JSON_BODY` and `E2E_TEST_ONLY_WALLET_JSON_BODY`, we could put both of these with the corresponding env variables inside a function (if not `onboardWithJSON`)

e2e-tests/fork-based/transactions.spec.ts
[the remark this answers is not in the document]
hyphenized: > What I mean is that having similar implementation in tests and in tested code may lead to introducing same bug in both places and not catching it by the tests

> Another issue is that in tests we sometimes want to use RegExp, sometimes string, sometimes value with all letters capitalized, sometimes only with the first letter captalized etc, depending on the context. We'll either need a separate variable for each different version of the label we use or we'll need to use some method that modifies given variable into the formatting we need.

Both are great points, I'm not entirely onboard with abstracting simple names as it reduces readability on written test cases, which should remain as simple as possible and, are also unlikely to change unless there are changes in the implementation. 

Still, I don't think there's much harm if it's a simple string that's repeated multiple times and it's a simple change. Considering that does not seem to be the case here given the multiple variants, I'd keep the hardcoded strings.

e2e-tests/utils/assets.ts
```diff
@@ -0,0 +1,468 @@
...
+    await this.popup.waitForTimeout(2000) // We need to wait for the background elements to disappear
+    await expect(asset.getByText(assetSymbol)).toBeVisible()
+    /**
+     * Make sure the asset is not listed among unverified assets.
+     */
+    await expect(
+      this.popup
+        .getByTestId("hidden_assets_container")
+        .filter({ has: asset.getByText(assetSymbol) })
+    ).not.toBeVisible()
```
hyphenized, line 38: I think this won't work if we have some spam asset with the same name as the one we just verified

e2e-tests/utils/assets.ts
```diff
@@ -0,0 +1,468 @@
...
+    let asset: ReturnType<typeof this.popup.locator> | undefined
+    if (assetType === "base") {
+      asset = this.popup.getByTestId("asset_list_item").first() // We use `.first()` because the base asset should be first on the list
+    } else {
+      asset = this.popup.getByTestId("asset_list_item").filter({
+        has: this.popup.locator("span").filter({ hasText: assetSymbol }),
+      })
+    }
+    await this.popup.waitForTimeout(2000) // We need to wait for the background elements to disappear
+    await expect(asset.getByText(assetSymbol)).toBeVisible()
```
hyphenized, line 30: We can set the timeout on the assertion itself too (unless for some reason it does not work)
```suggestion
    await this.popup.waitForTimeout(2000) // We need to wait for the background elements to disappear
    await expect(asset.getByText(assetSymbol)).toBeVisible({ timeout: 2000 })
```

e2e-tests/utils/walletPageHelper.ts
hyphenized: We can add a test id here like `accounts_list_slide_up_menu` https://github.com/tahowallet/extension/blob/085b99ca2bcf9f47fd1705b17ee8444e5aeb948d/ui/components/TopMenu/TopMenu.tsx#L127-L133

Then use playwright assertions so we don't have to throw up errors if the locator doesn't return anything

```ts
const accountsSlideUp = this.popup.getByTestId(
      "accounts_list_slide_up_menu"
    )

    await expect(accountsSlideUp).toBeVisible()
```

This slide up is indeed trickier because there are nested slide ups so to get the right close button we can just ask playwright to return the first one it finds, we'll get the right one because we start looking from the right slide up menu.

```ts
    await accountsSlideUp
      .getByRole("button", { name: "Close menu" })
      .first()
      .click()
```

This way playwright throws if it can't find the element to click so we can skip the manual `if` check as well.

e2e-tests/utils/walletPageHelper.ts
hyphenized: This one looks useful and generic enough that we could benefit from keeping it but I'd suggest that we move here the code from `closeAccountsPopup` and separate it later if we need the logic somewhere else.

hyphenized (commented):
Looks good overall :muscle:! Left a few comments

Shadowfiend:
@hyphenized do you consider any of your notes above blockers?

hyphenized:
> @hyphenized do you consider any of your notes above blockers?

I don't consider these as blockers for merging, but I do see these as concerns regarding code quality and potential future technical debt.

Shadowfiend:
Cool, since Michalina is back on Monday we'll keep this fellow open for visibility and merge after she's had a look.

e2e-tests/utils/walletPageHelper.ts
michalinacienciala: It looks to me that this the test id set in this place does not get reflected in the final DOM. I've set this in the `extension/ui/components/TopMenu/TopMenu.tsx`:
```ts
      <SharedSlideUpMenu
        isOpen={isNotificationsOpen}
        allowOverflow
        close={() => {
          setIsNotificationsOpen(false)
        }}
        data-testid="accounts_list_slide_up_menu"
      >
```
Then I rebuilt the code and run a test that does this:
```ts
      await popup.getByTestId("top_menu_profile_button").last().click()
      await walletPageHelper.closeAccountsSlideUp()
```
where `closeAccountsSlideUp()` looks like this:

```ts
  async closeAccountsSlideUp(): Promise<void> {
    const accountsSlideUp = await this.popup.getByTestId(
      "accounts_list_slide_up_menu"
    )
    await expect(accountsSlideUp).toBeVisible()
    await accountsSlideUp
      .getByRole("button", { name: "Close menu" })
      .first()
      .click()
```

I've got a failure of the `await expect(accountsSlideUp).toBeVisible()` assertion. I checked the DOM of the app and there was no element with the `accounts_list_slide_up_menu` test id. The Accounts slide up had the `slide_up_menu` test id, which is set in

michalinacienciala: It looks to me that this the test id set in this place does not get reflected in the final DOM. I've set this in the `extension/ui/components/TopMenu/TopMenu.tsx`:
```ts
      <SharedSlideUpMenu
        isOpen={isNotificationsOpen}
        allowOverflow
        close={() => {
          setIsNotificationsOpen(false)
        }}
        data-testid="accounts_list_slide_up_menu"
      >
```
Then I rebuilt the code and run a test that does this:
```ts
      await popup.getByTestId("top_menu_profile_button").last().click()
      await walletPageHelper.closeAccountsSlideUp()
```
where `closeAccountsSlideUp()` looks like this:

```ts
  async closeAccountsSlideUp(): Promise<void> {
    const accountsSlideUp = await this.popup.getByTestId(
      "accounts_list_slide_up_menu"
    )
    await expect(accountsSlideUp).toBeVisible()
    await accountsSlideUp
      .getByRole("button", { name: "Close menu" })
      .first()
      .click()
```

I've got a failure of the `await expect(accountsSlideUp).toBeVisible()` assertion. I checked the DOM of the app and there was no element with the `accounts_list_slide_up_menu` test id. The Accounts slide up had the `slide_up_menu` test id, which is set in
https://github.com/tahowallet/extension/blob/6b8e4b0d77c1d65451738d8922abb2377253a933/ui/components/Shared/SharedSlideUpMenu.tsx#L51.
Maybe we can't overwrite one test id with another?

michalinacienciala: This works:
```ts
  async closeAccountsSlideUp(): Promise<void> {
    const accountsSlideUp = await this.popup.getByTestId("slide_up_menu").nth(2)
    await expect(accountsSlideUp).toBeVisible()
    await accountsSlideUp
      .getByRole("button", { name: "Close menu" })
      .first()
      .click()
  }
```
However I'm not a fan of using `.nth(2)` to locate the right slide up menu.
Playing a bit more with this, I was able to simplify the code by using filtering:
```ts
  async closeAccountsSlideUp(): Promise<void> {
    const accountsSlideUp = await this.popup
      .getByTestId("slide_up_menu")
      .filter({ hasText: "Accounts" })
    await expect(accountsSlideUp).toBeVisible()
    await accountsSlideUp
      .getByRole("button", { name: "Close menu" })
      .first()
      .click()
  }
```

hyphenized: > It looks to me that this the test id set in this place does not get reflected in the final DOM. I've set this in the `extension/ui/components/TopMenu/TopMenu.tsx`:
> ```ts
>       <SharedSlideUpMenu
>         isOpen={isNotificationsOpen}
>         allowOverflow
>         close={() => {
>           setIsNotificationsOpen(false)
>         }}
>         data-testid="accounts_list_slide_up_menu"
>       >
> ```

On this code, it seems we passed a `data-testid` prop instead of `testid` to `SharedSlideUpMenu`. If that's the case, it would explain why the test id was not set correctly on the element.

e2e-tests/utils/assets.ts
michalinacienciala: It seems that some of the recent changes made the timeout completely obsolete. I removed it in c21299e8369701e116968a3b17fc57d4ab626e62.

e2e-tests/regular/token-trust.spec.ts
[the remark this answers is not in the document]
michalinacienciala: Is something like this what you had in mind: 5f45f7f55769a8ae872d6a228b1f6493eafe3494?

e2e-tests/utils/walletPageHelper.ts
michalinacienciala: Implemented in c562952f4480cac6be22bdb14dad4fb1983d39e7.

e2e-tests/utils/walletPageHelper.ts
michalinacienciala: See 1cf604ec8d48f3a441befe776bc60154a41fdfe8.

e2e-tests/utils/assets.ts
michalinacienciala: I've refactored the `assertVerifiedAssetOnWalletPage()` method to handle such case. See 1265763b9b710dea43c81a39f8f49ec6fa87e20d.

hyphenized (approved):
Comments addressed

PR #3477, opened 2023-06-14
Allow setting custom auto-lock timer

### LINKED ISSUE #3471: Custom auto lock
https://www.figma.com/file/q8gfu9XTbYBF07KMnDCtsD/UI---Dark?type=design&node-id=19917-393091&t=fyoGFeYaBK5ATloJ-4

This config value will replace the constants `MAX_KEYRING_IDLE_TIME` and `MAX_OUTSIDE_IDLE_TIME`:

https://github.com/tahowallet/extension/blob/e3cad978d45a3528d3df9e87bb772810486c003d/background/services/keyring/index.ts#L25-L26

For now, the two should end up with the same value.

It would make sense for `KeyringService` to gain a dependency on `PreferencesService` and update its internal lock time based on configuration adjustments via event emissions. When that setting is updated, it would probably also make sense to call `autolockIfNeeded`, so that if the user drops the max idle time to an amount that is < the current idle time, the keyring locks immediately.

hyphenized:
Closes #3471

Adds a dropdown on settings to allow changing the maximum amount of time the wallet can remain idle before locking signing. This new setting is updated through an async thunk which triggers an event to also handle the change on the keyring service. The data is persisted on preferences db and exposed through redux to the UI.

## Testing Env

```
SUPPORT_CUSTOM_AUTOLOCK=true
```

## To Test
- [x] Import an account
- [x] Lock and unlock your wallet
- [x] Head to settings, change auto-lock time
- [x] Change your system's date or change the options [here](https://github.com/tahowallet/extension/blob/be6fae78bd7949cd5e4d8f37c882152d00e8141d/ui/pages/Settings.tsx#L308), test that different time options work correctly
- [x] If the wallet has been idle for 10 minutes and the current auto lock setting changes to from 15 minutes to 5, the wallet should lock itself.

Latest build: [extension-builds-3477](https://github.com/tahowallet/extension/suites/13795656037/artifacts/764974743) (as of Thu, 22 Jun 2023 16:11:15 GMT).


background/redux-slices/ui.ts
```diff
@@ -267,6 +279,20 @@ export const addNetworkUserResponse = createBackgroundAsyncThunk(
   }
 )
 
+export const updateAutoLockTimer = createBackgroundAsyncThunk(
+  "ui/updateAutoLockTimer",
+  // TODO: Switch to fixed set of values?
```
jagodarybacka, line 284: I think it is fine, we have the "switch" implemented on the UI side anyway

jagodarybacka (commented):
Haven't tested yet, some quick comments

background/services/keyring/index.ts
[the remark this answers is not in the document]
Shadowfiend: Yes let’s use that as base please. Should have said that earlier and not referenced main in the issue, sorry folks 😕

background/redux-slices/ui.ts
Shadowfiend: Correct, I think the view is the only place we should keep these fixed, the rest of the code should be flexible. We can drop the TODO.

background/services/keyring/index.ts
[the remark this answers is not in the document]
hyphenized: All good diff isn't that big

jagodarybacka:
Hmm, Github isn't running checks for PRs to non-main branches, there is a problem with nonexisting variable reference, I'll fix it quickly

jagodarybacka:
if we are dropping a flag for private keys then I think we can release this thing in the same release as well and we can omit the feature flag here 🤔

jagodarybacka (approved):
Updated with the newest main changes so we can QA it here. Overall looks wonderful! I think we can drop the flag 🏳️

background/services/preferences/defaults.ts
```diff
@@ -1,7 +1,9 @@
-import { ETHEREUM, USD } from "../../constants"
+import { ETHEREUM, MINUTE, USD } from "../../constants"
 import { storageGatewayURL } from "../../lib/storage-gateway"
 import { Preferences } from "./types"
 
+export const DEFAULT_AUTOLOCK_TIMER = 60 * MINUTE
```
Shadowfiend, line 5: Let's use the term `interval` or `time` (as the previous variables used), a timer is a thing that keeps time, whereas this is just the interval we want to use for our timer.

background/services/internal-signer/index.ts
```diff
@@ -170,12 +169,17 @@ export default class InternalSignerService extends BaseService<Events> {
    */
   lastOutsideActivity: UNIXTime | undefined
 
-  static create: ServiceCreatorFunction<Events, InternalSignerService, []> =
-    async () => {
-      return new this()
-    }
+  #internalAutoLockTimer: UNIXTime = DEFAULT_AUTOLOCK_TIMER
```
Shadowfiend, line 172: This isn't a timestamp, it's an interval, so we can let the type be inferred.

ui/pages/Settings.tsx
```diff
@@ -302,6 +318,63 @@ export default function Settings(): ReactElement {
     ),
   }
 
+  const autoLockOptions = [
+    { label: "5", value: String(5 * MINUTE) },
+    { label: "15", value: String(15 * MINUTE) },
+    { label: "30", value: String(30 * MINUTE) },
+    { label: "60", value: String(60 * MINUTE) },
+  ]
```
Shadowfiend, line 326: Let's lift this to the top level and ALL_CAPS it as we've done with other constants we're referencing that don't rely on scope.

ui/pages/Settings.tsx
```diff
@@ -324,7 +397,7 @@ export default function Settings(): ReactElement {
     ),
   }
 
-  const settings = Object.values({
+  const settings: SettingsList = {
```
Shadowfiend, line 400: Any reason this isn't just an array? Doesn't look like we're using the key for anything.

It actually feels like we could straightforwardly use the key for i18n purposes, iterating `Object.entries` and skipping the `title` field, so we have a `{ [titleKey: string]: SettingsItem[] }`.

Not blocking ofc.

Shadowfiend (changes requested):
Left a couple of notes around naming, typing, and data flow.

ui/pages/Settings.tsx
hyphenized: No reason but also didn't consider it blocking. We could use `Object.entries` but we'd have to do some casting since translations are typed and expect a valid dictionary key.

jagodarybacka: This is already changed here and on [Flashbots PR](https://github.com/tahowallet/extension/blob/39d1c9b3bf7b0479ce1d6652f8e4ff6642043fea/ui/pages/Settings.tsx) as it was causing conflicts, I think we should unify it separately from both PRs to make them all compatible

Shadowfiend: > We could use Object.entries but we'd have to do some casting since translations are typed and expect a valid dictionary key.

I hate this about `Object.entries` 🙃

jagodarybacka:
@hyphenized could you resolve conflicts? I can test it one more time and I think it will be good to go

jagodarybacka:
I realized we are not mentioning what unit we are using here, pretty obvious that these are minutes but maybe not 🤔 cc @VladUXUI 
image

jagodarybacka (approved):
Works fine ✅

Issue #3325, opened 2023-04-27
Add custom asset copy tweak


mr-michael:
Let's change to:

✅ **Asset already added.**
If you're not seeing it, that means your balance is 0. Have a positive balance but still not seeing it? Make sure you've disabled the "Hide asset balances under $2" options from Settings and check the Wallet tab again.

<img width="375" alt="Screen Shot 2023-04-26 at 8 52 24 PM" src="https://user-images.githubusercontent.com/7005061/234739006-1ba7ddbd-e02c-449c-9381-b897056c18e7.png">

hyphenized:
<img width="377" alt="image" src="https://user-images.githubusercontent.com/28708889/234751207-5acbaa74-8ca1-4796-b29e-dacd9dca0b5a.png">

I feel that the previous message conveyed this information better, although the copy was still wrong for the related setting. Note that the asset might not have been added, but already exists because it's part of a token list.

<img width="377" alt="image" src="https://user-images.githubusercontent.com/28708889/234760305-9a27fdbe-2431-4456-8a00-c71f059b6d49.png">

jagodarybacka:
Yeah I think that message is too long, can we maybe split it into paragraphs or add a tooltip or Notion page about it? 
cc @VladUXUI

VladUXUI:
Users are trying to add an asset that already exists, this can be for 1 of two reasons:
1 - They have 0 balance so they don't see it 
2 - They only have dust, need to toggle off "hide assets"

I feel like the previous version had a bit better context. The "title" reinforced the state of the asset and the body text provided a fix for the edgecases.

@mr-michael can you go provide a bit more context as to why you would like to change?

Shadowfiend:
Is there a reason we wouldn't make this reactive to what we know? That is, tell them that they should see it/they can't see it because we know the balance is below $2 and they need to drop it/tell them they can't see it because balance is 0?

Additionally, it feels like we should forcibly display all assets the user has added explicitly. I'm sure that conversation already happened and I'm a little late to it but--did we explicitly decide not to do that, or did we avoid it for technical reasons?

VladUXUI:
> @VladUXUI Are there any UX concerns over making this reactive?
> 
Nope, having them be more specific should be awesome

kkosiorowska:
> > @VladUXUI Are there any UX concerns over making this reactive?
> 
> Nope, having them be more specific should be awesome

[...]

PR #3478, opened 2023-06-14
Make specific warnings for adding custom asset

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3325

## What

We should display more detailed warnings to the user when adding assets. Users are trying to add an asset that already exists, this can be for 1 of 3 reasons:

1 - They have 0 balance so they don't see it
2 - They only have dust, need to toggle off "hide assets"
3 - Asset is already on the list but they didn't notice it.

## UI
Before
Screenshot 2023-06-14 at 10 17 46

After
Screenshot 2023-06-16 at 10 32 13
Screenshot 2023-06-16 at 10 32 51
Screenshot 2023-06-16 at 10 31 58
Screenshot 2023-06-16 at 10 31 36

## Testing
Check that the messages are displayed correctly:
- [ ] Try adding an asset whose balance is greater than $2
- [ ] Try adding an asset whose balance is smaller than $2 when `Hide asset balances under $2` is on
- [ ] try adding an asset whose balance is smaller than $2 when `Hide asset balances under $2` is off
- [ ] Try adding an asset with a balance of 0. (Add the token and then try to do it again)
   - Address: `0x841fad6eae12c286d1fd18d1d525dffa75c7effe`
   - Token:`0x841fad6eae12c286d1fd18d1d525dffa75c7effe` 
   -  Network: Fantom network

Latest build: [extension-builds-3478](https://github.com/tahowallet/extension/suites/13651616650/artifacts/753745497) (as of Fri, 16 Jun 2023 08:37:14 GMT).


ui/_locales/en/messages.json
```diff
@@ -550,8 +550,12 @@
       "asset.label.balance": "Balance",
       "asset.label.symbol": "Name",
       "networkSelect.title": "Select Network",
-      "warning.alreadyExists.title": "Asset already exists, you will see it once your balance is over 0",
-      "warning.alreadyExists.desc": "If you still have issues seeing an asset, turn off “Hide asset balances under 2$” from Settings",
+      "warning.alreadyExists.title": "Asset already exists",
+      "warning.alreadyExists.desc": {
+        "dust": "Your balance is bellow $2, you will see it once your turn off “Hide asset balances under 2$” in Settings",
```
Shadowfiend, line 555:
```suggestion
        "dust": "Your balance is below $2. You will see it once you turn off “Hide asset balances under 2$” in Settings.",
```

Also can we interpolate both $2 (see below) and `Hide asset balances under 2$` (so that if we change the phrasing in Settings it can be updated immediately in this message as well)?

ui/_locales/en/messages.json
```diff
@@ -550,8 +550,12 @@
...
+        "noBalance": "Your balance is 0, you will see it once your balance is over $2",
```
Shadowfiend, line 556:
```suggestion
        "noBalance": "Your balance is 0, you will see it once your balance is over $2.",
```

Let's also interpolate that $2 if possible, so we're set up to make this threshold configurable in the future (more or less necessary if we ever make the user's currency configurable). I'm fine with the interpolation being a constant "$2" in the component, to be clear, just want it to be interpolated so we're not having to revise i18n keys later.

ui/_locales/en/messages.json
```diff
@@ -550,8 +550,12 @@
...
+        "visibility": "You should see in it in your asset list"
+      },
       "warning.dust.title": "Asset balance is under <$2, you wil see it once your balance is over $2",
```
Shadowfiend, line 559: Is this still being used?

Shadowfiend (commented):
Some quick notes on the messages, haven't tested or actually looked beyond that messages file yet. Love this change though 🙌

ui/_locales/en/messages.json
kkosiorowska: Yes because this message is displayed when the asset does not exist. 

Screenshot 2023-06-15 at 17 10 35

ui/_locales/en/messages.json
```diff
@@ -552,11 +552,11 @@
...
       "warning.alreadyExists.desc": {
-        "dust": "Your balance is bellow $2, you will see it once your turn off “Hide asset balances under 2$” in Settings",
-        "noBalance": "Your balance is 0, you will see it once your balance is over $2",
-        "visibility": "You should see in it in your asset list"
+        "dust": "Your balance is below {{sign}}{{amount}}. You will see it once you turn off “{{settings}}” in Settings.",
+        "noBalance": "Your balance is 0, you will see it once your balance is over {{sign}}{{amount}}.",
+        "visibility": "You should see in it in your asset list."
       },
-      "warning.dust.title": "Asset balance is under <$2, you wil see it once your balance is over $2",
+      "warning.dust.title": "Asset balance is under <{{sign}}{{amount}}, you wil see it once your balance is over {{sign}}{{amount}}",
```
Shadowfiend, line 559: Well while we're in here… Can we change this to either `under {{sign}}{{amount}}` or `<{{sign}}{{amount}}`? Probably should be `under` and the `dust` key a few lines above should be the `under` as well.

Shadowfiend (dismissed):
Code looks good, gotta test drive it (hoping for tomorrow).

ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -370,7 +383,20 @@ export default function SettingsAddCustomAsset(): ReactElement {
...
+                    ...warningOptions,
+                    settings: sharedT(
+                      "settings.hideSmallAssetBalance",
+                      warningOptions
+                    ),
+                  })}
+                {showWarningAboutNoBalance &&
+                  t("warning.alreadyExists.desc.noBalance", warningOptions)}
+                {showWarningAboutVisibility &&
+                  t("warning.alreadyExists.desc.visibility")}
```
hyphenized, line 398: We can move these to a function and structure by priority with either a `switch` statement or early `returns`, this would also allow us to get rid of the `!showOtherWarning` checks.

e.g.
```tsx
 const renderWarningText = () => {
    const showWarningAboutDust = ...something

    if (showWarningAboutDust) {
      return t("warning.alreadyExists.desc.dust", {
        ...warningOptions,
        settings: sharedT("settings.hideSmallAssetBalance", warningOptions),
      })
    }

    // showWarningAboutNoBalance
    if (hideDustEnabled && ...) {
      return t("warning.alreadyExists.desc.noBalance", warningOptions)
    }

    // showWarningAboutVisibility
    return t("warning.alreadyExists.desc.visibility")
  }
```

hyphenized (commented):
Looks great! Found one minor thing

ui/pages/Settings/SettingsAddCustomAsset.tsx
```diff
@@ -178,9 +180,20 @@ export default function SettingsAddCustomAsset(): ReactElement {
   const hideDustEnabled = useBackgroundSelector(selectHideDust)
   const showWarningAboutDust =
     hideDustEnabled &&
+    assetData?.mainCurrencyAmount !== 0 &&
     assetData?.mainCurrencyAmount !== undefined &&
     assetData?.mainCurrencyAmount < userValueDustThreshold
 
+  const showWarningAboutNoBalance =
+    !showWarningAboutDust && assetData?.mainCurrencyAmount === 0
```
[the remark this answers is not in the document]
kkosiorowska: In fact, this message should be displayed when the balance is zero. 
https://github.com/tahowallet/extension/assets/23117945/2fee5d65-33d1-4483-a573-7614e9eedfbb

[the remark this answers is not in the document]
hyphenized: Sorry I missed a word, comment should've said "Do we want to show this warning only if hide dust is enabled and display the visibility message otherwise?"

My question was: if hide dust is not enabled should we display "Asset already exists, you should see it in your asset list "?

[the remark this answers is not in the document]
kkosiorowska: If its balance is not zero then yes.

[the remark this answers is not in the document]
hyphenized: Ah I see! We should check that setting's value here too then. In the video, after disabling hide dust, the same warning shows up.

kkosiorowska:
After changing the copy, there was a problem with the right padding.  IMO, it didn't look good.  @VladUXUI Could you please confirm if 24 px is fine?

Before
Screenshot 2023-06-16 at 10 40 44

After
Screenshot 2023-06-16 at 10 41 10

VladUXUI:
> 

24px looks good @kkosiorowska !
Thank you

hyphenized (approved):
🚢 🚢 🚢

PR #3479, opened 2023-06-15
Exclude `article` and `result` types for abilities

kkosiorowska:
## What

Currently, all abilities are fetched from one endpoint. In about two weeks' time, the abilities type the `article` and the `result` should be fetched from another endpoint. Let's exclude these types and then, in the next steps, retrieve them from another endpoint. 

## UI
https://github.com/tahowallet/extension/assets/23117945/9586609d-3719-42c6-889c-1dcff932bf4a

## Testing
- [x] Check that abilities of this type are not fetched from the API. 
- [x] In the filter list, the `article` and `result` types should not be present.

Latest build: [extension-builds-3479](https://github.com/tahowallet/extension/suites/13624732258/artifacts/751620264) (as of Thu, 15 Jun 2023 10:02:31 GMT).

### DIFF
--- a/background/abilities.ts
+++ b/background/abilities.ts
@@ -31,8 +31,11 @@ export const ABILITY_TYPES_ENABLED = [
 export const ABILITY_TYPES = [
   ...ABILITY_TYPES_ENABLED,
   "event",
-  "article",
-  "result",
+  // Abilities type `article` and `result` will be fetched from the new endpoint.
+  // Let's exclude this type for a moment.
+  // TODO Fetch abilities from the correct endpoint.
+  // "article",
+  // "result",
   "misc",
 ] as const
 
--- a/background/services/abilities/index.ts
+++ b/background/services/abilities/index.ts
@@ -106,6 +106,7 @@ export default class AbilitiesService extends BaseService<Events> {
 
   protected override async internalStartService(): Promise<void> {
     await super.internalStartService()
+    await this.fetchAbilities()
   }
 
   // Should only be called with ledger or imported accounts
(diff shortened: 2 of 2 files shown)


jagodarybacka (changes requested):
Hmmm I'm not so sure about removing these from filters while we are keeping articles and results cached. If someone had articles filter turned on the that user will be left with a lot of articles that he can no longer filter out.

But if we plan to fetch articles in the future then we should't remove them from cache IMO. I would suggest just leaving these filters for now.

That is just one account with articles filter left turned on:
image

Question - what if we will still ask for articles and results when they change the endpoint - are they warning us that the endpoint will throw error or will it just ignore these?

kkosiorowska:
> Hmmm I'm not so sure about removing these from filters while we are keeping articles and results cached. If someone had articles filter turned on the that user will be left with a lot of articles that he can no longer filter out.

This is not quite true. Because our update mechanism works in such a way that we catch removed abilities from the API and then remove them from the cache. This means that the next time the abilities are fetched (without `article` and `result`), they will be removed from the wallet. 

Let me see if we can force abilities to refresh after migration.

> But if we plan to fetch articles in the future then we should't remove them from cache IMO. I would suggest just leaving these filters for now.

I don't entirely agree with this. I think this is a fairly non invasive way of solving this problem. This would require filtering such abilities and then updating the cache correctly.   I'm not sure we really want to restructure that [logic](https://github.com/tahowallet/extension/blob/main/background/services/abilities/index.ts#L125-L169). The temporary lack of support for this seems a simpler solution.  

> That is just one account with articles filter left turned on: image
> 
> Question - what if we will still ask for articles and results when they change the endpoint - are they warning us that the endpoint will throw error or will it just ignore these?

At the moment there is no problem yet, but in some time we will not be able to ask about these types and it should return the error. Daylight wants to make sure it doesn't break anything for us.

kkosiorowska:
@jagodarybacka  Please check again.

jagodarybacka (approved):
Alright, looks good 🚀

PR #3486, opened 2023-06-16
Fix throwing SharedToggleItem on item without label

jagodarybacka:
### What

Solve throwing NFTs filters page when some collection got `undefined` label. Seems like it makes `SharedToggleItem` throw an unexpected error. As we don't know yet how it happened that the collection had an `undefined` label let's make a fix in `isProbablyEVMAddress`

image

Latest build: [extension-builds-3486](https://github.com/tahowallet/extension/suites/13700610101/artifacts/757499670) (as of Mon, 19 Jun 2023 10:13:43 GMT).

### DIFF
--- a/ui/components/Shared/SharedToggleItem.tsx
+++ b/ui/components/Shared/SharedToggleItem.tsx
@@ -8,7 +8,7 @@ import SharedToggleButton from "./SharedToggleButton"
 export const STARS_GREY_URL = "./images/stars_grey.svg"
 
 type SharedToggleItemProps = {
-  label: string
+  label?: string
   thumbnailURL?: string
   checked: boolean
   onChange: (toggleValue: boolean) => void
@@ -25,7 +25,9 @@ export default function SharedToggleItem({
       <div className="text_wrap">
         <div className="thumbnail" role="img" />
         <label className="label ellipsis">
-          {isProbablyEVMAddress(label) ? truncateAddress(label) : label ?? ""}
+          {typeof label === "string" && isProbablyEVMAddress(label)
+            ? truncateAddress(label)
+            : label ?? ""}
         </label>
       </div>
       <SharedToggleButton onChange={onChange} value={checked} />
--- a/ui/components/Shared/SharedToggleItem.tsx
+++ b/ui/components/Shared/SharedToggleItem.tsx
@@ -25,7 +25,7 @@ export default function SharedToggleItem({
       <div className="text_wrap">
         <div className="thumbnail" role="img" />
         <label className="label ellipsis">
-          {isProbablyEVMAddress(label) ? truncateAddress(label) : label}
+          {isProbablyEVMAddress(label) ? truncateAddress(label) : label ?? ""}
         </label>
       </div>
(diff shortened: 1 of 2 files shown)


ui/components/Shared/SharedToggleItem.tsx
```diff
@@ -25,7 +25,7 @@ export default function SharedToggleItem({
       <div className="text_wrap">
         <div className="thumbnail" role="img" />
         <label className="label ellipsis">
-          {isProbablyEVMAddress(label) ? truncateAddress(label) : label}
+          {isProbablyEVMAddress(label) ? truncateAddress(label) : label ?? ""}
```
Shadowfiend, line 28: I think what we're missing here is updating the type of `label` to include the possibility of `undefined` (and same for `isProbablyEVMAddress`).

ui/components/Shared/SharedToggleItem.tsx
```diff
@@ -8,7 +8,7 @@ import SharedToggleButton from "./SharedToggleButton"
 export const STARS_GREY_URL = "./images/stars_grey.svg"
 
 type SharedToggleItemProps = {
-  label: string
+  label?: string
```
kkosiorowska, line 11: Non-blocking comment
I wonder if this is the correct place for the fix. :thinking:  My point is that the problem is not with the `SharedToggleItem` component but with the `undefined` name value for collection. Therefore, I don't know whether we shouldn't fix this issue [here](https://github.com/tahowallet/extension/blob/main/ui/components/NFTs/Filters/FilterList.tsx#L37).  Because now the changes apply to the shared component what is a bit unclear is why the label can be optional. What I mean is that the `thumbnailURL` is already optional and with the optional `label` it reduces this component to a simple `SharedToggleButton`.  Maybe we could add a comment for this?

jagodarybacka: Tbh I'm now hesitant to even merging this fix, as I was not able to reproduce it and in theory the name should be replaced with empty string if needed while the collection is fetched from SH. Unless we are able to find a way to reproduce it 🤔 
cc @michalinacienciala can you try to reproduce that problem with NFTs filters?

michalinacienciala: I was able to reproduce it. A bit different steps than before: I reproduced it in Playwright tests when switching networks quickly (Ethereum -> Optymism -> Polygon -> Arbitrum) and then going straight away to NFTs filtering.

jagodarybacka: Do we have this test already in the codebase? if not can you create a draft PR with it?

michalinacienciala: Added temporary test in https://github.com/tahowallet/extension/pull/3512.

jagodarybacka:
Ok, I've found where the problem was, it is not actually because of NFTs, it happens because the account is not fully loaded and it has no name yet but we are not providing a fallback to address. Will close this PR and open a new one.

PR #3487, opened 2023-06-19
Run NFTs e2e tests on a controlled wallet address

michalinacienciala:
Previously we run tests by importing `bravonaver.eth` address as a read-only account. We don't control that address and can't guarantee that it's state won't change (changes may lead to unstable tests). Now we're switching to importing the third address of the `testertesting.eth` account (`0x6f1b1f1feb01235e15a7962f16c389c7f8218ed6`) as a read-only. We control this address and can decide on its state.

Latest build: [extension-builds-3487](https://github.com/tahowallet/extension/suites/13715479293/artifacts/758632877) (as of Mon, 19 Jun 2023 21:31:19 GMT).


michalinacienciala:
I re-run the workflow with tests a couple of times to make sure it's not randomly failing - all executions were green: https://github.com/tahowallet/extension/actions/runs/5308837549.

e2e-tests/nfts.spec.ts
```diff
@@ -36,10 +36,30 @@ test.describe("NFTs", () => {
         }
       })
 
-      await walletPageHelper.onboarding.addReadOnlyAccount("bravonaver.eth")
+      await walletPageHelper.onboarding.addReadOnlyAccount(
+        "0x6f1b1f1feb01235e15a7962f16c389c7f8218ed6"
```
jagodarybacka, line 40: Do we plan to reuse that address for other e2e tests? Then we can make it a variable.

e2e-tests/nfts.spec.ts
```diff
@@ -36,10 +36,30 @@ test.describe("NFTs", () => {
...
       await walletPageHelper.goToStartPage()
       await walletPageHelper.setViewportSize()
+
+      // Switch to Optimism, then to Polygon and then to Arbitrum (to load NFTs
+      // on those chains).
+      await page.getByTestId("top_menu_network_switcher").last().click()
+      await page.getByText(/^Optimism$/).click()
+      await expect(
+        page.getByTestId("top_menu_network_switcher").last()
+      ).toHaveText(/^Optimism$/)
```
jagodarybacka, line 52: What do you think about making it a util function where you just specify what network you want to switch to?

jagodarybacka (dismissed):
Works great, some ideas for improvements but non-blocking, lmk if you want to apply it here or later 😄

e2e-tests/nfts.spec.ts
michalinacienciala: I don't have a particular test in mind yet where we would reuse it. Let's leave it as it is for now - if we introduce the test which uses the same account, we can parametrize this then.

e2e-tests/nfts.spec.ts
```diff
@@ -70,7 +90,9 @@ test.describe("NFTs", () => {
 
       await page
         .getByTestId("nft_account_filters")
-        .filter({ hasText: "bravonaver.eth" })
+        .filter({
+          hasText: /^(Phoenix|Matilda|Sirius|Topa|Atos|Sport|Lola|Foz)$/,
```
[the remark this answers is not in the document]
michalinacienciala: Yeah, good point, displaying the label based on UNS/ENS is a functionality related to NFTs, so it makes sense to test it in this test. Will do.

e2e-tests/nfts.spec.ts
michalinacienciala: Done in 602e12bd9e2aaac9cdf34f84d90f72ff54364061.

e2e-tests/nfts.spec.ts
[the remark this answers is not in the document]
michalinacienciala: I thought I will be able to buy the domain myself, but had to ask Ops. Let's maybe not wait for that to get processed and merge the PR as it is... I'll update the account name in a separate PR once we have the domain secured - wdyt?

e2e-tests/nfts.spec.ts
Shadowfiend: We should put it in a named constant regardless—no magic strings please!

Shadowfiend (approved):
Reapproving because I did it too soon <_<

Issue #3467, opened 2023-06-12
Add assets when trust/untrust is enabled/disabled 


VladUXUI:
Right now if trusted/untrusted is enabled or disabled, it doesn't allow for an asset on that list to be added.
We should allow users to add an asset that is in the untrusted list.

Feedback from @beemeeupnow 

"For unverified assets:

- If the user for some reason enables the setting to show them, then later decides to disable it again in the settings page, it seems that they are blocked from adding any of the unverified assets by contract address using "Add custom asset" in the wallet view. (the UI shows "Asset already exists, you will see it once your balance is over 0", and you are unable to proceed)
- If the user tried to use "Add custom asset" with the unverified assets setting enabled and the contract matches an existing unverified asset, I feel it would make sense to allow adding it or take them to the verification page for the asset. (same UI result as above currently)
- If you hide a verified/unverified asset then try to add it with "Add custom asset" it also blocks with the same message"

PR #3491, opened 2023-06-21
Untrusted assets should not block the addition of custom tokens

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3467

## What

Currently, the user isn't able to add an asset that is in the untrusted list. We display a message about the existence of the token and block the button. We should allow users to add an asset that is in the untrusted list.

### Changes
- Ability to manually add unverified assets.
- Simplification of utils function for untrusted/trusted tokens.
- When adding an unverified asset, don't display a dust warning.
- Don't update the metadata of assets in the token list and network base assets.

## UI

https://github.com/tahowallet/extension/assets/23117945/f5dd9dea-e8c5-4bf1-b535-73edbd93a86c

## Testing
- [x] Select an unverified token from the list and then try adding it manually. The asset should be added correctly.
- [x] Try adding an unverified token which is dust. There should be a dust warning.
- [x] Verify the asset from the list and then try to add it manually. A message should appear saying that the asset is already added.
- [x] Try adding a trusted asset that already exists in the wallet. A message should appear saying that the asset is already added.
- [x] Check that adding tokens works for assets outside the unverified list. 
  - Test address: `testertesting.eth`
  - Network: Fantom Opera
  - Token: `0x841fad6eae12c286d1fd18d1d525dffa75c7effe`
- [x] Ensure that you continue to be unable to send and swap unverified assets until verified. After verification, the token should appear in the list of assets to be sent and swapped.
  - [x] Verification by manually importing the token
  - [x] Verification by the add to list button (asset warning slide-up menu)

Latest build: [extension-builds-3491](https://github.com/tahowallet/extension/suites/14149590642/artifacts/792592880) (as of Sat, 08 Jul 2023 04:51:18 GMT).


background/redux-slices/utils/asset-utils.ts
```diff
@@ -377,6 +377,23 @@ export function isUnverifiedAssetByUser(asset: AnyAsset | undefined): boolean {
...
+ * Only verified assets can take part in wallet actions.
+ * By actions is meant:
+ * - doing an swap with this asset
+ * - sending this asset to another address
+ */
+export function isVerifiedAsset(asset: AnyAsset): boolean {
+  if (!isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET)) {
+    return true
+  }
+  return isUntrustedAsset(asset) ? !isUnverifiedAssetByUser(asset) : true
```
Shadowfiend, line 394: Doesn't `isUnverifiedAssetByUser` already take `isUntrustedAsset` into account?

https://github.com/tahowallet/extension/blob/546f8c579bb0fe87ac9ffa7eb047f3be68acfa97/background/redux-slices/utils/asset-utils.ts#L371-L374

I think if we dig through and do a bunch of [really ugly K Maps](https://en.wikipedia.org/wiki/Karnaugh_map) of this[^1], we can simplify this assertion to:

```typescript
  return isNetworkBaseAsset(asset) || !isUntrustedAsset(asset)
```

I'm having a lot of difficulty reasoning my way through these though and I think it's because we're u8sing a lot of negatives, and we've also muddled three concepts I want to circle back to: *baseline trusted* assets, *trusted* assets, and *verified* assets. I think we can define a very clear, stackable set of conditions that are easier to reason about with these:

- **Baseline trusted** means “the wallet ~trusts this”—i.e., the asset is in a token list OR the asset is a network base asset. Currently `!isUnverifiedAssetByUser` sort of tries to do this.
- **Verified** means the user has explicitly verified the asset.
- **Trusted** means the asset is baseline trusted OR verified. Currently that is what `isVerifiedAsset` reports.

In this scenario the negatives also have very clear meanings as the opposites of the above (`!...`):

- **Baseline untrusted** means the asset is `!baseline trusted`—i.e. “the wallet does not trust this without an explicit user signal”. Currently this is roughly equivalent to `isUntrustedAsset`. It could still be verified.
- **Unverified** means the asset is `!verified`—i.e., the user has not explicitly verified the asset. It can still be baseline trusted.
- **Untrusted** means the asset is `!trusted`—i.e. the asset is neither baseline trusted NOR verified. Currently this is roughly equivalent to `isUnverifiedAssetByUser`.

I strongly believe these three concepts are enough to handle all of what we're trying to address, and can all be stated in terms of positives (`verified`, `trusted`, `baselineTrusted`) most of the time and only as negatives `untrusted`, `untrustedByUser`, etc) when we need to check the negative. They also feel straightforward to unit test, explain, and document.

Do you agree? Have I missed an edge case? I found about 4 while I was writing and rewriting this comment, so I wouldn't be surprised if I missed another 😅

[^1]: image

Shadowfiend (changes requested):
Left some suggestions on simplifying boolean logic here—notably, there's a few other places in these code paths but outside of this PR where we should also ruthlessly simplify the boolean logic, e.g.:

https://github.com/tahowallet/extension/blob/546f8c579bb0fe87ac9ffa7eb047f3be68acfa97/background/services/indexing/index.ts#L708

background/redux-slices/selectors/0xSwapSelectors.ts
```diff
@@ -29,7 +29,7 @@ export const selectSwapBuyAssets = createSelector(
       ): asset is SwappableAsset & {
         recentPrices: SingleAssetState["recentPrices"]
       } => {
-        if (!canBeUsedForTransaction(asset)) {
+        if (!isVerifiedAsset(asset)) {
```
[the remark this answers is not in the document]
kkosiorowska: https://github.com/tahowallet/extension/pull/3491/commits/c83cf0eb8dd5b6c93deaf6b4083524fa890491cf

background/redux-slices/utils/asset-utils.ts
kkosiorowska: This has been brilliantly explained. Let's simplify this logic to make it clear and unambiguous. PR with a new concept https://github.com/tahowallet/extension/pull/3528

kkosiorowska:
Improvements have been done to parts of the logic. Terminology changes and simplifications will be included in the https://github.com/tahowallet/extension/pull/3528. Simplification of the asset removal logic has reported in https://github.com/tahowallet/extension/issues/3533, and will be resolved in the near future.  

The PR is ready to be reviewed and tested again. One note we currently don't fetch prices for untrusted tokens. There is a discussion on this topic on discord.

Shadowfiend (approved):
Left a couple of notes, can be followed up in #3528. Let's ship it :ship:

PR #3492, opened 2023-06-21
Flip updated dApp connections flag

### LINKED ISSUE #2500: MM impersonation problems with beta Polygon Wallet dapp
### Background
There is a beta version of [Polygon Wallet app](https://wallet-beta.polygon.technology/) and if user once selected the beta version then it is always redirecting from stable version. I see no easy way to go back to stable version and cleaning cache is not working so we can assume our users will be stuck there.

### What
On beta version TallyHo is not popping up after selecting the Metamask option even if "default wallet" is selected.

### LINKED ISSUE #3430: Cannot connect wallet to coinmarketcap.com
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/1114179647987523614

Chrome

Connect the wallet through the Metamask option on https://coinmarketcap.com/

### What did not work?

Instead of connecting, the site redirects you to metamask install page

### Version

v0.34.0

### Relevant log output

_No response_

### LINKED ISSUE #3453: Can't connect to Eigenlayer
Metamask option is no longer working regardless if Taho is default wallet or not.
https://www.eigenlayer.xyz/
image

Shadowfiend:
cc @VladUXUI for design QA

e2e test was also added for the initial display of the Connect to Taho hint popover.

## Testing

- With Taho as default set to on, the first time:
  - [x] Ensure the toggle in the Taho popup is pointed at Taho.
  - [x] Trying to connect to Uniswap, ensure the dApp connection popup displays the informational hint, and that the toggle in the popup is pointed at Taho.
  - [x] Click the Taho as default toggle and ensure Taho auto-closes and MetaMask immediately appears, and allows connections normally.
  - [x] Ensure the toggle in the Taho popup is pointed at MetaMask.
  - [x] Switch the default toggle back to Taho and reload the page.
  - [x] Ensure you can connect normally with Taho.
- [x] Check that coinmarketcap.com allows connections (with Taho set as default).
- [x] Check that https://wallet-beta.polygon.technology/ allows connections as Taho with Taho set as default and not. 
- [x] Check that https://app.eigenlayer.xyz allows connections (with Taho set as default).

Fixes #3430.
Fixes #2500. (Sort of—Taho is now directly an option, but if we hack `window.tally.isTally = false` and try with Taho set as default, Taho works correctly. Only important insofar as the bug is no longer relevant.)
Fixes #3453.

Latest build: [extension-builds-3492](https://github.com/tahowallet/extension/suites/14093969281/artifacts/788218649) (as of Thu, 06 Jul 2023 02:41:44 GMT).


VladUXUI:
Couple of non blocking small bits. Happy to have these as a ticket for next sprint.

- [ ] Dapp connection box is still using the old image. For consistency let's use the one that we are using in sign tx information box.

image
image
 
 It doesn't need to be clickable
 

- [ ]  Couple of copy updates for this box

- Dap Connection = Website connections (title)
- How to connect to dapps = How to connect to websites
- Set Taho as default = Connect using Taho
- Click “connect to wallet” in a dapp = Click “connect” on a website

Shadowfiend:
This fellow should be ready.

Shadowfiend:
Trying to fix the failing e2e tests here led me to an issue that @hyphenized caught @ https://github.com/tahowallet/extension/pull/3462#discussion_r1230296367 . Need to fix that one as it will trigger anytime the user tries to connect to Taho as Taho (rather than as default) if they have Taho-as-default turned off (by auto-closing the dApp connection popup whoops).

Shadowfiend:
Ok, this should be ready again.

michalinacienciala:
@Shadowfiend, what about that removal of the connection banner from the swap screen? Are we leaving this at it is?

Shadowfiend:
Yep, remaining for now and left for future cleanup.

VladUXUI:
@hyphenized since you are feature buddy here, can you please review this PR?

e2e-tests/dapp-connect.spec.ts
```diff
@@ -1,28 +1,94 @@
...
+    })
 
-  await walletPageHelper.goToStartPage()
+    expect(await connectingPopupTitle.count()).toBe(1)
+    await expect(connectingPopupTitle).toBeVisible()
 
-  await page.locator('text="Settings"').click()
-  await page.locator("text=Connected websites").click()
+    // Clear the popover.
+    const bgLocator = popupPage.locator(".bg")
```
hyphenized, line 33: Can we target the "void space" button instead? Something like this should work:

```ts
        popupPage
        .getByRole("button", { name: "Background close" })
        .click()
```

e2e-tests/dapp-connect.spec.ts
```diff
@@ -1,28 +1,94 @@
...
+  test("should work and add an entry to the connected websites list", async ({
+    page,
+    context,
+    walletPageHelper,
+  }) => {
+    await walletPageHelper.onboarding.addReadOnlyAccount("testertesting.eth")
+    await walletPageHelper.hideDappConnectPopup()
+
+    const dappPage = await context.newPage()
+    await dappPage.goto("https://swap.cow.fi/")
```
hyphenized, line 72: Probably worth running one small dApp along hardhat for some standard tests of this kind

hyphenized (changes requested):
Works well! Found a small issue: when the wallet is not set as default, using the default wallet toggler twice in the connection popup opens a duplicate connect page.

Shadowfiend:
Updated the last bullet to reference app.eigenlayer.xyz instead of .com.

e2e-tests/dapp-connect.spec.ts
Shadowfiend: Would really love to avoid more stuff we have to maintain tbh, but agree this is a pretty kludgey way to do it.

e2e-tests/dapp-connect.spec.ts
Shadowfiend: I think that was my original approach and then I ran into issues that I thought were related but ultimately weren't, and then never changed it back. I'll poke at it.

Shadowfiend: https://github.com/tahowallet/extension/pull/3492/commits/220a0b8dadd72adbe3798605efce19033f4c029f

Shadowfiend:
> when the wallet is not set as default, using the default wallet toggler twice in the connection popup opens a duplicate connect page

This is fixed. I also made it so the wallet will correctly auto-close in this scenario.

The remaining edge case here is:
- If Taho is the only installed wallet.
- And Taho starts as non-default.
- And the user sets Taho as default in the dApp Connection popup.
- And the user then switches Taho back to non-default.
- Then Taho will auto-close and automatically reject the connection request, instead of continuing to wait for the user to hit Connect or Reject.

This seems vanishingly unlikely, and I suggest we don't spend any time trying to fix it unless we get a lot of user confusion, even though my engineering brain wants to solve it.

hyphenized (approved):

PR #3494, opened 2023-06-22
Change expected account name in NFTs tests to ENS domain

michalinacienciala:
We have configured a ENS name for the
`0x6f1b1f1feb01235e15a7962f16c389c7f8218ed6` address. We need to update the tests to reflect that.

Latest build: [extension-builds-3494](https://github.com/tahowallet/extension/suites/13792525506/artifacts/764721892) (as of Thu, 22 Jun 2023 14:20:28 GMT).

### DIFF
--- a/e2e-tests/nfts.spec.ts
+++ b/e2e-tests/nfts.spec.ts
@@ -80,7 +80,7 @@ test.describe("NFTs", () => {
       await page
         .getByTestId("nft_account_filters")
         .filter({
-          hasText: /^(Phoenix|Matilda|Sirius|Topa|Atos|Sport|Lola|Foz)$/,
+          hasText: /^e2e\.testertesting\.eth$/,
         })
         .getByRole("checkbox")
         .click()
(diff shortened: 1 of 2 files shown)


e2e-tests/nfts.spec.ts
```diff
@@ -80,7 +80,7 @@ test.describe("NFTs", () => {
       await page
         .getByTestId("nft_account_filters")
         .filter({
-          hasText: /^(Phoenix|Matilda|Sirius|Topa|Atos|Sport|Lola|Foz)$/,
+          hasText: /^e2e\.testertesting\.eth$/,
```
jagodarybacka, line 83: let's add a `const` variable for it so we can reuse that name easily

jagodarybacka: I would move it to the utils file so every test will have access to that name 😄

michalinacienciala: [5821c55](https://github.com/tahowallet/extension/pull/3494/commits/5821c559d022ae7265adcbb8458df79c643f9d77)

jagodarybacka (approved):
✅

Issue #3470, opened 2023-06-12
Argon2 integration


Shadowfiend:
Upgrade the use of PBKDF2 to use Argon2 instead using https://github.com/antelle/argon2-browser .

Considerations:
- Upgrades should be transparent to the user; the first time the user enters their passphrase in the extension after upgrading, all historical vaults should be decrypted using the PBKDF2-derived key and then re-encrypted using the Argon2 key.
- We should bump the encrypted vault version to 2 for vaults that were encrypted using an Argon2-derived key. This will make it easier to determine later on whether an upgrade was or was not performed.
- We should register an analytics event when this occurs, so we can have a sense of how many users who have opted into analytics have been upgraded vs how many might remain on an older implementation.

Specific example references:
- Basic example in the argon2 repo: https://github.com/antelle/argon2-browser/blob/master/examples/node/argon2-demo-node.js
- More advanced example from an app using it: https://github.com/keeweb/keeweb/blob/develop/app/scripts/util/kdbxweb/kdbxweb-init.js#L11

jagodarybacka:
https://github.com/tahowallet/extension/pull/3502 merged to feature branch

PR #3415, opened 2023-05-27
Versions that go bump in the night: Bump versions across a few different core dependencies

Shadowfiend:
This brings us up to latest Ethers 5-series, React, and @reduxjs/toolkit.

We'll eventually want to poke at Ethers 6, but that will have to wait as it has breaking API changes.

Most of the resulting code changes were typing-related, and the outcome here is real nice—background async thunks have fully functional typing!

Leaving this as draft until we're ready to test rather heavily for the release.

Latest build: [extension-builds-3415](https://github.com/tahowallet/extension/suites/15651065061/artifacts/893364443) (as of Wed, 30 Aug 2023 15:18:14 GMT).

PR #3502, opened 2023-06-26
Use Argon2 for encrypted vaults

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3470

### What

Let's use Argon2 instead of PBKDF2 🔑 

What was already done:

- [x] added required packages
- [x] configured webpack config to work with webAssembly following [example from the docs](https://github.com/antelle/argon2-browser/tree/master/examples/webpack)
- [x] add necessary config to `content_security_policy` to allow webAssembly - without `'wasm-eval'` we are not able to use argon implementation in the extension
- [x] migration from old vaults to new vaults
- [x] handling errors during migration and allowing users to continue with old implementation if needed
- [x] added one time analytics event after successful migration 
- [x] updated Jest to allow unit tests to work with WebAssembly, updated Typescript as this was necessary to make new version of Jest work correctly and fixed a bunch of Typescript issues - most fixes are copied from https://github.com/tahowallet/extension/pull/3415

### Testing

- [x] test migration from main - install extension on `main`, add some HD wallets, checkout this branch, reload and unlock the wallet, make sure you don't see the error about failed migration in the background console, check if analytics event has been emitted, lock and unlock more than one time
- [x] install extension on this branch, test locking and unlocking the wallet
- [x] try to make webassembly part break (throw Error [here](https://github.com/tahowallet/extension/blob/migrate-to-argon/background/services/internal-signer/encryption.ts#L118)) and make sure user is able to continue with old vaults

Latest build: [extension-builds-3502](https://github.com/tahowallet/extension/suites/14011994578/artifacts/782054890) (as of Sun, 02 Jul 2023 21:28:25 GMT).


background/services/internal-signer/encryption.ts
```diff
@@ -101,6 +102,34 @@ export async function deriveSymmetricKeyFromPassword(
...
+
+  // Argon2 returns hash which is 24 bytes long, we need 16 bytes for AES-GCM
+  const { hash } = await argon2.hash({
+    pass: password,
+    salt,
+  })
+
+  const key = await crypto.subtle.importKey(
+    "raw",
+    hash.slice(0, 16),
```
jagodarybacka, line 121: we need either length of 16 or 32, we have 24 bytes currently, do we want to add padding or are fine with slicing the hash?

background/services/internal-signer/encryption.ts
```diff
@@ -64,7 +65,7 @@ function requireCryptoGlobal(message?: string) {
  *          material using AES GCM mode, as well as the salt required to derive
  *          the key again later.
  */
-export async function deriveSymmetricKeyFromPassword(
+export async function depricatedDeriveSymmetricKeyFromPassword(
```
Shadowfiend, line 68:
```suggestion
export async function deprecatedDerivePbkdf2KeyFromPassword(
```

background/services/internal-signer/encryption.ts
Shadowfiend: You should be able to pass `hashLen: 32` to get 32 bytes (see “other parameters” under https://github.com/antelle/argon2-browser#js-library ).

jagodarybacka:
For fixing tests we need to update Jest to the newest version as they apparently added [support for webassembly](https://github.com/jestjs/jest/pull/13505#discussion_r1018498749) some time ago. There is a bunch of errors after I've upgraded `jest` version unfortunately 🙈

Shadowfiend:
I went through a lot of the jest upgrade pain in #3415 . https://github.com/tahowallet/extension/pull/3415/commits/8ea287bb299bebcc0cae75a684ea7dab2e44e37c is a (not so clean) commit with those changes. There's Ledger stuff in there as well, but might be straightforward to pick it apart.

background/services/internal-signer/index.ts
```diff
@@ -265,21 +273,29 @@ export default class InternalSignerService extends BaseService<Events> {
       return true
     }
 
+    const { vaults, version } = await migrateVaultsToArgon(password)
```
hyphenized, line 276: Might be nitpicking here but I think we could improve our approach if we kept the concern for migrating vaults separate from retrieving them. So, rather than returning vaults here, I propose we just return a boolean indicating success or failure and continue using `getEncryptedVaults`.

background/services/internal-signer/index.ts
```diff
@@ -121,6 +126,7 @@ interface Events extends ServiceLifecycleEvents {
   // TODO message was signed
   signedTx: SignedTransaction
   signedData: string
+  migratedToArgon2: never
```
Shadowfiend, line 129: Argon2 is an internal storage detail--it should not escape the service IMO. If we ever want to do a migration that isn't transparent to the user (eg because we want to encourage users to upgrade), we can revisit, but right now we should treat it as an implementation detail.

background/services/internal-signer/index.ts
Shadowfiend: This should be true internally in the encryption layer, but for the caller it's the opposite--encryption layer should be completely in charge of making sure that the data is in the latest format, just like the db objects migrate before returning data.

Vault version is basically a constant across a given run of the extension. All retrievals of encrypted data should automatically reencrypt to the latest version. All encryptions should use the latest version. That fact should be transparent to almost everyone. Migrations should not be even a little bit optional (as skipping a function call might make them). This is how we ensure calling code doesn't make weird decisions about not wanting to upgrade, or weird mistakes about sometimes not doing so, etc: we give no one a choice or a hook. 

To put this in terms of separation of concerns: the encryption layer is concerned with encrypting and decrypting data, and ensuring to the extent possible that all encrypted data is in the latest format.

background/main.ts
```diff
@@ -1151,6 +1151,12 @@ export default class Main extends BaseService<never> {
       }
     })
 
+    this.internalSignerService.emitter.on("migratedToArgon2", async () => {
+      this.analyticsService.sendOneTimeAnalyticsEvent(
```
Shadowfiend, line 1155: This is one of the few cases where we should just call this directly on the service IMO. Piping analytics through events makes sense when we're observing an existing action, but less so when we're wanting to track an action that is internal to a service.

background/services/internal-signer/index.ts
```diff
@@ -149,6 +155,8 @@ const isKeyring = (
 export default class InternalSignerService extends BaseService<Events> {
   #cachedKey: SaltedKey | null = null
 
+  #cachedVaultVersion: VaultVersion = VaultVersion.PBKDF2
```
Shadowfiend, line 158: This should never change, right? All vaults should be argon2 by the time the internal signer service sees them.

Shadowfiend (commented):
Left a few comments. Pondering how we can push more of the migration out of the service core and into the encryption layer... Right now the two are very entwined (really an old decision, nothing to do with the Argon2 stuff other than that it adds complexity) and it's leading to a gnarly `unlock` method that's starting to coordinate too much. 

Not sure if it's worth refactoring further right now or not, will try to give more thoughts later today.

background/services/internal-signer/index.ts
jagodarybacka: That was my initial approach but then we have to call `getEncryptedVaults` 2/3 times:
1. to get the version and see if it was already migrated, if it wasn't then we can either:
2. pass vaults to migration method or call `getEncryptedVaults` inside the migration method, then
3. to get migrated encrypted vaults in the unlock once again (if we don't want to get them from `migrate...()`

which is not the worst but also seems like a lot of repetition 🤔 will think about it more

background/services/internal-signer/index.ts
jagodarybacka: well initially for our existing users they are `VaultVersion.PBKDF2`, once they are migrated successfully to Argon then yeah, this should never change again but if the migration fails then I want to be able to use old vaults so these users are not stuck with locked keys - that's why I wanted to keep that info here

.eslintrc.js
Shadowfiend, line 1: This is going to be the one that makes me fix all our linting heh.

.eslintrc.js
```diff
@@ -94,7 +94,11 @@ module.exports = {
     ],
     "@typescript-eslint/no-unused-vars": [
       "error",
-      { argsIgnorePattern: "^_", varsIgnorePattern: "^_" },
+      {
+        argsIgnorePattern: "^_",
+        varsIgnorePattern: "^_",
+        ignoreRestSiblings: true,
```
Shadowfiend, line 100: Let's not turn this one on. This makes something like `const { a, b, ...rest } = c` not error if `a` or `b` is unused. The pattern we use for these cases is `const { a: _, b: __, ...rest } = c`. This assigns `a` and `b` to an `_`-prefixed variable, which is our “this isn't intended to be used” indicator, and should be allowed through by the `_` prefix test.

background/services/internal-signer/index.ts
Shadowfiend: Ah, I see the thinking here. Here's how I think we handle that possibility:

- Every time we unlock, we should try to migrate. The migration should check that once an Argon2 vault is encrypted, it can also be decrypted and produce the same data as the PBKDF2 vault had, before writing the migration to `localStorage`.
- If the migration fails, we should log an analytics event. The unlocking process should continue normally (which will use the vault's current version, PBKDF2 if there was an error with the Argon2 encryption).
- The migration process should tell us if it succeeded, i.e. we should be able to know “the internal signer data is all on the latest version of encryption”—this is just a boolean.
- When we save the vault, we should use the strongest encryption we have (i.e., Argon2). If that fails, we should log an analytics event. At this point, if all vaults are on the latest version of encryption, we should fail the same way we would have failed if PBKDF2 blew up. If all vaults are not on the latest version, we should fall back to PBKDF2. In practice, this means we never want to end up with a newer vault on an older version of encryption than we have used until now—we should treat that as a full system failure.
- After a set amount of time (maybe 6 months?) without errors in analytics, we should consider dropping the PBKDF2 fallbacks for encryption (but never for the migration for decryption).

This is more complex, but it never winds back security, and it doesn't leak encryption implementation details (vault version) outside of the encryption code.

Thoughts

background/services/internal-signer/encryption.ts
```diff
@@ -10,6 +11,11 @@ export type EncryptedVault = {
   cipherText: string
 }
 
+export enum VaultVersion {
```
Shadowfiend, line 14: Re: my note about not leaking vault version information: that basically means we should try to be able to not export this.

background/services/internal-signer/index.ts
Shadowfiend: Thought about this some more. The gap between this and how the internal signer was already implemented is not huge but it's gnarly. I don't think we should deal with it right now, unfortunately.

The way this would manifest is that we would have `getEncryptedVaults` transparently handle vault migration, and `writeLatestEncryptedVault` would be redone as a `writeLatestVault` and would transparently handle encryption and fallbacks.

background/services/internal-signer/index.ts
jagodarybacka: Sounds good, let me try to implement it and see how it will go.

jagodarybacka: Implemented:
- [x] on every unlock we try to migrate
- [x] test if the vault encrypted with Argon produces the same data as before the migration
- [x] log analytics event on migration fail
- [x] migration process should return a boolean if it was successful

What is not clear to me:
> When we save the vault, we should use the strongest encryption we have (i.e., Argon2)....

So this point - if I understand correctly what you want to achieve here - is something I would rather avoid putting into code, because it should never be the case where we are writing new vault encrypted with Argon while old vaults are on the PBKDF2. 
That is because to save new vault we have to have the service unlocked and we are trying to migrate during unlocking. If the migration fails then most likely writing a new vault with Argon will fail as well. I would probably go with saving the current encryption algorithm in the service as this is implemented right now and then we can use it to add new vaults without adding more complexity 🤔

background/services/internal-signer/encryption.ts
jagodarybacka: I see but if we need to know and save (in the service) on which version we are then we need this exported, right?

background/services/internal-signer/index.ts
Shadowfiend: A couple of notes on the reasoning, but ultimately I think we're good here ->

> it should never be the case where we are writing new vault encrypted with Argon while old vaults are on the PBKDF2.

I agree this isn't ideal, but it is strictly better than having all vaults on PKDF2. The goal of having version alongside the vault is that we can have mixed versions if absolutely necessary.

> If the migration fails then most likely writing a new vault with Argon will fail as well.

I don't know why argon2 would fail, so there's nothing to indicate to me whether such a failure would be transient or not. It could fail during migration due to issues loading wasm, but it could also run into transient resource limitations--again, it's unclear to me.

All that said: **you are right**! The piece I forgot about here is that we (correctly) don't keep the password around after the initial `unlock`. This means if the initial `unlock` fails to use argon2 to derive the key, and we want to try to encrypt with an argon2-derived key later… We would have to keep the password around/cached instead of just the key. A definite no! So let's put this one to rest.

background/services/internal-signer/encryption.ts
Shadowfiend: Ultimately discussed in https://github.com/tahowallet/extension/pull/3502#discussion_r1247829024, since what I was pushing for is precisely *not* saving in the service which version we're on—but the need for that has been clarified.

Shadowfiend:
Pushed the above changes to accelerate a little; we can always walk the commit back if necessary.

The main thing I noticed while adding a test for the error messages is that (and this was true before my changes) if we track migration failures, we track migration failures due to incorrect passwords as well as ones that might be due to an internal technical issue. There are a few ways we could mitigate this, but my instinctive reaction is that we roll it out as-is and see how it goes.

Shadowfiend (approved):
All righty, good to go here. Since I pushed some code, going to let @jagodarybacka do a final sanity check and then merge.

Shadowfiend:
> > if we track migration failures, we track migration failures due to incorrect passwords
> 
> ugh 😞 so what exactly do we want to measure with these events? I think info on how many users are migrated is the most important one anyway so we are good here

Secondarily, I think it'll be useful to know potentially know what we need to look at if they aren't being migrated, or at least how many failures in migrations we're seeing. I'm comfortable shipping without being sure we can get that info though.

jagodarybacka (commented):
QA went fine, let's 🚢

Issue #3501, opened 2023-06-26
Fix issue with OETH populating on Taho Wallet


sumpg41:
> how high does it need to get to appear in the list? In my experience any amount of tokens in the wallet should cause it to populate on the list

@sumpg41 I have checked the code more carefully and will try to explain it more clearly. The decimal amount of this token is close to zero. The token will be visible when the amount has grown enough to make sense to show it in the wallet. It is about correctly displaying the token amount in the wallet with some precision. It is aimed at hiding the dust from the sale of my assets. Otherwise, we would show a long list of tokens with a zero balance like in the image below. This is an example of what it would look like.

<img width="250" alt="Screenshot 2023-06-26 at 11 43 05" src="https://github.com/tahowallet/extension/assets/23117945/25a5671e-aeff-4baa-9fef-943b3f5006e2">

_Originally posted by @kkosiorowska in https://github.com/tahowallet/extension/issues/3497#issuecomment-1607094625_
            
![Screenshot 2023-06-26 at 10 08 10 AM](https://user-images.githubusercontent.com/80375331/248830809-6b90bace-3b09-4a07-9f25-5b6e5182d43c.png)

@kkosiorowska thanks for explaining, but the issue is still present. In this wallet there are plenty of OETH tokens but the token still does not show on the wallet tab

sumpg41:
"Hide asset balances under 2$" is off by the way

Issue #3391, opened 2023-05-18
Token list is not getting updates


jagodarybacka:
Token list saved in extension is not updated with new content

[Extension's token list ](https://ipfs.io/ipfs/bafybeigtlpxobme7utbketsaofgxqalgqzowhx24wlwwrtbzolgygmqorm)

[New token list](https://ipfs.io/ipfs/bafybeianu4bww63ayktccmrwqzyvabgbrqwerswxm5ycn5txyfyoakt3ta) - +30 new tokens

Shadowfiend:
Updating the token list right now requires updating the underlying IPFS address in the extension, as it's content-addressed so updates in content change the address.

PR #3506, opened 2023-06-29
Update Taho token list reference

kkosiorowska:
Closes https://github.com/tahowallet/extension/issues/3391
Closes https://github.com/tahowallet/extension/issues/3501

## What
The current list of tokens in the application is not updated. We should update it every time new tokens are added.

## UI

Before
Screenshot 2023-06-29 at 11 39 16

After
Screenshot 2023-06-29 at 11 39 56

## Testing
Test address - `0x23Dd978A22f81e29941358Dbb0eAde9F53389Eb2`
- [x] Install the new app from the `main`. 
- [x] Do migration to the PR branch.
- [x] The OETH token should be verified by default because it belongs to the new token list.
- [x] Check if the URL has been updated in db.

Latest build: [extension-builds-3506](https://github.com/tahowallet/extension/suites/13972850715/artifacts/779111374) (as of Fri, 30 Jun 2023 08:19:57 GMT).

### DIFF
--- a/background/services/preferences/db.ts
+++ b/background/services/preferences/db.ts
@@ -369,6 +369,7 @@ export class PreferenceDatabase extends Dexie {
             "bafybeihufwj43zej34itf66qyguq35k4f6s4ual4uk3iy643wn3xnff2ka"
           )
 
+          // Param reassignment is the recommended way to use `modify` https://dexie.org/docs/Collection/Collection.modify()
           // eslint-disable-next-line no-param-reassign
           storedPreferences.tokenLists = {
             ...storedPreferences.tokenLists,
--- a/background/services/preferences/db.ts
+++ b/background/services/preferences/db.ts
@@ -20,6 +20,24 @@ const getSignerRecordId = (signer: AccountSignerWithId): SignerRecordId => {
   return `${signer.type}/${id}`
 }
 
+/**
+ * Update Taho token list reference.
+ * Returns an updated URLs for the token list.
+ */
+const getNewUrlsForTokenList = (
+  storedPreferences: Preferences,
+  oldPath: string,
+  newPath: string
+): string[] => {
+  // Get rid of old Taho URL
+  const newURLs = storedPreferences.tokenLists.urls.filter(
+    (url) => !url.includes(oldPath)
+  )
+  newURLs.push(`https://ipfs.io/ipfs/${newPath}`)
+
+  return newURLs
+}
+
 // The idea is to use this interface to describe the data structure stored in indexedDb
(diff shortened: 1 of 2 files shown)


background/services/preferences/db.ts
```diff
@@ -346,6 +356,27 @@ export class PreferenceDatabase extends Dexie {
...
+        .modify((storedPreferences: Preferences) => {
+          const newURLs = getNewUrlsForTokenList(
+            storedPreferences,
+            // Old path
+            "bafybeigtlpxobme7utbketsaofgxqalgqzowhx24wlwwrtbzolgygmqorm",
+            // New path
+            "bafybeihufwj43zej34itf66qyguq35k4f6s4ual4uk3iy643wn3xnff2ka"
+          )
+
+          // eslint-disable-next-line no-param-reassign
```
Shadowfiend, line 372: I know we've made this a pattern, but let's add a comment above each of these eslint-disables explaining why we're assigning to the param. I know it seems repetitive, but… Well, that's all the more reason to make it so we don't have to keep updating a manual reference. Will follow up on that piece in Discord.

jagodarybacka (dismissed):
Token list updated ✅ There is a problem with fetching the price for that asset unfortunately but I assume that's a different bug.

Shadowfiend:
Gotta follow that approval with a merge! 😁

jagodarybacka:
As we reviewed that at the same time I wanted to let @kkosiorowska add these comments you requested 😄

background/services/preferences/db.ts
```diff
@@ -346,6 +356,27 @@ export class PreferenceDatabase extends Dexie {
...
+      return tx
+        .table("preferences")
+        .toCollection()
+        .modify((storedPreferences: Preferences) => {
+          const newURLs = getNewUrlsForTokenList(
+            storedPreferences,
+            // Old path
+            "bafybeigtlpxobme7utbketsaofgxqalgqzowhx24wlwwrtbzolgygmqorm",
+            // New path
+            "bafybeihufwj43zej34itf66qyguq35k4f6s4ual4uk3iy643wn3xnff2ka"
```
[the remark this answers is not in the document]
kkosiorowska: yeah, I agree. Maybe it should be good to create separate a link where will be hosted actual token list and avoid these migrations. It's just a thought.

background/services/preferences/db.ts
kkosiorowska: Param reassignment is the recommended way to use `modify` https://dexie.org/docs/Collection/Collection.modify() I will add a correct information comment for it.

jagodarybacka (approved):

PR #3508, opened 2023-06-30
Fix custom asset price fetching

hyphenized:
- ~Fixes price fetching of assets that were first discovered as custom assets and then retrieved from token lists~ Now in #3534. 
- ~Fixes a bug that caused token list assets to have a discovery transaction hash and to also be included as custom assets.~
- Enables price fetching for all trusted assets: base network, token list, and custom assets that have been verified.

## To Test
- [ ] Verify that some custom assets have price information (an easy way to test this would be to remove all token lists from preference service defaults and reinstall the wallet)

Latest build: [extension-builds-3508](https://github.com/tahowallet/extension/suites/14225221756/artifacts/798442312) (as of Tue, 11 Jul 2023 20:36:43 GMT).


background/services/chain/tests/index.integration.test.ts
```diff
@@ -57,15 +57,15 @@ describe("ChainService", () => {
 
       const initializeBaseAssets = sandbox.spy(
         chainServiceInstance.db,
-        "initializeBaseAssets"
+        "initializeBaseAssets" as keyof ChainDatabase
```
kkosiorowska, line 60: hmm I wonder how we can improve this. 🤔 I'm thinking about using something like below. But I'm not sure what is better.

```
type ChainDatabaseExternalized = Omit<ChainDatabase, ""> & {
  initializeBaseAssets: () => Promise<void>
}

type ChainServiceExternalized = Omit<ChainService, ""> & {
  db: ChainDatabaseExternalized
}
```

kkosiorowska (commented):
I tested the solution and it looks like the prices load correctly after changing the branch. 🚀 In my case, it took quite a long time to reload the prices. I left some small thoughts.

background/services/chain/db.ts
```diff
@@ -39,6 +39,12 @@ type AccountAssetTransferLookup = {
 // TODO keep track of transaction replacement / nonce invalidation
 
 export class ChainDatabase extends Dexie {
+  static defaultSettings = {
```
jagodarybacka, line 42: Why do we need this?

background/redux-slices/utils/asset-utils.ts
```diff
@@ -391,6 +402,19 @@ export const getAssetID = (
   return `erc20/${asset.contractAddress}`
 }
 
+/**
+ * Returns a string that can be used as a unique identifier for an asset
+ */
+export const getFullAssetID = (
```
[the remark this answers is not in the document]
hyphenized: This is still used for balances, so we'd need a migration to remove it completely. The difference is that `getFullAssetID` includes the chainID. This wasn't necessary at the time since balances are already indexed by `chainID`. However, we can remove this in #3526

background/services/chain/tests/index.integration.test.ts
hyphenized: Would be easier to just ditch access modifiers in favour of dangling underscores cc @Shadowfiend

background/redux-slices/utils/asset-utils.ts
```diff
@@ -342,14 +342,18 @@ export function heuristicDesiredDecimalsForUnitPrice(
...
+export function isTokenListAsset(asset?: AnyAsset): boolean {
+  return !!asset?.metadata?.tokenLists?.length
+}
+
 /**
  * Check if the asset has a list of tokens.
  * Assets that do not have it are considered untrusted.
  *
  */
 export function isUntrustedAsset(asset: AnyAsset | undefined): boolean {
```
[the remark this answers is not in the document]
hyphenized: This pattern allows us to do `isUntrustedAsset(mightBeUndefined)` instead of `mightBeUndefined && isUntrustedAsset(...)` within components.

background/services/chain/db.ts
hyphenized: It's much easier to stub this during testing if it's exposed as a property on the class

background/services/chain/tests/index.integration.test.ts
Shadowfiend: If we're needing private access for integration tests we are either writing bad tests or not providing correct access modifiers. My guess is we're writing bad integration tests (because integration tests are notoriously difficult to write at the correct abstraction level).

I'll try to look in more detail tomorrow. In the meantime, I both agree that the instinct to avoid `as` is correct, and I think we can bear the `as` in the context of the test since it has lower implication on production behavior.

background/redux-slices/utils/asset-utils.ts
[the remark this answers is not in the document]
jagodarybacka: I know but still, it seems like additional logic this function shouldn't be responsible for, also the function name and comment above don't indicate that we have this kind of logic here.

background/redux-slices/utils/asset-utils.ts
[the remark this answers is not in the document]
jagodarybacka: I would say it would be clearer if we could rename it to something like `getAssetIDwithChainID` or so - agreed that `full` is not explaining much and if someone will decide to use it elsewhere then it can be confusing why this and not the other one. But if we plan to remove one of the functions soon then it's fine.

background/redux-slices/utils/asset-utils.ts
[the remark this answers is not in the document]
hyphenized: Yeah also this makes it harder to reuse this function as `!isUntrustedAsset` to determine trust because if something is `undefined` it will return `true`

background/services/indexing/index.ts
```diff
@@ -701,19 +704,8 @@ export default class IndexingService extends BaseService<Events> {
...
-      const existingDiscoveryTxHash = addressForDiscoveryTxHash
-        ? knownAsset.metadata?.discoveryTxHash?.[addressForDiscoveryTxHash]
-        : undefined
-      // If the discovery tx hash is not specified
-      // or if it already exists in the asset, do not update the asset
-      if (!newDiscoveryTxHash || existingDiscoveryTxHash) {
-        await this.addAssetToTrack(knownAsset)
-        return knownAsset
-      }
+      await this.addAssetToTrack(knownAsset)
```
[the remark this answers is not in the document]
kkosiorowska: Please take a look at this https://github.com/tahowallet/extension/pull/3534

background/services/indexing/index.ts
```diff
@@ -823,19 +826,25 @@ export default class IndexingService extends BaseService<Events> {
...
-      if (customAssetsById.has(getAssetId(asset))) {
+      if (
+        customAssetsById.has(getFullAssetID(asset)) &&
+        // Only filter custom assets which do not appear in token lists
+        !isTokenListAsset(
+          this.getKnownSmartContractAsset(
+            asset.homeNetwork,
+            asset.contractAddress
+          )
+        )
```
kkosiorowska, line 846: I'm not sure how best to solve this. Because https://github.com/tahowallet/extension/pull/3491 improves the logic of trusted assets and is also in review. But we should also fetch prices for untrusted but verified assets. We will probably fix this in the next PR when both changes are merged.

Shadowfiend: All trusted assets should get price-checked, whether they were on token lists or not (I made this remark in 3491 as well).

What I'm not 100% sure about is—does that mean this PR is unnecessary?

hyphenized: This PR was made to address the same issue in #3534 which caused price fetching problems for token list assets. 

Since we should fetch prices for verified assets, it makes sense that this fix should address that now too.

background/services/indexing/index.ts
```diff
@@ -823,19 +826,25 @@ export default class IndexingService extends BaseService<Events> {
...
     const customAssets = await this.db.getActiveCustomAssetsByNetworks(
       trackedNetworks
     )
 
-    const customAssetsById = new Set(customAssets.map(getAssetId))
+    const customAssetsById = new Set(customAssets.map(getFullAssetID))
 
     // Filter all assets based on supported networks
     const activeAssetsToTrack = assetsToTrack.filter((asset) => {
       // Skip custom assets
```
Shadowfiend, line 837: Ok I have found the most specific version of my confusion! The answer may be due to recent changes making this unnecessary. 

The question is: why are we skipping custom assets here at all? Any custom assets? We're burning a lot of logic where it feels like we could just be filtering by isTrusted and calling it a day--no looking up the custom asset list, no adding a new id type, etc--just `assetsToTrack.filter((asset) => isTrusted(asset) && trackedNetworks.some((trackedNetwork) => isAssetOnNetwork(asset, trackedNetwork)))`.

@kkosiorowska also just updated [a TODO on CustomAssetListItem](https://github.com/tahowallet/extension/pull/3528/files#diff-18fa31f65c43e5d1ee48dd39838ed8d58657f986558126123c153b920b3f5639R89) that seems like it could be nuked with that change.

hyphenized: > The question is: why are we skipping custom assets here at all? Any custom assets? We're burning a lot of logic where it feels like we could just be filtering by isTrusted and calling it a day--no looking up the custom asset list, no adding a new id type, etc--just assetsToTrack.filter((asset) => isTrusted(asset) && trackedNetworks.some((trackedNetwork) => isAssetOnNetwork(asset, trackedNetwork))).

This change was made before the decision to start fetching prices for verified assets

Shadowfiend: Amazing!

background/redux-slices/utils/asset-utils.ts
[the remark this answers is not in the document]
hyphenized: Removed from this PR in favor of improving it in https://github.com/tahowallet/extension/pull/3526

Shadowfiend (approved):
I haven't had a chance to do extensive testing here, but everything about this seems right and everything about the smells right, so I suggest we do some light smoke testing merge and then let the release testing pick everything else up.

kkosiorowska:
> Enables price fetching for all trusted assets: base network, token list, and custom assets that have been verified.

@hyphenized  This point wasn't marked as being done. It seems that after verification we still don't have prices for such an asset. I'm not sure what the next steps are for this and whether we shouldn't create a task to not lose this.

PR #3513, opened 2023-06-30
Fix throwing NFT filters panel

jagodarybacka:
### What

Follow up for https://github.com/tahowallet/extension/pull/3486

Add fallback from name to address in NFT filters. To avoid NFT filters throwing error when the account is not yet loaded let's add a fallback from name to address to put in the SharedToggle label.

### Testing

- [x] install extension and throttle network so it is super slow 🐌 
- [x] add account and while it is loading quickly go to NFTs and try to click on the filters panel as soon as it is unblocked
- [x] filters panel should not throw

Latest build: [extension-builds-3513](https://github.com/tahowallet/extension/suites/14022443935/artifacts/782867040) (as of Mon, 03 Jul 2023 10:15:53 GMT).

### DIFF
--- a/background/redux-slices/utils/nfts-utils.ts
+++ b/background/redux-slices/utils/nfts-utils.ts
@@ -38,7 +38,7 @@ export const getAdditionalDataForFilter = (
   accounts: AccountData[]
 ): { name?: string; thumbnailURL?: string } => {
   const a = accounts.find(({ address }) => address === id)
-  return a ? { name: a.name, thumbnailURL: a.avatarURL } : {}
+  return a ? { name: a.name ?? a.address, thumbnailURL: a.avatarURL } : {}
 }
 
 /* Items are sorted by price in USD. All other elements are added at the end. */
--- a/background/redux-slices/utils/nfts-utils.ts
+++ b/background/redux-slices/utils/nfts-utils.ts
@@ -7,15 +7,10 @@ import {
   NFTCollectionCached,
   SortType,
 } from "../nfts"
+import { AccountTotal } from "../selectors/accountsSelectors"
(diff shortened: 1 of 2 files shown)


background/redux-slices/utils/nfts-utils.ts
```diff
@@ -38,7 +38,7 @@ export const getAdditionalDataForFilter = (
   accounts: AccountData[]
 ): { name?: string; thumbnailURL?: string } => {
   const a = accounts.find(({ address }) => address === id)
-  return a ? { name: a.name, thumbnailURL: a.avatarURL } : {}
+  return a ? { name: a.name ?? a.address, thumbnailURL: a.avatarURL } : {}
```
Shadowfiend, line 41: Here's the typing miss:

https://github.com/tahowallet/extension/blob/main/background/redux-slices/selectors/nftsSelectors.ts#L49

Looks like that `as` can go away if we update `AccountData` to reflect the possibility of `undefined` that we're handling in this change:

https://github.com/tahowallet/extension/blob/main/background/redux-slices/utils/nfts-utils.ts#L16

```
TS2345: Argument of type 'AccountTotal[]' is not assignable to parameter of type 'AccountData[]'.
  Type 'AccountTotal' is not assignable to type 'AccountData'.
    Types of property 'name' are incompatible.
      Type 'string | undefined' is not assignable to type 'string'.
        Type 'undefined' is not assignable to type 'string'.
    47 |       const additionalData = getAdditionalDataForFilter(
    48 |         filter.id,
  > 49 |         accountTotals
       |         ^^^^^^^^^^^^^
    50 |       )
```

Shadowfiend (approved):
Looks good, shipping it 🚢

Issue #3515, opened 2023-07-03
High gas fees on Optimism


jagodarybacka:
These txs paid way too high gas fee for a swap, to be investigated.

https://optimistic.etherscan.io/tx/0x01236edd7388d6306e0920bd7a410732f52e3c6f2d4f3a73509f40991b910aeb
https://optimistic.etherscan.io/tx/0xb5621da06d967ccb7e804a8f460e9e75024f0cd18ad9fe6bec77052b115caeef

am1993421:
https://optimistic.etherscan.io/tx/0x129560153ddb0d7f4de31af9870bb217084a8adf184cf98787d632d32cc616a1
https://optimistic.etherscan.io/tx/0x82d083afaeca0a3f94d07e337ece3cd90864439bc4fbd8b61da2e757495851c5

Additions to the same issue

am1993421:
https://optimistic.etherscan.io/tx/0x129560153ddb0d7f4de31af9870bb217084a8adf184cf98787d632d32cc616a1
https://optimistic.etherscan.io/tx/0x82d083afaeca0a3f94d07e337ece3cd90864439bc4fbd8b61da2e757495851c5

Additions to the same issue

PR #3516, opened 2023-07-03
Use gas price from 0x quote to setup swap transaction

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3515

### What
Let's not override gas price that 0x gave us for Optimism transactions, as our existing fee handling for Optimism has been broken by the EIP-1559 functionality added in the last network upgrade.

### Testing

This should be tested on all networks that are supporting swaps. This can potentially result in high gas fees so you can block broadcasting transaction to avoid loosing money for testing (for example throwing error [here](https://github.com/tahowallet/extension/blob/swaps-on-op/background/services/chain/index.ts#L1190) and logging transaction instead) You should check if gas prices are matching market values.

Latest build: [extension-builds-3516](https://github.com/tahowallet/extension/suites/14029858033/artifacts/783404247) (as of Mon, 03 Jul 2023 15:11:30 GMT).

### DIFF
--- a/ui/pages/Swap.tsx
+++ b/ui/pages/Swap.tsx
@@ -347,16 +347,11 @@ export default function Swap(): ReactElement {
       )) as unknown as AsyncThunkFulfillmentType<typeof fetchSwapQuote>
 
       if (finalQuote) {
-        const { gasPrice, ...quoteWithoutGasPrice } = finalQuote
-
         await dispatch(
           executeSwap({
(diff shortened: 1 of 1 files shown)


Shadowfiend (approved):
Fix here is perfectly targeted, let's get it done. Will do a test drive shortly.

Shadowfiend:
Tested and confirmed our use of the 0x-provided gas price. Let's rock and roll :guitar:

PR #3527, opened 2023-07-05
Fix migration for cleaning account balances

jagodarybacka:
### What
Migration should be ready to handle accounts that are in a loading state without throwing and breaking the extension.

image

Latest build: [extension-builds-3527](https://github.com/tahowallet/extension/suites/14071112936/artifacts/786522923) (as of Wed, 05 Jul 2023 09:15:07 GMT).


jagodarybacka:
I think we should be using types from redux slices in the migrations to avoid errors like this that are hard to notice without manually comparing types 🤔

kkosiorowska (approved):

PR #3528, opened 2023-07-05
Refactor of terms for verified/unverified assets

kkosiorowska:
## What

Currently, some of the terms and conditions for untrusted assets have become unclear. This should be simplified and described. There are three concepts baseline trusted assets, trusted assets, and verified assets.

- Baseline trusted means “the wallet ~trusts this”—i.e., the asset is in a token list OR the asset is a network base asset.
- Verified means the user has explicitly verified the asset.
- Trusted means the asset is baseline trusted OR verified.

The negatives:

- Baseline untrusted means the asset is !baseline trusted—i.e. “the wallet does not trust this without an explicit user signal”.
- Unverified means the asset is !verified—i.e., the user has not explicitly verified the asset. It can still be baseline trusted.
- Untrusted means the asset is !trusted—i.e. the asset is neither baseline trusted NOR verified.

More information can be found [here](https://github.com/tahowallet/extension/pull/3491#discussion_r1243052806).

## Testing
The changes should not affect the behavior of the app, it should be the same as before the changes. 

Wallet page

- [x] Only trusted assets should be on the list on the wallet page. 
- [x] Enable the display of unverified assets in the settings. Only unverified assets have verification options on the wallet page.
- [x] Untrusted assets should not be on the list of send assets.
- [x] Untrusted assets should not be on the list of assets to be swapped.
- [x] Verify the untrusted asset. The asset should appear on the list of assets to be swapped and send.
- [x] Warning slide-up menu for unverified/verified is showing the correct buttons. 

Single asset page

- [x] The basic network asset - visible send and swap buttons
- [x] Token list asset - visible send and swap buttons
- [x] Untrusted asset - visible verify status label and button
- [x] Verified asset - visible send and swap buttons and verify status label

Feature flag disabled - `SUPPORT_UNVERIFIED_ASSET=false`
- [ ] Untrusted assets should be on the list of send assets.
- [ ] Untrusted assets should be on the list of assets to be swapped.
- [ ] Enable the display of unverified assets in the settings. Only unverified assets have a verification warning on the wallet page.
- [ ] All assets should have visible send and swap buttons on the single asset page

Latest build: [extension-builds-3528](https://github.com/tahowallet/extension/suites/14211360622/artifacts/797370444) (as of Tue, 11 Jul 2023 12:05:49 GMT).


background/redux-slices/selectors/0xSwapSelectors.ts
```diff
@@ -29,21 +27,15 @@ export const selectSwapBuyAssets = createSelector(
...
-        if (
-          // Explicitly add a network's base asset.
-          isBuiltInNetworkBaseAsset(asset, currentNetwork)
-        ) {
-          return true
-        }
-        return false
+        return (
+          // When the flag is disabled all assets can be sent and swapped
+          (!isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET) ||
```
jagodarybacka, line 32: Can we set up a PR that will remove the flag? That will further simplify the logic in many places

jagodarybacka (commented):
I would vote to write down in an RFB or in some other obvious place, in plain English, what we mean by `verified` `trusted` `baselineTrusted` or so. I know there are already comments in the `asset-utils` but I think it would be nice to have a glossary somewhere.

background/redux-slices/utils/asset-utils.ts
```diff
@@ -343,37 +342,72 @@ export function heuristicDesiredDecimalsForUnitPrice(
...
 
+/**
+ * Checks if the asset is untrusted.
+ * Untrusted means the asset is neither baseline trusted NOR verified.
+ *
+ */
+export function isUntrustedAsset(asset?: AnyAsset): boolean {
+  if (asset) {
+    return !isTrustedAsset(asset)
+  }
```
hyphenized, line 410: Going to suggest that we remove `undefined` from these functions as it's going to be confusing and bug prone the fact that `!isUntrustedAsset` can't be used interchangeably as `isTrustedAsset`.

e.g.
```ts
  isUntrustedAsset(trustedAsset) => false
  !isUntrustedAsset(trustedAsset) => true
  isUntrustedAsset(undefined) => false
  !isUntrustedAsset(undefined) => true // `undefined` is not a trusted asset
```

ui/components/Wallet/AssetListItem/CommonAssetListItem.tsx
```diff
@@ -52,7 +52,7 @@ export default function CommonAssetListItem(
       ? assetAmount.asset.contractAddress
       : undefined
 
-  const isUnverified = isUnverifiedAssetByUser(assetAmount.asset)
+  const isUntrusted = isUntrustedAsset(assetAmount.asset)
```
hyphenized, line 55: If untrusted assets are now custom assets that haven't been verified, then this will show prices for verified custom assets right?

ui/pages/Send.tsx
```diff
@@ -112,7 +112,9 @@ export default function Send(): ReactElement {
       (assetAmount): assetAmount is CompleteAssetAmount<FungibleAsset> =>
         isFungibleAssetAmount(assetAmount) &&
         assetAmount.decimalAmount > 0 &&
-        canBeUsedForTransaction(assetAmount.asset)
+        // When the flag is disabled all assets can be sent and swapped
+        (!isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET) ||
+          isTrustedAsset(assetAmount.asset))
```
hyphenized, line 117: This looks much better than `canBeUsedForTransaction` 👌

ui/utils/swap.ts
```diff
@@ -271,7 +272,9 @@ export function getOwnedSellAssetAmounts(
         (isSmartContractFungibleAsset(assetAmount.asset) ||
           assetAmount.asset.symbol === currentNetwork.baseAsset.symbol) &&
         assetAmount.decimalAmount > 0 &&
-        canBeUsedForTransaction(assetAmount.asset)
+        // When the flag is disabled all assets can be sent and swapped
+        (!isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET) ||
+          isTrustedAsset(assetAmount.asset))
```
hyphenized, line 277: We're allowing swapping for custom assets that are verified?

ui/pages/SingleAsset.tsx
```diff
@@ -98,14 +98,23 @@ export default function SingleAsset(): ReactElement {
...
 
   const showActionButtons = isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET)
-    ? !isUnverifiedByUser
+    ? isTrusted
     : true
+  const showVerificationButton = isEnabled(
+    FeatureFlags.SUPPORT_UNVERIFIED_ASSET
+  )
+    ? !isVerified && !isTrusted
+    : false
```
hyphenized, line 113: If an asset is not trusted, then it must be a custom asset that has not been verified, hence we should be able to refactor this to:

```suggestion
const showVerificationButton =
    isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET) && !isTrusted
```

ui/pages/SingleAsset.tsx
```diff
@@ -98,14 +98,23 @@ export default function SingleAsset(): ReactElement {
...
+
+  const showVerifiedLabel = isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET)
+    ? isVerified
+    : false
```
hyphenized, line 117:
```suggestion
  const showVerifiedLabel =
    isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET) && isVerified
```

Would also like to suggest inlining these checks if they're not too complex. Non-blocking

hyphenized (changes requested):
Code looks great, left some suggestions. Will give it a test run shortly

background/redux-slices/selectors/0xSwapSelectors.ts
Shadowfiend: Given we've had the feature flag kicking around, I'd be comfortable just killing it in this PR tbh.

background/redux-slices/accounts.ts
```diff
@@ -190,9 +190,9 @@ function updateCombinedData(immerState: AccountState) {
       /**
        * Asset amounts can be aggregated if the asset is a base network asset
        * or comes from a token list, e.g. ETH on Optimism, Mainnet
+       * or has been verified by the user
        */
-      const canBeAggregated =
-        isNetworkBaseAsset(asset) || !isUntrustedAsset(asset)
+      const canBeAggregated = isNetworkBaseAsset(asset) || isTrustedAsset(asset)
```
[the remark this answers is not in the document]
Shadowfiend: It feels like the whole concept of `canBeAggregated` can be nuked—do we need to compute a combined account balance for untrusted assets? It feels like before we run the `reduce` we should just do a `.filter(({ asset }) => isTrustedAsset(asset))`.

As far as I can tell, we use combined data to feed the overview page (which makes sense, since it's the only place where we would aggregate by symbol across chains—though that's probably not a strictly safe strategy), and the overview page never shows unverified assets.

background/redux-slices/selectors/0xSwapSelectors.ts
```diff
@@ -29,21 +27,15 @@ export const selectSwapBuyAssets = createSelector(
...
+            isTrustedAsset(asset)) &&
+          // Only list assets for the current network.
+          (isBuiltInNetworkBaseAsset(asset, currentNetwork) ||
+            (isSmartContractFungibleAsset(asset) &&
+              sameNetwork(asset.homeNetwork, currentNetwork)))
+        )
```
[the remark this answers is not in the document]
Shadowfiend: I'd be fine moving part after `&&` to another named constant (i.e. `const assetIsOnCurrentNetwork = isNetworkBase(...) || isSmartContractFungibleAsset(asset) && ...`) or even adding an `assetIsOnNetwork` helper, but nested ifs for simple boolean checks are going to be a no from me.

background/redux-slices/utils/asset-utils.ts
```diff
@@ -343,37 +342,72 @@ export function heuristicDesiredDecimalsForUnitPrice(
...
-    if (asset.metadata?.verified !== undefined) {
-      // If we have verified metadata return it
-      return !asset.metadata.verified
-    }
+export function isVerifiedAsset(asset?: AnyAsset): boolean {
+  if (asset?.metadata?.verified !== undefined) {
+    // If we have verified metadata return it
+    return asset.metadata.verified
+  }
+  return false
```
Shadowfiend, line 375: This really just means `return asset?.metadata?.verified !== undefined && asset.metadata.verified`.

background/redux-slices/utils/asset-utils.ts
Shadowfiend: Strong agreement with of this. Even though it's convenient to be able to pass `undefined`, these functions are security-sensitive and should do the minimum work we can squeeze into them.

background/redux-slices/utils/asset-utils.ts
```diff
@@ -343,37 +342,72 @@ export function heuristicDesiredDecimalsForUnitPrice(
...
 
 /**
- * Check if the asset has a list of tokens.
- * Assets that do not have it are considered untrusted.
+ * Check if the asset is from a token list.
  *
  */
-export function isUntrustedAsset(asset: AnyAsset | undefined): boolean {
+export function isTokenListAsset(asset?: AnyAsset): boolean {
+  return !!asset?.metadata?.tokenLists?.length
```
Shadowfiend, line 349: `!!` really bugs me. I really feel like this would be clearer as:

```typescript
  const tokenListCount = asset?.metadata?.tokenLists?.length ?? 0

  return tokenListCount > 0
```

That said, this both isn't blocking and if you disagree we don't have to discuss it further haha.

background/redux-slices/utils/tests/asset-utils.unit.test.ts
```diff
@@ -1,50 +1,77 @@
...
-          tokenLists: [],
-          websiteURL: "",
-        },
-      }
-      expect(isUnverifiedAssetByUser(asset)).toBeTruthy()
+  describe("isBaselineTrustedAsset", () => {
+    test("should return true if is a token list asset", () => {
+      const asset = createSmartContractAsset()
+
+      expect(isBaselineTrustedAsset(asset)).toBeTruthy()
```
Shadowfiend, line 16: For all of these, we're specifically looking for `toBe(true)` and `toBe(false)`, not truthy/falsy.

background/redux-slices/utils/tests/asset-utils.unit.test.ts
```diff
@@ -1,50 +1,77 @@
...
-              name: "",
-              url: "",
-            },
-          ],
-          websiteURL: "",
-          verified: true,
-        },
-      }
-      expect(isUnverifiedAssetByUser(asset)).toBeFalsy()
+    test("should return false if is not a token list asset", () => {
```
Shadowfiend, line 19:
```suggestion
    test("should return false if is neither a token list asset nor a network base asset", () => {
```

background/redux-slices/utils/tests/asset-utils.unit.test.ts
```diff
@@ -1,50 +1,77 @@
...
+      const asset = createSmartContractAsset({ metadata: {} })
+
+      expect(isBaselineTrustedAsset(asset)).toBeFalsy()
+    })
+
+    test("should return true if is a network base asset", () => {
```
Shadowfiend, line 25: Should add a test that we get `true` if it's both.

background/redux-slices/utils/tests/asset-utils.unit.test.ts
```diff
@@ -1,50 +1,77 @@
...
+      expect(isBaselineTrustedAsset(ETH)).toBeTruthy()
+    })
+  })
+
+  describe("isVerifiedAsset", () => {
+    test("should return true if is a verified asset", () => {
```
Shadowfiend, line 31:
```suggestion
    test("should return true if asset is explicitly verified", () => {
```

background/redux-slices/utils/tests/asset-utils.unit.test.ts
```diff
@@ -1,50 +1,77 @@
...
+      const asset = createSmartContractAsset({ metadata: { verified: true } })
+
+      expect(isVerifiedAsset(asset)).toBeTruthy()
+    })
+
+    test("should return false if is a unverified asset", () => {
```
Shadowfiend, line 37:
```suggestion
    test("should return false if asset is explicitly unverified", () => {
```

background/redux-slices/utils/tests/asset-utils.unit.test.ts
```diff
@@ -1,50 +1,77 @@
...
+
+    test("should return false if is a network base asset", () => {
+      expect(isVerifiedAsset(ETH)).toBeFalsy()
+    })
+
+    test("should return false if is a token list asset", () => {
+      const asset = createSmartContractAsset()
+
+      expect(isVerifiedAsset(asset)).toBeFalsy()
+    })
```
Shadowfiend, line 51: These are really testing around “if it has no explicit verification state”. I would make one `test(...)` that checks network base asset, token list asset, and perhaps more directly one with `createSmartContractAsset({ metadata: { verified: undefined } })`.

background/redux-slices/utils/tests/asset-utils.unit.test.ts
```diff
@@ -1,50 +1,77 @@
...
-    test("should return false if an asset is undefined", () => {
-      expect(isUnverifiedAssetByUser(undefined)).toBeFalsy()
+    test("should return true if is a network base asset", () => {
+      expect(isTrustedAsset(ETH)).toBeTruthy()
+    })
+
+    test("should return true if is a token list asset", () => {
+      const asset = createSmartContractAsset()
+
+      expect(isTrustedAsset(asset)).toBeTruthy()
```
Shadowfiend, line 74: We should add a check that it also returns false for an asset with no verification state, no token list, and that isn't a network base asset.

ui/components/Wallet/AssetListItem/CommonAssetListItem.tsx
Shadowfiend: I hope the answer is yes! :)

ui/pages/Send.tsx
Shadowfiend: Wait 'til the feature flag is killed!

ui/pages/SingleAsset.tsx
Shadowfiend: I actually think we should skip our easy booleans and tunnel in here. We should show the verification button only if `verified` is `undefined`: `asset?.metadata?.verified === undefined`.

My reasoning here is that, on the off chance that the user somehow finds themselves (through a bug!) viewing the single asset page for an asset that they have explicitly marked as not verified, we don't want to make it look exactly like a page whose trust level we don't actually know.

In a perfect world we would have some sort of “⚠️ Asset is not trusted” thing in the UI, but that's (1) out of scope here and (2),again, if they're viewing this asset page it means we've had a bug since we remove explicitly unverified assets from all lists that might lead you to this page.

ui/pages/SingleAsset.tsx
Shadowfiend: Agree, once the flag is dropped we can just drop `showVerifiedLabel` and use `isVerified` instead.

ui/pages/SingleAsset.tsx
```diff
@@ -117,21 +126,14 @@ export default function SingleAsset(): ReactElement {
...
-                <AssetVerifyToggler
-                  text={t("assets.verifiedByUser")}
-                  icon="notif-correct"
-                  color="var(--green-20)"
-                  hoverColor="var(--white)"
-                  onClick={() => setWarnedAsset(asset)}
-                />
-              )}
-          </>
+        {showVerifiedLabel && asset && isSmartContractFungibleAsset(asset) && (
```
Shadowfiend, line 129: I can't believe `asset` can be `undefined` here. We should really return early or show an error or throw or something if `asset` is undefined, way up at the top of the page rendering. I do now understand what led to `isVerifiedAsset` & co taking `undefined` though (let's still not do it!).

ui/utils/swap.ts
Shadowfiend: No reason we wouldn't? These assets should be equivalent to token list assets in every way (in fact, as I've mentioned a few times, the right long-term way to track custom assets is probably *as* a token list, which would simplify a lot of the logic we're throwing around even further).

background/redux-slices/selectors/0xSwapSelectors.ts
```diff
@@ -29,21 +27,15 @@ export const selectSwapBuyAssets = createSelector(
...
-        ) {
-          return true
-        }
-        return false
+        return (
+          // When the flag is disabled all assets can be sent and swapped
+          (!isEnabled(FeatureFlags.SUPPORT_UNVERIFIED_ASSET) ||
+            isTrustedAsset(asset)) &&
+          // Only list assets for the current network.
+          (isBuiltInNetworkBaseAsset(asset, currentNetwork) ||
```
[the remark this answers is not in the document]
Shadowfiend: `isNetworkBaseAsset` is too broad—we want to explicitly filter out base assets for networks other than the current one.

That said, `isBuiltInNetworkBaseAsset` is a misnomer. We should rename it to `isBaseAssetForNetwork(asset, network)`.

Shadowfiend (commented):
Left some thoughts in four broad categories:

- Extra/slightly tweaked tests.
- Adjusted/clarified ways of expressing boolean expressions.
- Old man yells at cloud.
- Agreeing with @hyphenized lol.

background/redux-slices/selectors/0xSwapSelectors.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/22ee963102b1f2248c53f0160fda380c7d4db899

ui/pages/SingleAsset.tsx
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/22ee963102b1f2248c53f0160fda380c7d4db899

ui/pages/Send.tsx
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/22ee963102b1f2248c53f0160fda380c7d4db899

background/redux-slices/selectors/0xSwapSelectors.ts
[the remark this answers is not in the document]
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/6c0b10c259a9f8f422ed3032e05e24e31cb14342

background/redux-slices/selectors/0xSwapSelectors.ts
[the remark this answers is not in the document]
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/6c0b10c259a9f8f422ed3032e05e24e31cb14342

ui/pages/SingleAsset.tsx
Shadowfiend: (Btw this is 100% out of scope for the PR.)

ui/pages/SingleAsset.tsx
```diff
@@ -181,7 +167,7 @@ export default function SingleAsset(): ReactElement {
             )}
           </div>
           <div className="right">
-            {showVerificationButton && isSmartContractFungibleAsset(asset) && (
+            {!isVerified && !isTrusted && isSmartContractFungibleAsset(asset) && (
```
Shadowfiend, line 170: `!isTrusted` implies `!isVerified`, right?

background/redux-slices/utils/tests/asset-utils.unit.test.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/9a5cf2e6b7c2417deb66ed807c2f4fe3a46b72b3

background/redux-slices/utils/tests/asset-utils.unit.test.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/9a5cf2e6b7c2417deb66ed807c2f4fe3a46b72b3

background/redux-slices/utils/tests/asset-utils.unit.test.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/9a5cf2e6b7c2417deb66ed807c2f4fe3a46b72b3

background/redux-slices/utils/tests/asset-utils.unit.test.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/9a5cf2e6b7c2417deb66ed807c2f4fe3a46b72b3

background/redux-slices/utils/tests/asset-utils.unit.test.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/9a5cf2e6b7c2417deb66ed807c2f4fe3a46b72b3

background/redux-slices/utils/asset-utils.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/c351ab0e88638ba2ae20943af1c35e536a7df049

background/redux-slices/utils/asset-utils.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/c351ab0e88638ba2ae20943af1c35e536a7df049

background/redux-slices/utils/asset-utils.ts
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/c351ab0e88638ba2ae20943af1c35e536a7df049

ui/pages/SingleAsset.tsx
kkosiorowska: Let's fix it later in https://github.com/tahowallet/extension/issues/3543.

ui/components/Wallet/AssetListItem/CommonAssetListItem.tsx
kkosiorowska: Yes, the price should be shown for verified assets as well. [Here](https://github.com/tahowallet/extension/blob/4e83cca1e85901a2b666fc2d40e98c663d3ec956/ui/components/Wallet/AssetListItem/CommonAssetListItem.tsx#L88-L104) is the logic for displaying the price for the asset.  However, this should also be included in the fix for fetching prices https://github.com/tahowallet/extension/pull/3508.

ui/utils/swap.ts
kkosiorowska: > We're allowing swapping for custom assets that are verified?

Yes, after verification, the user should be able to swap and send such an asset. Confirmed with Vlad.

background/redux-slices/accounts.ts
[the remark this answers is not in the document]
kkosiorowska: https://github.com/tahowallet/extension/pull/3528/commits/07e7f6fd797042970474da08b9acffa7eb4c3ed2

hyphenized (commented):
Tested and works well, there are still a couple of outstanding comments but otherwise looks good to go:

https://github.com/tahowallet/extension/pull/3528#discussion_r1256872328
https://github.com/tahowallet/extension/pull/3528#discussion_r1258658032

hyphenized (approved):

Shadowfiend:
I think neither of those are blocking and both see improvements already from the changes here; let's merge this fellow for the overall gains.

ui/pages/SingleAsset.tsx
kkosiorowska: There you are right. I overlooked that after the earlier changes.  :see_no_evil:

Issue #3466, opened 2023-06-12
Caching assets in redux is blocking UI


jagodarybacka:
Caching assets in redux is blocking UI and it can take even 1-2s.

https://github.com/tahowallet/extension/blob/main/background/main.ts#L1065

I've changed that function to:

```javascript
    this.indexingService.emitter.on("assets", async (assets) => {
      console.log(assets.length)
      const t1 = performance.now()
      await this.store.dispatch(assetsLoaded(assets))
      const t2 = performance.now()
      console.log("asssets took", t2 - t1)
    })
```

And I got these results:
<img width="300" alt="image" src="https://github.com/tahowallet/extension/assets/20949277/058b4495-5621-424a-8750-c9db5ebf63cb">
Sometimes it's even >2s:
<img width="344" alt="image" src="https://github.com/tahowallet/extension/assets/20949277/c46e46fe-814d-478b-af41-938d82e27a2a">

Let's try to refactor the code so at least UI is not getting blocked for so long.

_Originally posted by @jagodarybacka in https://github.com/tahowallet/extension/issues/3264#issuecomment-1584568809_

hyphenized:
Seems to be no longer an issue after #3561 and #3526, further improved by #3579

PR #3526, opened 2023-07-04
Refactor prices caching

hyphenized:
Relocates cached asset prices to a new redux slice. This reduces performance hits after price fetching by avoiding nested array diffs, which Immer struggles to manage. Consequently, it prevents unnecessary updates being serialized/deserialized.

## To Test
This clears cached data for both account balances and assets, after switching from `main` to this branch:
- [x] Check that prices and balances are loading and displaying normally
- [x] Check Swaps (Quotes)
- [x] Check Send page (Estimates)

Latest build: [extension-builds-3526](https://github.com/tahowallet/extension/suites/14603122227/artifacts/827664213) (as of Thu, 27 Jul 2023 06:33:36 GMT).

PR #3530, opened 2023-07-05
Refactor assets slice to fix performance problems

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3466
### What
As Redux was having serious performance problems when we were using an array to store all assets let's refactor that slice so it is now storing assets as an object that is indexed by assets' symbols. Seems like this is working great with redux and we are able to reduce CPU usage as we were updating assets very often.

### DIFF
--- a/background/redux-slices/assets.ts
+++ b/background/redux-slices/assets.ts
@@ -32,9 +32,9 @@ export type AssetWithRecentPrices<T extends AnyAsset = AnyAsset> = T & {
 
 export type SingleAssetState = AssetWithRecentPrices
 
-export type AssetsState = SingleAssetState[]
+export type AssetsState = { [symbol: string]: SingleAssetState[] }
 
-export const initialState = [] as AssetsState
+export const initialState = {} as AssetsState
 
 const assetsSlice = createSlice({
   name: "assets",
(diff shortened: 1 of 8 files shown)


Shadowfiend:
Looks like we have two competing approaches here with #3526 .

I directionally prefer the approach in 3526, because the question of “what do we use as the key for an asset” is one that seems easy and then inevitably gets complicated, even though we're trivializing it right now with `getAssetID`. We can see this already because we've had to introduce the `getFullAssetID` alternative, and future changes to supported chains or asset types are going to make things even more complicated. We're also in a scenario here where displaying “assets for this network” is a more taxing flow than with a flat list, a price that we're likely going to pay in unexpected and hard-to-detect ways down the line.

With that said, I'm interested in seeing how the two perform relative to each other, and, if this performs better, developing a little bit of a deeper understanding of *why* to guide future decisions for how redux is structured. My intuition would expect this approach to have the same performance characteristics of the current flat list, because I would expect any change of the `assets` object (even for a subkey) to trigger updates in the per-network assets list and full asset list, which in turn I would expect to trigger most of the same selectors as an update of the assets array does.

Obviously you're seeing performance improvements, so I would love to understand what *isn't* triggering here that *is* in the current approach. Maybe I can do some digging on my own there :)

jagodarybacka:
I'll dig further today but I think https://github.com/tahowallet/extension/pull/3526 should work fine with this PR - my current theory is that keeping assets as a flat list is making redux work really hard while merging state. Obviously, I have to confirm it to be sure. 
But overall if my theory is correct we can think what will be the best structure to store assets state, I'm far from claiming that indexing by the symbol is the best 😉

Shadowfiend:
The reason I didn't expect that to be the case is that if you look at the package of updates that comes across the wire it's a diff, so should be relatively few changes to the list, and I believe the changes should be in place. That might not be how it's happening in practice though!

jagodarybacka:
Ok so I've investigated it a little bit more and the problem is actually in the `webext-redux` package. We are using `deepDiff` strategy there and it is taking many repetition of the same function (mostly [this one](https://github.com/tshaddix/webext-redux/blob/master/src/strategies/deepDiff/arrayDiff/diff/lcs.js#L64)) to get proper diff for an array that is about 1k+ nested objects. 

https://github.com/tahowallet/extension/blob/load-assets-perf-fix/background/index.ts#L33

The second problem is that we are dispatching the `"assets"` event multiple times in a row which is unnecessary and we can probably batch these updates. Every time it is dispatched we give `webext-redux` another state to calculate diff for.

Seems like changing the structure of the redux slice is already making it easier to calculate diff but I'm looking for other options on how to handle it best.

The last part of solving this problem would be to make sure react is not triggered unnecessarily by these assets slice changes.

Shadowfiend:
> The second problem is that we are dispatching the "assets" event multiple times in a row which is unnecessary and we can probably batch these updates. Every time it is dispatched we give webext-redux another state to calculate diff for.

I… Have an idea here. We've looked at auto-batching actions and subscribers before and it's seemed very complicated and had some potential unintended consequences (see https://github.com/tahowallet/extension/pull/1296). But what I just realized is what want here isn't to batch *actions*, it's to batch *store updates*, because that's the expensive thing here. I think we can (ab)use the patch function to debounce… #3535 has a quick draft attempt.

jagodarybacka:
Ok, migration was added, this is good to go to QA.

Problems with debouncing state updates I think can be done in another PR as they are not that simple - let's see if this thing alone is making a visible difference.

hyphenized:
Ran a few tests on this, noted a few things:

The following tests were ran using tiny-bench and benchmarkjs, the later was dropped due to needing manual patching to work.

>- Loading state (hydrating)
Fills an empty state with the assets array
>- Updating state
Updates state by inserting 1 new asset
>- Diffing empty -> full
>Performs deepDiff over an empty state and an loaded state
>- Diffing loaded -> insert 1
Performs deepDiff over a loaded state and an updated state
>- Encoding/Decoding
Performs encodeJSON/decodeJSON




From all these tests the more important to us is the one that says `diffing (loaded -> insert 1)`, because state "hydration" is only done once and we're frequently dispatching small updates. 

- There's some benefit from this approach over the past implementation, which is seen in diffing hydration and diffing updates.
- Performance was best when using an object indexed by asset ids (20x improvement for diffing updates, 2x for state updates), for diff hydration the performance was actually worse but, since it's done only once I think it's worth the trade off.
- Major benefit came from using `produce` when updating nested objects and when working over an object instead of an array, though that is expected.


We're also keeping assets from all networks in state, which is not ideal.

These tests can be run with this patch:
[diff.zip](https://github.com/tahowallet/extension/files/12023739/diff.zip)

hyphenized:
Closed until reprioritization

PR #3534, opened 2023-07-07
Don't treat assets in the token list as unverified

kkosiorowska:
## What

During the new application installation, some of the assets in the token list were treated as unverified assets. 

Removal of the `discoveryTxHash` refresh logic. It is no longer needed. The change was to update the `discoveryTxHash` after the data migration, where `discoveryTxHash` was removed for all custom assets. This also resulted in the wrong adding of a `discoveryTxHash` for the asset in the token list. The change solves this issue.

## UI

**Before**

https://github.com/tahowallet/extension/assets/23117945/c8867a3f-0886-4b9a-a606-30eeba3bbd6b

**After**

https://github.com/tahowallet/extension/assets/23117945/3af9111f-22c9-4f61-a9a8-f6d0bbf2632e

## Testing
- [x] Install the app from `mian` branch. Assets in the token list should tread as unverified for a while time. Check the `assetsToTrack` and `customAssets` tables in the db. Find there, the asset which is in the token list. 

Change the branch. Check the `assetsToTrack` and `customAssets` tables in the db. 
 - [x] Assets in the token list should be removed. - `customAssets` table
 - [x] Assets in the token list should have updated metadata, no `discoveryTxHash`. - `assetsToTrack` table

Install a completely new app from the current branch. Check the `assetsToTrack` and `customAssets` tables in the db. 
 - [x] Assets in the token list shouldn't be in the `customAssets` table.
 - [x] Assets in the token list shouldn't have `discoveryTxHash` in  `assetsToTrack` table.

Latest build: [extension-builds-3534](https://github.com/tahowallet/extension/suites/14136247295/artifacts/791600079) (as of Fri, 07 Jul 2023 14:28:48 GMT).

### DIFF
--- a/background/services/indexing/db.ts
+++ b/background/services/indexing/db.ts
@@ -14,6 +14,7 @@ import {
 import { DeepWriteable } from "../../types"
 import { fixPolygonWETHIssue, polygonTokenListURL } from "./token-list-edit"
 import { normalizeEVMAddress } from "../../lib/utils"
+import { isNetworkBaseAsset } from "../../redux-slices/utils/asset-utils"
 
 /*
  * IndexedPricePoint extends PricePoint to expose each asset's ID directly for
@@ -255,14 +256,17 @@ export class IndexingDatabase extends Dexie {
     })
 
     // Assets in the token list shouldn't be in the customAssets table, let's remove them
-    // Also, such tokens shouldn't have a discoveryTxHash defined
+    // Also, such tokens and network base assets shouldn't have a discoveryTxHash defined
     this.version(7).upgrade(async (tx) => {
       const contractAddresses: string[] = []
       await tx
         .table("assetsToTrack")
         .toCollection()
         .modify((asset: SmartContractFungibleAsset) => {
-          if (Object.keys(asset.metadata?.discoveryTxHash ?? {}).length !== 0) {
+          if (
+            Object.keys(asset.metadata?.discoveryTxHash ?? {}).length !== 0 &&
+            (isNetworkBaseAsset(asset) || !!asset.metadata?.tokenLists?.length)
+          ) {
             // param reassignment is the recommended way to use `modify` https://dexie.org/docs/Collection/Collection.modify()
             // eslint-disable-next-line no-param-reassign
             delete asset.metadata?.discoveryTxHash
--- a/background/services/indexing/db.ts
+++ b/background/services/indexing/db.ts
@@ -253,6 +253,29 @@ export class IndexingDatabase extends Dexie {
           delete customAsset.metadata?.discoveryTxHash
         })
     })
+
+    // Assets in the token list shouldn't be in the customAssets table, let's remove them
+    // Also, such tokens shouldn't have a discoveryTxHash defined
+    this.version(7).upgrade(async (tx) => {
+      const contractAddresses: string[] = []
+      await tx
+        .table("assetsToTrack")
+        .toCollection()
+        .modify((asset: SmartContractFungibleAsset) => {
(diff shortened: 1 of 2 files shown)


background/services/indexing/db.ts
```diff
@@ -253,6 +253,29 @@ export class IndexingDatabase extends Dexie {
...
+        .modify((asset: SmartContractFungibleAsset) => {
+          if (Object.keys(asset.metadata?.discoveryTxHash ?? {}).length !== 0) {
+            // param reassignment is the recommended way to use `modify` https://dexie.org/docs/Collection/Collection.modify()
+            // eslint-disable-next-line no-param-reassign
+            delete asset.metadata?.discoveryTxHash
+            contractAddresses.push(asset.contractAddress)
+          }
+        })
+
+      tx.table("customAssets")
```
jagodarybacka, line 273: We want `await` here right?

jagodarybacka (changes requested):
Tested it manually and all is fine regarding token list's assets but none of the custom assets have discover tx hash after migration

background/services/indexing/db.ts
kkosiorowska: I'm not sure how best to save assets that we should remove from `customAssets`. `await` is needed here to get this info. If you have any suggestions on how to improve this, please let me know.

background/services/indexing/db.ts
```diff
@@ -253,6 +254,32 @@ export class IndexingDatabase extends Dexie {
           delete customAsset.metadata?.discoveryTxHash
         })
     })
+
+    // Assets in the token list shouldn't be in the customAssets table, let's remove them
+    // Also, such tokens and network base assets shouldn't have a discoveryTxHash defined
+    this.version(7).upgrade(async (tx) => {
+      const contractAddresses: string[] = []
+      await tx
+        .table("assetsToTrack")
```
hyphenized, line 263: Something to keep in mind here is that custom assets can also appear (and most should be due to balances syncing) in `assetsToTrack` so this is likely cleaning up custom assets as well. Additionally there should be no harm in leaving these assets in the custom assets table because, as mentioned in #3508, an asset might be discovered as a custom asset before being included in a token list. On the other hand if we still want to remove them, we should look at stored token list data and use that as the reference for removal.

Shadowfiend:
I'm struggling a little bit with the core motivation for this PR. With #3528 in place, we should be treating items as unverified only if they either have `verified: false | undefined` or they are not in a token list. All custom assets should have `verified: true` forcibly set, so it should be impossible for a custom *or* token list asset to be treated as unverified.

Is all of that right? Is the current behavior effectively fixed by #3528?

background/services/indexing/db.ts
kkosiorowska: @hyphenized  Yeah, good point. So what do you think about just removing the `discoveryTxHash` from assets that shouldn't have it from the `assetsToTrack` table?  The `customAssets` table will then be left as it is.

kkosiorowska:
> I'm struggling a little bit with the core motivation for this PR. With https://github.com/tahowallet/extension/pull/3528 in place, we should be treating items as unverified only if they either have verified: false | undefined or they are not in a token list. All custom assets should have verified: true forcibly set, so it should be impossible for a custom or token list asset to be treated as unverified.
> Is all of that right? Is the current behavior effectively fixed by https://github.com/tahowallet/extension/pull/3528?

Yes, everything is fine. The issue has been fixed [here](https://github.com/tahowallet/extension/pull/3491/files#diff-b0e4a7036b74ef74bb2061264a7470aa9dc941d5bb0b326556464bb28480fc80L703-R726). There isn't need to hold this PR open. Let's close it and, if there is a need for possible changes, let's do it in another PR.

PR #3535, opened 2023-07-07
WIP proposal: debounce background store updates to UI

Shadowfiend:
This is in many ways a much more targeted version of #1296 .

The observation feeding this PR is that the most expensive part of redux work right now is:

- Diffing.
- Serializing.
- Deserializing.
- Patching.

Doing this whole cycle on every action means that we can't structure actions freely, because we trigger the whole cycle on every action. We've experimented in the past with debouncing actions and subscribers at the cross-redux store level, but very quickly ran into edge cases with what happens on unsubscribe, timing issues, etc.

This approach eliminates all of that by *only* debouncing the state updates being pushed out by webext-redux. We do this by abusing the diff function on the background script side. The diff function now branches in two ways:
- If we have no time for a next pending update, set a marker for when the next pending update should trigger, and trigger a noop action on the store to make sure the update goes out.
- If we have a time for the next pending update and we haven't gotten to that time yet, return an empty diff.
- If we have a time and we've crossed that time, clear the marker for the next pending update, and return the full state diff between old and new.

The diff only affects the UI states, not the background state—the background state is wholly managed by redux. The old state tracked by webext-redux only changes when the diff is non-empty. We use these two facts to create a single batch of changes at most every 300ms (hardcoded right now).

This commit needs a little cleanup and a message that looks more like this PR description.

Latest build: [extension-builds-3535](https://github.com/tahowallet/extension/suites/14135695857/artifacts/791567013) (as of Fri, 07 Jul 2023 14:11:40 GMT).

### DIFF
--- a/background/main.ts
+++ b/background/main.ts
@@ -496,10 +496,33 @@ export default class Main extends BaseService<never> {
     // Start up the redux store and set it up for proxying.
     this.store = initializeStore(savedReduxState, this)
 
+    let nextUpdate: number | undefined
+
+    function debouncedDeepDiffFunction(
+      store: ReturnType<typeof initializeStore>
+    ) {
+      return (oldState: unknown, newState: unknown): unknown | [] => {
+        if (nextUpdate === undefined) {
+          nextUpdate = Date.now() + 300
+          window.setTimeout(() => store.dispatch({ type: "noop" }), 301)
+        }
+
+        if (Date.now() < nextUpdate) {
+          return []
+        }
+
+        // Sending a full diff, clear the next update marker.
+        nextUpdate = undefined
+
+        // @ts-expect-error booyan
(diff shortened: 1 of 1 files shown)


mhluongo:
Excited about this  — decoupling the UI update performance issue from the data structure issue is :ok_hand:

Shadowfiend:
Unfortunately poked at it a bit and the improvements aren't what I wanted to see, and when things are performing smoothly the 300ms delay is noticeable (so we can't batch even more). Needs more digging, as always with performance stuff. 

So far the biggest win I've seen has come from #3530, seemingly due to the cost of deep diffing arrays.

Shadowfiend:
Going to close this for now. Path seems like it could be promising, but measured outcomes seem less so.

PR #3541, opened 2023-07-10
Remove waiting for Loading Doggo component in E2E tests

jagodarybacka:
### What

If the NFT page is loading really fast then the "loading doggo" animation may not be visible for long enough to allow the e2e test to catch that element. As the loader itself is not necessary for the rest of the test let's omit testing for loader visibility, we can just test if the loader disappeared.

Latest build: [extension-builds-3541](https://github.com/tahowallet/extension/suites/14183949363/artifacts/795270419) (as of Mon, 10 Jul 2023 14:09:54 GMT).

### DIFF
--- a/e2e-tests/nfts.spec.ts
+++ b/e2e-tests/nfts.spec.ts
@@ -50,10 +50,9 @@ test.describe("NFTs", () => {
 
       await walletPageHelper.navigateTo("NFTs")
 
-      await expect(page.getByTestId("loading_doggo")).toBeVisible()
-
       // Wait until load finishes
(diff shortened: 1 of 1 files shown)


Shadowfiend (approved):
Is there any chance the loader might show up after the check? That could lead to intermittency as well. Meantime, shipping it as it's obviously helping rn.

hyphenized:
The doggo should appear because of the request intercept made [before](https://github.com/tahowallet/extension/pull/3541/files#diff-0c26d5cc022663ccdec363c4962989e260476470a63d34012131ce0a9ce95a51L15-L37). However, it is possible that this test is failing to intercept the request.

PR #3556, opened 2023-07-19
Switch default webext-redux deepDiff for jsondiffpatch

hyphenized:
On certain nested properties such as arrays, `deepDiff` generates a patch even if the contents are the same e.g. empty arrays. Since we're not relaying on referential identity to determine state updates, this causes a bunch of unnecessary updates after the diff gets patched into the UI state.

`jsondiffpatch` on the other hand only generates patches for structural changes, which minimizes the amount of patches processed. In order to have redux correctly process updates, we still need to update parent nodes to detect deeply nested changes, this can be achieved by applying generated patches through immer's produce function, which recursively updates parent nodes for applied changes.

Results can be reviewed within the attached traces:
[traces.zip](https://github.com/tahowallet/extension/files/12090817/traces.zip)

These can be generated using the still available chrome CPU profiler(though it is hidden under experiments within the chrome devtools settings)

Latest build: [extension-builds-3556](https://github.com/tahowallet/extension/suites/14422871896/artifacts/813717076) (as of Wed, 19 Jul 2023 20:15:36 GMT).


Shadowfiend (approved):
Checks out, let's do ittttt.

background/main.ts
```diff
@@ -508,10 +508,31 @@ export default class Main extends BaseService<never> {
...
+        }
+
+        const diff = deepDiff(lastState, newState)
+
+        if (diff !== undefined) {
+          updateFn(newState, [diff])
+        }
+      },
+      30,
+      { maxWait: 30, trailing: true }
```
Shadowfiend, line 524: `30` here should be a named constant, as a side note.

PR #3557, opened 2023-07-19
Do not open uninstall URL on development builds

hyphenized:
This is probably creating a bunch of false positives on posthog

Latest build: [extension-builds-3557](https://github.com/tahowallet/extension/suites/14399503257/artifacts/811880891) (as of Wed, 19 Jul 2023 04:11:40 GMT).


jagodarybacka:
> false positives

I don't think so as dev builds should be using DEV library (or how is it called on posthog) so we should be able to filter them, right? If not then probably we should do something about it on the posthog side as this is affecting other events as well.

But tbh I'm glad to avoid closing that window every time so 👍

hyphenized:
> > false positives
> 
> I don't think so as dev builds should be using DEV library (or how is it called on posthog) so we should be able to filter them, right? If not then probably we should do something about it on the posthog side as this is affecting other events as well.
> 
> But tbh I'm glad to avoid closing that window every time so 👍

That's true but we're opening the uninstall url for production, so the uuid passed is sent in an event with production parameters.

Shadowfiend (approved):
Would be better to have some sort of tweak to the URL to report it in dev mode, mostly so we can check that the website side of things is working as expected, but it's a small change and we might as well get on with it heh.

PR #3558, opened 2023-07-20
Buggity fixes: two small bug fixes

Shadowfiend:
Two fixes here, one for a small edge case in internal signer emission
that meant that if for some reason the internal signers data was cleared
in redux, it would never be correctly repopulated (expectation is that
there's no production scenario that would really trigger this).

The other is an issue where adding a network to the extension that didn't
have an associated spot price oracle listed would cause price lookups for
that network to fail, and this would cascade into failing all price lookups.

The correct fix for the latter is probably to make price lookups less sensitive
to individual failure, but for now I'm making the spot price oracle bit be
safe for networks that don't have an entry in the spot price oracle list.

Latest build: [extension-builds-3558](https://github.com/tahowallet/extension/suites/14438222069/artifacts/814921375) (as of Thu, 20 Jul 2023 10:37:18 GMT).

### DIFF
--- a/background/lib/priceOracle.ts
+++ b/background/lib/priceOracle.ts
@@ -152,6 +152,10 @@ const getRatesForTokens = async (
     }
   }[]
 > => {
+  if (SPOT_PRICE_ORACLE_CONSTANTS[network.chainID] === undefined) {
+    return []
+  }
+
   const multicall = new ethers.Contract(
     MULTICALL_CONTRACT_ADDRESS,
     MULTICALL_ABI,
--- a/background/services/internal-signer/index.ts
+++ b/background/services/internal-signer/index.ts
@@ -334,8 +334,6 @@ export default class InternalSignerService extends BaseService<Events> {
         this.#hiddenAccounts = {
           ...plainTextVault.hiddenAccounts,
         }
-
-        this.#emitInternalSigners()
       }
     }
 
@@ -347,6 +345,8 @@ export default class InternalSignerService extends BaseService<Events> {
     }
 
     this.#unlock()
+    this.#emitInternalSigners()
+
     return true
   }
 


background/services/internal-signer/index.ts
```diff
@@ -347,6 +345,8 @@ export default class InternalSignerService extends BaseService<Events> {
     }
 
     this.#unlock()
+    this.#emitInternalSigners()
```
jagodarybacka, line 348: Isn't it making `ignoreExistingVaults` argument useless? As we are always emitting signers after this change. RIght now we are only using it during password creation anyway which means there are no keys stored yet anyway. To keep this code clear I would remove `ignoreExistingVaults` argument.

background/lib/priceOracle.ts
```diff
@@ -152,6 +152,10 @@ const getRatesForTokens = async (
     }
   }[]
 > => {
+  if (SPOT_PRICE_ORACLE_CONSTANTS[network.chainID] === undefined) {
```
jagodarybacka, line 155: This looks safe but I couldn't really trigger this line from testing UI, what are the testing steps here?

jagodarybacka (approved):
Nothing really blocking, some questions:

background/services/internal-signer/index.ts
Shadowfiend: That argument would result in cleared vaults here, which would clear the signers in redux--which is what we would want in a scenario where we were resetting a password, which is the non-onboarding reason that argument is in place, see the function docs:

https://github.com/tahowallet/extension/blob/7543954c840545a9db716c484ea101869766caea/background/services/internal-signer/index.ts#L262-L263

We haven't implemented that yet but I don't think it's useful to remove it.

background/lib/priceOracle.ts
Shadowfiend: Whoops, did it in a hurry and forgot to add test instructions 🤦 Adding a custom asset on a custom chain and then waiting for the price alarm to trigger price checks I think will do the trick here. Background script should log an error without this, no error with it.

Shadowfiend:
Because the scope of the internal signers change is pretty tight here, I'm going to bypass review by @mhluongo in the interest of shipping.

PR #3559, opened 2023-07-20
Use JSON onboarding in e2e tests of transactions

michalinacienciala:
Prior to this change we were onboarding using a seed phrase, which was not ideal, as it could lead to a seed phrase leakage in the Playwright report added to the artifacts of CI jobs (that's why we've temporarily switched off storing of the test artifacts).
Now that we support in Taho the onboarding with a JSON keystore file (which stores an encrypted private key), we are good to use this method of account import in our E2E tests. This method is safer, as the sensitive data entered in the form during onboarding is obfuscated, so there's no risk it will leak in the Playwright report (and even if so, what will leak will be just a password, not a corresponding to it JSON file).
As we're switching to this onboarding method, we can bring back storing of the Playwright reports in the CI jobs artifacts.

Apart from the above, as part of this commit/PR we're also introducing a script which can be used to generate a JSON keystore file based on the provided private key and password.

TODO:

- [x] Add `TEST_WALLET_JSON_BODY ` and `TEST_WALLET_JSON_PASSWORD` secrets in the GitHub Settings
- [x] Rerun the failing CI jobs and make sure they're passing
- [ ] Delete the `TEST_WALLET_RECOVERY_PHRASE` secret

Latest build: [extension-builds-3559](https://github.com/tahowallet/extension/suites/14560436128/artifacts/824331784) (as of Tue, 25 Jul 2023 19:09:50 GMT).


scripts/key-generation/export-key-as-json.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
michalinacienciala, line 1: I decided to add a linter exception here, as I couldn't make the linter happy without braking the script (can't say confidently that this can't be done though...).

michalinacienciala:
Hmm, I don't why the lint check is failing... When I run `yarn eslint scripts/key-generation/` locally I don't get an error.

scripts/key-generation/export-key-as-json.js
Shadowfiend: Choice looks correct, just drop a note above it that since this is a script we can’t use import yet.

scripts/key-generation/export-key-as-json.js
```diff
@@ -0,0 +1,17 @@
...
+const fs = require("fs")
+const wallet = require("ethereumjs-wallet").default
```
Shadowfiend, line 3: For the lint error below—we need to yarn install in the parent dir in the actions workflow.

michalinacienciala: Added in [24b20bc](https://github.com/tahowallet/extension/pull/3559/commits/24b20bcaf7e69da7fcd2d4357243991dfef58c47), it helped, thanks!

scripts/key-generation/export-key-as-json.js
michalinacienciala: [1cce0db](https://github.com/tahowallet/extension/pull/3559/commits/1cce0db585d3fa74b49148d49f9211c78501bf0d)

michalinacienciala:
@Shadowfiend, ready for an another round of review.

Shadowfiend:
The fork build still seems to be failing here, so once we get that sorted we'll be good to go.

Which is good because the main e2e tests are failing atm 😅

michalinacienciala:
@Shadowfiend, since merging the `main` to the feature branch of this PR the tests are passing. I think we can merge 🚀 

Let's remove `TEST_WALLET_RECOVERY_PHRASE` secret after the merge.

Shadowfiend:
Ok, I think we're good here. Going to see if the e2e tests run as expected and then will land this fellow.

Shadowfiend (approved):
All righty let's rock and roll!

PR #3561, opened 2023-07-20
Add bigint filter for jsondiffpatch

hyphenized:
In the current implementation, `bigints` are treated as objects and are incorrectly excluded from generated patches. This fixes `bigint` diffing by adding a handler for these cases to the `jsondiffpatch` pipeline.

## To Test
- [x] Send assets
- [x] Check that balance changes after transaction succeeds

Latest build: [extension-builds-3561](https://github.com/tahowallet/extension/suites/14449081920/artifacts/815732866) (as of Thu, 20 Jul 2023 17:02:04 GMT).

### DIFF
--- a/background/differ.ts
+++ b/background/differ.ts
@@ -1,0 +1,18 @@
+import { DiffContext, Filter, create } from "jsondiffpatch"
+
+const differ = create()
+
+const bigintDiffFilter: Filter<DiffContext> = (context) => {
+  if (typeof context.left === "bigint" && typeof context.right === "bigint") {
+    if (context.left !== context.right) {
+      context.setResult([context.left, context.right])
+    }
+  }
+}
+bigintDiffFilter.filterName = "bigint"
+
+differ.processor.pipes.diff.before("objects", bigintDiffFilter)
+
+export const diff = differ.diff.bind(differ)
+export const patch = differ.patch.bind(differ)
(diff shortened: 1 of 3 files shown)


background/differ.ts
```diff
@@ -0,0 +1,18 @@
+import { DiffContext, Filter, create } from "jsondiffpatch"
+
+const differ = create()
+
+const bigintDiffFilter: Filter<DiffContext> = (context) => {
+  if (typeof context.left === "bigint" && typeof context.right === "bigint") {
+    if (context.left !== context.right) {
+      context.setResult([context.left, context.right])
+    }
```
Shadowfiend, line 9: Do we get correct diffs if left is bigint and right isn't or vice versa right now?

hyphenized: Yes, before there was an empty diff ( no result ). I manually tested this to check if it would handle unions well but I didn't test the opposite ( object -> bigint ) as I was trying to get the release unblocked. I think this deserves some tests too.

Shadowfiend:
(Also confirmed the balance updates.)

background/differ.ts
Shadowfiend: Agree, both with the trade off choice and the benefit of adding some tests here now that release is unblocked.

PR #3564, opened 2023-07-21
Fix incorrect build asset uploaded on release

hyphenized:
Added a suffix to the fork-related assets and changed the workflow steps to publish the correct production build as the release asset.

Latest build: [extension-builds-3564](https://github.com/tahowallet/extension/suites/14480268296/artifacts/818192057) (as of Fri, 21 Jul 2023 19:49:40 GMT).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,6 +57,14 @@ jobs:
           COMMIT_SHA: ${{ github.sha }}
           POAP_API_KEY: ${{ secrets.POAP_API_KEY }}
           DAYLIGHT_API_KEY: ${{ secrets.DAYLIGHT_API_KEY }}
+      - name: Create Release and Upload Artifacts
+        uses: softprops/action-gh-release@1e07f4398721186383de40550babbdf2b84acfc5 # v1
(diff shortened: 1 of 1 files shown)


webpack.config.ts
```diff
@@ -164,42 +164,47 @@ const modeConfigs: {
...
-            compress: browser === "firefox",
-            output:
-              browser === "firefox"
-                ? undefined
-                : {
-                    beautify: true,
-                    indent_level: 2, // eslint-disable-line camelcase
-                  },
-          },
+  production: (browser) => {
```
Shadowfiend, line 167: Thoughts on keeping the fork stuff in the workflow instead of further complexifying our webpack config? Could add after `yarn build` for `Production build on mainnet fork`:

```bash
    find dist -name "*.zip" -exec sh -c 'mv "$1" "${1%.zip}-fork.zip"' _ {} \;
```

hyphenized: Thanks, seems like a better move

Shadowfiend (approved):
Looks good, let's do it :rocket:

PR #3571, opened 2023-07-25
Revolintion: Apply most of the `@thesis-co/eslint-config` lint styling to the extension

Shadowfiend:
A few notes here:
- This definitively moves us from my preferred conditional React rendering of `condition ? <Element /> : <></>` to the more common null-punning approach of `condition && <Element />`. The style I've suggested in the past is disallowed by [`react/no-jsx-useless-fragment`](https://github.com/TopDev-727/React-Eslint-Plugin/blob/main/docs/rules/jsx-no-useless-fragment.md), which is part of the default airbnb style and therefore inherited through the Thesis config. Obviously I don't love it, but the point of these is to eliminate relatively pointless debate, so I'm not debating :grin:
- That rule also shifted a few places where we were returning `<></>` to return `null` instead, and updated return types accordingly. In one or two places where doing that would have a very far-reaching impact, I've dropped an eslint-ignore for now (with clarifying comment ofc), which can be revisited in the future if we feel strongly about it.
- This also pushes us away from the pattern of having `() => {}`-style functions with a single return statement inside, favoring instead making these return implicitly. A lot of the changes are around that.
- There were also a few cases where undefineds were being left to dangle that got flagged and now get alternate empty states. Delightful!

Last but not least, I've disabled one whole rule so that we can get the core of this merged: [`react/no-unstable-nested-components`](https://github.com/TopDev-727/React-Eslint-Plugin/blob/main/docs/rules/no-unstable-nested-components.md) has flagged a bunch of places where another pattern I think I suggested (defining anonymous functions that return elements lazily) is being used in a way that's creating a lot of unnecessary diffs. Probably not impactful in the large, but regardless it would be good to fix this. It's a big enough hunk of work that I decided to fire this off before fixing it; it'll be the next thing to circle back and snipe.

## Testing

Pretty much all the changes here are functionally identical, except for:

- https://github.com/tahowallet/extension/pull/3571/files?diff=split&w=1#diff-bb2ba232569016c28a3e1afeb301ce20420881034e715b0c87f88c6fd5db410cR247 swaps the order of parameters to the redux store initializer; checking that you don't lose data when loading this PR's build up should be sufficient to prove this is working fine.
- https://github.com/tahowallet/extension/pull/3571/files?diff=split&w=1#diff-5e1ac590a204b4721bc5f808cb274d5dddf75db557363c9e413717cf9247bb0fL23 drops `!!` usage in a case where we were already using booleans. Checking NFT pricing and filtering should be enough to prove this is working as intended.
- https://github.com/tahowallet/extension/pull/3571/files?diff=split&w=1#diff-e372e177a149ddb49522f9ce9f59956d30d40d221a3f287d8c2d78bbc2380c53R48-R60 reworks the `SharedDropdown` toggle function/context to avoid some invalidation issues.
- https://github.com/tahowallet/extension/pull/3571/files?diff=split&w=1#diff-18fa31f65c43e5d1ee48dd39838ed8d58657f986558126123c153b920b3f5639R94-R98 reworked some price display logic enough to be worth double-checking.
- https://github.com/tahowallet/extension/pull/3571/files?diff=split&w=1#diff-67347e171ecce754b7e623a056ca13d1b49d7a75fbcffefdcadf2ea80399a0b2R215-R222 changes how we include dApp popup body height: 100% CSS in styled-jsx, and it's worth double-checking that the popup actually does get that.

I suggest any review turn off whitespace in the GitHub diff view as the whitespace changes are fairly noisy.

Latest build: [extension-builds-3571](https://github.com/tahowallet/extension/suites/14833893452/artifacts/844637430) (as of Fri, 04 Aug 2023 18:54:53 GMT).


mhluongo:
I would just love a follow-up issue to deal with `react/no-unstable-nested-components` :)

.eslintrc.js
```diff
@@ -1,56 +1,13 @@
...
-  extends: [
-    "airbnb",
-    "airbnb-typescript",
-    "airbnb/hooks",
-    "plugin:import/typescript",
-    "plugin:@typescript-eslint/recommended",
-    "plugin:prettier/recommended",
-  ],
-  plugins: ["no-only-tests"],
+  extends: ["@thesis-co"],
```
mhluongo, line 2: :tada:

background/redux-slices/earn.ts
```diff
@@ -348,31 +348,25 @@ const earnSlice = createSlice({
...
-        availableVaults: state.availableVaults.map((availableVault) => {
-          const currentVault = payload.find(
-            (updatedVault) =>
-              availableVault.vaultAddress === updatedVault.vaultAddress
-          )
-          return currentVault || availableVault
-        }),
-      }
-    },
+    inputAmount: (state, { payload }: { payload: string }) => ({
```
mhluongo, line 351: Unfortunate that this still in the codebase. C'est la vie

background/services/enrichment/utils.ts
```diff
@@ -40,7 +40,7 @@ export async function enrichEIP2612SignTypedDataRequest(
   // If we have a corresponding asset - use known decimals to display a human-friendly
   // amount e.g. 10 USDC.  Otherwise just display the value e.g. 10000000
   const formattedValue = asset
-    ? `${Number(value) / 10 ** asset?.decimals} ${asset.symbol}`
+    ? `${Number(value) / 10 ** (asset?.decimals ?? 0)} ${asset.symbol}`
```
mhluongo, line 43: Glad this got caught.

Shadowfiend:
> I would just love a follow-up issue to deal with `react/no-unstable-nested-components` :)

https://github.com/tahowallet/extension/issues/3594

mhluongo (approved):
I manually reviewed all `background` changes, but could only spot-check `ui` and various tests. I think it's worth thinking about adding an automated security scan as part of our CI — we're big enough that even PRs from trusted devs deserve more scrutiny.

That said... I'm right next to @Shadowfiend and am confident in how's he's gotten here, so we can kick the can a little bit longer.

PR #3573, opened 2023-07-27
Fix assertions made on the latest transaction

michalinacienciala:
There is a stage in the tests when we want to verify the attributes of the latest transaction. By mistake we did some of the checks on an objet that wasn't a latest transaction, but on an object of 'any' transaction. The assertion was failing if more than one object was detected.


jagodarybacka (approved):
LGTM 🚢

PR #3575, opened 2023-07-27
Update release-test-list.md: changed outdated terminology and added tests for new features

andreachapman:
- Changed "overview" to "portfolio" to match the new name of the page in the extension
- Added a test to read-only accounts so that we're explicitly checking that these accounts don't have export options for private key and recovery phrase.

Latest build: [extension-builds-3575](https://github.com/tahowallet/extension/suites/14642864832/artifacts/830631345) (as of Fri, 28 Jul 2023 13:08:31 GMT).

### DIFF
--- a/.github/workflows/test-list/release-test-list.md
+++ b/.github/workflows/test-list/release-test-list.md
@@ -28,28 +28,31 @@ This release checklist should be performed before release is published.
    - [ ] check NFTs
    - [ ] check abilities (abilities should not be displayed)
    - [ ] check activities
-   - [ ] check overview page
+   - [ ] check portfolio page
+   - [ ] check export options (export recovery phase and export private key should not be available)
 2. Add read-only account with UNS
    - [ ] check assets
    - [ ] check balance
    - [ ] check NFTs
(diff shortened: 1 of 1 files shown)


.github/workflows/test-list/release-test-list.md
```diff
@@ -58,38 +61,38 @@ This release checklist should be performed before release is published.
...
+   - [ ] check portfolio page
    - [ ] check private key export for second account
 6. Add account with a Ledger
    - [ ] check assets
    - [ ] check balance
    - [ ] check NFTs
    - [ ] check abilities
    - [ ] check activities
-   - [ ] check overview page
+   - [ ] check portfolio page
```
jagodarybacka, line 72: I think we may add `check export options (export recovery phase and export private key should not be available)` for Ledger as well

andreachapman: Ok, I edited the file and committed to this same PR. Hopefully that worked but let me know if not. Thanks for the suggestion, totally agree!

PR #3579, opened 2023-07-29
Revert "Revert "Debounce dispatched state diffs""

hyphenized:
Reimplements debounced/aggregated diffs dispatched on store updates and fixes awaiting of dispatched thunks from the UI.

Added some logic to track pending store updates on the dispatch responder. This is done so that async thunk responses are postponed until state changes are committed to the UI's proxied store. This fixes cases where thunks cause unexpected behaviour by resolving too early, before the aggregated updates have been sent to the UI.

This relies on the subscribed callback from webext-redux for diffing to have executed before the dispatched thunk has finished executing in the background script.

## To Test
A consistent way to reproduce this issue is to attempt to initiate a transfer through the Send Asset page and then reject it. 

Internally the route is meant to change once the thunk completes. Thus, if we change the route before the transaction state has been cleared by the thunk completion, we are incorrectly sent back to the sign transaction screen. This PR also addresses similar issues.

Latest build: [extension-builds-3579](https://github.com/tahowallet/extension/suites/15065445898/artifacts/860358513) (as of Mon, 14 Aug 2023 06:15:02 GMT).


background/main.ts
```diff
@@ -508,27 +508,82 @@ export default class Main extends BaseService<never> {
...
+
+    const queueUpdate = debounce(
+      (lastState, newState, updateFn) => {
+        const clearPendingUpdates = () => {
+          pendingUpdates.forEach((resolve) => resolve())
+
+          // Clear all currently tracked updates
+          while (pendingUpdates.length) {
+            pendingUpdates.pop()
+          }
```
Shadowfiend, line 528: If I'm not mistaken and the below remark about not needing an array doesn't apply, I think this can be rewritten as:

```suggestion
          pendingUpdates
            .splice(0)
            .forEach((resolve) => resolve())
```

background/main.ts
```diff
@@ -508,27 +508,82 @@ export default class Main extends BaseService<never> {
...
       ) => {
+        const isThunk = dispatchResult instanceof Promise
         try {
+          // if dispatch is a thunk, wait for the result
+          const result = await dispatchResult
+
+          // by this time, all pending updates should've been tracked.
+          // since we're dealing with a thunk, we need to wait for
+          // the store to be updated
+          if (isThunk) await storeUpdateLock
```
Shadowfiend, line 572: Do we suffer if we do this for all dispatches?

background/main.ts
```diff
@@ -508,27 +508,82 @@ export default class Main extends BaseService<never> {
...
-            value: encodeJSON(await dispatchResult),
+            value: encodeJSON(result),
           })
         } catch (error) {
           logger.error(
             "Error awaiting and dispatching redux store result: ",
             error
           )
+
+          // Store could still have been updated if there was an error
```
Shadowfiend, line 584: With this in mind and assuming we don't suffer if we wait on the lock for all dispatches, I wonder if the data flow might be clearer with `Promise.allSettled`… Something like:

```typescript
  const [result] = await Promise.all([
    dispatchResult,
    // Block on the update lock when relevant.
    storeUpdateLock
  ])

  if (result.status === "fulfilled") {
    send({
      error: null,
      value: encodeJSON(result.value)
    })
  } else {
    send({
      error: encodeJSON(result.reason),
      value: null,
    })
  }
```

Even if we do want the lock to be unnecessary for non-promise dispatches, if this data flow seems clearer, I think we can do `dispatchResult instanceof Promise ? storeUpdateLock : Promise.resolve()` or so, because `Promise.all` doesn't yield if none of its parameters are unresolved promises, it just runs right through.

Thoughts?

background/main.ts
```diff
@@ -508,27 +508,82 @@ export default class Main extends BaseService<never> {
...
+
     wrapStore(this.store, {
       serializer: encodeJSON,
       deserializer: decodeJSON,
-      diffStrategy: (oldObj, newObj) => {
-        const diffWrapper = deepDiff(oldObj, newObj)
-        return diffWrapper === undefined ? [] : [diffWrapper]
+      diffStrategy: (oldObj, newObj, forceUpdate) => {
+        storeUpdateLock = new Promise((resolve) => {
+          pendingUpdates.push(resolve)
```
Shadowfiend, line 553: Do we need this to be an array? It feels like `storeUpdateLock` can be one `Promise<void> | undefined` with one associated unlocker function:

```typescript
    // let storeUpdateLock: Promise<void> | undefined
    // let unlockStoreUpdate: (_: void) => void | undefined

    if (storeUpdateLock === undefined) {
      // Set read lock so others can block on it.
      storeUpdateLock = new Promise((resolve) => unlockStoreUpdate = resolve)
   }

   queueUpdate(oldObj, newObj, forceUpdate)

   // Return no diffs.......
   return []
```

Then `queueUpdate`, when it's called debounced, can just resolve (and clear) the one promise and its resolver, and everyone who has blocked in the meantime will be awaiting that one promise. I might not be playing it all out correctly in my brain though, I'm not at 100%.

Shadowfiend (commented):
Solution feels strong. Left a couple of notes on how we might tighten it up just a touch. I wish we could go ahead and do our own fork of webext-redux already, but I just remembered when I tried to do this I had to face down some dependency stuff because the package hasn't seen a lot of updates the last couple of years… For another time.

Shadowfiend:
None of the above seem blocking, so I'm going to give this a test whirl and plan to merge it for next week's release whether or not we get a chance to work on them.

background/main.ts
hyphenized: I think this could work but it might cause unexpected behaviour if a thunk performs a dispatch and takes longer than 30ms to complete 🤔

Shadowfiend: I'm wondering if that's functionally different than the current version. In both cases, `storeUpdateLock`'s value becomes relevant once the thunk is resolved, and in both cases the `storeUpdateLock` promise is resolved as soon as the diff starts running. My read is that means the array and non-array versions would be identical in behavior, but these are def complicated interactions to reason about so I could totally be wrong.

Shadowfiend: Ahhh I maybe see what you're saying. I think *this change* from array of promises to single promise would be functionally identical, but the `Promise.all` approach in the previous comment would *not* be, because using `Promise.all` reads the lock before the thunk has completed, while the behavior we want here is to let the thunk finish, *then* wait until the latest debounced state update is pushed before responding with the thunk outcome.

background/main.ts
```diff
@@ -549,10 +536,16 @@ export default class Main extends BaseService<never> {
...
-          queueUpdate(oldObj, newObj, forceUpdate)
-        })
+        if (!storeUpdateLock) {
+          storeUpdateLock = new Promise((resolve) => {
+            releaseLock = () => {
+              resolve()
+              storeUpdateLock = null
+            }
+          })
+        }
```
Shadowfiend, line 546: Golfing at this point, but:

```suggestion
        storeUpdateLock ??= new Promise((resolve) => {
          releaseLock = () => {
            resolve()
            storeUpdateLock = null
          }
        }
```

Shadowfiend (approved):
Looks like when we force update after the thunk is done, during dApp signing, we can try to push something through a disconnected port (since the popover has closed out). Shouldn't be a huge issue in practice, but might be nice to catch it and log it at a lower level than error.

Screenshot 2023-08-15 at 15 22 47

Not blocking, wallet seems to be behaving correctly across the board, tried dApps and swaps as well as settings toggles. Let's roll it!

background/main.ts
hyphenized: Yup they are identical, refactored the array out in favour of a single lock (forgot to submit these comments 🤦‍♂️ )

background/main.ts
hyphenized: Good point! We don't because this is only relevant to dispatch responses

PR #3595, opened 2023-08-04
Underbridge: Drop updated dApp connections feature flag, further Taho renames

Shadowfiend:
This PR does two things:
- Drops the `ENABLE_UPDATED_DAPP_CONNECTIONS` feature flag completely, as well as all related code paths.
- Updates all remaining references to `tally` in the window provider to say `taho` instead, except if they were part of a publicly-visible API. In the latter case, adds a `taho` equivalent, so that we can phase `tally` out at an unspecified later date.

Latest build: [extension-builds-3595](https://github.com/tahowallet/extension/suites/15119822565/artifacts/863994512) (as of Tue, 15 Aug 2023 20:01:43 GMT).


Shadowfiend:
Did some light testing here with dApp connections across a few places, including stake.lido.fi which has direct Taho support. All working well, so going to ship it.

src/window-provider.ts
```diff
@@ -202,21 +207,28 @@ if (!window.walletRouter) {
...
+      tallyProvider: window.taho,
+      tahoProvider: window.taho,
       providers: dedupedProviders,
       shouldSetTallyForCurrentProvider(
         shouldSetTally: boolean,
         shouldReload = false
       ) {
-        if (shouldSetTally && this.currentProvider !== this.tallyProvider) {
-          this.currentProvider = this.tallyProvider
+        this.shouldSetTahoForCurrentProvider(shouldSetTally, shouldReload)
```
hyphenized, line 219: Do we have to keep `shouldSetTallyForCurrentProvider` as an alias?

Shadowfiend: I went back and forth on this one--it was designed for internal calls, and I'm not sure why a third party would call it. But, it is strictly speaking public API, because it needed to be for us to access it in the first place. I felt safe nixing it since the integrations that do exist are provider-style and wouldn't need this call...

With that said, if you'd feel more comfortable with an alias sticking around, I'll slip it in.

PR #3597, opened 2023-08-09
Fix invalid react fragment

hyphenized:
This was causing `TransactionDataPanelSwitcher` to fail, blocking transaction signing.


Shadowfiend:
Puzzled as to how this passed e2e tests, thought we had some for signing now 👀

PR #3613, opened 2023-09-01
Proxy Wars: More QoL improvements for dApp connections

### LINKED ISSUE #3395: Support EIP 2255 (`wallet_requestPermissions`)
### Discord Discussion Link

https://discord.com/channels/808358975287722045/888500174685614090/1108852143030534244

N/A

### LINKED ISSUE #3599: Disconnecting from a dapp does not remove it's stored "current network"
To reproduce:
- Connect to a dapp 
- Open background inspector
- Check indexed db / internal-ethereum-provider / currentNetwork
- Remove the dapp from connected websites
- Reload indexed db entries

#### Correct behaviour: 
- The dapp entry should have been removed
#### Current behaviour:
- The dapp entry is not removed

### LINKED ISSUE #3604: infinite loop on guild.xyz only when Taho as default is selected
### Discord Discussion Link

_No response_

Chrome

Trying to use guild.xyz with Taho after having logged in to the same page on another browser/machine with the same account.

This is something I've been encountering recently but could not find the cause of. (did not happen often)

During testing I finally found a way to force the issue to happen.

Note: Taho is the _only_ wallet installed during this issue

### What did not work?

When Taho is the default (despite being the only wallet installed) some looping issue is encountered.

Setting to not be default resolves the issue.

* connect to https://guild.xyz/ with Taho while the Taho as default is enabled, sign the login message
* in a different browser or machine connect to https://guild.xyz/ with the same wallet address, sign the login message
* on the first browser/machine refresh the page
* it should indicate that you need to log in again and then pop up the wallet selection, which may appear to flicker
* the browser console will be flooded with messages if the issue occurs and you will _not_ get a popup from the wallet to sign the login message
* open the extension and toggle the default at the top of the UI to not be Taho
* the looping should stop, but you must then refresh the page to have things work again
* you should now be presented with the login prompt ("Verify account")

### Version

v0.47.0

### Relevant log output

_No response_

### LINKED ISSUE #3612: Curve DAO dapp not able to connect to Taho
### Discord Discussion Link

I have a dumb q: I just try to connect Taho onto curve without metamask installed and I'm not sure how I can get that running

Chrome

Connecting to the Curve DAO dApp : https://dao.curve.fi/

I am on Chrome desktop and do not have MetaMask installed.

### What did not work?

The wallet connect overlay doesn't show Taho, and when trying to connect via MetaMask it would ask me to download MetaMask instead.

### Version

v0.47.0

### Relevant log output

_No response_

Shadowfiend:
This PR:

- Fixes #3612, which stemmed from issues impersonating MetaMask.
- Fixes #3604, which stemmed from bad `this` assignment during proxy calls when impersonating MetaMask.
- Fixes #3395 with a vestigial `wallet_requestPermissions`/`wallet_getPermissions` implementation.
- Fixes #3599 by clearing current network state when a disconnect on a matching account and network occurs. This can be tested by connecting to a dApp on a network, disconnecting, switching network, and attempting to connect again. Previously, the new connection would be on the original network, now it should be on the currently-selected network.
- Likely fixes additional dApp connection issues.

Additionally, it introduces logging for MetaMask-impersonated requests that includes timing information and parameter/return value/error value information:

Screenshot 2023-09-01 at 00 36 14

This is similar to existing request/response logging in dApps, but it:
- Only happens on development builds.
- Distinguishes between calls handled by Taho and calls handled by MetaMask or the MetaMask mock (if MetaMask is not installed).
- Intercepts all function calls and return values, not just JSON-RPC requests, allowing us to identify diverging behaviors with MM more easily.
- Highlights errored calls.
- Includes call and return value in one line.
- Includes how long the call took to complete.
- Only reports the call without the return value if it takes >1s to complete.
- Reports long-running calls every second with a note on how long the call has been pending.

Latest build: [extension-builds-3613](https://github.com/tahowallet/extension/suites/17865889143/artifacts/1025428882) (as of Thu, 02 Nov 2023 22:01:18 GMT).


beemeeupnow:
anything needed for this? testing?

Shadowfiend:
> anything needed for this? testing?

Testing, yes! Hoping to ship this fella next week.

andreachapman:
> > "#3599 by clearing current network state when a disconnect on a matching account and network occurs. This can be tested by connecting to a dApp on a network, disconnecting, switching network, and attempting to connect again. Previously, the new connection would be on the original network, now it should be on the currently-selected network.":
> 
> > > I connected to opensea.io on Polygon. I then disconnected in opensea.io and in the extension switched to Ethereum. I connected in opensea.io again but it connected on Polygon instead of Ethereum.
> 
> Did you disconnect via the lightning bolt button in the wallet? That is the action that should clear the association since it is revoking the site permissions. Disconnecting in the dapp UI isn't expected to clear it, from what I understand.

Yea, should have noted that. I tried disconnecting from within the extension and from within the dapp (just to try it all) - neither case gave me the results I expected **until** I cleared the entry from the currentNetwork index and then when I do that, it works like a charm.

Shadowfiend:
I might need to ride shotgun with you to understand this one better. Clearing the index is what should be happening in the first place (if, as beem notes, you disconnect from the lightning bolt in the extension), so having to do it manually first is puzzling.

andreachapman:
> I might need to ride shotgun with you to understand this one better. Clearing the index is what should be happening in the first place (if, as beem notes, you disconnect from the lightning bolt in the extension), so having to do it manually first is puzzling.

I have to leave soon but here's another video showing me using the lightning bolt to disconnect so hopefully that helps clarify to some degree and tomorrow I can definitely sit together to look at it. In this video, you can see that after I manually delete the indexed network, the dapp switches from Optimism to Polygon automatically. Polygon is what is was on when I connected the 2nd time and what it **should** have connected on. But instead, I had to manually delete that entry before it switched. 

https://github.com/tahowallet/extension/assets/135037801/86013d46-79d3-4b76-b23f-01b3ee70e685

And I think this will make it less confusing: it's still always doing the indexing for me. I had said it only does it once but now I can't recreate that scenario. Now, it's always indexing and the automatic switch when connecting on a different network won't happen unless I delete that entry. 

Should it be storing anything here at all after this change? Because these are all ones that have been stored since I started testing on this version today (ones that I haven't deliberately cleared yet). Or is it supposed to be storing them but then deleting/re-creating when we disconnect? I'm trying to understand if I should or should not see anything there with this change.
Screenshot 2023-10-30 at 12 51 05 PM

Shadowfiend:
Yeah on disconnect, it should be deleting them. I'll poke at it on my end as well. Odd that we're not seeing it…

background/services/provider-bridge/index.ts
```diff
@@ -369,11 +397,25 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
+        resolve("Time to move on")
+      })
     })
+
+    const result = await permissionPromise
+
+    if (this.#pendingPermissionsRequests[permissionRequest.origin]) {
+      // Just in case this is a different promise, go ahead and resolve it with
+      // the same result.
+      this.#pendingPermissionsRequests[permissionRequest.origin](result)
```
jagodarybacka, line 414: How is this case triggered? We are allowing only one pending request per origin, right?

src/window-provider.ts
```diff
@@ -47,6 +51,117 @@ const metaMaskMock: WalletProvider = {
...
+      return (...args: unknown[]) => {
+        const safeSerialize = (object: unknown) => {
+          if (object instanceof Error) {
+            return `${object.name}[${object.message}]${
+              object.stack ? `: \n${object.stack}` : ""
+            }`
+          }
+
+          try {
+            return JSON.stringify(object)
```
jagodarybacka, line 81: do we care here about serializing bigints? Or are they already stringified here?

background/services/provider-bridge/index.ts
```diff
@@ -252,46 +257,69 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
-        chainID: dAppChainID,
-        faviconUrl: faviconUrl || tab?.favIconUrl || "", // if favicon was not found on the website then try with browser's `tab`
-        title,
-        state: "request",
-        accountAddress,
-      }
+      const existingPermission = await this.checkPermission(origin, dAppChainID)
+      if (
+        // If there's an existing permission record and it's not an explicit
+        // allow, immedhttps://doc.rust-lang.org/std/primitive.char.html#representationiately return a rejection.
```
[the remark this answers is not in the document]
Shadowfiend: 🙉

src/window-provider.ts
Shadowfiend: Yeah no bigints make it across the provider bridge.

background/services/internal-ethereum-provider/db.ts
```diff
@@ -53,6 +53,22 @@ export class InternalEthereumProviderDatabase extends Dexie {
...
+  /**
+   * Clear origin's current network state if it is the same as the passed
+   * chainId.
+   */
+  async unsetCurrentNetworkForOrigin(
+    origin: string,
+    chainId: string,
+  ): Promise<void> {
+    const originMatches =
+      (await this.currentNetwork.where({ origin, chainId }).count()) > 0
```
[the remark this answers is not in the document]
Shadowfiend: Sure enough, that probably explains the issues showing up during testing. Unclear why this worked when I tested it though…

jagodarybacka (approved):
Code looks ok, @andreachapman up to you if you prefer to merge it and do release testing or test it here again.

Shadowfiend:
Looks like the latest build here doesn't include the fix from main for some reason…

Shadowfiend:
Rolling this in due to positive feedback from a couple of testers. We'll test the whole shebang together in a release PR.

Issue #3608, opened 2023-08-28
PK import issues


mr-michael:
Logs posted on discord:
https://discord.com/channels/808358975287722045/888500174685614090/1146697575169273916
https://discord.com/channels/808358975287722045/888500174685614090/1146365094171774987
https://discord.com/channels/808358975287722045/888500174685614090/1146029538984333354

Shadowfiend:
Fixed by #3614.

PR #3614, opened 2023-09-01
Attempt to import private key to validate it

jagodarybacka:
Ref https://github.com/tahowallet/extension/issues/3608

### What

Let's validate private key by trying to import it

## Testing

- [x] Try to import the private key `0c1cd323ff8fd5b34f260d580c3a843f3e06c3e1e31c357d38a4a22e8b26809b`, which has a leading 0 (more than one leading 0 should produce the same result). It should work without failing.
- [x] Try to import the private key `1cd323ff8fd5b34f260d580c3a843f3e06c3e1e31c357d38a4a22e8b26809b` (same as above but with first two characters missing entirely). It should fail.

Latest build: [extension-builds-3614](https://github.com/tahowallet/extension/suites/15728481611/artifacts/898042290) (as of Fri, 01 Sep 2023 13:37:44 GMT).


ui/pages/Onboarding/Tabbed/ImportPrivateKey.tsx
```diff
@@ -34,23 +33,22 @@ export default function ImportPrivateKey(props: Props): ReactElement {
...
-        setIsImporting(false)
-      }
+    setIsImporting(true)
+
+    const { success } = (await dispatch(
+      importSigner({
+        type: SignerSourceTypes.privateKey,
+        privateKey: trimmedPrivateKey,
+      }),
+    )) as unknown as AsyncThunkFulfillmentType<typeof importSigner>
```
Shadowfiend, line 44: Non-blocking: you should be able to drop the `as` chain here now :) The dependency bumps included a typing rework that properly types `dispatch`.

Shadowfiend (approved):
Code looks good, will do a spot check here soon.

Shadowfiend:
Confirmed the bug fix here. I used the private key `0c1cd323ff8fd5b34f260d580c3a843f3e06c3e1e31c357d38a4a22e8b26809b` which I generated some-random-where. It fails on `main` and works on this branch. I also tested a few other private keys to make sure they still worked 💪

PR #3616, opened 2023-09-02
Unsettled Networks: Clear an origin's current network when revoking dApp permissions

### LINKED ISSUE #3599: Disconnecting from a dapp does not remove it's stored "current network"

Shadowfiend:
Draft until I fill in some testing details.

The actual implementation is a little more subtle: permissions are for an origin AND a network. If the permission being cleared is for the origin's current network, as tracked in the `InternalEthereumProvider`, then that current network is cleared so that future interactions can set it.

Latest build: [extension-builds-3616](https://github.com/tahowallet/extension/suites/15762123848/artifacts/899755021) (as of Sat, 02 Sep 2023 20:14:11 GMT).

### DIFF
--- a/background/services/internal-ethereum-provider/db.ts
+++ b/background/services/internal-ethereum-provider/db.ts
@@ -53,6 +53,22 @@ export class InternalEthereumProviderDatabase extends Dexie {
     return this.currentNetwork.put({ origin, network })
   }
 
+  /**
+   * Clear origin's current network state if it is the same as the passed
+   * chainId.
+   */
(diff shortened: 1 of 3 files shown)


Shadowfiend:
Going to turbo this one so we can look at everything in #3613 .

PR #3623, opened 2023-09-05
Joint Privacy: Rework analytics service event management and add signing tracking

Shadowfiend:
This PR does two core things:
- Moves responsibility for more event tracking into the AnalyticsService itself.
- Adds tracking events for when data is signed and when there is an error signing data or transactions. The data signed and signature are not sent to Posthog, and in the case of errors only a generic set of errors is sent (rather than full error data).

Note that there is still a lot of unnecessary event management happening in `Main`, which should just be event observations from `AnalyticsService`. These can be handled at a later time.

Latest build: [extension-builds-3623](https://github.com/tahowallet/extension/suites/15873783826/artifacts/905833489) (as of Wed, 06 Sep 2023 10:24:01 GMT).


jagodarybacka (approved):
Code looks good, haven't tested if events are actually getting into posthog correctly but let's test it during release testing

🚀

PR #3626, opened 2023-09-14
Run pledge sync daily

michalinacienciala:
We want to update the pledge sync job to run daily rather than every three days so that no new pledge signers (or old community who havent signed, but now need to) are missed.

Latest build: [extension-builds-3626](https://github.com/tahowallet/extension/suites/16161486375/artifacts/922815423) (as of Thu, 14 Sep 2023 10:48:13 GMT).


Shadowfiend (approved):
:shipit:

PR #3633, opened 2023-09-21
Add support for Mainnet Sepolia testnet

michalinacienciala:
Although it was previously possible to add Sepolia testnet as a custom network, we want to add it to the list of networks automatically added to the networks list upon switching on the 'Show testnet networks' setting. We want that because the only testnet currently added to that list is Goerli, which will become deprecated at the end of 2023 (see
https://ethereum-magicians.org/t/proposal-predictable-ethereum-testnet-lifecycle/11575). So far we want to still keep Goerli on the list, as it is still widely used. We'll need to remove it once it'll reach end of its life.

As at the moment `0x` project which we use for swaps does not support Sepolia, the sweep functionality will be disabled on this testnet. We need to enable it once the support is added.

Latest build: [extension-builds-3633](https://github.com/tahowallet/extension/suites/16709744384/artifacts/954942099) (as of Fri, 29 Sep 2023 16:39:01 GMT).


background/constants/networks.ts
```diff
@@ -162,6 +175,7 @@ export const CHAIN_ID_TO_0X_API_BASE: {
   [POLYGON.chainID]: "polygon.api.0x.org",
   [OPTIMISM.chainID]: "optimism.api.0x.org",
   [GOERLI.chainID]: "goerli.api.0x.org",
+  // TODO: Add Swap API for Sepolia once 0x supports it.
```
michalinacienciala, line 178: I haven't found any official confirmation they plan to do that. But I saw a couple of people requesting this support on their Discord. [I've asked them there](https://discord.com/channels/435912040142602260/936366257521954857/1154359872217952277) if they plan to support it since Goerli becomes deprecated, but I didn't get a response yet.

background/constants/networks.ts
```diff
@@ -172,6 +186,8 @@ export const NETWORKS_SUPPORTING_SWAPS = new Set(
 )
 
 export const ALCHEMY_SUPPORTED_CHAIN_IDS = new Set(
+  // TODO: Add `SEPOLIA` once `ethersproject` adds support for it to v5 (see
+  // https://github.com/ethers-io/ethers.js/issues/4374).
```
michalinacienciala, line 190: According to the linked issue ethersproject plans to release a patch to v5 with a couple of small changes, including support for Sepolia, but it's not high on their priority list, if I understood correctly.

If we don't want to wait, we can upgrade `@ethersproject/providers` to the `6.0.0-beta.8` version, which does support Sepolia (see https://www.npmjs.com/package/@ethersproject/providers/v/5.7.2?activeTab=code vs https://www.npmjs.com/package/@ethersproject/providers/v/6.0.0-beta.8?activeTab=code).

But I'm not sure how big of a deal is it to upgrate to v6 - would that be problematic / have potential to break things?

background/constants/networks.ts
```diff
@@ -203,6 +219,7 @@ export const CHAIN_ID_TO_RPC_URLS: {
   ],
   [ARBITRUM_NOVA.chainID]: ["https://nova.arbitrum.io/rpc	"],
   [GOERLI.chainID]: ["https://ethereum-goerli-rpc.allthatnode.com"],
+  [SEPOLIA.chainID]: ["https://ethereum-sepolia-rpc.allthatnode.com"],
```
michalinacienciala, line 222: I'm not sure it that's a problem with the RPC or our code, but I see some problems with accessing historical data:
* When a wallet gets imported to the extension and network gets set to Sepolia, no historical transactions get displayed - the Activity tab shows "All your new transactions will show up here."
* Sometimes after making transaction on the Sepolia testnet the result screen shows us updated balance but does not show any transactions (again, "All your new transactions will show up here." message displayed). There are also errors in the logs:
```
index.js:224 Uncaught (in promise) Error: failed response (requestBody="{\"method\":\"eth_getCode\",\"params\":[\"0x0581470a8b62bd35dbf121a6329d43e7edd20fc7\",\"latest\"],\"id\":180,\"jsonrpc\":\"2.0\"}", requestMethod="POST", url="https://ethereum-sepolia-rpc.allthatnode.com", code=SERVER_ERROR, version=web/5.7.1)
    at Logger.makeError (index.js:224:23)
    at Logger.throwError (index.js:233:20)
    at index.js:275:27
    at Generator.next (<anonymous>)
    at fulfilled (index.js:5:58)
```
and
```
POST https://ethereum-sepolia-rpc.allthatnode.com/ 429
```
Only after exiting the result view and navigating to the Activity view we see the performed transaction.
* Not all Sepolia transactions made on a given instalation of the extention are shown on the Activity list.
* Also, I once hit an issue when Send transaction was categorized by extension as Contract Interaction (it was transaction during which I had that `failed response` error).

I tried using different RPC (https://rpc2.sepolia.org), but it worked worse than the one from AllThatNode.

background/lib/alchemy.ts
```diff
@@ -69,7 +69,7 @@ export async function getAssetTransfers(
   const category = ["external", "erc20"]
 
   if (addressOnNetwork.network.name === "Ethereum") {
-    // "internal" is supported only on Ethereum Mainnet and Goerli atm
+    // "internal" is supported only on Ethereum Mainnet, Goerli and Sepolia atm
     // https://docs.alchemy.com/alchemy/enhanced-apis/transfers-api#alchemy_getassettransfers-testnets-and-layer-2s
```
michalinacienciala, line 73: Actually, according to the link it is also supported on Polygon Mainnet, but I didn't want to do anything about that observation, as I don't know the system well enough...

background/constants/networks.ts
Shadowfiend: Yeah, 6.x is a breaking change.

Ethers is set up to resolve static functions like `getUrl` recursively through the inheritance hierarchy. Let's create a `TahoAlchemyProvider` in our codebase that does something like this:

```typescript
export class AlchemyProvider extends UrlJsonRpcProvider {
  static getUrl(network: Network, apiKey: string): ConnectionInfo {
        static getUrl(network: Network, apiKey: string): ConnectionInfo {
        let host = null;
        switch (network.name) {
            case "sepolia":
                host = "eth-sepolia.alchemyapi.io/v2/";
                break;
            default:
               return AlchemyProvider.getUrl(network, apiKey)
        }

        return {
            allowGzip: true,
            url: ("https:/" + "/" + host + apiKey),
            throttleCallback: (attempt: number, url: string) => {
                if (apiKey === defaultApiKey) {
                    showThrottleMessage();
                }
                return Promise.resolve(true);
            }
        };
    }
  }
```

And see if that fixes Alchemy.

Note that I believe all transactions are forcibly sent via Alchemy, so probably transaction submission won't work without this... But not 100% sure about that.

michalinacienciala: @Shadowfiend, where in the codebase should I add this? In some separate file, or maybe in `networks.ts`? (Sorry, my knowledge of the codebase and TS is fragmentary 😞).

hyphenized: A good place would probably be in a separate file under the same folder as the serial fallback provider (I think)

ui/utils/constants.ts
```diff
@@ -23,6 +25,14 @@ export const blockExplorer = {
...
   [POLYGON.chainID]: { title: "Polygonscan", url: "https://polygonscan.com" },
   [GOERLI.chainID]: { title: "Etherscan", url: "https://goerli.etherscan.io/" },
+  [SEPOLIA.chainID]: {
+    title: "Etherscan",
+    url: "https://sepolia.etherscan.io/",
+  },
+  [ARBITRUM_SEPOLIA.chainID]: {
+    title: "Arbiscan",
+    url: "https://sepolia.arbiscan.io/",
+  },
```
michalinacienciala, line 35: There is no block explorer for Sepolia Arbitrum yet. The address I provided is the expected one, once ArbiScan adds a support. I've sent them a question whether and when they plan to add the support - no response yet.
With this entry, the links to scan website on Arbitrum Sepolia will currently lead to non-existing page. An alternative would be to not have the scan website entry for this testnet, but then there is an `Unexpected Error` when user opens details of transaction.
There is probably also an option to not display the link to the scan website for this testnet. Should I go with this approach?

background/constants/networks.ts
Shadowfiend: Yeah that would be fine enough 👆

ui/utils/constants.ts
Shadowfiend: Since we're not needing to ship this like, tomorrow, let's wait and see if they answer. If it doesn't exist, we should drop the scanner until it does.

background/constants/networks.ts
michalinacienciala: I tried adding `background/services/chain/taho-alchemy-provider.ts` file with the following contents (similar to what Antonio provided, but with some changes to make IDE/linter happy):

```
import { AlchemyProvider, UrlJsonRpcProvider } from "@ethersproject/providers"
import { ConnectionInfo } from "@ethersproject/web"
import { Network as EthersNetwork } from "@ethersproject/networks"

export default class TahoAlchemyProvider extends UrlJsonRpcProvider {
  static override getUrl(
    network: EthersNetwork,
    apiKey: string,
  ): string | ConnectionInfo {
    let host = null
    switch (network.name) {
      case "sepolia":
        host = "eth-sepolia.alchemyapi.io/v2/"
        break
      default:
        return AlchemyProvider.getUrl(network, apiKey)
    }

    return {
      allowGzip: true,
      url: `https://${host}${apiKey}`,
      throttleCallback: () => Promise.resolve(true),
    }
  }
}
```

I've tested the extension with and without this file and didn't notice a difference in behavior. In both versions I was able to make transactions on Ethereum Sepolia. Sometimes they were performed with no errors and were correctly recognized as transactions of `Send` type. And sometimes there was a ` POST https://ethereum-sepolia-rpc.allthatnode.com/ 429` error in logs and transaction was performed, but got recognized as of `Contract Interaction` type.
The reading of the past (pre-extension-installation) transactions does not work on both versions.

michalinacienciala: > I've tested the extension with and without this file and didn't notice a difference in behavior. 
Ah, sorry, I forgot to add `SEPOLIA` to list of `ALCHEMY_SUPPORTED_CHAIN_IDS` when testing with the `taho-alchemy-provider.ts` file. Now I tested it locally again and it looks something is still off with the Alchemy config. I'm getting
```
Uncaught (in promise) Error: unsupported network (argument="network", value={"name":"sepolia","chainId":11155111,"ensAddress":null}, code=INVALID_ARGUMENT, version=providers/5.7.2)
    at _ethersproject_logger_lib_esm_Logger.makeError (index.js:224:23)
    at _ethersproject_logger_lib_esm_Logger.throwError (index.js:233:20)
    at _ethersproject_logger_lib_esm_Logger.throwArgumentError (index.js:236:21)
    at getUrl (alchemy-provider.js:67:24)
    at new UrlJsonRpcProvider (url-json-rpc-provider.js:56:28)
    at new AlchemyProvider (alchemy-provider.js:26:8)
    at SerialFallbackProvider.creator [as alchemyProviderCreator] (serial-fallback-provider.ts:1100:15)
    at new SerialFallbackProvider (serial-fallback-provider.ts:314:35)
    at makeSerialFallbackProvider (serial-fallback-provider.ts:1128:10)
    at index.ts:373:11
```
It looks like that the `taho-alchemy-provider.ts` file isn't taken under consideration. Am I missing something?

Also, maybe we can live without SEPOLIA` added to the list of `ALCHEMY_SUPPORTED_CHAIN_IDS`? The transactions are successful without it.

michalinacienciala: > And sometimes there was a  POST https://ethereum-sepolia-rpc.allthatnode.com/ 429 error in logs and transaction was performed, but got recognized as of Contract Interaction type.

That I managed to fix by changing RPC in https://github.com/tahowallet/extension/pull/3633/commits/c039fb1e965b08ff19672f8a78ac689ce615c3ae.

hyphenized: > It looks like that the `taho-alchemy-provider.ts` file isn't taken under consideration. Am I missing something?

Yes but updating the serial fallback provider to reference the new custom provider should do the trick.

Using the example shared, I ended up with something like this (Also noticed the URL was incorrect and fixed that.):
https://github.com/tahowallet/extension/commit/a245e8af55ed34fc4d77ad5f38090e2261a4fa66

Unless it has already been mentioned, would like to point out that it would be easier to patch the ethers library and add the missing URL handler there.

michalinacienciala: Thanks for the help @hyphenized! I confirmed your setup works and fixes the problem with reading historical data from the Ethereum Sepolia chain. Committed in [eaf5023](https://github.com/tahowallet/extension/pull/3633/commits/eaf502300ad4f905fd0acd90d44fd63d1288cd87).

We still have the problem with reading historical data from Arbitrum Sepolia, but Alchemy does not have yet an RPC for this testnet. [I've asked them on Discord](https://discord.com/channels/735965332958871634/1002527639489232906/1157338058362994840) if they plan to add it. Once it's added, I'm going to add a case for `arbitrum-sepolia` to `TahoAlchemyProvider` (but I think we don't need to do it in this PR).

background/constants/networks.ts
michalinacienciala: The problems described above are no longer occurring. We fixed the problem with reading historical data on Ethereum Sepolia in [eaf5023](https://github.com/tahowallet/extension/pull/3633/commits/eaf502300ad4f905fd0acd90d44fd63d1288cd87) and the problem with Send transactions marked as Contract Interaction in [c039fb1](https://github.com/tahowallet/extension/pull/3633/commits/c039fb1e965b08ff19672f8a78ac689ce615c3ae).

michalinacienciala:
Ready for review.
There's a couple of things not working at the moment for Arbitrum Sepolia due to the testnet not being supported yet by some of the external projects:
* _0x_ (they have an open feature request for it) - affects Swaps
* _ArbiScan_ ("it should be ready soon") - affects scan website
* _OpeanSea_ ("we are aware of this and we will add to opensea-js as soon as it's added on OpenSea") - affects NFTs
* _Alchemy_ (no response whether they plan to add it yet) - affects discovery of historical data

beemeeupnow:
Did some testing of the available build
* imported a testnet account I have with some Sepolia ETH
* enabled testnets in the settings
* switched to Sepolia and the balance was correctly displayed
* verified that Activity pane showed the incoming testnet ETH transaction
* sent small amount with destination address same as source (to myself)
* the Activity pane updated as the transaction progressed
* viewing the transaction details worked
* link in that view to the block explorer worked
* used https://bridge.arbitrum.io/ to bridge to Arbitrum Sepolia
* waited the ~10 minutes it took
* as soon as it was confirmed, switched to the L2 testnet and the balance was correct

I will note that I did not enjoy having to scroll on the network selector. Not sure if I would change it, but thought I would mention it in case someone has an idea.

Once more external factors are addressed to support the features mentioned I will proceed with verifying those as well.

background/constants/networks.ts
michalinacienciala: They [posted an answer](https://discord.com/channels/735965332958871634/1157338058362994840/1157800066229223424):

> Yes, Alchemy is aware of the upcoming deprecation of the Goerli testnet for Arbitrum. We are actively working on adding support for the Arbitrum Sepolia RPC to ensure a smooth transition for our users. Keep an eye on our ⁠📢│announcements updates for the official announcement!

background/constants/networks.ts
```diff
@@ -79,13 +81,29 @@ export const OPTIMISM: EVMNetwork = {
...
+  family: "EVM",
+  coingeckoPlatformID: "ethereum",
+}
+
+export const ARBITRUM_SEPOLIA: EVMNetwork = {
+  name: "Arbitrum Sepolia",
+  baseAsset: ARBITRUM_SEPOLIA_ETH,
+  chainID: "421614",
+  family: "EVM",
+  coingeckoPlatformID: "ethereum",
```
Shadowfiend, line 104: Should we leave these Coingecko references here and/or in the asset definitions? Wondering if these will give false prices for assets.

background/services/chain/taho-provider.ts
```diff
@@ -0,0 +1,30 @@
+import {
+  AlchemyProvider,
+  Network,
+  showThrottleMessage,
+} from "@ethersproject/providers"
+import { ConnectionInfo } from "@ethersproject/web"
+
+export default class TahoAlchemyProvider extends AlchemyProvider {
```
Shadowfiend, line 8: Let's drop a comment noting why this exists.

ui/_locales/en/messages.json
```diff
@@ -545,6 +545,7 @@
     "beta": "Mainnet (beta)",
     "testnet": "Test Network",
     "l2": "L2 scaling solution",
+    "l2Testnet": "L2 Test Network",
```
Shadowfiend, line 548: I would just use `testnet`, fwiw.

Shadowfiend (approved):
Looks good. Cross-checked chain ids, left a few small notes, and need to do a quick test drive. Will do that + merge tomorrow.

Shadowfiend:
Tested by moving stuff from old test wallet to new one, let's :shipit:

background/constants/networks.ts
michalinacienciala: I must admit, the `coingeckoPlatformID` value slipped my attention here and I copied the value from `GOERLI`/`SEPOLIA` networks. If I understand correctly, the `coingeckoPlatformID` value is used to fetch a price of the asset, so that we could show how much is the asset worth in USD. The list of supported values is here: https://api.coingecko.com/api/v3/asset_platforms. As there are no testnets listed there, I think in the extension we can either show a mainnet price (which is not a real value of the testnet asset) or not show the price at all. I prefer showing the mainnet value in that case - this allows us to test if the USD price is being shown in correct place, etc.

So what I think would make sense here is changing the `coingeckoPlatformID`'s value to `arbitrum-one` for `ARBITRUM_SEPOLIA` and leaving the rest unchanged.

Unless there's something I'm missing?

michalinacienciala:
@Shadowfiend , I addressed your comments in https://github.com/tahowallet/extension/pull/3636.

PR #3636, opened 2023-10-05
Sepolia support - small fixes

michalinacienciala:
In this PR we address the leftover comments from the review of https://github.com/tahowallet/extension/pull/3633 (_Add support for Mainnet Sepolia and Arbitrum Sepolia testnets_).

Latest build: [extension-builds-3636](https://github.com/tahowallet/extension/suites/16906558298/artifacts/965879218) (as of Thu, 05 Oct 2023 11:15:11 GMT).

### DIFF
--- a/background/services/chain/taho-provider.ts
+++ b/background/services/chain/taho-provider.ts
@@ -15,7 +15,7 @@ import { ConnectionInfo } from "@ethersproject/web"
 // upcoming patch, but as it's release date is unknown, we're adding the Sepolia
 // support by creating `TahoAlchemyProvider` class that handles this case.
 // In the future we may want to add there another case, for `arbitrum-sepolia`,
-// but we can't do that at this moment, as Alchemy dos not offers an RPC for
+// but we can't do that at this moment, as Alchemy does not offer an RPC for
 // Arbitrum Sepolia yet.
 export default class TahoAlchemyProvider extends AlchemyProvider {
   static override getUrl(network: Network, apiKey: string): ConnectionInfo {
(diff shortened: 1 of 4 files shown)


background/services/chain/taho-provider.ts
```diff
@@ -5,6 +5,18 @@ import {
...
+// allow the extension to fill the list of historical transactions on that
+// chain.
+// We can't just add `SEPOLIA` to `ALCHEMY_SUPPORTED_CHAIN_IDS` because the
+// `@ethersproject/providers@5.7.2` package that we use as a project dependency
+// does not support Sepolia yet (no `sepolia` case in the `getUrl` function of
+// `./src.ts/alchemy-provider.ts`). The support is planned to be added in the
+// upcoming patch, but as it's release date is unknown, we're adding the Sepolia
+// support by creating `TahoAlchemyProvider` class that handles this case.
+// In the future we may want to add there another case, for `arbitrum-sepolia`,
+// but we can't do that at this moment, as Alchemy dos not offers an RPC for
```
jagodarybacka, line 18:
```suggestion
// but we can't do that at this moment, as Alchemy does not offer an RPC for
```

jagodarybacka (approved):

background/services/chain/taho-provider.ts
michalinacienciala: Fixed in [cfb1708](https://github.com/tahowallet/extension/pull/3636/commits/cfb1708887a67b20971465307a4238a958d79f36).

michalinacienciala:
The e2e test in this PR will fail, but this is unrelated to the changes in the PR - the tests failures are caused by the fact that we are in the middle of switching to different testing accounts.

jagodarybacka (approved):
✅

PR #3638, opened 2023-10-06
Implement EIP-7524 (PLUME Signatures)

orenyomtov:
## Explanation
We want Taho to be the first wallet to support private voting and private airdrops on Ethereum and other EVM chains.

This PR adds new `eth_getPlumeSignature` RPC method that implements a [novel ECDSA nullifier scheme](https://aayushg.com/thesis.pdf) as [described](https://blog.aayushg.com/posts/nullifier) in [EIP-7524](https://github.com/ethereum/EIPs/blob/53683e7716ff7c962e5a30a087ced51a4c60951b/EIPS/eip-7524.md).

The `eth_getPlumeSignature` method takes in two parameters, a **message** and an **address**, then generates a deterministic signature (PLUME) and several other inputs. The plume can be used as a nullifier to prevent double-spending in an anonymity set. This capability unlocks novel on-chain behavior, such as [private DAO voting](https://prop.house/nouns/private-voting-research-sprint), [fair, non-doxxing airdrops](https://github.com/stealthdrop/stealthdrop), and more.

## Screenshot
plume confirmation window

## Manual Testing Steps

After building and running Taho locally, enter this into the browser console

```js
await window.ethereum.request({
  "method": "eth_requestAccounts",
  "params": []
});

accountAddress = (await window.ethereum.request({
  "method": "eth_accounts",
  "params": []
}))[0];

await window.ethereum.request({
  "method": "eth_getPlumeSignature",
  "params": [
    "this is a test message - hi aayush",
    accountAddress
  ]
});
```

A confirmation screen should open up. After clicking "Sign", you will see the plume and other signals outputted into the console.


mhluongo:
Umm, this is really cool! @orenyomtov could we get some more context on what you want to support here first?

Divide-By-0:
in the last month, 3 more protocols have voiced a clear demand for this:

gitcoin passport wants this feature asap so they can have nullifiers with eth addresses to avoid passport attestations involving ZK from being double-used [happy to link to the PM there or send a screenshot of a chat from him]
summa (proof of solvency/reserves) for CEXs requires PLUME to work without the exchanges revealing their balances, and is the main blocker
zksnap (private voting) is what ameen called the "holy grail" of private voting tradeoffs and requires plume as a critical component of the proofs

as these zk apps and usecases hit mainnet, we are going to be pushing harder on wallets for support -- we have open convos with metamask, rabby, and ledger now as well as PRs in review for all of them, and will open more convos in the coming months. we want to give taho a fighting chance to be at the forefront of zk innovation, so definitely let us know how you are feeling about this PR. happy to jump on a call or meet at ethdenver as well!

Divide-By-0:
> Umm, this is really cool! @orenyomtov could we get some more context on what you want to support here first?

@mhluongo Not exactly sure by what you mean about what we want to 'support here first', but the idea is that we want Taho users to be able to generate PLUME signatures using existing wallets. This PR adds the cryptographic, RPC, and UI functionality to do so. This would let them subsequently generate the needed ZK proofs on applications of their choosing (gitcoin passport for instance is blocked on this problem for their zk attestations).

mhluongo:
What happens here when a user is using a Ledger?

Divide-By-0:
> What happens here when a user is using a Ledger?

Hey! We have a specific PR available to Ledger devices as well right now, and are working on getting that approved as well in parallel. So we would simply call something like eth_getPlumeSignature on the Ledger, but we haven't added that yet because we haven't merged the Ledger PR yet.

PR #3645, opened 2023-10-20
Show a insufficient funds warning

kkosiorowska:
## Testing Env

```
SUPPORT_THE_ISLAND_ON_TENDERLY=true
```

### DIFF
--- a/background/services/chain/index.ts
+++ b/background/services/chain/index.ts
@@ -581,15 +581,25 @@ export default class ChainService extends BaseService<Events> {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         const anyError: any = error
 
-        if (
-          "code" in anyError &&
-          anyError.code === Logger.errors.UNPREDICTABLE_GAS_LIMIT
-        ) {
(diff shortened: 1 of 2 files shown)


background/services/chain/index.ts
```diff
@@ -581,15 +581,25 @@ export default class ChainService extends BaseService<Events> {
...
+          if (anyError.code === Logger.errors.INSUFFICIENT_FUNDS) {
+            gasEstimationError = anyError.error ?? "Insufficient funds error."
+
+            if (
+              "annotation" in transactionRequest &&
+              transactionRequest.annotation !== undefined
+            ) {
+              transactionRequest.annotation.warnings ??= []
+              transactionRequest.annotation.warnings.push("insufficient-funds")
+            }
```
Shadowfiend, line 598: We shouldn't be pushing annotations anywhere other than the enrichment service…

We also unfortunately have this estimation handling duplicated here and below in `populatePartialEIP1559TransactionRequest`.

Will think about the best way to tie these together.

mhluongo:
Closing as stale.

PR #3650, opened 2023-10-31
Sepo-Ledger: Patch @ledgerhq/hw-app-eth for Arbitrum Sepolia chain id handling

### LINKED ISSUE #3647: `transaction.chainId/signature.v mismatch` error when signing transactions using Ledger
I did some tests of staking in the Subscape dapp (I used the stage-live deployment) on Ethereum Sepolia and although some staking transactions were successful, some failed with the following errors in the wallet logs:
```
logger.ts:208 Error encountered! ledgerID: LEDGER_NANO_S_0x43866d77672dd19252e8ea2274ecece9c62d3262 transactionRequest: [object Object] error: Error: transaction.chainId/signature.v mismatch (argument="signature", value={"r":"0x2b069505b96a272c34d3820efdec59442b4424d4362bec7af2e368150df6a9c3","s":"0x4efc2c085a3688549d2ede91647168de8dbdd45434ab602fcefcaf69c906aaf6","v":843518}, code=INVALID_ARGUMENT, version=transactions/5.7.0)
```
and
```
Error signing transaction Error: transaction.chainId/signature.v mismatch (argument="signature", value={"r":"0x2b069505b96a272c34d3820efdec59442b4424d4362bec7af2e368150df6a9c3","s":"0x4efc2c085a3688549d2ede91647168de8dbdd45434ab602fcefcaf69c906aaf6","v":843518}, code=INVALID_ARGUMENT, version=transactions/5.7.0)
    at Logger.makeError (index.js:224:23)
    at Logger.throwError (index.js:233:20)
    at Logger.throwArgumentError (index.js:236:21)
    at _serialize (index.js:204:20)
    at serialize (index.js:221:16)
    at index.ts:415:35
```
and
```
ProviderBridgeService.routeContentScriptRPCRequest
Error processing request
```

I used version `0.50.0c1e7552` of the wallet and used Ledger Nano S (`2.60.0`) with Ethereum app (`1.10.1`) for signing.

Here are the tests that I did:
TEST 1. Stake 2 TAHO. Staking failed with a different error (`Transfer amount exceeds spender allowance`) - not in scope of this issue
TEST 2. Stake 2 TAHO. Both transactions successful
TEST 3. Stake 3 TAHO. `Token approval` successful, `Contract interaction` failed with `transaction.chainId/signature.v mismatch`
TEST 4. Stake 3 TAHO. `Contract interaction` failed with `transaction.chainId/signature.v mismatch`
TEST 5. Stake 3 TAHO. `Contract interaction` successful
TEST 6: Stake 1 TAHO. Both transactions successful
TEST 7: Stake 2 TAHO. Approve asset spend successful, Contract interaction failed with `transaction.chainId/signature.v mismatch`
TEST 8: Stake 2 TAHO. Contract interaction failed with `transaction.chainId/signature.v mismatch`
TEST 9: Stake 2 TAHO. Contract interaction failed with `transaction.chainId/signature.v mismatch`
TEST 10: Stake 2 TAHO. Contract interaction failed with `transaction.chainId/signature.v mismatch`
TEST 11: Stake 2 TAHO. Contract interaction failed with `transaction.chainId/signature.v mismatch`
TEST 12: Stake 2 TAHO. Contract interaction failed with `transaction.chainId/signature.v mismatch`

Video covering tests 6-12:

https://github.com/tahowallet/extension/assets/78352137/c299681f-74b5-4ce4-ba4f-09b7552a8e5c

Looking at the results, there's no clear rule about which transactions fail and which are successful.

I wonder if the problem may be caused by similar issue as discussed here: https://github.com/ethers-io/ext-signer-ledger/issues/6#issuecomment-1646196739. In this issue, bumping ledger libraries in `ether.js` (https://github.com/ethers-io/ethers.js/pull/1370/files) has helped. But we already use higher versions than those...

Shadowfiend:
It currently seems like there is an issue with Ledger's handling of high chain ids for certain transactions, where the Ledger can provide a first byte that breaks the expectations of the chain id adjustment done for legacy transactions to conform to EIP-155. The resulting `v` value returned from the signing flow is invalid and Ethers throws because of this.

This fix *may* be correct, but it may not be. To adjust for this fact, the patch limits adjustments of the underlying parity value in Ledger to happen only when the chain id matches Arbitrum Sepolia's (412614), as that's the only place we've seen consistent reports of the issue.

See also https://github.com/LedgerHQ/ledger-live/issues/5265 .

Fixes #3647 .

## Testing

- Easiest way to test here is really to just hammer https://app.taho.xyz/ . In reproducing, we've found that some transactions trigger the issue and others don't, depending on what the first byte of the Ledger response to the signing request is.

Latest build: [extension-builds-3650](https://github.com/tahowallet/extension/suites/17798752935/artifacts/1021248068) (as of Wed, 01 Nov 2023 09:45:37 GMT).

### DIFF
--- a/patches/@ledgerhq+hw-app-eth+6.33.4.patch
+++ b/patches/@ledgerhq+hw-app-eth+6.33.4.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/@ledgerhq/hw-app-eth/lib/Eth.js b/node_modules/@ledgerhq/hw-app-eth/lib/Eth.js
+index 4a7cb64..bd7051b 100644
+--- a/node_modules/@ledgerhq/hw-app-eth/lib/Eth.js
++++ b/node_modules/@ledgerhq/hw-app-eth/lib/Eth.js
+@@ -228,7 +228,7 @@ class Eth {
+                 }
+                 else {
+                     // Legacy type transaction with a big chain ID
+-                    v = chainId.times(2).plus(35).plus(ecc_parity).toString(16);
++                    v = chainId.times(2).plus(35).plus(chainIdTruncated === 412614 ? ecc_parity % 2 : ecc_parity).toString(16);
+                 }
+             }
+             else {


patches/@ledgerhq+hw-app-eth+6.33.4.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@ledgerhq/hw-app-eth/lib/Eth.js b/node_modules/@ledgerhq/hw-app-eth/lib/Eth.js
+index 4a7cb64..bd7051b 100644
+--- a/node_modules/@ledgerhq/hw-app-eth/lib/Eth.js
++++ b/node_modules/@ledgerhq/hw-app-eth/lib/Eth.js
+@@ -228,7 +228,7 @@ class Eth {
+                 }
+                 else {
+                     // Legacy type transaction with a big chain ID
+-                    v = chainId.times(2).plus(35).plus(ecc_parity).toString(16);
++                    v = chainId.times(2).plus(35).plus(chainIdTruncated === 412614 ? ecc_parity % 2 : ecc_parity).toString(16);
```
jagodarybacka, line 10: What is a type of `chainIdTruncated` here? Always number? Would it be safer to ensure `412614` and `chainIdTruncated` are using the same type?

jagodarybacka (commented):
Looks good, only this comparison is worrying me a bit. I'll let @andreachapman and @michalinacienciala QA it before merging (unless we want to QA it all in one batch on the release)

patches/@ledgerhq+hw-app-eth+6.33.4.patch
Shadowfiend: It's always a number, yeah. You can convince yourself by looking at the source TS file for `decodeTxInfo` (https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledgerjs/packages/hw-app-eth/src/utils.ts#L35 ) and asking for the function type from the compiler:

Screenshot 2023-11-01 at 11 49 29

Even if it isn't, it means we won't apply the adjustment—since the goal is to limit the application of the fix, I think that's the right behavior.

PR #3651, opened 2023-11-02
Fix e2e tests

michalinacienciala:


e2e-tests/regular/signing.spec.ts
```diff
@@ -26,8 +26,11 @@ test.describe("Signing", () => {
     await signDataPromptOpens.then(async (prompt) => {
       await prompt.getByRole("button", { name: "Sign" }).click()
     })
-
-    // If we see this then it means we were able to sign in
+    // There's a bug on login.xyz that makes the test fail
+    // (https://discord.com/channels/862419652286218251/886997073650655232/1173226370776694794).
```
jagodarybacka, line 30: I doubt storing the discord links in a code is a good way to document it 😅 Is the login.xyz website open source? Can we add an issue somewhere?

jagodarybacka (approved):
✅

jagodarybacka:
Ugh I see there is something else failing again on e2e flow, could you take a look?

e2e-tests/regular/signing.spec.ts
Shadowfiend: Unfortunately not, btw. I checked on Friday :)

michalinacienciala:
> Ugh I see there is something else failing again on e2e flow, could you take a look?

I wasn't able to reproduce the failure locally, but I think it was caused by the https://github.com/tahowallet/extension/issues/3675 bug. As we will likely want to fix the #3675 soon, I did not do any changes to the tests in relation to this bug.

I did a different change though ([f0441b1](https://github.com/tahowallet/extension/pull/3651/commits/f0441b13e01e76cf95c5480db7ec1d2cad1ce3ad)), so I need a new review.

e2e-tests/regular/signing.spec.ts
michalinacienciala: Yep, I also prefer documenting such issues in GH, but did not find a repo for the project, hence I contacted them via Discord and documented that.

Btw, I did get a response - they will look into the issue.

jagodarybacka (approved):
Sorry for the double update, Github glitched 
image

Looks good but we will need @Shadowfiend to approve because I'm not in a code owners group

Shadowfiend (approved):
Lgtm let’s do it. Sorry to block here!

Issue #3669, opened 2023-11-13
Subscape access in wallet extension


Shadowfiend:
Goal is to be able to access Subscape easily from the Taho wallet. Should be tucked away, and only appear if the user has beta/testnet TAHO.

Figma link:
https://www.figma.com/file/bEY6fqdX6LiPKP7J1eg7Ag/TokenLaunch-Dapp?type=design&node-id=5392-10346&mode=design&t=c4t7SlkfyTG4R6ro-4

VladUXUI:
@Shadowfiend i was thinking this would be open to everybody, not only to users that have testnet Taho.
It simplifies mechanics, and we don't get a "were did that button dissapear" if users is on another address that doesn't have TAHO.

It also might give us a small boost in visibility across people that aren't aware/active, but can still sign up for waitlist

VladUXUI:
Discussed with Jagoda, but also adding here, it should have an expiration date. Dec 14th

PR #3670, opened 2023-11-14
Add static subcape link to the wallet

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3669

### What

Subscape link in the wallet:
- hidden on the right
- visible on hover
- visible for everyone, no matter the TAHO/veTAHO balance

Button will be hidden later after the beta season will end (https://github.com/tahowallet/extension/issues/3673)

### Testing

- [x] make sure link is only overlapping the wallet page UI, slideups (accounts list, networks list) should be displayed over the link
- [x] make sure you can open the Subscape using the link
- [x] make sure you can use wallet page UI - no regressions

https://github.com/tahowallet/extension/assets/20949277/6bbc8a2c-3675-4e6b-b63d-7424065df440

Latest build: [extension-builds-3670](https://github.com/tahowallet/extension/suites/18276596797/artifacts/1056726491) (as of Fri, 17 Nov 2023 09:58:43 GMT).


ui/public/index.css
```diff
@@ -36,6 +36,9 @@
   --trading-outgoing: #ef6194;
   --trading-incoming: #11bea9;
 
+  --subscape: #e4eeee;
+  --subscape-background: #315A5A;
```
xpaczka, line 40: Let's stay consistent with lowercase hex values here

```suggestion
  --subscape-background: #315a5a;
```

ui/components/Wallet/WalletSubscapeLink.tsx
```diff
@@ -0,0 +1,59 @@
...
+        icon="new_tab@2x.png"
+        width={16}
+        color="var(--trophy-gold)"
+        customStyles="margin-left: 5px;"
+      />
+      <style jsx>{`
+        .subscape_link {
+          cursor: pointer;
+          position: absolute;
+          z-index: 998; // Above the UI, below the menu
```
xpaczka, line 35: Maybe we could set it as a CSS variable (not sure how reusable it'd be, up to you)

jagodarybacka: I would love to but this is a bigger refactor because we have a lot of z-indexes to take care of, let's do it later

xpaczka (approved):
:shipit:

ui/components/Wallet/WalletSubscapeLink.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export default function WalletSubspaceLink(): ReactElement {
+  const [isVisible, setIsVisible] = useState(false)
+
+  return (
+    <button
+      type="button"
+      className={classNames("subscape_link", { visible: isVisible })}
+      onClick={() => {
+        window.open("https://app.taho.xyz/", "_blank")?.focus()
```
ioay, line 13: Generally, we should avoid passing the arrow function directly as a prop to improve performance, because the arrow function in prop creates a new function each time the component renders. In this case, it is a small issue, but in the future, it's good to create separate` const openDapp = () => {}` (also using useMemo) and only then pass it as a prop.

ui/components/Wallet/WalletSubscapeLink.tsx
```diff
@@ -0,0 +1,59 @@
...
+      }}
+      onMouseEnter={() => setIsVisible(true)}
```
ioay, line 15: Also here we can create const showButton/hideButton and based on the state `isVisible` we should return null or button because I don't see the reason to return the button when it's not needed.

ui/components/Wallet/WalletSubscapeLink.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { ReactElement, useState } from "react"
+import classNames from "classnames"
+import SharedIcon from "../Shared/SharedIcon"
+
+export default function WalletSubspaceLink(): ReactElement {
+  const [isVisible, setIsVisible] = useState(false)
+
+  return (
+    <button
```
ioay, line 9: Based on the state `isVisible` we should return null or button because I don't see the reason to return the button when it's not needed.

jagodarybacka: it is needed all the time, if `isVisible = false` then the button is tucked away but the icon is still visible

ui/components/Wallet/WalletSubscapeLink.tsx
jagodarybacka: I see but in this repo it is a common practice and I believe that if arrow function is assigned to the low-level DOM node (not other React component) the optimisation it provides is super small so we generally were not concerned about it in the past

ui/components/Wallet/WalletSubscapeLink.tsx
Shadowfiend: I think this points to a tweak we can make to the naming. `isVisible` should probably be `isCompact`, `isIconOnly`, `onlyShowIcon` or something similar to indicate that we're not hiding the whole link.

ui/components/Wallet/WalletSubscapeLink.tsx
jagodarybacka: https://github.com/tahowallet/extension/issues/3676

ui/components/Wallet/WalletSubscapeLink.tsx
ioay: We can leave it as it is here, because it's small issue, but generally in the future we must pay special attention to this considering the general performance problems of the application.

ioay (approved):
LGTM, we can merge it after changing the name of `isFullyVisible` to `isIconOnly` or `showIcon` which was suggested by @Shadowfiend.

ui/components/Wallet/WalletSubscapeLink.tsx
Shadowfiend: Hadn't seen variables used for z-indices before, I like that!

ui/components/Wallet/WalletSubscapeLink.tsx
Shadowfiend: We should def do some digging. For the most part we've found that performance issues have come in places that aren't strictly related to rendering (in particular, they've come from the JSON serialization/deserialization that happens for state sync between background and foreground scripts). Could be that we're compounding that, too.

We've reduced a lot of that, but we still do a full JSON serialization of redux state to local storage every 50ms or so during busy times, which I'm guessing is still at the core of a lot of the pain here.

ui/components/Wallet/WalletSubscapeLink.tsx
Shadowfiend: Want to be clear that I was open to any name, these were just suggestions. So `isFullyVisible` (https://github.com/tahowallet/extension/pull/3670/commits/55024face2591993083e008d22dbcbc4f6ba781a) and `isIconOnly` (https://github.com/tahowallet/extension/pull/3670/commits/1de5e109ff9ce7c312fde87af60f7b98fb874bb9) are both fine.

Shadowfiend (approved):
Let's :shipit:

Shadowfiend (approved):
(That's still a yes post-merge lol.)

PR #3689, opened 2023-12-04
Z-index refactor

### LINKED ISSUE #3676: Z-index refactor
We don't have specified z-indexes used in the wallet and it becomes increasingly harder to add new z-indexes. 
Let's refactor z-indexes - name the layers and create css variables for each layer.

xpaczka:
Resolves #3676 

## What has been done:
- Added `z-index` variables and renamed them
- Changed `z-index` values (no need for f.e. 9999999999)


jagodarybacka:
Seems like some of the e2e tests are failing because elements they expect are not visible - could you run e2e tests locally and manually check if the correct elements are visible when they are expected?

Issue #3685, opened 2023-11-29
Unable to remove custom networks


beemeeupnow:
### What browsers are you seeing the problem on?

Chrome

### What were you trying to do?

Added custom network via ChainList, then tried to remove it from the Custom networks page in Settings. (in this particular case, Base with Llamanodes RPC)

### What did not work?

The network is not successfully removed.

### Version

v0.51.0

### Relevant log output

```shell
{
    "_e": {},
    "name": "NotFoundError",
    "message": "Failed to execute 'objectStore' on 'IDBTransaction': The specified object store was not found.\n NotFoundError: Failed to execute 'objectStore' on 'IDBTransaction': The specified object store was not found.",
    "inner": {
        "stack": "Error: Failed to execute 'objectStore' on 'IDBTransaction': The specified object store was not found.\n    at chrome-extension://[REDACTED]/background.js:100478:29\n    at Ie (chrome-extension://[REDACTED]/background.js:98897:11)\n    at new Ce (chrome-extension://[REDACTED]/background.js:98867:58)\n    at Object.count (chrome-extension://[REDACTED]/background.js:100477:22)\n    at h.<computed> [as count] (chrome-extension://[REDACTED]/background.js:101347:31)\n    at Object.count (chrome-extension://[REDACTED]/background.js:100919:31)\n    at chrome-extension://[REDACTED]/background.js:99917:33\n    at i (chrome-extension://[REDACTED]/background.js:99295:20)\n    at chrome-extension://[REDACTED]/background.js:100209:25\n    at Ie (chrome-extension://[REDACTED]/background.js:98897:11)"
    }
}
```

EggsyOnCode:
```js
removeChainBalances: (
      immerState,
      { payload: chainID }: { payload: string },
    ) => {
      delete immerState.accountsData.evm[chainID]
    },
```

there seems to be a problem with this reducer of the accountSlice in accounts.ts. The DB is not being updated ; that's why its not being reflected in the UI

Shadowfiend:
This is probably failing in https://github.com/tahowallet/extension/blob/main/background/main.ts#L1974 -> so in one of these db calls:

- https://github.com/tahowallet/extension/blob/main/background/services/provider-bridge/db.ts#L142
- https://github.com/tahowallet/extension/blob/main/background/services/internal-ethereum-provider/db.ts#L81
- https://github.com/tahowallet/extension/blob/main/background/services/chain/db.ts#L288

PR #3699, opened 2023-12-13
Fix removing custom networks

jagodarybacka:
Resolves https://github.com/tahowallet/extension/issues/3685

### What

Add database transaction dependency to fix removing items from `customRpcConfig` table.

### Testing

- [x] add custom network
- [x] succesfully remove it

Latest build: [extension-builds-3699](https://github.com/tahowallet/extension/suites/19016838762/artifacts/1114194310) (as of Thu, 14 Dec 2023 09:34:36 GMT).


michalinacienciala (approved):
Tested, looks OK. Was able to add custom network and remove it. Tested removing network that was selected and one that wasn't. Also tested adding a network that was previously removed.

Issue #3675, opened 2023-11-16
Removed assets reappear on the token list


michalinacienciala:
The once removed assets re-appear in the wallet soon after their removal. Happens for all removed assets (no matter if removed from the list of verified or unverified assets).

1. Import the account with some unverified assets
2. Go to the unverified asset ant click `Verify asset -> Don't show` (either from token list or token details)
3. Observe token being removed from the list (OK) and soon being listed again (NOOK).

https://github.com/tahowallet/extension/assets/78352137/5a6fdea0-e112-49a3-91d6-f47f44ab57d2

Tested on:
`bnbspider2.crypto` and `0x0581470a8b62bd35dbf121a6329d43e7edd20fc7` accounts
Taho 0.51.0 (from Chrome store) and a local build from `fix-e2e-tests`

PR #3700, opened 2023-12-14
Support video avatars

### LINKED ISSUE #3688: Support video user avatars
If user avatar is a video then the wallet doesn't know how to display it correctly (as it assumes that will be an image)
Let's support video avatars as well.

image

xpaczka:
Resolves #3688

## What has been done

Added video support for wallet avatars

## Testing
- [x] Test with video avatar (example user `0xab7375daf14bc661a0c9aff8800d49a34b037a98`)
- [x] Test with image avatar

Latest build: [extension-builds-3700](https://github.com/tahowallet/extension/suites/19135090954/artifacts/1121162389) (as of Mon, 18 Dec 2023 11:20:52 GMT).


michalinacienciala:
Looks that something is off about the default doggos avatars - they look zoomed in:
image

jagodarybacka (changes requested):

michalinacienciala:
The failures in the tests are caused by a known issue: https://github.com/tahowallet/extension/issues/3675.
In the tests we're first mark `BANANA` token as trusted and later remove it from the token list. The test expects to not see the BANANA token on the list after that, but due to the bug the token reappears on the list.
This may not always be failing, as sometimes it takes longer and sometimes shorter for the deleted asset to reappear.

background/redux-slices/migrations/to-36.ts
```diff
@@ -0,0 +1,82 @@
...
+    accountsData: {
+      evm: {
+        [chainID: string]: {
+          [address: string]:
+            | "loading"
+            | {
+                ens: {
+                  name?: DomainName
+                  avatarURL?: URI
+                  avatarType?: string
```
jagodarybacka, line 15: old state shouldn't have `avatarType` because this is actually the field we just added in this PR right?

background/redux-slices/migrations/to-36.ts
```diff
@@ -0,0 +1,82 @@
...
+
+          accountData.avatarType = avatarType ?? undefined
+        }
+      }
+    }),
+  )
+
+  return {
+    ...typedPrevState,
+    assets: [],
```
jagodarybacka, line 80: why we are resetting assets?

background/redux-slices/migrations/to-36.ts
```diff
@@ -0,0 +1,82 @@
...
+      const account = accountsData.evm[chainID][address]
+
+      if (account !== "loading") {
+        const accountData = account.ens
+        const { avatarURL } = accountData
+
+        if (!avatarURL) {
+          accountData.avatarType = undefined
+        } else {
+          const fileTypeResponse = await fetch(avatarURL, { method: "HEAD" })
```
jagodarybacka, line 69: Hmm I don't think this will work correctly because we are not waiting for results here at all, script will continue sync execution and will deserialize the preloaded state without waiting for these fetches. Actually, now I'm even thinking that we shouldn't even do it here at all - migrations should be working fast and be deterministic 🤔 Sorry about making it more confusing than it should be 😓 
Can we try to load the `avatarType` when the actual avatar is being loaded on the frontend side?

jagodarybacka (changes requested):

jagodarybacka (approved):
Looks 🔥 
image

PR #3704, opened 2023-12-15
Check DevTools when executing tests on release testlist

michalinacienciala:
We want to make sure that during the release testing the tests are executed with an open DevTools window and that the logs and network traffic are monitored.

Latest build: [extension-builds-3704](https://github.com/tahowallet/extension/suites/19073390611/artifacts/1116830471) (as of Fri, 15 Dec 2023 09:04:07 GMT).


jagodarybacka (approved):
❤️‍🩹

PR #3713, opened 2023-12-19
Show wallet address after hovering on wallet name

### LINKED ISSUE #3171: Show wallet address after hovering on wallet name
There is a couple of places in the extension where we show the names associated with the wallet addresses. It would be nice to show fa full or shortened wallet address when hovering over that name.

Proposed places:
1. At the right top corner of the `Wallet` view. Currently we show `Copy address` and the copy icon on hover. Maybe we could replace the `Copy address` text with the shortened address and leave the copy icon next to it.
image
2. In the `Show/hide accounts` list on `Filter collections` view. Currently we don't show anything on hover.
image
3.  In the `Show/hide accounts` list on `Filter abilities` view. Currently we don't show anything on hover.
image

xpaczka:
Resolves #3171 

## What has been done

- Show address to be copies in "Copy address" tooltip
- In filters on account list hover, show the truncated address (instead of ens/uns name)

## Testing
- [ ] Hover over item on accounts list in filters - address shows

Zrzut ekranu 2023-12-19 o 11 29 19

Latest build: [extension-builds-3713](https://github.com/tahowallet/extension/suites/19199860149/artifacts/1128487044) (as of Thu, 21 Dec 2023 09:31:19 GMT).

### DIFF
--- a/ui/components/NFTs/Filters/FilterList.tsx
+++ b/ui/components/NFTs/Filters/FilterList.tsx
@@ -1,5 +1,7 @@
 import { Filter } from "@tallyho/tally-background/redux-slices/nfts"
-import React from "react"
+import React, { useCallback } from "react"
+import { setSnackbarMessage } from "@tallyho/tally-background/redux-slices/ui"
+import { useDispatch } from "react-redux"
 import SharedSkeletonLoader from "../../Shared/SharedSkeletonLoader"
 import SharedToggleItem from "../../Shared/SharedToggleItem"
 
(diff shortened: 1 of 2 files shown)


jagodarybacka:
image

Oh I think that it doesn't make sense to show the tooltip with the actual address here - it makes UI less readable IMO compared to having `copy address` CTA. Other places that were mentioned in the issue make more sense. We also have a component that will allow to copy address on click - like here:
image

I think we should be using this component wherever possible and show ENS/UNS instead of the short address - it makes the UI more user friendly and easier to work with.

Shadowfiend:
Not relevant atm, let's clean it up. We can restore later if needed.

PR #3733, opened 2024-01-30
Changed the build for the firefox browser

kogeletey:
In new versions the extension was simply not built for this browser.
Changed the build scripts and made them cross-managers.
Now it can be compiled not only by docker.

In general, everything works, except probably the main thing - sign in at dapps.
If it ever gets fixed it will be great

Latest build: [extension-builds-3733](https://github.com/tahowallet/extension/suites/20259618358/artifacts/1209187521) (as of Wed, 31 Jan 2024 15:25:58 GMT).


Shadowfiend (commented):
This is fairly well-contained. In general we haven't been supporting Firefox, and I don't expect we'll have the bandwidth to do it moving forward—are you willing to maintain this longer term? We'll probably merge it regardless, but want to add a note that it doesn't have an assigned maintainer.

kogeletey:
As I understand it, you want to compete with metamask, which means youu need to be on firefox.
I'll try to come up with something with authorization and send you path

Shadowfiend (approved):
Looking more closely, there's really no change here between what we had and what we have now other than it actually works. So going to unblock here.

PR #3734, opened 2024-02-16
Disable abilities extension-wide

Shadowfiend:
Due to the way that abilities are currently implemented, load on the Daylight API side is very high. We don't have the cycles right now to implement the underlying changes that would improve API load, so for now we disable abilities fetching, as well as access to abilities in the UI.

The code that is commented out is targeted to only remove the parts that link to the abilities UI and the parts that call out to the Daylight API; the rest remains in place so that turning things back on once the fetching strategy can be improved is straightforward.

Latest build: [extension-builds-3734](https://github.com/tahowallet/extension/suites/20839694050/artifacts/1252426921) (as of Fri, 16 Feb 2024 21:10:27 GMT).

### DIFF
--- a/background/services/abilities/index.ts
+++ b/background/services/abilities/index.ts
@@ -5,7 +5,7 @@ import {
   createSpamReport,
   DaylightAbility,
   DaylightAbilityRequirement,
-  getDaylightAbilities,
+  // getDaylightAbilities,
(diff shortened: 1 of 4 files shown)


Shadowfiend:
Turboing this for release.

PR #3737, opened 2024-02-29
Support for Wallet Test Framework

SamWilsn:
Hey!

I'm one of the developers on the [Wallet Test Framework](https://wtf.allwallet.dev/about/) project. As part of our grant from the Ethereum Foundation, we've built some integrations with wallets&mdash;Taho included.

This pull request contains the changes we needed to make to automate your wallet: adding a few `id` attributes, and introducing an attribute to differentiate between signing and sending a transaction. Neither myself nor @gaudren are experienced TypeScript developers, so please forgive any obvious mistakes here.

If you're curious what Taho+WTF looks like:

https://github.com/tahowallet/extension/assets/57262657/303d1df2-6458-440d-8d9e-3abbc302f720

And the glue code that makes this work: https://github.com/wallet-test-framework/glue-taho

Assuming you folks are on board, we'd like to eventually discuss integrating WTF with your testing strategy!

Latest build: [extension-builds-3737](https://github.com/tahowallet/extension/suites/21568125918/artifacts/1339626984) (as of Tue, 19 Mar 2024 16:33:25 GMT).

### DIFF
--- a/background/services/internal-ethereum-provider/index.ts
+++ b/background/services/internal-ethereum-provider/index.ts
@@ -109,6 +109,7 @@ type Events = ServiceLifecycleEvents & {
     Partial<EnrichedEVMTransactionRequest> & {
       from: string
       network: EVMNetwork
+      broadcastOnSign: boolean
     },
     SignedTransaction
   >
@@ -245,6 +246,7 @@ export default class InternalEthereumProviderService extends BaseService<Events>
             ...(params[0] as JsonRpcTransactionRequest),
           },
           origin,
+          true,
         ).then(async (signed) => {
           await this.chainService.broadcastSignedTransaction(signed)
           return signed.hash
@@ -253,6 +255,7 @@ export default class InternalEthereumProviderService extends BaseService<Events>
         return this.signTransaction(
           params[0] as JsonRpcTransactionRequest,
           origin,
+          false,
         ).then((signedTransaction) =>
           serializeEthersTransaction(
             ethersTransactionFromSignedTransaction(signedTransaction),
@@ -394,6 +397,7 @@ export default class InternalEthereumProviderService extends BaseService<Events>
   private async signTransaction(
     transactionRequest: JsonRpcTransactionRequest,
     origin: string,
+    broadcastOnSign: boolean,
   ): Promise<SignedTransaction> {
     const annotation =
       origin === TAHO_INTERNAL_ORIGIN &&
@@ -457,6 +461,7 @@ export default class InternalEthereumProviderService extends BaseService<Events>
           from,
           network: currentNetwork,
           annotation,
+          broadcastOnSign,
         },
         resolver: resolve,
         rejecter: reject,
--- a/ui/components/Shared/SharedCurrentAccountInformation.tsx
+++ b/ui/components/Shared/SharedCurrentAccountInformation.tsx
@@ -5,6 +5,7 @@ import { useAreInternalSignersUnlocked } from "../../hooks/signing-hooks"
 import SharedAvatar from "./SharedAvatar"
 
 type Props = {
+  address: string
   shortenedAddress: string
   name: string | undefined
   avatarURL: string | undefined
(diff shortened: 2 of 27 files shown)


ui/components/Signing/index.tsx
```diff
@@ -37,8 +37,11 @@ function SigningLoaded<T extends SignOperationType>({
     getAccountTotal(state, signingAddress),
   )
 
+  const broadcastOnSign =
+    "broadcastOnSign" in request && request.broadcastOnSign
+
   return (
-    <section>
+    <section data-broadcast-on-sign={broadcastOnSign}>
```
Shadowfiend, line 44: The fact that we have to drill the property through so many layers makes me feel weird, but I haven't come up with a better approach off the top of my head. I'm going to sit with it for a little bit :)

Meantime though, is there any chance we can assign this `data-` attribute at https://github.com/tahowallet/extension/blob/1b9fdcfec7e8d9d4ad10af75add7ca0d6966acc5/ui/components/Signing/SignatureDetails/TransactionSignatureDetails/index.tsx#L19 , or does it have to be available while the skeleton loader displays? I'm asking because in `TransactionSignatureDetails` we already know that we're dealing with a transaction request, which makes it a better place to handle transaction-request-specific code vs here where we're trying to treat it as a generic signing request.

Shadowfiend (commented):
This is super cool, thank you so much for the contribution! I have two aspects of the `data-broadcast-on-sign` piece that I'm grappling with—one is for me to chew on some more, one is a question for you in terms of where we're setting the attribute.

We're likely going to be shipping a small release next week, so I expect we can bundle this in with that!

SamWilsn:
I moved the attribute as suggested. Let me know if there's anything else you'd like to see changed!

background/services/internal-ethereum-provider/index.ts
```diff
@@ -109,6 +109,7 @@ type Events = ServiceLifecycleEvents & {
     Partial<EnrichedEVMTransactionRequest> & {
       from: string
       network: EVMNetwork
+      broadcastOnSign: boolean
```
Shadowfiend, line 112: Noticing this type is the same as the enrichment type that's being updated here, could reuse that. Just a note to self though 😁

ui/components/Shared/SharedCurrentAccountInformation.tsx
```diff
@@ -23,7 +25,7 @@ export default function SharedCurrentAccountInformation({
   const icon = areInternalSignersUnlocked ? "unlock" : "lock"
   return (
     <div className={classNames("account_info_wrap", { hover: showHoverStyle })}>
-      <span className="account_info_label ellipsis">
+      <span title={address} className="account_info_label ellipsis">
```
Shadowfiend, line 28: Glorious, thank you for this one 💪

Shadowfiend (approved):
Took a deeper look and actually the way `broadcastOnSign` was added is perfect, thank you so much! Running the build shortly and then we'll merge and hopefully ship in a release soon.

ui/components/SignData/EIP191Info.tsx
```diff
@@ -17,7 +17,7 @@ export default function EIP191Info({
     <>
       <div className="message">
         <div className="message-title">{t("message")}</div>
-        <div className="light">{`${signingData}`}</div>
+        <div id="message-content" className="light">{`${signingData}`}</div>
```
hyphenized, line 20: Not sure if this will cause issues [here](https://github.com/wallet-test-framework/glue-taho/blob/master/src/glue.ts#L204). Might be better to stay consistent with using either `data-testid` or `id`.

SamWilsn: I'd prefer to just use `id` if we can. The `data-testid` stuff is how we automate Coinbase (and I think this is a leftover from when we ported our glue to Taho :sweat_smile:)

Shadowfiend:
Ah, missing secrets again. Let me see what I can do here.

Shadowfiend:
Ok we're just going to see how these do on main heh.

PR #3743, opened 2024-04-21
Move autolock timer

### LINKED ISSUE #3572: Settings: Move auto lock timer
Auto lock timer should be in General and not in Networks & Assets
Can be the last item

radchukd:
Resolves #3572 

## What

Move auto-lock timer
autolock

Fix SharedSelect trigger z-index
settings

Fix OnboardingDerivationPathSelect width
width

Latest build: [extension-builds-3743](https://github.com/tahowallet/extension/suites/25993316338/artifacts/1700067915) (as of Mon, 15 Jul 2024 01:43:14 GMT).


ui/components/Shared/SharedSelect.tsx
```diff
@@ -167,7 +167,6 @@ export default function SharedSelect(props: Props): ReactElement {
 
           .button {
             position: relative;
-            z-index: 1;
```
Shadowfiend, line 170: Worried that this misses a place where the z-index does matter. Are we sure it doesn't?

ui/pages/Onboarding/Tabbed/ImportSeed.tsx
```diff
@@ -180,6 +180,7 @@ export default function ImportSeed(props: Props): ReactElement {
           margin-top: 16px;
         }
         .select_wrapper {
+          width: 100%;
```
Shadowfiend, line 183: 👌

ui/pages/Settings.tsx
```diff
@@ -412,6 +412,7 @@ export default function Settings(): ReactElement {
           FeatureFlags.SUPPORT_ACHIEVEMENTS_BANNER,
           notificationBanner,
         ),
+        autoLockSettings,
```
Shadowfiend, line 415: Solid catch!

Shadowfiend (commented):
Not quite ready to merge here as I'm worried that the z-index will have unintended side-effects. In particular, this z-index was introduced in https://github.com/tahowallet/extension/commit/8cf22e4824c8e61d6f98a63d28acb7943bea204d, which handled the slippage tolerance dropdown for swaps—wondering if removing it creates problems there.

ui/components/Shared/SharedSelect.tsx
radchukd: <SharedSelect /> is being used in 2 places as of now:

- `<Settings />`, where it actually overflows another one, hence the removal of `z-index` (screenshot 2)
- `<OnboardingDerivationPathSelect />`, where `z-index` does not affect anything.
 
IMO, this select trigger is a button and there is no reason for it to have an abnormal `z-index`, especially if it's higher than select menu.

Shadowfiend (approved):
All righty let's land this, sorry for the delay here.

PR #3744, opened 2024-06-04
Feat/add gridplus lattice1

mrcnk:
## Changes

- Adds a new onboarding entry for GridPlus Lattice1.
- Adds a new service to communicate with Lattice1.
- Adds a new signer that signs with Lattice1.
- Allows to sign with Lattice1 in window.ethereum.
- Adds an E2E test for the onboarding with Lattice1.

## Screenshots

CleanShot 2024-06-04 at 15 21 29
CleanShot 2024-06-04 at 15 22 00
CleanShot 2024-06-04 at 15 22 16
CleanShot 2024-06-04 at 15 23 21
CleanShot 2024-06-04 at 15 24 19
CleanShot 2024-06-04 at 15 25 20


background/services/gridplus/index.ts
```diff
@@ -0,0 +1,278 @@
...
+  ) {
+    if (
+      typeof typedData !== "object" ||
+      !(typedData.types || typedData.primaryType || typedData.domain)
+    ) {
+      throw new Error("unsupported typed data version")
+    }
+    const accounts = await this.readAddresses()
+    const accountData = accounts.find((account) => account.address === address)
+    const { domain, types, primaryType, message } = TypedDataUtils.sanitizeData(
```
Shadowfiend, line 174: By the time we reach this point we've already passed everything through ethers's typed message data handling—does this provide any material functionality that ethers doesn't already do?

background/services/gridplus/index.ts
```diff
@@ -0,0 +1,278 @@
+import { storage } from "webextension-polyfill"
+import {
+  fetchAddresses,
+  pair,
+  setup,
+  signMessage,
```
Shadowfiend, line 6: Let's rename this since right now it's a duplicate of the `signMessage` within, leading to hard-to-follow code.

An easy option is to `import * as GridPlus from "gridplus-sdk"` and then call `GridPlus.signMessage`, etc.

background/main.ts
```diff
@@ -790,6 +805,67 @@ export default class Main extends BaseService<never> {
...
+    return this.gridplusService.setupClient({ deviceId, password })
+  }
+
+  async pairGridplusDevice({ pairingCode }: { pairingCode: string }) {
+    return this.gridplusService.pairDevice({ pairingCode })
+  }
+
+  async fetchGridPlusAddresses({
+    n = 10,
+    startPath = [0x80000000 + 44, 0x80000000 + 60, 0x80000000, 0, 0],
```
Shadowfiend, line 824: Methods in this service should be dumb relays, so no default values—default values should happen in the GridPlusService instead.

Shadowfiend (commented):
A few early comments, more to come this week.

background/services/gridplus/index.ts
mrcnk: Just tested it, and it works fine without the additional sanitization. Resolved.

background/services/preferences/db.ts
```diff
@@ -21,6 +21,8 @@ const getSignerRecordId = (signer: AccountSignerWithId): SignerRecordId => {
       return `${signer.type}/${signer.keyringID}`
     case "private-key":
       return `${signer.type}/${signer.walletID}`
+    case "gridplus":
+      return `${signer.type}/${signer.path}`
```
Shadowfiend, line 25: Note that the purpose here is to distinguish “signers” in the sense of accounts that would be derived from the same place. For Ledger, each “device” would get its own signer. I'm worried that `path` here refers to a full derivation path, which would mean each address would get its own “signer”. That results in poor UX because each signer gets a grouping in the selector list, with accounts getting entries under the signer; for example, here, “Taho 1” is a signer, and `testertesting.eth` is an account:

Screenshot 2024-10-05 at 20 59 47

If every account added from a GridPlus device corresponds to a grouping, that'll potentially create a lot of clutter—instead.

Ultimately right now this is only used to manage that category title (“Taho 1” in the example above), so it's not a huge blocker—but it may be worth considering if it would be better to funnel the device id out of the service so that it can be used for differentiation here like it is for Ledgers.

background/services/gridplus/index.ts
```diff
@@ -0,0 +1,267 @@
...
+    this.readClient()
+    this.readAddresses()
+  }
+
+  static create: ServiceCreatorFunction<Events, GridplusService, []> =
+    async () => new this()
+
+  async readClient() {
+    this.client =
+      (await storage.local.get(CLIENT_STORAGE_KEY))?.[CLIENT_STORAGE_KEY] ??
```
Shadowfiend, line 67: Services generally don't read or write from local storage—that's considered only for ephemeral data. Services instead own an IndexedDB database (see e.g. `ledger/db.ts`) that handles querying, etc.

I don't think this is a blocker, but it's a relatively important part of the underlying architecture as local storage has aggressive size quotas and can fail in unexpected or silent ways when those size quotas are exceeded, while IndexedDB has a much more generous quote and clearer failure modes.

background/services/gridplus/index.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  async readAddresses() {
+    const persistedAddresses =
+      (await storage.local.get(ADDRESSES_STORAGE_KEY))?.[
+        ADDRESSES_STORAGE_KEY
+      ] ?? "[]"
+    const activeAddresses = JSON.parse(persistedAddresses)
+    this.activeAddresses = activeAddresses
+    return this.activeAddresses
+  }
```
Shadowfiend, line 80: Currently this happens in `Popup.tsx` via a hook, but instead it should happen during the `internalStartService` call that triggers when the service starts up (as the service's lifecycle is typically longer than the lifecycle of the popup).

background/services/gridplus/index.ts
```diff
@@ -0,0 +1,267 @@
...
+      deviceId,
+      password,
+      name: APP_NAME,
+      getStoredClient: () => this.client ?? "",
+      setStoredClient: writeClient,
+    })
+  }
+
+  async pairDevice({ pairingCode }: { pairingCode: string }) {
+    await this.readClient()
```
Shadowfiend, line 99: What's the purpose of running this every time `pairDevice`, `fetchAddresses`, etc are run? It seems like it just reads data that's been serialized, but that data remains in-memory, no? I don't see it cleared anywhere.

Also, is `this.client` secret/sensitive data or is it irrelevant if someone else gets access to it?

Shadowfiend (commented):
Did a final pass here, I think a couple of questions remain on my end:
- Am I reading this right that this does not accommodate multiple Lattice1s? The data model in the service does not seem to allow for that, but I could be misreading. It needs to be massaged anyway (from local storage to IndexedDB), so it's not a massive deal.
- Does the current setup assume that the Lattice1 will always be connected? In particular, it seems like it doesn't handle connection status, and my understanding is the Lattice1 is communicated with over the internet—so a failure in the wifi router, having the Lattice1 off, etc, could result in the Lattice1 not actually being available.

There are two notes below that requires some work I'll be doing (one is about renaming from `gridplus` to `grid-plus` or `gridPlus` and `Gridplus` to `GridPlus`, depending on where we are, just for consistency; the other is about a specific aspect of internal-ethereum-provider). The rest I don't think is a blocker, and we can merge & release with future iterations of work addressing the remainder.

background/services/internal-ethereum-provider/index.ts
```diff
@@ -503,10 +503,13 @@ export default class InternalEthereumProviderService extends BaseService<Events>
     // Ethers does not want to see the EIP712Domain field, extract it.
     const { EIP712Domain, ...typesForSigning } = params.typedData.types
 
+    const { domain } = params.typedData
+    domain.chainId = params.account.network.chainID
```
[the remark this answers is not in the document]
Shadowfiend: Hm, reading further, we actually catch this in the `SigningService` right now, where we check the account network chainID against the chainID passed in the data:

https://github.com/tahowallet/extension/blob/34d13502be9d5fd592912ca6be88505aeac657ab/background/services/signing/index.ts#L244-L260

With this in mind, the above code actually circumvents this explicit error in favor of an implicit one (an invalid signature), so I'll be removing this code and leaving the current paths unchanged; please let me know if there was a strong reason for this adjustment, though.

PR #3752, opened 2024-11-08
Fix MV3 provider injection

hyphenized:
Manifest v3 CSP changes (`unsafe-inline`) prevent injecting the in-page provider source through a script element. Injection has been moved to a content script so that execution time and context remain consistent. File name was also renamed for clarity.

An additional workaround was added to keep the service worker alive as a safeguard against missing store updates due to termination.


Shadowfiend:
Finally properly integrated this into #3757 . Closing!

PR #3756, opened 2024-12-14
Bump playwright to 1.48.0

mhluongo:
The new version should support Ubuntu 24.04 in Github Actions

https://github.com/microsoft/playwright/issues/30368

Latest build: [extension-builds-3756](https://github.com/tahowallet/extension/suites/32386348469/artifacts/2356688184) (as of Mon, 23 Dec 2024 17:03:36 GMT).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,9 @@ on:
     tags:
       - v*
 
+env:
+  NODE_OPTIONS: --openssl-legacy-provider
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


Shadowfiend:
That node bump requires setting some SSL env variables or bumping webpack and possibly other deps 😬

Shadowfiend (approved):
Ok we now have running playwright tests, though they are failing. They already were before, but these failures look a bit different.

Nonetheless, incremental progress is progress, so we're going to go ahead and land this.

I ended up adding the legacy OpenSSL flag to get us moving here.

PR #3757, opened 2024-12-23
Redux Redux: A few QOL fixes + Manifest v3

### LINKED ISSUE #3397: Ledger address list page does not scroll
### Discord Discussion Link

_No response_

Chrome

Adding a new Ledger, looking at the page that lists addresses for adding to Taho.

### What did not work?

The window was relatively short, making the Continue button invisible, and could not be scrolled, meaning it was impossible to continue without resizing the page.

### Version

v0.33.0

### Relevant log output

_No response_

Shadowfiend:
Sundry:
- Manifest v3 support by moving window provider to an injected content script instead of the previous injection (h/t @hyphenized) as well as a version bump on webext-redux and a variety of other small changes to deal with related dependency updates.
- Improved logging details across the board.
- Common address copying logic.
- Added Open in Explorer to the account popup menu in the account slideover.
- Main is refactored to ReduxService, with store logic in its own file.
- Support for certain invocations of wallet_requestPermissions, the lack of which was breaking some sites.
- Fixes for tracking networks that were marked as inactive but were reactivated.
- Improvements for fallback behavior that was unnecessarily retrying connections.
- Fixes for RPC batching to avoid issues with excessive batch sizes.
- Improved redux store performance for certain actions that can be batched.
- Fixes #3397.

Latest build: [extension-builds-3757](https://github.com/tahowallet/extension/suites/32443165539/artifacts/2362331963) (as of Thu, 26 Dec 2024 02:45:08 GMT).


background/features.ts
```diff
@@ -73,7 +73,7 @@ export function isEnabled(
   }
 
   if (checkBrowserStorage) {
-    const state = localStorage.getItem(flagName)
+    const state = "" as string // localStorage.getItem(flagName)
```
mhluongo, line 76: This looks like a mistake?

background/index.ts
```diff
@@ -4,8 +4,8 @@ import { Store as ProxyStore } from "webext-redux"
 import { produce } from "immer"
 import { Action, AnyAction, ThunkAction, ThunkDispatch } from "@reduxjs/toolkit"
 
-import { Delta, patch as patchDeepDiff } from "./differ"
-import Main from "./main"
+import { Delta, patch as patchDeepDiff } from "./services/redux/differ"
+import ReduxService from "./services/redux"
```
mhluongo, line 8: Glad to see the rename

background/lib/logger.ts
```diff
@@ -6,10 +6,10 @@ const HOUR = 1000 * 60 * 60
 
 const store = {
   get(key: string): string {
-    return window.localStorage.getItem(key) ?? ""
+    return "" // return window.localStorage.getItem(key) ?? ""
```
mhluongo, line 9: Also looks like a mistake

background/lib/logger.ts
```diff
@@ -6,10 +6,10 @@ const HOUR = 1000 * 60 * 60
...
   },
   set(key: string, value: string): void {
-    window.localStorage.setItem(key, value)
+    // window.localStorage.setItem(key, value)
```
mhluongo, line 12: Boo

background/lib/posthog.ts
```diff
@@ -69,7 +69,8 @@ export function createPosthogPayload(
       $lib: USE_ANALYTICS_SOURCE,
       // properties[$current_url] is a convention used by posthog
       // Let's store the URL so we can differentiate between the sources later on.
-      $current_url: window.location.href,
+      $current_url:
```
mhluongo, line 72: Interesting, does Posthog know what `service-worker` means?

background/lib/priceOracle.ts
```diff
@@ -213,6 +220,27 @@ export async function getUSDPriceForTokens(
 ): Promise<{
   [contractAddress: string]: UnitPricePoint<FungibleAsset>
 }> {
+  if (assets.length > BATCH_SIZE) {
```
mhluongo, line 223: V nice

background/lib/token-lists.ts
```diff
@@ -146,8 +146,12 @@ export function mergeAssets<T extends FungibleAsset>(
         metadata: {
           ...matchingAsset.metadata,
           ...asset.metadata,
-          tokenLists: (matchingAsset.metadata?.tokenLists || [])?.concat(
-            asset.metadata?.tokenLists ?? [],
+          tokenLists: Array.from(
```
mhluongo, line 149: I shudder to think how many there were before the unique

background/services/abilities/index.ts
```diff
@@ -196,7 +196,7 @@ export default class AbilitiesService extends BaseService<Events> {
   }
 
   async fetchAbilities(): Promise<void> {
-    localStorage.setItem(this.ABILITY_TIME_KEY, Date.now().toString())
+    // localStorage.setItem(this.ABILITY_TIME_KEY, Date.now().toString())
```
mhluongo, line 199: Another one @Shadowfiend

background/services/abilities/index.ts
```diff
@@ -210,7 +210,7 @@ export default class AbilitiesService extends BaseService<Events> {
   }
 
   async refreshAbilities(): Promise<void> {
-    const lastFetchTime = localStorage.getItem(this.ABILITY_TIME_KEY)
+    const lastFetchTime = Date.now() // localStorage.getItem(this.ABILITY_TIME_KEY)
```
mhluongo, line 213: Mistake

background/services/chain/asset-data-helper.ts
```diff
@@ -89,7 +89,9 @@ export default class AssetDataHelper {
       )
     } catch (error) {
       logger.debug(
-        "Problem resolving asset balances; network may not support it.",
+        "Problem resolving asset balances; network",
```
mhluongo, line 92: I'd rather see this in a template string but :shrug:

background/services/indexing/index.ts
```diff
@@ -76,7 +76,21 @@ interface Events extends ServiceLifecycleEvents {
...
+     * - `all`: the payload contains all tracked assets and should be used to
+     *          replace any lists of tracked assets elsewhere in the system.
+     * - `network`: the payload contains all tracked assets for the given
+     *              networks in the payload, and should be used to replace any
+     *              network-specific lists.
+     * - `incremental`: the payload makes no representation of completeness and
+     *                  should only be used to update or add assets from the
+     *                  payload.
+     */
+    loadingScope: "all" | "network" | "incremental"
```
mhluongo, line 92: I hate that we had to make this a thing, but it's a good solution

background/services/provider-bridge/index.ts
```diff
@@ -254,8 +269,15 @@ export default class ProviderBridgeService extends BaseService<Events> {
...
       ).toString()
 
-      // these params are taken directly from the dapp website
-      const [title, faviconUrl] = event.request.params as string[]
+      // these params are taken directly from the dapp website and injected by
+      // /provider-bridge/index.ts . The last two parameters are guaranteed
+      // to be this, and no parameters that might be passed before these should
+      // be read as injected (which can happen in certain invocations of these
+      // methods).
+      const [title, faviconUrl] = (event.request.params as unknown[]).slice(
```
mhluongo, line 277: Strange formatting here, not a fan

src/background.ts
```diff
@@ -19,10 +20,18 @@ browser.runtime.onInstalled.addListener((obj) => {
...
+    Object.keys(RuntimeFlag).forEach(
+      (flagName) => "", // localStorage.removeItem(flagName),
     )
   }
 })
 
-startMain()
+startRedux()
+
+// FIXME: Temporary workaround to prevent the service worker from being suspended
```
mhluongo, line 31: We could use a follow-up issue for this as well as the other FIXME

mhluongo (changes requested):
Found a couple major issues that look like leftover local testing kludges.

Concept ACK, though, and around to review a fix

background/features.ts
Shadowfiend: Not per se. I forgot to call out I disabled client side feature flags as storage access is completely different.

background/lib/logger.ts
Shadowfiend: This one is going to cost us logging for now but same deal, we have to use the async store which breaks most of the assumptions in here atm.

background/lib/posthog.ts
Shadowfiend: No more than anything else, this is just a string in posthog. Previously it was just the same URL always for the background script.

background/services/chain/asset-data-helper.ts
Shadowfiend: Template strings stringify, separate params generally get interactive objects in the console.

background/lib/token-lists.ts
Shadowfiend: I cut my store in 4 I think by fixing this via asset updates.

manifest/manifest.json
```diff
@@ -4,19 +4,37 @@
...
+    {
+      "matches": ["file://*/*", "http://localhost/*", "https://*/*"],
+      "js": ["window-provider.js"],
+      "run_at": "document_start",
+      "world": "MAIN",
       "all_frames": true
     }
   ],
+  "sandbox": {
+    "pages": ["provider-bridge.js"]
```
hyphenized, line 34: I couldn't find if there's any benefit to adding this script under [sandbox](https://developer.chrome.com/docs/extensions/reference/manifest/sandbox)

background/services/chain/serial-fallback-provider.ts
```diff
@@ -398,6 +452,42 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
       // Awful, but what can ya do.
       const stringifiedError = String(error)
 
+      if (
+        stringifiedError.match(/Batch size too large/) &&
+        (pendingBatchSize === undefined || pendingBatchSize === 0)
+      ) {
+        this.forcedBatchMaxPreviousSize =
+          pendingBatch?.length ?? pendingBatchSize ?? 1
```
hyphenized, line 460: A bit confused here, would appreciate if you could explain why we use the reference here to grab the most current length, instead of the stored pending batch size

hyphenized (commented):
Mostly looks good, left a couple of comments but I definitely need to take a deeper look into the batch call tweak. Merry Christmas doggos

mhluongo:
Okay, I better understand where we're at now — we're going to lose some functionality in the MV3 upgrade until we work through service worker issues, but time is of the essence.

Doing some local testing of our release candidate. Random things I've noticed

* Alt-T isn't working to open the extension. This might be because I have another (disabled) install
* Connecting to Curve (eg crvusd.curve.fi) is rough, though is always has been  
* Connecting to Aave, Threshold, and other dApps I use is way faster! I was a little shocked
* Connecting to Uniswap doesn't work at all :disappointed: and leads to an error page in the Taho UI

image

mhluongo (approved):
Would love :pray: follow up issues for the features that were effectively disabled by the MV3 upgrade. While I've seen a couple wallet connection issues, I think we should release and dp some follow-up QA to figure those out.

Shadowfiend:
> Alt-T isn't working to open the extension. This might be because I have another (disabled) install

This has been wobbly for a long time, for reasons that are still very unclear to me. It should just be set-it-and-forget-it.

> Connecting to Uniswap doesn't work at all 😞 and leads to an error page in the Taho UI

Will have a look at that. Would be good to implement EIP-6963 sooner rather than later, may or may not help. This may be related to https://github.com/tahowallet/extension/pull/3757/commits/5c35455459e61e47d8fa6a8685c9b9c46095c3e3 though.

PR #3758, opened 2024-12-24
Redux Redux Patch

mhluongo:
Attempt to fix a couple issues in https://github.com/tahowallet/extension/pull/3757

Latest build: [extension-builds-3758](https://github.com/tahowallet/extension/suites/32399595008/artifacts/2358101407) (as of Tue, 24 Dec 2024 03:37:53 GMT).

### DIFF
--- a/background/features.ts
+++ b/background/features.ts
@@ -73,7 +73,7 @@ export function isEnabled(
   }
 
   if (checkBrowserStorage) {
-    const state = "" as string // localStorage.getItem(flagName)
+    const state = localStorage.getItem(flagName)
     return state !== null ? state === "true" : RuntimeFlag[flagName]
   }
 
--- a/background/services/abilities/index.ts
+++ b/background/services/abilities/index.ts
@@ -196,7 +196,7 @@ export default class AbilitiesService extends BaseService<Events> {
   }
 
   async fetchAbilities(): Promise<void> {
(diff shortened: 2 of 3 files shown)


background/features.ts
```diff
@@ -73,7 +73,7 @@ export function isEnabled(
   }
 
   if (checkBrowserStorage) {
-    const state = "" as string // localStorage.getItem(flagName)
+    const state = localStorage.getItem(flagName)
```
Shadowfiend, line 76: `localStorage` and `window` aren't available in service workers, that's how we got ourselves here unfortunately :grimacing:

background/services/abilities/index.ts
```diff
@@ -196,7 +196,7 @@ export default class AbilitiesService extends BaseService<Events> {
   }
 
   async fetchAbilities(): Promise<void> {
-    // localStorage.setItem(this.ABILITY_TIME_KEY, Date.now().toString())
+    localStorage.setItem(this.ABILITY_TIME_KEY, Date.now().toString())
```
Shadowfiend, line 199: Oh also we're not using this whole service at the moment. See line 155, which is where the magic would usually happen <_<

background/features.ts
mhluongo: I see :upside_down_face:

mhluongo:
sadpanda.gif

PR #3766, opened 2025-01-17
Update e2e tests for MV3

hyphenized:
A few tests still need to be updated but this seems to be working for now.

Latest build: [extension-builds-3766](https://github.com/tahowallet/extension/suites/33257255362/artifacts/2454391856) (as of Mon, 20 Jan 2025 05:20:06 GMT).


Shadowfiend:
Oof incidentally looks like our upload-action usage is about to be killed on Jan 30 🙃 

Looks like we're still seeing some light failures in the tests, might be timing-related though and seems like a better outcome than previously. I'd be down to merge this & keep rolling once the lint issues are resolved.

PR #3768, opened 2025-01-17
Correct dApp info passed on eth_requestAccounts

hyphenized:
This was making one of the e2e test fail as an empty array was passed as dApp connection info.

## To test
- [ ] Head to https://swap.cow.fi, connect and check title and logo displayed under connection info

Latest build: [extension-builds-3768](https://github.com/tahowallet/extension/suites/33209975358/artifacts/2449542737) (as of Fri, 17 Jan 2025 22:18:50 GMT).

### DIFF
--- a/background/services/provider-bridge/index.ts
+++ b/background/services/provider-bridge/index.ts
@@ -234,7 +234,6 @@ export default class ProviderBridgeService extends BaseService<Events> {
       // methods).
       const [title, faviconUrl] = (event.request.params as unknown[]).slice(
         -2,
-        0,
       ) as string[]
 
       const existingPermission = await this.checkPermission(origin, dAppChainID)


background/services/provider-bridge/index.ts
```diff
@@ -234,7 +234,6 @@ export default class ProviderBridgeService extends BaseService<Events> {
       // methods).
       const [title, faviconUrl] = (event.request.params as unknown[]).slice(
         -2,
-        0,
```
Shadowfiend, line 237: 🤔 -2,0 and -2 alone should be synonymous. What's the difference in outcome?

hyphenized: I thought that too but -2 means start from (array length + -2). So setting the end as 0 makes it return an empty array because the ending position is before the starting position

Shadowfiend: Well that's embarrassing 😬

Shadowfiend (approved):

PR #3771, opened 2025-01-31
Rework RPC caching

hyphenized:
This streamlines RPC caching by consolidating balance and hasCode caches into a single unified map. Caching TTL can now be configured more flexibly per method inside `#cacheSettings`.

The current caching mechanism does not prevent request duplication within batches but that can be implemented afterwards

## To test
- [ ] Update extension, check cache cleanups are being logged

Latest build: [extension-builds-3771](https://github.com/tahowallet/extension/suites/34887267581/artifacts/2652426123) (as of Tue, 25 Feb 2025 23:29:41 GMT).


hyphenized:
I think this one is relatively safe to merge, but the next set of changes will have to wait until e2e's are passing

background/services/chain/serial-fallback-provider.ts
```diff
@@ -95,6 +93,39 @@ function waitAnd<T, E extends Promise<T>>(
   })
 }
 
+function getCacheKey(method: string, params: unknown) {
+  return `${method}::${JSON.stringify(params, (_k, val) => {
```
Shadowfiend, line 97: Could use a comment here explaining what in the world we're doing hehe.

background/services/chain/serial-fallback-provider.ts
```diff
@@ -95,6 +93,39 @@ function waitAnd<T, E extends Promise<T>>(
...
+    if (val === null) {
+      return null
+    }
+
+    if (typeof val === "bigint") {
+      return `bigint:${val}`
```
Shadowfiend, line 103: Since we never deserialize, can this just be `${val}n`?

Shadowfiend (commented):
Need to read a little more here; I basically hate this as much as I hated the previous cache but still don't have a better idea :laughing:

Will try to wrap up review tomorrow.

Shadowfiend (approved):
Gave this a shot, seems to work. Let's 🚢

PR #3772, opened 2025-02-05
Refactor RPC error handling

hyphenized:
Reworks error handling for more precise error handling when retrying requests or switching between fallback providers.

The gas limit check will improve reliability of eth_calls made to the price oracle/multicall. We should consider removing this method from alchemy provider routed methods as requests also fail intermittently on Alchemy

Latest build: [extension-builds-3772](https://github.com/tahowallet/extension/suites/34374323610/artifacts/2590024395) (as of Fri, 14 Feb 2025 02:57:07 GMT).


background/services/chain/errors.ts
```diff
@@ -0,0 +1,91 @@
+/**
+ * These are from geth, nethermind, and ankr
+ */
+const BATCH_LIMIT_REGEXP =
+  /batch size (is )?too large|batch size limit exceeded|batch too large/i
+
+/**
+ * - WebSocket is already in CLOSING
+ * We are reconnecting
```
Shadowfiend, line 9: Just for clarity, would love to indent the descriptions a bit, or doing `- error: description` or something—right now this is a bit hard to read.

background/services/chain/errors.ts
```diff
@@ -0,0 +1,91 @@
...
+const INVALID_RESPONSE_REGEXP = /bad response|bad result from backend/i
+
+const RATE_LIMIT_REGEXP =
+  /too many requests|call rate limit exhausted|exceeded the quota usage/i
+
+const GAS_LIMIT_REGEXP = /out of gas/i
+
+export type RPCErrorType =
+  // Batch size too large
+  | "batch_limit_exceeded"
```
Shadowfiend, line 44: We generally use `kebab-case` unless we can't in the codebase, probably we stick to that here?

background/services/chain/serial-fallback-provider.ts
```diff
@@ -452,8 +453,10 @@ export default class SerialFallbackProvider extends JsonRpcProvider {
       // Awful, but what can ya do.
       const stringifiedError = String(error)
 
+      const errorType = getErrorType(stringifiedError, method)
+
       if (
-        stringifiedError.match(/Batch size too large/) &&
+        errorType === "batch_limit_exceeded" &&
```
Shadowfiend, line 459: Wondering if we want to move to `switch`/`case` here with these changes.

Shadowfiend (approved):
This is a :fire: refactor, thank you.

Shadowfiend:
Hm… I'm not sure why this is causing provider timeouts in the unit tests. Any thoughts?

background/services/chain/serial-fallback-provider.ts
hyphenized: Let's revisit this later, the fallback logic still needs some work done

hyphenized:
> Hm… I'm not sure why this is causing provider timeouts in the unit tests. Any thoughts?

Looks like it was a bug uncovered by this refactor :grimacing:

background/services/chain/errors.ts
hyphenized: Added some space and indentation between errors and descriptions

Shadowfiend (approved):
Yeh yeh.

PR #3773, opened 2025-02-05
Fix extension mocking during tests

hyphenized:
Fixed testing setup of `jest-webextension-mock` and, removed the custom `webextension-polyfill` mock as it was no longer necessary.

Missing methods not supplied by `jest-webextension-mock` are initialized during setup through `setup-extension-mock.ts`. Also fixed an `await` omission in a test...

Latest build: [extension-builds-3773](https://github.com/tahowallet/extension/suites/34000366804/artifacts/2544398379) (as of Thu, 06 Feb 2025 00:21:17 GMT).


setupJest.ts
```diff
@@ -11,7 +11,7 @@ const IS_CI = process.env.CI === "true"
 // CI workflows to get an overview of both failed expectations and possible errors.
 // This is not the case during development, hence, we set the minimum log level to warning
 // as it helps with debugging while writing new tests.
-logger.logLevel = IS_CI ? LogLevel.error : LogLevel.warn
+logger.logLevel = IS_CI ? LogLevel.off : LogLevel.warn
```
Shadowfiend, line 14: Do we not want error logs?

Shadowfiend (approved):
Big improvement and unblocks CI, left one question for later. Let's :shipit:

setupJest.ts
hyphenized: Yes, but I removed it because jest already points at where a test failed

PR #3775, opened 2025-02-06
Update e2e tests

hyphenized:
This also changes the workflow config so there's an output even if the job is cancelled, helping with debugging when playwright hangs indefinitely.

Latest build: [extension-builds-3775](https://github.com/tahowallet/extension/suites/34404092171/artifacts/2593695375) (as of Fri, 14 Feb 2025 16:45:49 GMT).

### DIFF
--- a/background/services/indexing/index.ts
+++ b/background/services/indexing/index.ts
@@ -360,7 +360,7 @@ export default class IndexingService extends BaseService<Events> {
    ******************* */
 
   private acceleratedTokenRefresh: {
-    timeout: number | undefined
+    timeout: ReturnType<typeof setTimeout> | undefined
     assetLookups: {
       asset: SmartContractFungibleAsset
       addressOnNetwork: AddressOnNetwork
@@ -430,7 +430,7 @@ export default class IndexingService extends BaseService<Events> {
           )
 
           this.acceleratedTokenRefresh.assetLookups.push(...assetLookups)
-          this.acceleratedTokenRefresh.timeout ??= window.setTimeout(
+          this.acceleratedTokenRefresh.timeout ??= setTimeout(
             this.handleAcceleratedTokenRefresh.bind(this),
             ACCELERATED_TOKEN_REFRESH_TIMEOUT,
           )
--- a/e2e-tests/regular/dapp-connect.spec.ts
+++ b/e2e-tests/regular/dapp-connect.spec.ts
@@ -10,8 +10,8 @@ test.describe("dApp Connections", () => {
     const dappPage = await context.newPage()
     await dappPage.goto("https://swap.cow.fi/")
     await dappPage
-      .locator("#swap-button")
       .getByRole("button", { name: "Connect Wallet" })
+      .first()
       .click()
 
     // Get page after a specific action (e.g. clicking a link)
@@ -44,8 +44,8 @@ test.describe("dApp Connections", () => {
     const dappPage2 = await context.newPage()
     await dappPage2.goto("https://swap.cow.fi/")
     await dappPage2
-      .locator("#swap-button")
       .getByRole("button", { name: "Connect Wallet" })
+      .first()
       .click()
 
     // Get page after a specific action (e.g. clicking a link)
@@ -73,8 +73,8 @@ test.describe("dApp Connections", () => {
     const dappPage = await context.newPage()
     await dappPage.goto("https://swap.cow.fi/")
     await dappPage
-      .locator("#swap-button")
       .getByRole("button", { name: "Connect Wallet" })
+      .first()
       .click()
 
     // Get page after a specific action (e.g. clicking a link)
--- a/e2e-tests/regular/dapp-connect.spec.ts
+++ b/e2e-tests/regular/dapp-connect.spec.ts
@@ -14,30 +14,15 @@ test.describe("dApp Connections", () => {
       .first()
       .click()
 
+    const popupPage = walletPageHelper.getConnectPopup()
     // Get page after a specific action (e.g. clicking a link)
-    const [popupPage] = await Promise.all([
-      context.waitForEvent("page"),
-      await dappPage.locator("text=Injected").click(), // Opens a new tab
-    ])
-    await popupPage.waitForLoadState()
+    await dappPage.locator("text=Injected").click() // Opens a new tab
 
-    // Clear the one-time informational popup, if present.
-    const connectingPopupTitle = popupPage.locator("h3", {
-      hasText: "Connecting with Taho",
-    })
+    await popupPage.ready()
 
-    expect(await connectingPopupTitle.count()).toBe(1)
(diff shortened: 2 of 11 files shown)


e2e-tests/utils/connect-popup.ts
```diff
@@ -0,0 +1,65 @@
+import { BrowserContext, Page, expect } from "@playwright/test"
+
+export default class SiteConnectionHelper {
```
Shadowfiend, line 3: This class name is wrong, right? <_<

e2e-tests/regular/dapp-connect.spec.ts
```diff
@@ -10,53 +10,38 @@ test.describe("dApp Connections", () => {
...
-      name: "Background close",
-    })
+    await popupPage.ready()
 
-    await popupCloseLocator.click()
-    await popupCloseLocator.waitFor({ state: "detached", timeout: 1000 })
+    await popupPage.hideDappConnectPopup()
 
-    await popupPage.locator("button", { hasText: "Reject" }).click()
+    await popupPage.closeWindow()
```
Shadowfiend, line 25: Should we explicitly call `rejectConnection` to get more surface area on the test? (Namely: ensure that rejecting also closes the popup.)

e2e-tests/regular/nfts.spec.ts
```diff
@@ -59,9 +72,8 @@ test.describe("NFTs", () => {
 
       // Wait until load finishes
       await expect(page.getByTestId("loading_doggo")).toBeVisible()
+      stall.end("Stop stalling simplehash")
```
Shadowfiend, line 75: I'mma say it: this test is wayy too convoluted lol. Someday perhaps we'll have time to figure out how to test what we want here without craziness...

e2e-tests/regular/signing.spec.ts
```diff
@@ -4,33 +4,30 @@ test.describe("Signing", () => {
...
 
     await signDataPromptOpens.then(async (prompt) => {
+      await prompt.waitForLoadState()
       await prompt.getByRole("button", { name: "Sign" }).click()
     })
-    // There's a bug on login.xyz that makes the test fail
-    // (https://discord.com/channels/862419652286218251/886997073650655232/1173226370776694794).
-    // We're adding the expectation of failure. Playwright will throw `Expected
-    // to fail, but passed.` when bug is fixed.
-    test.fail()
```
Shadowfiend, line 33: :fire:

src/background.ts
```diff
@@ -28,10 +27,9 @@ browser.runtime.onInstalled.addListener((obj) => {
   }
 })
 
-let redux: Promise<ReduxService>
+const redux = startRedux()
 
 browser.runtime.onConnect.addListener(async (port) => {
-  ;(await redux).connectPort(port)
+  const service = await redux
```
Shadowfiend, line 33: Won't this start redux every time a popup is created, and not kick off redux initialization until the popup is created in the worst case? https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onConnect

ui/components/NFTs/NFTListPortfolio.tsx
```diff
@@ -34,7 +35,18 @@ export default function NFTListPortfolio(props: {
   const isLoading = useBackgroundSelector(selectIsReloadingNFTs)
 
   if (isEmptyPortfolio) {
-    return <NFTsExploreBanner type={type} />
+    return (
+      <>
+        {isLoading && (
```
Shadowfiend, line 40: Should be `isLoading && nftCount === 0`? Since this is set by reloading NFTs, it feels like any time we reload (esp: when an account is added) we'll show the whole thing in a loading state, vs previously we would show the cached list of NFTs and then update with more later.

It feels like in the case where we've already got NFTs and we're in a loading state, we should have something more subtle than replacing the whole list. Thoughts?

Shadowfiend (approved):
Mostly game. I have one question I think is mildly blocking about the NFT loader and a couple of small nonblocking notes.

ui/components/NFTs/NFTListPortfolio.tsx
hyphenized: Agree, just checked and it seems that's exactly what `isEmptyPortfolio` tracks. Wasn't sure why the dog wasn't appearing until I found this early return :sweat_smile:

e2e-tests/regular/nfts.spec.ts
hyphenized: Yeah tbh I was just about to remove the loading state test because it was too flaky with the timeout, but somehow found a better way to set it up :smile:

src/background.ts
Shadowfiend: Ahhh, the service worker itself is guaranteed to be a singleton, of course. Good deal.

background/services/indexing/index.ts
```diff
@@ -430,7 +430,7 @@ export default class IndexingService extends BaseService<Events> {
           )
 
           this.acceleratedTokenRefresh.assetLookups.push(...assetLookups)
-          this.acceleratedTokenRefresh.timeout ??= window.setTimeout(
+          this.acceleratedTokenRefresh.timeout ??= setTimeout(
```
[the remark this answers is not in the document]
hyphenized: Linter didn't like it but `globalThis` works though, should I do the change?

`Unexpected use of 'self'.eslint[no-restricted-globals](https://eslint.org/docs/latest/rules/no-restricted-globals)`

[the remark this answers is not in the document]
Shadowfiend: Looks like that comes from https://github.com/facebook/create-react-app/tree/b532a58792ea79ba7b0e85c1cb42b2a36dafe276/packages/confusing-browser-globals, and doesn't apply to us because `window` isn't available in a service worker context. I would disable next line with a note that window can't be used in the service worker context 🤷

e2e-tests/utils/connect-popup.ts
Shadowfiend, line 1: Yak shaving note: this really should be `connect-popup-helper`, onboarding should be `onboarding-helper`, etc.

For someday in the future heh.

ui/components/NFTs/NFTListPortfolio.tsx
Shadowfiend: You wanting to make this change or nah?

Shadowfiend: (If not I think we're good to merge here.)

hyphenized: I think it's good like this? This block only runs if the nftCount is 0 though I feel like I misunderstood something

Shadowfiend: OH. I see. Yep. I'm blind. Ty sir.

Shadowfiend (approved):
🚢

PR #3776, opened 2025-02-07
Fix duplicate message handling in webext-redux

hyphenized:
For some reason, messages sent from the extension service worker unexpectedly trigger attached listeners twice. While they have distinct origins—one directly from the script and the other from a generic extension ID hostname—they still share the same UUID, confirming they are indeed the same message. This happens even though `sendMessage` is called only once. 

image

This adds a UUID to every state update sent by webext-redux, ensuring that patches are processed only once by the receiving store. This fixes issues where the extension ends up in an invalid state, especially when handling arrays. 

Latest build: [extension-builds-3776](https://github.com/tahowallet/extension/suites/34344687334/artifacts/2586128780) (as of Thu, 13 Feb 2025 15:10:17 GMT).


patches/webext-redux+4.0.0.patch
```diff
@@ -150,7 +208,25 @@ index 69de40f..5ea4e4b 100644
...
+     serializedMessagePoster({
+       type: _constants.STATE_TYPE,
+       payload: currentState,
+-      channelName: channelName // Notifying what store is broadcasting the state changes
++      channelName: channelName, // Notifying what store is broadcasting the state changes
++      messageId: crypto.randomUUID()
+     });
+ 
+     /**
+@@ -185,7 +196,7 @@ var _default = exports["default"] = function _default() {
```
Shadowfiend, line 229: So I hate to do this, but...

- Currently we do track our changes to webext-redux, just at https://github.com/shadowfiend/webext-redux (I don't remember why :grimacing: need to transfer that over here). Then generate the diff from there. Any chance you can open a PR for these changes there so we don't lose the change?
- Feels like crypto.randomUUID might introduce some performance bottlenecks, but I think we can take that before moving forward.

Shadowfiend:
F, any chance we can yak shave that upload-action build issue?

hyphenized:
> F, any chance we can yak shave that upload-action build issue?

Hmmm I think I've already fixed this on a different branch. Let me pull that here

Shadowfiend (approved):
A'ight let's go.

PR #3779, opened 2025-02-13
Reduce price lookups to tracked assets

hyphenized:
Improves price fetching by performing lookups only for tracked assets. Another gain here is the reduced load on the RPCs used, quicker price updates and better service worker startup time.

Also renamed `getAssetsByAddress` to `indexAssetsByAddress` for clarity and removed `getActiveAssetsByAddressForNetwork` as it was just filtering assets by network.

## To test
- [ ] Update extension, check prices are being updated and there are fewer requests made in the background

Latest build: [extension-builds-3779](https://github.com/tahowallet/extension/suites/34372093155/artifacts/2589757365) (as of Fri, 14 Feb 2025 01:31:24 GMT).


background/services/indexing/index.ts
```diff
@@ -95,7 +95,7 @@ interface Events extends ServiceLifecycleEvents {
   removeAssetData: SmartContractFungibleAsset
 }
 
-const getAssetsByAddress = (assets: SmartContractFungibleAsset[]) => {
+const indexAssetsByAddress = (assets: SmartContractFungibleAsset[]) => {
```
Shadowfiend, line 98: But this doesn't index anything, it just fetches them right?

background/services/indexing/index.ts
```diff
@@ -863,28 +852,34 @@ export default class IndexingService extends BaseService<Events> {
    * Loads prices for all tracked assets except untrusted/custom network assets
    */
   private async getTrackedAssetsPrices() {
+    logger.debug("Fetching tracked asset prices...")
     // get the prices of all assets to track and save them
+    const assetsToTrack = await this.db.getAssetsToTrack()
```
Shadowfiend, line 857: Unless I'm misunderstanding the pieces here, this change is problematic. As a starting point, see my note here:

https://github.com/tahowallet/extension/blob/d108f08a1368fc2eb42a6df6bb1085d780073fda/background/services/indexing/index.ts#L302-L305

So I think this change here *should* be pulling all assets we want to track, but in practice I think that might skip custom assets due to how we've been handling those?

Additionally, the cached assets list we were using before contains assets from the token lists—which we need to pull because otherwise we're impacting asset discovery for users To wit, if the user receives a known token, they won't see it in their list, they'll have to explicitly add it as a custom asset.

hyphenized: This is pulling both custom assets and token list assets

Custom assets are tracked when added here:
https://github.com/tahowallet/extension/blob/d8df56358b1181ab75bf208470d5cad95546e743/background/services/indexing/index.ts#L788-L792

Assets from token lists are tracked if they have a positive balance, after their balance is queried, here:
https://github.com/tahowallet/extension/blob/price-tracking/background/services/indexing/index.ts#L1057-L1060

One thing I noticed near that comment is that we're not filtering from price fetching custom assets that were hidden ("removed")  :thinking:

background/services/indexing/index.ts
hyphenized: I thought indexing could be applied here because it created a map of (address => asset), that's why I named it that way, another option I had was `keyAssetsByAddress`, similar to what lodash `keyBy` does. If you don't find this a meaningful change I can revert it :eyes:

background/services/indexing/index.ts
Shadowfiend: Awesome re: custom assets, let's update `cacheAssets` to not additionally do the custom asset lookup then.

Re: assets that were hidden/removed—we should be untracking them, but perhaps are not? I don't think that's pants-on-fire but could be a handy optimization for sure. We should be able to do a check on startup that marks all hidden as untracked and then untracked in the future? Realizing now there are no db functions for untracking an asset lol—separate PR I think.

Re: token lists: amazing! I forgot that this is price lookup, not balance lookup. Huge optimization not to check all token list assets, awesome find!

Shadowfiend: I wonder then if cached assets should not include token list assets and we should explicitly add token list assets to balance lookups, rather than the current approach where we cache assets including the token list and make an exception for price lookups.

To be figured out later.

background/services/indexing/index.ts
Shadowfiend: `keyAssetsByAddress` sounds like an awesome adjustment 💪

Shadowfiend: Actually we have lodash… Why don't we use it instead of having a named function? I think what we're really trying to do here is `_.keyBy(assets, ({ contractAddress }) => normalizeAddress(contractAddress))`, which IMO is better even if repeated than having this extra wrapper.

Anyway that's probably shaving the yak a little too long. Let's do that rename you flagged and merge this puppy 🐶

background/services/indexing/index.ts
hyphenized: Cached assets include all assets related to a network. They are cached because, previously, accessing all network assets required running `mergeAssets` over token list assets, network assets, and discovered assets (iirc). This was very expensive and significantly slowed down the extension.

For example when a erc20 transfer is enriched, the indexing service checks cached assets to see if the asset is known. These are also broadcast with the `assetsLoaded` event when the chain service starts and once again after fetching token lists.

background/assets.ts
```diff
@@ -226,6 +227,11 @@ export function isSmartContractFungibleAsset<T extends AnyAsset>(
   return "homeNetwork" in asset && isFungibleAsset(asset)
 }
 
+export function keyAssetsByAddress(assets: SmartContractFungibleAsset[]) {
+  const key: keyof SmartContractFungibleAsset = "contractAddress"
+  return keyBy(assets, key)
```
Shadowfiend, line 232: We do need to normalize the address to be safe I think.

hyphenized: Should be good now, there are a couple of instances where contract addresses are not normalized but I'd rather not dig into that here

Shadowfiend (approved):
Lfg 🚀

PR #3781, opened 2025-02-14
Replace e2e fork node with anvil

hyphenized:
This updates fork tests to run on anvil instead of hardhat. The change significantly improves test speed and prevents hangups. Even for simple cases anvil performs better without a cache.


mhluongo:
You interested in help getting this over the line ser? Seems worthwhile

Issue #3778, opened 2025-02-13
Persist enrichments in the `EnrichmentService`


mhluongo:
We're re-running enrichments constantly. They should be cached with a TTL at a minimum.

PR #3783, opened 2025-02-17
Reuse existing activity state data

hyphenized:
This adds a check into the account activity initialization process so we re-use already-enriched activity data if it corresponds to a finalized transaction.

Improves worker startup and prevents requests bursts by optimizing activity initialization for tracked accounts.

## To Test
- [ ] Load an account e.g. `testertesting.eth`
- [ ] On initial extension load, check that there are some RPC requests made for transaction enrichment
- [ ] Reload extension, check that there aren't as many requests as before

Latest build: [extension-builds-3783](https://github.com/tahowallet/extension/suites/34503181381/artifacts/2605178562) (as of Mon, 17 Feb 2025 19:35:53 GMT).

### DIFF
--- a/background/redux-slices/activities.ts
+++ b/background/redux-slices/activities.ts
@@ -63,55 +63,6 @@ const addActivityToState =
     }
   }
 
-const initializeActivitiesFromTransactions = ({
-  transactions,
-  accounts,
-}: {
-  transactions: Transaction[]
-  accounts: AddressOnNetwork[]
-}): Activities => {
-  const activities: {
-    [address: string]: {
-      [chainID: string]: Activity[]
-    }
-  } = {}
-
-  const addActivity = addActivityToState(activities)
-
-  const normalizedAccounts = accounts.map((account) =>
-    normalizeAddressOnNetwork(account),
-  )
-
-  // Add transactions
-  transactions.forEach((transaction) => {
-    const { to, from, network } = transaction
-    const isTrackedTo = normalizedAccounts.some(
-      ({ address, network: activeNetwork }) =>
-        network.chainID === activeNetwork.chainID &&
-        sameEVMAddress(to, address),
-    )
-    const isTrackedFrom = normalizedAccounts.some(
-      ({ address, network: activeNetwork }) =>
-        network.chainID === activeNetwork.chainID &&
-        sameEVMAddress(from, address),
-    )
-
-    if (to && isTrackedTo) {
-      addActivity(to, network.chainID, transaction)
-    }
-    if (from && isTrackedFrom) {
-      addActivity(from, network.chainID, transaction)
-    }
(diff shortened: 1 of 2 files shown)


background/redux-slices/activities.ts
```diff
@@ -123,13 +74,59 @@ const activitiesSlice = createSlice({
...
+
+      const normalizedAccounts = accounts.map((account) =>
+        normalizeAddressOnNetwork(account),
+      )
+
+      // Add transactions
+      transactions.forEach((transaction) => {
+        const { to, from, network } = transaction
+
+        const trackedAccounts = normalizedAccounts.filter(
```
Shadowfiend, line 100: For a self-send (where the same address is `from` and `to`), could this create a duplicate?

hyphenized: Nope, because activities in this state tree are keyed by account address and the transaction's hash, so in cases where the `from` and `to` match the same tracked address the transaction is only added once.

background/redux-slices/activities.ts
```diff
@@ -123,13 +74,59 @@ const activitiesSlice = createSlice({
...
+            (sameEVMAddress(to, address) || sameEVMAddress(from, address)),
+        )
+
+        trackedAccounts.forEach(({ address, network: { chainID } }) => {
+          const accountData = existingActivities?.[address]?.[chainID] ?? []
+          const activityExists = accountData.find(
+            ({ hash }) => hash === transaction.hash,
+          )
+
+          // if activity transaction has finalized, reuse existing data
```
Shadowfiend, line 112: Note that for finalization in general, just having a block height isn't necessarily enough. It's enough to prove inclusion but not finality—there may be a temporary fork that resolves and leaves the transaction at a *different* block height than the one that was first seen. (See: https://ethereum.org/en/developers/docs/consensus-mechanisms/pos/faqs/#what-is-finality ). Right now Ethereum has 15-minute finality, though Mezo has instant finality since it uses Cosmos/Tendermint, so it's still a live concern.

That said, I think we're good to start here and then iterate by resolving this in the backend—the real thing we want to do long term is store enriched data in IndexedDB so we can minimize what needs to be rechecked; #3778 covers this.

Shadowfiend (approved):
:shipit:

background/redux-slices/activities.ts
hyphenized: Thanks for the in-depth explanation. I suppose when revisiting this we could use some kind of polling on `getBlockByNumber/finalized` though I'm not sure if that'd be good enough :thinking:

Shadowfiend: Something like that would get us most of the way there, yeah. We can align it with an existing lookup or something of the sort.

PR #3785, opened 2025-02-21
Mezoification: Add base state and UI elements

hyphenized:
Adds base state, notification handlers and banner elements for upcoming campaign...

Banners and notifications should change based on the state of the campaign.

## To Test
- [ ] With `SUPPORT_MEZO_NETWORK=true` open the wallet, after a minute, a banner should appear
- [ ] Clicking the banner should prompt for push notification permissions
- [ ] If push notifications are enabled a notification should appear

Latest build: [extension-builds-3785](https://github.com/tahowallet/extension/suites/35544139271/artifacts/2734391449) (as of Wed, 12 Mar 2025 01:15:33 GMT).


Shadowfiend:
Screenshot 2025-03-03 at 15 42 07

Alignment of the open-in-new-tab icon seems off; it also doesn't highlight on hover.

background/redux-slices/ui.ts
```diff
@@ -47,6 +54,13 @@ export type UIState = {
   routeHistoryEntries?: Partial<Location>[]
   slippageTolerance: number
   accountSignerSettings: AccountSignerSettings[]
+  activeCampaigns: {
+    "mezo-claim"?: {
+      dateFrom: string
```
Shadowfiend, line 59: Guessing if we make this `Date` it won't (de)serialize?

background/services/analytics/index.ts
```diff
@@ -134,6 +137,15 @@ export default class AnalyticsService extends BaseService<Events> {
     await super.internalStopService()
   }
 
+  get analyticsUUID() {
```
Shadowfiend, line 140: From below it looks like we can lose this?

background/services/redux/index.ts
```diff
@@ -1710,6 +1723,138 @@ export default class ReduxService extends BaseService<never> {
...
+          browser.tabs.create({ url: "https://mezo.org/matsnet/borrow" })
+          this.preferenceService.markDismissableItemAsShown(
+            "mezo-borrow-notification",
+          )
+        },
+      })
+    }
+
+    this.store.dispatch(updateCampaignState(["mezo-claim", campaignData]))
+  }
```
Shadowfiend, line 1856: Fine for our quick implementation here, but we really need to move this either into redux itself or out to a service—the ReduxService should be doing the minimum possible to coordinate between the two. More like what the island service hookup at https://github.com/tahowallet/extension/blob/main/background/services/redux/index.ts#L1567-L1569 does.

Need to do some deeper arch work to figure that out before we can clean up since these are more complex interactions though.

Shadowfiend (commented):
Left a few initial thoughts, haven't made it all the way through yet.

background/redux-slices/ui.ts
hyphenized: Yup, it will remain a string

background/services/analytics/index.ts
hyphenized: If you're referring to the error below, that's just to block access until service has initialized. This could happen during tests or if there's an attempt to get this data from within another service's initialization without awaiting `service.started()`

background/services/redux/index.ts
hyphenized: Thanks for giving this a look, wasn't confident enough to put this into it's own service without receiving some feedback on the approach first 🕵️

ui/components/TopMenu/TopMenuProtocolList.tsx
```diff
@@ -37,35 +39,26 @@ export const productionNetworkInfo = {
...
+  [ARBITRUM_SEPOLIA.chainID]: i18n.t("protocol.testnet"),
+}
 
 type TopMenuProtocolListProps = {
   onProtocolChange: (network: EVMNetwork) => void
 }
 
 /**
- * Places Ethereum and Mezo network above other networks
+ * Places Mezo network above other networks
```
Shadowfiend, line 53: Ngl I would love to kill this sort function. Its main purpose is to boost matsnet in the list if people have previously added it, is that right?

ui/components/TopMenu/TopMenuProtocolList.tsx
```diff
@@ -131,14 +125,13 @@ export default function TopMenuProtocolList({
...
+              {testNetworks.map((network) => (
                 <TopMenuProtocolListItem
-                  isSelected={sameNetwork(currentNetwork, info.network)}
-                  key={info.network.name}
-                  network={info.network}
-                  info={info.info}
+                  isSelected={sameNetwork(currentNetwork, network)}
+                  key={network.name}
+                  network={network}
+                  info={testNetworkInfo[network.chainID]}
```
Shadowfiend, line 133: I know this is descriptively named `info` lol—is this the description that lives under the network name? If so maybe let's rename to `description` (not a blocker).

ui/components/TopMenu/TopMenuProtocolList.tsx
hyphenized: Yup that's pretty much all it does. Currently testnets are hardcoded so we could remove it and use the hardcoded order but that seems a bit dirty 👀

background/lib/mezo.ts
```diff
@@ -0,0 +1,28 @@
...
+import { MEZO_TESTNET } from "../constants"
+import { sameEVMAddress } from "./utils"
+
+const BORROWER_CONTRACT_ADDRESS = "0x20fAeA18B6a1D0FCDBCcFfFe3d164314744baF30"
+
+const BorrowerABI = new Interface([
+  "function openTrove(uint256 _maxFeePercentage, uint256 debtAmount, uint256 _assetAmount, address _upperHint, address _lowerHint)",
+])
+
+// eslint-disable-next-line import/prefer-default-export
```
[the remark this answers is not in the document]
hyphenized: That's a good point, this is only used for the campaign. I initially put this under `/lib` because I thought this would be also used by the enrichment service.

Shadowfiend (commented):
Another round, still cookin'.

background/lib/mezo.ts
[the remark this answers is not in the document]
Shadowfiend: For sure. I think once we're enriching we'll probably export more anyway—and I wonder if the better approach here is actually making this file an ABI export file (`background/lib/mezo/musd-trovemanager-abi.ts` or so) and then we implement `checkIsBorrowingTx` directly where we need it and let enrichment do its own things with the ABI.

Anyway, none of that is blocking.

background/services/redux/index.ts
Shadowfiend: I think a few ways in my head we should factor it out:
- Mezo service that handles Mezo-related things including these kinds of campaigns.
- Notification service is directly responsible for stuff like this (not how we originally envisioned it, but may make sense).
- Campaigns service to handle these kinds of campaigns, Mezo service would handle more network-level/API stuff like mats, Passport, etc.

ui/_locales/en/messages.json
```diff
@@ -550,7 +550,8 @@
     "l2": "L2 scaling solution",
     "compatibleChain": "EVM-compatible blockchain",
     "avalanche": "Mainnet C-Chain",
-    "connected": "Connected"
+    "connected": "Connected",
+    "featuredNetwork": "new"
```
Shadowfiend, line 554: Is it featured or new? :wink:

ui/components/TopMenu/TopMenuProtocolList.tsx
Shadowfiend: Does it work even if the user had previously added it as a custom network? If so, then let's rely on the hardcoded order for now.

Shadowfiend:
> Screenshot 2025-03-03 at 15 42 07
> Alignment of the open-in-new-tab icon seems off; it also doesn't highlight on hover.

Want to bring this back to the fore, I think this is a blocker.

Also, I think if you hit the Login to Claim link, that should dismiss the notification—wdyt?

background/services/redux/index.ts
```diff
@@ -1676,11 +1690,10 @@ export default class ReduxService extends BaseService<never> {
     uiSliceEmitter.on(
       "shouldShowNotifications",
       async (shouldShowNotifications: boolean) => {
-        const isPermissionGranted =
-          await this.preferenceService.setShouldShowNotifications(
-            shouldShowNotifications,
-          )
-        this.store.dispatch(toggleNotifications(isPermissionGranted))
+        await this.preferenceService.setShouldShowNotifications(
```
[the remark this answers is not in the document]
hyphenized: I understand, I had to move permissions.request to the UI because in mv3 permission requests have to trigger in response to a user interaction and that detail seems to get lost on the way to the service worker 😕

background/services/preferences/index.ts
```diff
@@ -277,18 +276,10 @@ export default class PreferenceService extends BaseService<Events> {
   }
 
   async setShouldShowNotifications(shouldShowNotifications: boolean) {
-    if (shouldShowNotifications) {
```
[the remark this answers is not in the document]
hyphenized: Hmmm we can verify if permission has indeed been granted and guard that I think

hyphenized:
> > Screenshot 2025-03-03 at 15 42 07
> > Alignment of the open-in-new-tab icon seems off; it also doesn't highlight on hover.
> 
> Want to bring this back to the fore, I think this is a blocker.

Ok this should be fixed now

> 
> Also, I think if you hit the Login to Claim link, that should dismiss the notification—wdyt?

Good catch! That should also prevent additional "login to claim" notifications from appearing right?

background/services/analytics/index.ts
hyphenized: Hmmm I'm not following, we need this id to get the wallet's campaign state :thinking:

Shadowfiend (approved):
Ok let's go here.

PR #3786, opened 2025-02-24
Lower batch size for price discovery calls

hyphenized:
Helps reduce occurrences of `out of gas` errors when calling public RPCs. Won't help with calls made through multicall when they receive a valid RPC response with empty results. But, will reduce frequency of these scenarios.

Latest build: [extension-builds-3786](https://github.com/tahowallet/extension/suites/34820924724/artifacts/2644023557) (as of Mon, 24 Feb 2025 20:33:38 GMT).


Shadowfiend:
5?? 😿😿

Shadowfiend (approved):

PR #3787, opened 2025-02-25
Increase frequency of price fetching

hyphenized:
Prevents missed updates if a price fetch is requested shortly after after detecting a new token balance.

Latest build: [extension-builds-3787](https://github.com/tahowallet/extension/suites/34970189586/artifacts/2662917777) (as of Thu, 27 Feb 2025 12:20:43 GMT).

### DIFF
--- a/background/services/indexing/index.ts
+++ b/background/services/indexing/index.ts
@@ -124,7 +124,7 @@ export default class IndexingService extends BaseService<Events> {
    */
   private scheduledTokenRefresh = false
 
-  private lastPriceAlarmTime = 0
+  #pricesUpdateTimer: NodeJS.Timer | null = null
 
   private cachedAssets: Record<EVMNetwork["chainID"], AnyAsset[]> =
     Object.fromEntries(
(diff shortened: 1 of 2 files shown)


background/services/indexing/index.ts
```diff
@@ -178,9 +178,9 @@ export default class IndexingService extends BaseService<Events> {
       prices: {
         schedule: {
           delayInMinutes: 1,
-          periodInMinutes: 10,
+          periodInMinutes: 1,
```
Shadowfiend, line 181: Polling every minute seems excessive, especially given we pull prices when the service worker starts up, no?

In an ideal case we would poll visible prices more frequently—perhaps by having the popup monitor notify the indexing service when the UI is up/down, and tracking the current network—but not check everything everywhere all the time…

Shadowfiend (approved):
Primary change seems good, increased price check frequency makes me twitchy due to RPC throttling.

background/services/indexing/index.ts
hyphenized: That's true, I did this change on the basis that if the service worker is alive, then the user expects prices in the UI to be updated. That's assumption might be overkill though :thinking:

Shadowfiend: Since 10 is the status quo, how about we revert and then look at doing a smarter increase in update frequency based on visibility later?

hyphenized: Sounds good!

Shadowfiend (approved):

PR #3793, opened 2025-03-03
Force-toggle testnets visibility in network switcher

hyphenized:
This PR moves the testnet visibility setting to the preference service, and sets a migration to enable it on all existing and new installs.

## Testing
- [ ] On a new install, check testnets shows up by default on network switcher
- [ ] On existing install, after update testnets should show up

### DIFF
--- a/background/redux-slices/ui.ts
+++ b/background/redux-slices/ui.ts
@@ -15,7 +15,7 @@ import { DEFAULT_AUTOLOCK_INTERVAL } from "../services/preferences/defaults"
 export const defaultSettings = {
   hideDust: false,
   defaultWallet: false,
-  showTestNetworks: false,
+  showTestNetworks: true,
   showNotifications: undefined,
   collectAnalytics: false,
   showAnalyticsNotification: false,
@@ -77,6 +77,7 @@ export type Events = {
   updateAnalyticsPreferences: Partial<AnalyticsPreferences>
   addCustomNetworkResponse: [string, boolean]
   updateAutoLockInterval: number
+  toggleShowTestNetworks: boolean
 }
 
 export const emitter = new Emittery<Events>()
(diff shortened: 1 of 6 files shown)


ui/pages/Settings.tsx
```diff
@@ -186,8 +186,8 @@ export default function Settings(): ReactElement {
       )
   }
 
-  const toggleShowTestNetworks = (defaultWalletValue: boolean) => {
-    dispatch(toggleTestNetworks(defaultWalletValue))
```
Shadowfiend, line 190: We should probably kill this old action in the UI slice yeah?

Shadowfiend (commented):
Pretty shocked this wasn't in the preferences service to begin with :laughing:

ui/pages/Settings.tsx
hyphenized: Wait this old action is still used to update the "cached" store value 👀

background/redux-slices/ui.ts
```diff
@@ -392,6 +394,22 @@ export const setSelectedNetwork = createBackgroundAsyncThunk(
   },
 )
 
+export const toggleShowTestNetworks = createBackgroundAsyncThunk(
+  "ui/toggleShowTestNetworks",
+  async (value: boolean, { dispatch, getState }) => {
```
Shadowfiend, line 399:
```suggestion
  async (updatedValue: boolean, { dispatch, getState }) => {
```

Perhaps for clarity? Or even `updatedShowTestNetworksValue` or similar?

background/redux-slices/ui.ts
```diff
@@ -392,6 +394,22 @@ export const setSelectedNetwork = createBackgroundAsyncThunk(
...
+    const state = getState() as RootState
+
+    const currentNetwork = state.ui.selectedAccount.network
+
+    // If user is on one of the built-in test networks, don't leave them stranded
+    if (TEST_NETWORK_BY_CHAIN_ID.has(currentNetwork.chainID)) {
```
Shadowfiend, line 405: We should also guard on `!value` right? Like, if the user is toggling from off to on, we shouldn't push them into Ethereum—in theory they wouldn't be there anyway because they wouldn't be on a test network, but in fact they might be if they've added the testnet as a custom chain, for example.

PR #3794, opened 2025-03-07
Move active campaign concerns to campaign service

hyphenized:
This moves campaign related logic out of the main service into a dedicated service. It also causes dismissing campaign banners to clear and stop triggering notifications about the same topic. 

Campaign state is kept within the service db and updates are pushed to state in redux.

Latest build: [extension-builds-3794](https://github.com/tahowallet/extension/suites/35331061341/artifacts/2708368343) (as of Fri, 07 Mar 2025 03:08:56 GMT).

### DIFF
--- a/background/networks.ts
+++ b/background/networks.ts
@@ -422,3 +422,7 @@ export const isEnrichedEVMTransactionRequest = (
   transactionRequest: TransactionRequest,
 ): transactionRequest is EnrichedEVMTransactionRequest =>
   "annotation" in transactionRequest
+
+export const isTransactionWithReceipt = (
+  transaction: AnyEVMTransaction,
+): transaction is ConfirmedEVMTransaction => "status" in transaction
--- a/background/redux-slices/selectors/uiSelectors.ts
+++ b/background/redux-slices/selectors/uiSelectors.ts
@@ -37,7 +37,7 @@ export const selectShowingActivityDetail = createSelector(
 )
 
 export const selectActiveCampaigns = createSelector(
-  (state: RootState) => state.ui.activeCampaigns,
+  (state: RootState) => state.ui.campaigns,
   (campaigns) => campaigns,
 )
 
--- a/background/services/campaign/index.ts
+++ b/background/services/campaign/index.ts
@@ -1,0 +1,254 @@
+import dayjs from "dayjs"
+import isBetween from "dayjs/plugin/isBetween"
+import browser from "webextension-polyfill"
+import { isDisabled } from "../../features"
+import { checkIsBorrowingTx } from "../../lib/mezo"
+import AnalyticsService from "../analytics"
+import BaseService from "../base"
+import ChainService from "../chain"
+import EnrichmentService from "../enrichment"
+import NotificationsService from "../notifications"
+import PreferenceService from "../preferences"
+import { ServiceCreatorFunction, ServiceLifecycleEvents } from "../types"
+import { CampaignDatabase, getOrCreateDB } from "./db"
+import MEZO_CAMPAIGN, { MezoClaimStatus } from "./matsnet-nft"
+import { isTransactionWithReceipt } from "../../networks"
+import { Campaigns } from "./types"
+
+dayjs.extend(isBetween)
+
+interface Events extends ServiceLifecycleEvents {
+  /**
+   * Replaces UI state with active campaigns data
+   */
+  campaignStatusUpdate: Campaigns[]
+}
+
+export default class CampaignService extends BaseService<Events> {
+  static create: ServiceCreatorFunction<
+    Events,
+    CampaignService,
+    [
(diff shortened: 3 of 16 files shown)


background/networks.ts
```diff
@@ -422,3 +422,7 @@ export const isEnrichedEVMTransactionRequest = (
   transactionRequest: TransactionRequest,
 ): transactionRequest is EnrichedEVMTransactionRequest =>
   "annotation" in transactionRequest
+
+export const isTransactionWithReceipt = (
```
Shadowfiend, line 426: Should be `isConfirmedEVMTransaction`, since that's the type we're asserting.

background/redux-slices/selectors/uiSelectors.ts
```diff
@@ -37,7 +37,7 @@ export const selectShowingActivityDetail = createSelector(
 )
 
 export const selectActiveCampaigns = createSelector(
-  (state: RootState) => state.ui.activeCampaigns,
+  (state: RootState) => state.ui.campaigns,
```
Shadowfiend, line 40: Should above be `selectCampaigns` now?

background/services/campaign/index.ts
```diff
@@ -0,0 +1,254 @@
...
+    this.emitter.emit("campaignStatusUpdate", campaigns)
+  }
+
+  async checkMezoCampaignState() {
+    if (isDisabled("SUPPORT_MEZO_NETWORK")) {
+      return
+    }
+
+    await this.started()
+    const accounts = await this.chainService.getAccountsToTrack()
```
Shadowfiend, line 111: This implies we also need to monitor for new accounts and run this check on those, fwiw.

background/services/campaign/index.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    const hasSeenBorrowPush = shownItems.has(
+      MEZO_CAMPAIGN.notificationIds.canBorrow,
+    )
+
+    const hasSeenNFTNotification = shownItems.has(
+      MEZO_CAMPAIGN.notificationIds.canClaimNFT,
+    )
+
+    // fetch with uuid
```
Shadowfiend, line 143: Where are we fetching this from?

background/services/campaign/types.ts
```diff
@@ -0,0 +1,28 @@
...
+ *    data: {...}
+ *    disabled: boolean
+ *  }
+ * ```
+ */
+export type Campaigns = FilterValidCampaigns<MezoCampaign>
+
+export type CampaignIds = Campaigns["id"]
+
+export type FilterCampaignsById<T, K> = T extends { id: K } ? T : never
```
Shadowfiend, line 28: Would love to have types like this directly in the service rather than in a `types` file under it.

Separately, I think a better pattern for something like this might be:

```typescript
type Campaign<Id extends string> = {
   id: Id
   data: unknown
   /**
    * The campaign is disabled if the user is e.g. not eligible
    */
   enabled: boolean
 }

const booyan: Campaign<"blap"> = {
    id: "blap",
    data: "bammer",
    enabled: true
}

const booyak: Campaign<"bloop"> = {
    id: "bloop",
    data: "bammer",
    enabled: true
}

const allCampaigns = [booyan, booyak] as const
type Campaigns = typeof campaigns[number]["id"]
```

This avoids the relatively complex conditional types in favor of interaction with simple types. The downside is if a new campaign isn't explicitly declared as `Campaign<...>`, the `Campaigns` type collapses to `string`.

Anyway, this isn't a blocker.

background/services/notifications/index.ts
```diff
@@ -73,32 +76,39 @@ export default class NotificationsService extends BaseService<Events> {
...
           browser.notifications.onClicked.addListener(
             boundHandleNotificationClicks,
           )
+
           browser.notifications.onClosed.addListener(
-            boundCleanUpNotificationClickHandler,
+            boundHandleNotificationClose,
           )
         } else {
+          // CLear up notifications
```
Shadowfiend, line 91: Just to check, is the intent here to clear all notifications if the user turns notifications off?

ui/_locales/en/messages.json
```diff
@@ -563,7 +563,7 @@
     },
     "mainMenu": "Settings",
     "signing": "Signing",
-    "showNotifications": "Show Subscape notifications",
+    "showNotifications": "Show Mezo notifications",
```
Shadowfiend, line 566: Maybe just “Show notifications”?

Shadowfiend (approved):
Ok did a pass, nothing blocking. Going to test it as part of the parent.

background/services/campaign/index.ts
hyphenized: Endpoint on mezo API here: https://github.com/thesis/mezo-portal/pull/1591 👀

Issue #3692, opened 2023-12-08
Long raw version of a transaction doesn't fit on a screen


michalinacienciala:
When transaction that's being confirmed in the wallet has long hex, the view showing the raw format does not show the end of the hex value (even scrolling does not work).

![image](https://github.com/tahowallet/extension/assets/78352137/fb77e516-e4fd-4265-aeab-b70e5dd8cb36)

PR #3797, opened 2025-03-07
fix: ISSUES #3692

XiaoYu0807:
Snipaste_2025-03-07_18-11-19

Latest build: [extension-builds-3797](https://github.com/tahowallet/extension/suites/35345216617/artifacts/2803309260) (as of Sun, 23 Mar 2025 14:13:05 GMT).


Shadowfiend (approved):
Wish we could find a way to do this without the extra container but it'll do :grin:

PR #3798, opened 2025-03-07
Add content-script port reconnection

### LINKED ISSUE #3790: dApp loses connectivity if service worker goes to sleep
After the service worker stops, connected dApps can no longer use the port created through `chrome.runtime.connect` to communicate with the extension. This issue persists until a page reload.

hyphenized:
This one is weird, sometimes the port closes when alt-tabbing even if the service worker is still alive. 

Closes #3790

Sets an event handler on both the isolated context content-script and the taho `window.ethereum` provider content-script to reconnect and reset context if the extension messaging port closes.

Keep in mind connection happens from all frames in the page

## Testing

- [ ] Head to chrome://serviceworker-internals/ and stop the service worker,
- [ ] On an active tab's console, you should see something similar to the screenshot

Latest build: [extension-builds-3798](https://github.com/tahowallet/extension/suites/35377277650/artifacts/2714211159) (as of Fri, 07 Mar 2025 23:16:39 GMT).


provider-bridge/index.ts
```diff
@@ -74,8 +75,22 @@ export default function connectProviderBridge(): void {
...
 
+  port.onMessage.addListener(onPortMessageListener)
+
+  port.onDisconnect.addListener(() => {
+    window.removeEventListener("message", onWindowMessageListener)
+    port.onMessage.removeListener(onPortMessageListener)
+
+    // Log this for debugging
+    // eslint-disable-next-line no-console
+    console.log("Reconnecting port from contentScript")
```
Shadowfiend, line 90: Let's color this as we do other Taho stuff so it's clear where it's coming from.

Really we should be thinking about disabling this logging by default in prod builds, but one thing at a time.

Shadowfiend (approved):
Fixed the color thing, rest is ready to go.

PR #3801, opened 2025-03-12
Set analytics data on mezo campaign items

hyphenized:
Adds `src` query strings to identify banners clicked from Mezo, and sets analytics events to dispatch on banner click

## Testing
- [ ] Check posthog events are being sent on banner/notification click

Latest build: [extension-builds-3801](https://github.com/tahowallet/extension/suites/35630590546/artifacts/2745445456) (as of Thu, 13 Mar 2025 13:21:56 GMT).


Shadowfiend (approved):
I do wish we had a central path for these to go through so we weren't duping the tabs and such.

Anyway let's :shipit:

PR #3803, opened 2025-03-13
enhance swap quote functionality with error handling and auto refresh

Ernesto-tha-great:
- Debounce time for updating the swap quote was reduced from 500ms to 300ms.
- Added an error field in the QuoteUpdate type to store error messages.
- Introduced a retry system to handle network failures and transient errors.
- Ensures the user always sees the latest price without manually refreshing.
- Improved readability and maintainability by restructuring the fetchQuote logic

Latest build: [extension-builds-3803](https://github.com/tahowallet/extension/suites/36557107620/artifacts/2919065164) (as of Thu, 10 Apr 2025 12:03:33 GMT).


ui/utils/swap.ts
```diff
@@ -41,6 +43,8 @@ export type QuoteUpdate = {
   quoteRequest: SwapQuoteRequest
   timestamp: number
   amount: string
+  // Added error property to better handle and display error states
```
Shadowfiend, line 46: Would love to pull this kind of note into commit messages—it's specific to the change rather than a note for a developer looking at this later on.

ui/utils/swap.ts
```diff
@@ -148,12 +158,20 @@ export function useSwapQuote(useSwapConfig: {
   requestQuoteUpdate: DebouncedFunc<
     (config: RequestQuoteUpdateConfig) => Promise<void>
   >
+  // Added reset function to clear current quote
+  resetQuote: () => void
```
Shadowfiend, line 162: Do we use this anywhere?

ui/utils/swap.ts
```diff
@@ -174,6 +192,43 @@ export function useSwapQuote(useSwapConfig: {
...
+  // Added reset quote function
+  const resetQuote = useCallback(() => {
+    setQuoteRequestState({
+      quote: null,
+      loading: false,
+      loadingType: undefined,
+    })
+  }, [setQuoteRequestState])
+
+  // Auto-refresh quote every 30 seconds to ensure it doesn't get stale
```
Shadowfiend, line 206: Do we currently show any UI feedback that this happens, or is it just happening in the background?

ui/utils/swap.ts
```diff
@@ -41,6 +43,8 @@ export type QuoteUpdate = {
   quoteRequest: SwapQuoteRequest
   timestamp: number
   amount: string
+  // Added error property to better handle and display error states
+  error?: string
```
Shadowfiend, line 47: I wonder if the correct way to model this is something like:

```typescript
type QuoteUpdate = current

type QuoteUpdateResult = QuoteUpdate | { error: string, previousQuote: QuoteUpdate }
```

Or even having the error state not include the quote at all. In particular, while I love retries, what I don't love about just having an error alongside the quote update result is that the current version leaves the quote unchanged rather than return any made-up values (e.g. below we set a 0 gas price in case of failure).

ui/utils/swap.ts
```diff
@@ -174,6 +192,43 @@ export function useSwapQuote(useSwapConfig: {
   const mountedRef = useIsMounted()
 
   const requestId = useRef(0)
+  const retryCount = useRef(0)
```
Shadowfiend, line 195: Love the introduction of retries. I wonder if we might abstract this out for reuse in other places as well, as we have a few places where we could use better retry hygiene.

Shadowfiend (commented):
Left some notes! Thanks for contributing these changes!

ui/utils/swap.ts
hyphenized: I think this is good, we're currently fetching quotes through `fetchJson` @ https://github.com/tahowallet/extension/blob/main/background/redux-slices/0x-swap.ts#L243-L257 which comes with some retry options available but I'm not sure what it's defaults are set to. Moving the retry logic here gives us the choice to present this state in the UI

ui/utils/swap.ts
```diff
@@ -174,6 +192,43 @@ export function useSwapQuote(useSwapConfig: {
...
+        }
+      }, 30000) // Refresh every 30 seconds
+    }
+
+    return () => {
+      if (refreshInterval) {
+        clearInterval(refreshInterval)
+      }
+    }
+  }, [quoteRequestState.quote, quoteRequestState.loading])
```
hyphenized, line 231: This is already implemented but under a feature flag, see here: https://github.com/tahowallet/extension/blob/c6aa8be173771f1e0f1de6c7497baa8215dc10e0/ui/pages/Swap.tsx#L304-L326

Ernesto-tha-great: gotcha. thank you!

ui/utils/swap.ts
Ernesto-tha-great: done!

ui/utils/swap.ts
Ernesto-tha-great: actually happens in the background.

Shadowfiend:
The updated PR still does auto-refreshes but they're also being done on the frontend side, am I reading this right? Probably should stick to one or the other.

Ernesto-tha-great:
@Shadowfiend 

Please take a look once more. I have removed the refresh on the FE

Shadowfiend:
You may have had a rebase issue here—the history is looking gnarly. Maybe see if you can check out latest main and then cherry pick [c6aa8be](https://github.com/tahowallet/extension/pull/3803/commits/c6aa8be173771f1e0f1de6c7497baa8215dc10e0) and [203f85c](https://github.com/tahowallet/extension/pull/3803/commits/203f85cdaebe9b315d06213865dd9e10ba8a15cc) and force push over the current branch.

ui/pages/Swap.tsx
```diff
@@ -466,12 +416,9 @@ export default function Swap(): ReactElement {
               priceImpact={quote?.priceDetails?.priceImpact}
               isPriceDetailsLoading={isLoadingPriceDetails}
               showPriceDetails
-              // FIXME: Merge master asset list with account balances.
```
Shadowfiend, line 469: Is this no longer relevant?

ui/pages/Swap.tsx
```diff
@@ -357,8 +313,6 @@ export default function Swap(): ReactElement {
             sellAsset,
             buyAsset,
             gasPrice:
-              // Let's use the gas price from 0x API for Optimism
-              // to avoid problems with gas price on Optimism Bedrock.
```
Shadowfiend, line 361: This feels relevant still.

ui/pages/Swap.tsx
```diff
@@ -234,8 +227,6 @@ export default function Swap(): ReactElement {
       setBuyAsset(newBuyAsset)
       setBuyAmount("")
 
-      // Updating the target asset quotes the new target asset against the existing
-      // source amount.
```
Shadowfiend, line 238: This and previous notes that enlighten the semantic meaning of the conditionals still seem useful.

ui/pages/Swap.tsx
Ernesto-tha-great: yes. its been fixed

Ernesto-tha-great:
@Shadowfiend 

All done now.

Ernesto-tha-great:
> This doesn't seem to be auto refreshing quotes anymore?

yes. we removed the auto refresh from the frontend so its being handled on the backend.

hyphenized:
> > This doesn't seem to be auto refreshing quotes anymore?
> 
> yes. we removed the auto refresh from the frontend so its being handled on the backend.

Okay I see that the front end code for auto refresh was removed but I don't see anything in the backend handling that?

Ernesto-tha-great:
> > > This doesn't seem to be auto refreshing quotes anymore?
> > 
> > 
> > yes. we removed the auto refresh from the frontend so its being handled on the backend.
> 
> Okay I see that the front end code for auto refresh was removed but I don't see anything in the backend handling that?

@hyphenized  @Shadowfiend 
Take a look now if this is better. I have resolved the auto refresh to be handled in the hook (swap.ts) every 10 sec.

Ernesto-tha-great:
@hyphenized @Shadowfiend  bumping this up

ui/utils/swap.ts
```diff
@@ -250,6 +248,28 @@ export function useSwapQuote(useSwapConfig: {
   const loadingSellAmount = quoteRequestState.loadingType === "getSellAmount"
   const loadingBuyAmount = quoteRequestState.loadingType === "getBuyAmount"
 
+  useEffect(() => {
+    if (useSwapConfig.pauseAutoRefresh) return undefined
```
Shadowfiend, line 252:
```suggestion
    if (useSwapConfig.pauseAutoRefresh) {
      return
    }
```

Shadowfiend (commented):
One stylistic note, otherwise need to test this but looks good to me.

Ernesto-tha-great:
@Shadowfiend 

all done!

Shadowfiend:
We had to revamp swaps to make them work against 0x's v2 api as v1 was decommissioned. We need to test this with those changes and then I think it can roll in.

PR #3805, opened 2025-03-14
Resolve campaign settings remotely

hyphenized:
Implements in memory storage to override certain build time parameters. This can be used to e.g. flip a feature flag without having to make a new release. Since campaign settings are only checked during browser startup, user will need to perform another browser restart before the feature is flipped.

## Testing
- [ ] Change the JSON URL to point to some locally hosted JSON
- [ ] Check that feature flag has flipped, e.g. network and banners show up

Latest build: [extension-builds-3805](https://github.com/tahowallet/extension/suites/35857654300/artifacts/2773798632) (as of Tue, 18 Mar 2025 16:02:26 GMT).


background/features.ts
```diff
@@ -27,14 +77,16 @@ export const RuntimeFlag = {
...
-  SUPPORT_MEZO_NETWORK: process.env.SUPPORT_MEZO_NETWORK === "true",
 } as const
 
 type BuildTimeFlagType = keyof typeof BuildTimeFlag
 
 export type RuntimeFlagType = keyof typeof RuntimeFlag
 
-export type FeatureFlagType = RuntimeFlagType | BuildTimeFlagType
+export type FeatureFlagType =
+  | RuntimeFlagType
```
Shadowfiend, line 87: Isn't this roughly the same as the dynamic one?

Shadowfiend: Yeah I think restoring all the runtime flag code paths and then hooking them up to your lookup stuff is going to achieve what we're trying to do here... But I might be misremembering.

hyphenized: Hold on I think I got carried away with the types lol. We only need to check === "true" for boolean flags

Shadowfiend: My point here wasn't about types, my point was that dynamic flags and runtime flags are the same thing :) The only thing we're changing is adding a way to resolve certain runtime flags (and fixing the fact that runtime flags used to be tied to localStorage).

Not worth blocking on given the need to release, but we shouldn't have two different code paths for these.

src/background.ts
```diff
@@ -27,7 +30,59 @@ browser.runtime.onInstalled.addListener((obj) => {
...
+  }
+}
+
+/**
+ * Retrieves stored settings from storage and pushes them to
+ * in memory storage
+ */
+const tryRestoreDynamicSettings = async () => {
+  try {
+    const flags = await browser.storage.local
```
Shadowfiend, line 63: With this in place, we can probably drop https://github.com/tahowallet/extension/pull/3805/files#diff-138415fc15e8cf95ec7b136ffceffab1c8f79372e3f75f5ac80f39048ee2e496R103-R106 and lean on this process to restore local storage items.

background/services/campaign/matsnet-nft.ts
```diff
@@ -11,7 +13,9 @@ export type MezoCampaignState = {
   state: MezoClaimStatus
 }
 
-export const NFT_CONTRACT_ADDRESS = "0x2A22371b53E6070AF6e38dfFC4228496b469D7FA"
+export const NFT_CONTRACT_ADDRESS = getRuntimeFlagValue(
```
Shadowfiend, line 16: If this file loads before we finish restoring or updating campaign settings, it'll get locked in at an outdated value, right?

Wondering if we shouldn't just avoid caching this entirely, and instead resolve it via `getRuntimeFlagValue` every time. Obviously would require changing the downstream consumption, and now isn't the moment, but just a thought.

Shadowfiend (approved):
Overall the construction feels a little brittle; in particular, if we toggle the feature flag, we still need to force a reload of the service worker to actually pull the updated data. Granted this is less crazy than it was when the background script was persistent, but it'll still be a bit unpredictable.

That said, let's roll with this now :car:

Shadowfiend:
I'm going to ignore the e2e tests here for now and circle back on the holistic release cut.

src/background.ts
hyphenized: Yes, that is a nice extra we get from this

PR #3811, opened 2025-03-21
Allow for replacing EIP6963 MetaMask as well

Shadowfiend:
Note that EIP6963 should remove the need to do this, but instead sites like Etherscan still only present MetaMask as an option despite using a EIP6963-aware library under the covers. Malice, incompetence, or dirty dealings? That is left as an exercise for the reader.

## Testing
- [x] Go to https://etherscan.io/token/0xCdF7028ceAB81fA0C6971208e83fa7872994beE5 (or any contract).
- [x] Click contract & write contract
- [x] Click Connect to Web3 and then MetaMask
  - [x] With MetaMask enabled and Taho masquerading as MetaMask, Taho should come up.
  - [x] With MetaMask enabled and Taho NOT masquerading as MetaMask, MetaMask should come up.
  - [x] With MetaMask disabled and Taho masquerading as MetaMask, Taho should come up.
  - [x] With MetaMask disabled and Taho NOT masquerading as MetaMask, Taho should come up.

Latest build: [extension-builds-3811](https://github.com/tahowallet/extension/suites/36036334941/artifacts/2797168574) (as of Fri, 21 Mar 2025 16:15:20 GMT).

### DIFF
--- a/src/window-provider.ts
+++ b/src/window-provider.ts
@@ -514,7 +514,14 @@ function setupProviderWrapper() {
     "eip6963:announceProvider",
     (event) => {
       if (
-        "detail" in event &&
+        !("detail" in event) ||
+        typeof event.detail !== "object" ||
+        !event.detail
+      ) {
+        return
+      }
+
+      if (
         "provider" in event.detail &&
         (event.detail.provider === window.ethereum ||
           event.detail.provider === window.taho)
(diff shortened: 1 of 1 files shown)


hyphenized (approved):
> Malice, incompetence, or dirty dealings? That is left as an exercise for the reader.

Most likely "Not a priority". What's a priority then? Another exercise for the reader.

> * [ ]  With MetaMask disabled and Taho NOT masquerading as MetaMask, nothing should come up.

This one failed, Taho did show up but I don't see an issue with that as I believe it's just accessing `window.ethereum`

Shadowfiend:
Looks like I hit an edge case where Taho wouldn't come up (especially when you change from default to non-default in the connect popup itself), but usually it will. I changed the testing checkbox accordingly.

PR #3821, opened 2025-08-29
Setup a proxy for USB connect/disconnect events handlers

hyphenized:
Fixes these warnings

image


Shadowfiend:
Yah will see if we can get some more/better logs meantime.

PR #3833, opened 2026-02-24
Detect newer Ledger devices

Shadowfiend:
We improve device detection and add Stax and Flex to the supported device list. Connections with these are still a little wobbly, but successful signatures have happened!


mhluongo:
Works on my end!

mhluongo (approved):
🎸

PR #3838, opened 2026-03-10
chore: remove unnecessary useMemo for showHiddenAssets

Ocheretovich:
Removes an unnecessary useMemo wrapper around a simple boolean expression. The computation is trivial and does not benefit from memoization, so simplifying it improves readability without changing behavior.


mhluongo:
Looks an awful lot like you're farming for airdrops with small open PRs across multiple crypto repos... going to close this one as it's not worth a test or release.

PR #3841, opened 2026-05-10
Preserve Ledger device type on disconnect events

suyua9:
## Summary

- keep the Ledger device type detected during connection
- emit that same type when the current Ledger disconnects instead of always reporting Nano S

## Context

#3833 added detection for newer Ledger devices such as Stax and Flex. The connection events now carry the detected device type, but the disconnect event still hard-coded `LEDGER_NANO_S`, which could make downstream state consumers see a different device family after unplugging the same Ledger.

## Tests

- `git diff --check`
- Not run: `yarn eslint background/services/ledger/index.ts` (Corepack failed while downloading Yarn from `registry.yarnpkg.com` with `ECONNRESET`)

### DIFF
--- a/background/services/ledger/index.ts
+++ b/background/services/ledger/index.ts
@@ -174,6 +174,8 @@ async function generateLedgerId(
 export default class LedgerService extends BaseService<Events> {
   #currentLedgerId: string | null = null
 
+  #currentLedgerType: LedgerType | null = null
+
   transport: Transport | undefined = undefined
 
   #lastOperationPromise = Promise.resolve()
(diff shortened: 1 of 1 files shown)


background/services/ledger/index.ts
```diff
@@ -174,6 +174,8 @@ async function generateLedgerId(
 export default class LedgerService extends BaseService<Events> {
   #currentLedgerId: string | null = null
 
+  #currentLedgerType: LedgerType | null = null
```
Shadowfiend, line 177: No, I don't think this is the move.

We should capture id & type in a side-by-side object to reflect the fact that both are set or unset at the same time.

Do we have any confirmed bugs with this?

PR #3842, opened 2026-05-15
🐗 Boar turnover 

jagodarybacka:
## Summary

This PR replaces Alchemy as the RPC provider with [Boar](https://boar.network) and adds Mezo mainnet as a default network in the Taho extension.

### Provider infrastructure
- Replaced `TahoRPCProvider` (Alchemy-backed) with a new `TahoBoarProvider` class that accepts per-chain Boar RPC URLs
- Renamed relevant files, variables, types, and method names from "alchemy" to "boar" to reflect the new provider (e.g. `AlchemyProvider` → `BoarProvider`, `AlchemyError` → `BoarError`, etc.)

### Mezo mainnet
As Boar RPC provide native support for Mezo, we are adding it as a default network in the extension with the following changes:
- Added `MEZO` network constant with `MEZO_MAINNET_BTC` as the base asset
- Added Mezo to `DEFAULT_NETWORKS` so it appears in the extension out of the box
- Added Mezo fallback RPC URL (`https://rpc-internal.mezo.org`) to `CHAIN_ID_TO_RPC_URLS`

### Configuration & CI
- Replaced `ALCHEMY_KEY` env var with six per-chain vars `BOAR_RPC_URL_<CHAIN_NAME>`
- Updated env files accordingly with placeholder values
- Updated `.github/workflows/build.yml` to inject all Boar RPC URL secrets

## Notes for reviewers

- The Boar API is response-compatible with Alchemy's enhanced APIs (`alchemy_getAssetTransfers`, `alchemy_getTokenBalances`, `alchemy_getTokenMetadata`), so no response parsing logic was changed - only naming and configuration.
- Mezo mainnet intentionally skips Alchemy-enhanced calls via `BOAR_ALCHEMY_UNSUPPORTED_CHAIN_IDS`; standard ERC-20 fallbacks apply for token metadata and balances on that network.
- The CI workflow expects the following new repository secrets to be configured: `BOAR_RPC_URL_ETHEREUM`, `BOAR_RPC_URL_OPTIMISM`, `BOAR_RPC_URL_POLYGON`, `BOAR_RPC_URL_ARBITRUM_ONE`, `BOAR_RPC_URL_SEPOLIA`, `BOAR_RPC_URL_MEZO` (and their `DEV_` prefixed counterparts for dev builds).


background/services/chain/serial-fallback-provider.ts
```diff
@@ -41,18 +40,18 @@ const isWebSocketProvider = (
...
  * This allows us to use "namespaces" for providers eg `alchemy_...` or `qn_...`
  *
  * The structure is network specific with an extra `everyChain` option.
- * The methods in this array will be directed towards alchemy on every network.
+ * The methods in this array will be directed towards Boar on every network.
  */
-export const ALCHEMY_RPC_METHOD_PROVIDER_ROUTING = {
+export const BOAR_RPC_METHOD_PROVIDER_ROUTING = {
   everyChain: [
     "alchemy_", // alchemy specific api calls start with this
```
Shadowfiend, line 53: Does boar support these?

Shadowfiend (commented):
At first glance I'm worried this is more far reaching than ideal, treating Boar as an Alchemy equivalent instead of a standard RPC (alchemy gets special treatment because we use a number of Alchemy-specific calls). 

It feels like the ideal move here is to put Boar in the same place on the RPC instantiation pipelines that Alchemy currently holds, and remove the Alchemy instantiations only. I think it'll only be a handful of LOC changes--but if you went down this path and it didn't work out I'll defer to that!

background/services/chain/serial-fallback-provider.ts
jagodarybacka: Yes, Boar supports all of these! We actually built an Alchemy compatibility mode specifically for this integration, `alchemy_` methods all work as-is. It's a new addition on our end - not yet exposed in the dashboard, but fully functional.

That's also the reason for the broader rename since Boar is a true drop-in for these methods, keeping "Alchemy" names throughout the repository felt like it would confuse future contributors. Happy to discuss if you'd prefer the narrower approach anyway.

Shadowfiend: Oh holy moly!

holy-moly-amazing-wow-26y1a93v293f41ig

Shadowfiend: OK, I will re-review this soon with that in mind since that of course changes the whole mindset 😆 Thank you!

Shadowfiend (approved):
Ok been test driving this for a few days and it works great 😎

PR #3844, opened 2026-05-19
Fix LedgerService stop hook

luyua9:
## Summary

- Fix `LedgerService.internalStopService()` to call `super.internalStopService()`.
- Keep the existing USB listener cleanup unchanged.

The stop hook currently calls the base start hook, which bypasses the base stop lifecycle path when LedgerService shuts down.

## Tests

- `git diff --check`

Not run locally: `yarn eslint background/services/ledger/index.ts` could not run because this clean worktree does not have dependencies installed (`Command "eslint" not found`).

### DIFF
--- a/background/services/ledger/index.ts
+++ b/background/services/ledger/index.ts
@@ -302,7 +302,7 @@ export default class LedgerService extends BaseService<Events> {
   }
 
   protected override async internalStopService(): Promise<void> {
-    await super.internalStartService() // Not needed, but better to stick to the patterns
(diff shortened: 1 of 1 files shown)


Shadowfiend:
> Not run locally: yarn eslint background/services/ledger/index.ts could not run because this clean worktree does not have dependencies installed (Command "eslint" not found).

Uh-oh. We are definitely not taking a contribution whose contributor has not even bothered to run linting during their vibing workflow.

